PR #17, opened 2020-05-28
make help.md on top of doc folder

MichelML:
I wasn't able to find the *help.md* file until a colleague told me it was in the doc folder.

Making it appear on top of the folder across all the images would be a great idea for easy retrieval.

Also, I removed the check-epam-email task on pre-commit in this PR (why not accepting contribution from the outside world?)

small addition, but still, it was annoying.

Cheers


AndreiMazol:
The structure of Ketcher has been updated.  This moment it is based on CRA Typescript template. All docs have been moved to public folder.

PR #84, opened 2020-11-08
#7 Mount on DOM element

AndreiMazol:
- add WebpackReplacePlugin
- split project into example & lib folder
- init dependencies
- update example project  configuration
- export root component from the lib
- fix issue with svg icons
- import all svg as SVGR components
- load templates from {staticResourcesUrl} path


Vesdet (approved):
LGTM

Issue #83, opened 2020-11-06
surpressing window.alert popups


jaohbib:
I would like to surpress any window.alert calls from Ketcher currently i recive the following  when i load ketcher 
![sllert](https://user-images.githubusercontent.com/72259105/98367962-d0e62a80-2036-11eb-80cc-f43f91f4bad3.PNG)

AndreiMazol:
Dear @jaohbib ,

thank  you for your interest to Ketcher.

This issue has been marked as Bug and it will be fixed in the nearest release.

Best regards,
Ketcher team

PR #89, opened 2020-11-10
#83 surpress any window alert calls

AndreiMazol:


AndreiMazol:
This PR should be merged after #84

Vesdet (approved):

Issue #46, opened 2020-09-07
The structures with atoms with valences rendering incorrectly after saving


AlexanderSavelyev:
**Steps to reproduce**

1. Launch application
2. Create any structure, for example, chain.
3. Double click any structure carbon atom.
4. In the opened "Atom properties" window select Valence 0 (zero). Click OK.
5. Click "Save as" button.
6. In the Format field set "Daylight SMILE". Click "Save to file" button.
7. Clear the canvas.
8. Open saved in 6th step file.

**Actual result**
The atom with added valence is replaced with diradical (singlet).
Other variants show in the table:

Valence\Format | MOL | Daylight Smile | Extended Smile | Daylight SMART | Inchi | Inchi Auxinfo | CML
-- | -- | -- | -- | -- | -- | -- | --
OK | is replaced with diradical | disappears | Can't be saved | disappears | disappears | disappears
OK | is replaced with diradical | disappears | Can't be saved | disappears | disappears | OK
OK | is replaced with diradical | is replaced with diradical | Can't be saved | is replaced with triplet | is replaced with diradical | OK
OK | is replaced with monoradical | is replaced with monoradical | Can't be saved | is replaced with monoradical | is replaced with monoradical | OK
OK | disappears | disappears | Can't be saved | disappears | disappears | OK
OK | OK | OK | Can't be saved | OK | OK | OK
OK | OK | OK | Can't be saved | OK | OK | OK
OK | OK | OK | Can't be saved | OK | OK | OK
OK | OK | OK | Can't be saved | OK | OK | OK

This table is about carbon atom. For other atoms it can be different behavior.

**Expected result:**
The structures with added valences should saved and opened as they were created. In the case if such structure can't be saved the warning message should appear and warn user that the valences can be lost after saving.

KseniiaSemenishcheva:
Warning message:<<In Daylight SMILES valence is not supported>> appears, but if User save structure in .smi or .inchi format (according to the table) and then open it - atom with valence rendering still incorrectly as at actual result description.

AlexanderSavelyev:
it make sense to revalidate the task and return warning messages for specific formats if explicit valence is not supported

PR #164, opened 2020-12-10
#46 The structures with atoms with valences rendering incorrectly after saving

Aniuta907:


Aniuta907:
valence
valence2

AndreiMazol (changes requested):
warnings should be rendered one by one in a column (see 2th screenshot)

Aniuta907:
Untitled

AndreiMazol (approved):

Issue #160, opened 2020-12-09
Add delete button for user templates


Aniuta907:
**Steps to reproduce**

1. Launch Editor
2. Click the 'Custom Template' button.
3. Click the 'User Templates' button.

**Actual result**
The user cannot delete his own templates.

<img width="486" alt="Untitled" src="https://user-images.githubusercontent.com/40429134/101649597-205fb200-3a4c-11eb-93dd-7af15b9e7b63.png">

**Expected result**
The user can delete his own templates.

PR #176, opened 2020-12-15
#160 Add delete button for user templates

Aniuta907:


AndreiMazol (changes requested):
It does not work while filtering templates

AndreiMazol (approved):

AndreiMazol (approved):
please revert

Issue #174, opened 2020-12-15
It is impossible to choose zoom value from zoom drop down list


ElenaKlepova:
**Description:**
It is impossible for a user to change zoom choosing zoom value from zoom drop down.

**Steps to Reproduce:**
1. Click zoom drop down.
2. Choose zoom value from the zoom drop down list.

**Expected result:**
Zoom drop down is shown when clicking. It is possible to change zoom value using the zoom drop down list.

**Actual Result:**
Zoom drop down is shown and hidden at the same time when clicking. It is impossible to choose zoom value from the zoom drop down list. 

![2020-12-15_11h24_16](https://user-images.githubusercontent.com/75974843/102191622-ba02e580-3eca-11eb-9839-509ff82aa3dc.gif)

ElenaKlepova:
Status: **PASS**
Build: **Release 2.1**
OS/Browser: Win10/Chrome
![2020-12-17_00h13_34](https://user-images.githubusercontent.com/75974843/102407513-4a414780-3ffd-11eb-9871-f40b12df2b1e.gif)

ElenaKlepova:
The issue is reopened because cut, copy and paste functions do not work.
 
**Description:**
It is impossible for a user to cut, copy or paste structures.

**Steps to Reproduce:**
1. Draw a structure.
2. Select it with a selection tool.
3. Click "Cut" button (or use shortcut Ctrl+X).
4. Click "Paste" button (or use shortcut Ctrl+V).
5. Click "Copy" button (or use shortcut Ctrl+C).
6. 4. Click "Paste" button (or use shortcut Ctrl+V).

**Actual Result:**
The structure is not cut, copied or pasted.
Paste function is not available via menu. 

![2020-12-17_10h18_28](https://user-images.githubusercontent.com/75974843/102455900-6f5fa580-4051-11eb-9b63-ca7f5586dd3f.gif)

**Expected result:**
The structure is cut, copied and pasted.
Paste function is available via menu.

ElenaKlepova:
**Status:** PASS
**Build:** Release 2.1.0
**OS/Browser:** Win10/Chrome

**Actual result:**
The structure is cut, copied and pasted.

![2020-12-18_00h06_54](https://user-images.githubusercontent.com/75974843/102544872-82f92380-40c6-11eb-9358-ac712940c312.gif)

PR #179, opened 2020-12-15
#174 It is impossible to choose zoom value from zoom drop down list

Aniuta907:


AndreiMazol (changes requested):
Please remove form element

AndreiMazol (approved):

Issue #54, opened 2020-10-05
Fire custom event on structure change


Vesdet:
When user changes structure on canvas would be good to fire "change" event with information about performed operation (e.g. `{ op: "AtomMove", from: { x: 10, y: 10 }, to: { x: 20, y: 20 }`).
Now Ketcher is already firing empty "change" event. #3 
Needs to investigate possibility to pass information about performed action to fired event.
Create format of custom event and apply it on all operations if previous right.

AnastasiiaMiagkikh:
https://github.com/epam/ketcher/pull/205

PR #205, opened 2020-12-24
#54 fire custom event on sctructure change

AnastasiiaMiagkikh:
https://github.com/epam/ketcher/issues/54


AndreiMazol (changes requested):
Please look at use of 'switch - case' construction

packages/ketcher-react/src/script/editor/operations/op.js
```diff
@@ -845,7 +888,7 @@ function RxnPlusMove(id, d, noinvalidate) {
   this.data = { id, d, noinvalidate }
 }
 
-RxnPlusMove.prototype = new Base('Nove rxn plus')
+RxnPlusMove.prototype = new Base('Move rxn plus')
```
AndreiMazol, line 891: Please replace all magic string by OperationType

AndreiMazol (approved):

Issue #246, opened 2021-01-15
Track if origin file was changed


AndreiMazol:
It is necessary to provide functionality that  allows to get information if original  structure was changed after uploading. The following method should be added into ketcher public API:

- isDirty(): boolean -- return true if any changes were introduced in initial  structure
-  setOrigin():void -- set current state as initial

KseniiaSemenishcheva:
Issue is verified and closed.

PR #250, opened 2021-01-18
#246 Track if origin file was changed

softcat23:


packages/ketcher-react/src/script/index.ts
```diff
@@ -0,0 +1,169 @@
...
+  }
+  #origin = null
+
+  getSmiles(): string {
+    return smiles.stringify(ketcher.editor.struct(), { ignoreErrors: true })
+  }
+
+  saveSmiles(): Promise<any> {
+    const struct = ketcher.editor.struct()
+    return structformat
```
TomasLight, line 46: please use camel case

packages/ketcher-react/src/script/index.ts
```diff
@@ -0,0 +1,169 @@
...
+
+  setMolecule(molString: string): void {
+    if (!(typeof molString === 'string')) return
+    ketcher.ui.load(molString, {
+      rescale: true
+    })
+  }
+
+  addFragment(molString: string): void {
+    if (!(typeof molString === 'string')) return
```
TomasLight, line 63: Please, replace
```
!(typeof molString === 'string')
```
by
```
typeof molString !== 'string'
```

Because this look is more readable

packages/ketcher-react/src/script/index.ts
```diff
@@ -0,0 +1,169 @@
...
+  editor: any
+  server: any
+  ui: any
+  apiPath: any
+  readonly buildInfo = {
+    version: process.env.VERSION,
+    buildDate: process.env.BUILD_DATE,
+    buildNumber: process.env.BUILD_NUMBER
+  }
+  #origin = null
```
TomasLight, line 38: why u use `#` in field name?

AndreiMazol (approved):

Issue #215, opened 2020-12-28
If structure contains SRU-Group User can't click 'Calculated Values' button - nothing happens after clicking.


KseniiaSemenishcheva:
**Affected Version:** _Standalone_ _, Remote (Multiple S-Groups)_
**Steps to Reproduce:**
1. Create chain structure.
2. Add SRU-polymer S-Group.
3. Click 'Calculated Values' button.

**Expected Result:**
In 'Chemical Formula' field displays correct formula, all another fields contains message: "error: Cannot calculate mass for structure with repeating units"
![remote](https://user-images.githubusercontent.com/75974465/103208241-30630700-4911-11eb-961c-228e0f281ffb.jpg)

**Actual Result:**
If structure contains SRU-Group User can't click 'Calculated Values' button - nothing happens after clicking. The same behavior is with Multiple S-Groups that are on the end of Structure. (Multiple S-Groups behavior reproduces in _Remote mode_ too.)

ensemenova:
Here is 2 independent cases.
1. Standalone and SRU-Group in the middle.  Caused by  https://github.com/epam/Indigo/issues/299. When the linked ticket will be fixed, behavior would be the same as on Remote
2. Standalone and Remove and SRU-Group at the end of the structure. Caused by invalid molecule structure, in this case molecule even couldn't be saved. Enhancement suggested https://github.com/epam/ketcher/issues/251 

For now both cases  were improved with showing error message to user.

KseniiaSemenishcheva:
Issue is closed as partially fixed. 
When Structure contains SRU-polymer group on the end of Structure now error message appears.
When Structure in Standalone mode contains SRU-polymer group in the middle of Structure - 'Calculated Values' window opens with zero values.

PR #252, opened 2021-01-18
#215 Show error to user in case any error happened during calculating values

ensemenova:
#215


TomasLight (commented):
should we to add TODO comment here to replace this alert in future on something appropriated?

AndreiMazol (approved):

TomasLight (approved):

PR #256, opened 2021-01-19
Use css modules instead of global styles 

softcat23:


packages/ketcher-react/src/script/ui/component/view/accordion/accordion.jsx
```diff
@@ -37,7 +38,10 @@ class Accordion extends Component {
 
   static Group({ caption, isActive, onActive, index, children }) {
     return (
-      <li className={'ac_tab ' + (isActive(index) ? 'active' : 'hidden')}>
+      <li
+        className={`${styles.accordion_tab}${
+          isActive(index) ? '' : ` ${styles.hidden}`
+        }`}>
```
TomasLight, line 44: you can install `clsx` package, and your code will look like:
```
className={clsx(styles.accordion_tab, {
  [styles.hidden]: !isActive(index),
})}
```

softcat23: https://github.com/epam/ketcher/pull/256/commits/2876606261adb7ad64138ff55dd969394243060a

Issue #190, opened 2020-12-17
Add save to MDL molfile v3000 support


AlexanderSavelyev:
**Background** 
Ketcher is able to load V2000 and V3000 molfiles, but only V2000 is available to be saved. A requirement is to add V3000 molfile in save options
**Acceptance criteria:**
* User opens save menu.
* New item in drop down appears "MDL Molfile V3000"
* If user selects the "MDL Molfile v3000" option, then the Ketcher outputs V3000 output  for the drawn structure
* Old menu item should be renamed "MDL Molfile" -> "MDL Molfile V2000"

KseniiaSemenishcheva:
Issue is verified and closed.

Issue #173, opened 2020-12-14
When User creates S-Group, group selection doesn't disappear.


KseniiaSemenishcheva:
**Description:**
If User creates S-Group and then clicks on canvas - S-group creation modal window appears because S-Group selection doesn't disappears. 

**Steps to Reproduce:**
1. Click the "Chain" tool.
2. Create the Chain structure.
3. Choose "S-Group" tool. 
4. Create the Generic S-group.
5. Click on canvas.

**Expected Result:**
S-Group selection is disappeared.

**Actual Result:**
Modal window "S-Group Properties" for creation one more S-Group is opened. Not expected behavior.

KseniiaSemenishcheva:
Issue is closed as fixed.

PR #260, opened 2021-01-21
#190 add mol and rxn v3000 + refactor to ts

ensemenova:


packages/ketcher-react/src/script/ui/data/convert/struct.types.ts
```diff
@@ -0,0 +1,132 @@
...
+    options?: any
+  ) {
+    this.name = name
+    this.mime = mime
+    this.ext = ext
+    this.supportsCoords = supportsCoords || false
+    this.options = options || {}
+  }
+}
+export type SupportedFormatStrings = keyof typeof SupportedFormat
```
TomasLight, line 64: Do we need export for this type? I think - no

packages/ketcher-react/src/script/ui/data/convert/structConverter.ts
```diff
@@ -17,117 +17,76 @@
...
       const converting = server
         .then(() =>
           server.convert(
             {
               struct: moldata,
-              output_format: map[format].mime
+              output_format: formatProperties.mime
             },
-            serverOpts
+            Object.assign({}, serverOpts, formatProperties.options || {})
```
TomasLight, line 82: Can we use spread operator here instead of `Object.assign` ?

packages/ketcher-react/src/script/ui/data/convert/structConverter.ts
```diff
@@ -136,45 +95,56 @@ export function toString(struct, format, server, serverOpts) {
   })
 }
 
-export function fromString(structStr, opts, server, serverOpts) {
+export function fromString(
+  structStr: any,
```
TomasLight, line 99: if here is `Str`, so should we use `string` type instead of `any`? In this case you can remove `Str` postfix

packages/ketcher-react/src/script/ui/data/convert/structConverter.ts
```diff
@@ -190,46 +160,54 @@ export function fromString(structStr, opts, server, serverOpts) {
...
-    format === 'inchi' ||
-    format === 'inchi-aux' ||
-    format === 'cml'
+    format === SupportedFormat.Smiles ||
+    format === SupportedFormat.SmilesExt ||
+    format === SupportedFormat.Smarts ||
+    format === SupportedFormat.InChI ||
+    format === SupportedFormat.InChIAuxInfo ||
+    format === SupportedFormat.CML
   ) {
```
TomasLight, line 206: I'd prefer to use `switch` statement instead of two `if`, when we just compare variable value with several variants

TomasLight: + row #173

packages/ketcher-react/src/script/ui/dialog/mainmenu/save.jsx
```diff
@@ -43,9 +48,9 @@ const saveSchema = {
     },
     format: {
       title: 'Format',
-      enum: Object.keys(structFormat.map),
-      enumNames: Object.keys(structFormat.map).map(
-        fmt => structFormat.map[fmt].name
+      enum: Object.keys(SupportedFormatPropertiesMap),
+      enumNames: Object.keys(SupportedFormatPropertiesMap).map(
+        fmt => SupportedFormatPropertiesMap[fmt].name
```
TomasLight, line 53: Please, don't shortcut `format` word to `fmt`. It is not so long to do so.

packages/ketcher-react/src/script/ui/state/hotkeys.js
```diff
@@ -115,8 +115,8 @@ const rxnTextPlain = /\$RXN\n+\s+0\s+0\s+0\n*/
 
 /* ClipArea */
 export function initClipboard(dispatch, getState) {
-  const formats = Object.keys(structFormat.map).map(
-    fmt => structFormat.map[fmt].mime
+  const formats = Object.keys(SupportedFormatPropertiesMap).map(
+    fmt => SupportedFormatPropertiesMap[fmt].mime
```
TomasLight, line 119: + dont' shortcut

packages/ketcher-react/src/script/ui/dialog/mainmenu/save.jsx
TomasLight: Same for `options` -> `opts`. It doesn't make sence. Just 3 symbol less

packages/ketcher-react/src/script/ui/data/convert/struct.types.ts
ensemenova: Removed as not actual to have it separate

packages/ketcher-react/src/script/ui/data/convert/structConverter.ts
ensemenova: fixed

packages/ketcher-react/src/script/ui/data/convert/structConverter.ts
ensemenova: fixed both

packages/ketcher-react/src/script/ui/state/hotkeys.js
ensemenova: fixed

packages/ketcher-react/src/script/ui/dialog/mainmenu/save.jsx
ensemenova: fixed

TomasLight (approved):

AndreiMazol (approved):

Issue #196, opened 2020-12-21
Split Ketcher into Core and Presentations modules


AndreiMazol:
In order to  support different presentation layers (React/Angular/Vue & etc) it is necessary to move core functionality into  separate module. Every  module should be delivered as an independent npm package.

![Ketcher_Structure_Diagram](https://user-images.githubusercontent.com/72735611/102755414-6c312600-437f-11eb-8e26-f8a8bf4baa72.png)

Scope of this task:
1. Create core package
2. move shared interfaces into core package
3. add link to ketcher-react & ketscher-standalone
4. prepare ketcher-core for publishing to npm

ElenaKlepova:
The issue is verified.

PR #268, opened 2021-01-26
#196 create ketcher-core package

AndreiMazol:


packages/ketcher-core/src/infrastructure/services/struct/RemoteStructServiceProvider.ts
```diff
@@ -14,18 +14,30 @@
...
+} from './structServiceProvider.types'
+import RemoteStructService from './RemoteStructService'
 
 class RemoteStructServiceProvider implements StructServiceProvider {
+  private apiPath: string
   mode: ServiceMode = 'remote'
 
-  createStructService(baseUrl: string, options: any): StructService {
-    return new IndigoService(baseUrl, options)
+  constructor(baseUrl: string) {
```
ensemenova, line 29: It would be better to call as apiPath everywhere

packages/ketcher-core/src/infrastructure/services/struct/RemoteStructService.ts
```diff
@@ -77,12 +77,12 @@ function indigoCall(
...
+class RemoteStructService implements StructService {
+  private apiPath: string
   private defaultOptions: any
 
-  constructor(base: string, defaultOptions: any) {
-    this.baseUrl = !base || /\/$/.test(base) ? base : base + '/'
+  constructor(apiPath: string, defaultOptions: any) {
+    this.apiPath = apiPath
     this.defaultOptions = defaultOptions
   }
```
TomasLight, line 87: I'd prefer to add readonly for this fields:
```
  constructor(
    private readonly apiPath: string,
    private readonly defaultOptions: any
  ) {
  }
```

packages/ketcher-react/package.json
```diff
@@ -70,7 +70,8 @@
     "subscription": "^3.0.0",
     "url-search-params-polyfill": "^8.1.0",
     "w3c-keyname": "^2.2.4",
-    "whatwg-fetch": "^3.4.1"
+    "whatwg-fetch": "^3.4.1",
+    "ketcher-core": "^1.0.0"
```
TomasLight, line 74: I guess it should be sorted by alpabet there

packages/ketcher-react/src/index.tsx
```diff
@@ -31,9 +30,9 @@ function Editor(props: EditorProps) {
       ...props,
       element: rootElRef.current
     })
-  }, [])
+  }, [props])
```
TomasLight, line 33: I don't think, that we should add props to dependencies, because in this case `init` function can be called a few times. and it can broke something, isn't it?

packages/ketcher-core/src/infrastructure/services/struct/RemoteStructServiceProvider.ts
AndreiMazol: fixed

packages/ketcher-core/src/infrastructure/services/struct/RemoteStructService.ts
AndreiMazol: agreed

packages/ketcher-react/package.json
AndreiMazol: done

packages/ketcher-react/src/index.tsx
AndreiMazol: comments added

TomasLight (approved):

Issue #226, opened 2020-12-29
Selection for Rectangle sometimes select object incorrectly


KseniiaSemenishcheva:
**Affected Version:** _Release/2.1_
**Steps to Reproduce:**
1. Draw a Rectangle using "Shape Rectangle" tool.
2. Hover mouse over area near the Rectangle.

**Expected Result:**
Rectangle is not highlighted until mouse hovers over the rectangle' line.

**Actual Result:**
User can highlight and select rectangle when mouse hovers over area near rhe line (not on line only).
![2020-12-29_09h59_05](https://user-images.githubusercontent.com/75974465/103266323-657a6280-49c0-11eb-8ab4-fb966e8f6090.gif)

KseniiaSemenishcheva:
Issue is closed as fixed.

PR #272, opened 2021-01-28
#226 add common api for structure creating

TomasLight:


packages/ketcher-react/src/script/builders/ketcher/ButtonsConfig.ts
```diff
@@ -0,0 +1,6 @@
+import { ButtonConfig } from './ButtonConfig'
+import { ButtonName } from './ButtonName'
+
+export type ButtonsConfig = {
+  [buttonName in ButtonName]: ButtonConfig
+}
```
TomasLight, line 6: will be corerected in different PR

packages/ketcher-react/src/script/ui/state/modal/form.js
```diff
@@ -91,7 +90,7 @@ export const formsState = {
     valid: true,
     result: {
       filename: 'ketcher',
-      format: SupportedFormat.Mol
+      format: 'mol'
```
AndreiMazol, line 93: Why do you use magic string instead of enum type?

packages/ketcher-react/src/script/ui/dialog/mainmenu/save.jsx
```diff
@@ -176,7 +186,7 @@ class Save extends Component {
             schema={this.saveSchema}
             init={{
               filename,
-              format: this.isRxn ? SupportedFormat.Rxn : SupportedFormat.Mol
+              format: this.isRxn ? 'rxn' : 'mol'
```
AndreiMazol, line 189: Could you remove magic strings?

AndreiMazol (approved):

Issue #296, opened 2021-02-08
Ketcher API: Add possibility to save RXN files


AndreiMazol:
Ketcher API should be extended by the following method:
1. containsReaction(): boolean - return true if rxn file should be generated
2. getRxnAsync(): Promise<string> OR throw Error

KseniiaSemenishcheva:
Issue is verified and closed.

PR #298, opened 2021-02-08
#296 Ketcher API: Add possibility to save RXN files

AndreiMazol:
- update StructFormatter methods
- add containsReaction() method
- add getRxnAsync() method


softcat23 (approved):
Looks good as far as I can tell.

TomasLight (approved):

Issue #311, opened 2021-02-10
Error after Undo/Redo actions when User has edited S-Group


KseniiaSemenishcheva:
**Affected Version:** _Remote, Standalone_
**Steps to Reproduce:**
1. Launch Ketcher.
2. Create chain structure.
3. Create, for example, Generic S-Group.
4. Change Generic S-Group to Multiple/SRU-polymer S-Group.
5. Click 'Undo' button.
6. Click 'Redo' button.

**Expected Result:**
User can undo/redo all actions with S-Groups. Undo/Redo works correctly for each step.

**Actual Result:**
After Undo action part of Structure with S-Group disappears from canvas. When User clicks 'Redo' - part of structure appears without S-Group brackets.
![2021-02-10_10h43_31](https://user-images.githubusercontent.com/75974465/107480612-eb77f480-6b8d-11eb-9fb9-ece168d6e8c0.gif)

KseniiaSemenishcheva:
Issue is closed as fixed.

PR #331, opened 2021-02-17
#311 sgroup invert bugfix

TomasLight:


packages/ketcher-react/src/script/editor/operations/descriptors.ts
```diff
@@ -0,0 +1,78 @@
+/****************************************************************************
+ * Copyright 2020 EPAM Systems
```
AndreiMazol, line 2: 2021

AndreiMazol (approved):

packages/ketcher-react/src/script/editor/operations/descriptors.ts
TomasLight: fixed in [update copyrights](https://github.com/epam/ketcher/pull/331/commits/75d9332884c442c41a205b5bc673efae70972879)

Issue #145, opened 2020-12-06
Add multiple arrow support for the same canvas


AlexanderSavelyev:
**Background:**
See comment
https://github.com/epam/ketcher/issues/66#issuecomment-729587224
**Acceptance criteria**:
* User opens the reaction menu and selects type of reaction arrow (different types will be added in a separate task)
* User can add several arrows on the same canvas
* The only format to support several arrows on save/load is "ket" format
* If user selects a different format, then a warning message should appear "The ${selected_format} does not support drawn elements: reaction arrows will be lost"
* If user saves a different format, only one reaction arrow should be saved (left top one) with default reaction saving
* For the "ket" format save/load should work correctly (including arrow length, direction, coordinate and type)

softcat23:
Within this task will be added only one new reaction arrow type:
![image](https://user-images.githubusercontent.com/45556539/108380388-d8ad9180-7217-11eb-920b-bd8c2ad8db94.png)
Other types will be added in the separate tasks.

AndreiMazol:
Reopened. The structure should be saved as Mol if more than 1 arrow has been added.

PR #347, opened 2021-02-25
#145 Add multiple arrow support for the same canvas

softcat23:


packages/ketcher-react/src/script/chem/struct/rxnArrow.ts
```diff
@@ -0,0 +1,37 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ ***************************************************************************/
+import Vec2 from '../../util/vec2'
+
+class RxnArrow {
+  pp: Vec2
```
ensemenova, line 19: Is it possible to name pp as something more meaningfull?

softcat23: To avoid merge conflicts with @AndreiMazol changes, let's rename this prop later.
I agree that this prop name is not meaningful enough.

packages/ketcher-react/src/script/render/restruct/rerxnarrow.js
```diff
@@ -50,12 +51,17 @@ ReRxnArrow.prototype.makeSelectionPlate = function (restruct, paper, styles) {
...
+  var endPoint = new Vec2(centre.x + options.scale, centre.y)
+  var path
+  switch (this.item.mode) {
+    case RxnArrowMode.simple:
+      path = draw.arrow(render.paper, startPoint, endPoint, options)
+      break
+    case RxnArrowMode.equilibrium:
+      path = draw.equilibriumArrow(render.paper, startPoint, endPoint, options)
+      break
+  }
```
AndreiMazol, line 64: What about default statement?

AndreiMazol (approved):

packages/ketcher-react/src/script/render/restruct/rerxnarrow.js
softcat23: We define the default value in the RxnArrow constructor: https://github.com/epam/ketcher/blob/4cbc1c26a23cb2a73408332f871d6e7caed8b3db/packages/ketcher-react/src/script/chem/struct/rxnArrow.ts#L22
Thus, all arrows will have the mode.

Issue #338, opened 2021-02-19
Templates in template library are shifted down when Ketcher have less than 855px width


KseniiaSemenishcheva:
**Affected Version:** _Remote, Standalone_
**Steps to Reproduce:**
1. Launch Ketcher.
2. Open Templates.
3. Resize window to 854px width and lower.

**Expected Result:**
User can see templates with different size of Ketcher.

**Actual Result:**
Templates are shifted down under list of templates groups.
![templates](https://user-images.githubusercontent.com/75974465/108495382-c7b55c80-72b9-11eb-8092-d6403ed27cde.jpg)

AndreiMazol:
Reopened. Two scrolls are rendered:
![image](https://user-images.githubusercontent.com/72735611/110636045-dc03bf80-81bc-11eb-93d9-4679b7ced2fc.png)

Svetlana-Asinovskaia:
Issue is closed as fixed

PR #349, opened 2021-02-26
#338 Fix template table layout for small screens

irinatelnova:


ensemenova (approved):

packages/ketcher-react/src/style/dialogs.less
```diff
@@ -123,29 +123,40 @@ form[role='dialog'] {
       margin-left: 4px;
     }
   }
+  .table-group-wrap {
+    display: flex;
+    @media (max-width: @common-screen-tiny) {
+      flex-wrap: wrap;
+    }
```
TomasLight, line 130: please fix according our discussion

packages/ketcher-react/src/style/dialogs.less
```diff
@@ -123,29 +123,40 @@ form[role='dialog'] {
...
     width: 550px;
     height: 27em;
-    overflow: auto;
+    margin-top: 1em;
+    padding: 0;
+    overflow: hidden;
+    @media (max-width: @common-screen-tiny) {
+      width: auto;
+    }
+  }
```
TomasLight, line 150: same

TomasLight (changes requested):

packages/ketcher-react/src/style/dialogs.less
irinatelnova: @TomasLight fixed

packages/ketcher-react/src/style/dialogs.less
irinatelnova: @TomasLight fixed

TomasLight (approved):

AndreiMazol (changes requested):
delete master features please

Issue #315, opened 2021-02-11
Simple Objects selection and resizing mechanism 


KseniiaSemenishcheva:
**Affected Version:** _Remote_
**Steps:**
1. Launch Ketcher.
2. Draw an Ellipse using 'Shape Ellipse' tool.
3. Click on Ellipse line.
4. Drag and drop diagonal dot to change the Ellipse size.

**Actual Result:**
Diagonal dots is on Ellipse sides. User can resize ellipse without selection, when User hovers mouse over ellipse - dots appears.
![2021-02-11_14h17_18](https://user-images.githubusercontent.com/75974465/107630062-16d50f00-6c74-11eb-9a0b-6354a83cf6fd.gif)

**Expected Result:**

1. When User clicks ob Simple Object then dots appears and User can resize object by drag and drop option.
2. If User didn't click on Object he can only move it at another place on canvas and then selection with dots appears.
3. It should work for all Simple Object.
4. When User select ellipse or circle - dots appears as a rectangle in which the figure is enclosed.

KseniiaSemenishcheva:
Issue is related to https://github.com/epam/ketcher/issues/85
Need discussion with @ensemenova

PR #356, opened 2021-03-02
#315 fix reselect

ensemenova:


ensemenova:
Was opened with incorrect based fix branch

PR #366, opened 2021-03-03
#145 Add multiple arrow support for the same canvas

softcat23:
Fix saving issue

### DIFF
--- a/packages/ketcher-core/src/chemistry/entities/Struct.ts
+++ b/packages/ketcher-core/src/chemistry/entities/Struct.ts
@@ -75,7 +75,15 @@ export class Struct {
   }
 
   hasRxnArrow(): boolean {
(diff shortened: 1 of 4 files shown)


packages/ketcher-core/src/chemistry/entities/Struct.ts
```diff
@@ -75,7 +75,15 @@ export class Struct {
   }
 
   hasRxnArrow(): boolean {
-    return this.rxnArrows.size > 0
+    return this.rxnArrows.size === 1
+  }
+
+  hasRxnPlus(): boolean {
+    return this.rxnPluses.size === 1
```
AndreiMazol, line 82: I guess the condition  should be transformed to `rxnPluses.size > 0`

AndreiMazol (approved):

Issue #367, opened 2021-03-03
Allow select simple object by several points in selection area


ensemenova:
**Steps to reproduce:**
1. Create different simple object
2. Try to select them with selection tool

**Current behaviour:**
They will be selected only when p0 point  will turn into selection
**Expected Behavior:**
When at least any 3 points turns into selection from rectangle or ellipse  it should be selected. For line any of it's point.

Svetlana-Asinovskaia:
Issue is verified and closed.

PR #381, opened 2021-03-10
#367 change selection of simple objects

ensemenova:


packages/ketcher-react/src/script/render/restruct/ReObject.ts
```diff
@@ -0,0 +1,76 @@
+/****************************************************************************
+ * Copyright 2020 EPAM Systems
```
AndreiMazol, line 2: please update copyright (2021)

AndreiMazol (approved):

Issue #368, opened 2021-03-04
'Any Atom' tool incorrect behavior when selected with hotkey or typed in from keyboard


AleksandraTitova:
**Affected Versions**: Standalone, Remote
**Steps to Reproduce**:

1. Create a structure on the canvas
2. Press A hotkey
3. Click one of the structure's atoms --> Atom replaced with Any Atom
4. With Selection tool double-click the Any Atom

**Alternative flow**:

1. Create a structure on the canvas
2. Select an atom
3. Type an asterisk --> Modal for typing in Atom Label opens with asterisk typed in the field
4. Click Ok --> Atom replaced with Any Atom
5. With Selection tool double-click the Any Atom

**Expected result**: Any Atom displays as _A_; on double-click Periodic Table opens with Extended tab opened and Any Atom highlighted

**Actual result**: Any Atom displays as A (non-italic); on double-click Periodic Table opens with Table tab opened; nothing is highlighted on Extended tab.
![image](https://user-images.githubusercontent.com/79926367/109954519-7c0ba580-7cf2-11eb-8fca-b16f906244e8.png)

*NB*:
- Issue not reproducible when selecting Any Atom tool with mouse through Periodic Table tool;
- Issue does not reproduce if you add single atom on canvas;
- If you Erase the incorrect atom and then select Undo, it turns to look and work as expected

Svetlana-Asinovskaia:
Issue is verified and closed.

PR #390, opened 2021-03-15
#368 Fix any atom tool incorrect behavior

irinatelnova:


packages/ketcher-core/src/chemistry/entities/Atom.ts
```diff
@@ -200,6 +196,13 @@ export class Atom {
     this.atomList = params.atomList ? new AtomList(params.atomList) : null
     this.neighbors = [] // set of half-bonds having this atom as their origin
     this.badConn = false
+
+    Object.defineProperty(this, 'pseudo', {
```
AndreiMazol, line 200: Please use standard TS syntax to  define properties

Issue #60, opened 2020-10-12
SETTINGS for the STEREO FLAGS and STEREO MARKS


Vesdet:
As a Ketcher collaborator I'd like to implement "showStereoLabels" option in Settings menu.
It should work like "showChiralFlag" option, but for all fragments.

Part 3 in Epic #56

Svetlana-Asinovskaia:
Issue is verified and closed.

PR #410, opened 2021-03-19
#60 show/hide stereo flags base on option value + refactor to ts

ensemenova:


packages/ketcher-react/src/script/render/restruct/ReEnhancedFlag.ts
```diff
@@ -14,34 +14,41 @@
...
+  constructor(flag: StereoFlag | null, pos: Vec2 | null) {
     super('enhancedFlag')
     this.flag = flag
     this.pp = pos
   }
   static isSelectable() {
     return true
   }
-  highlightPath(render) {
+  highlightPath(render: any): any {
```
TomasLight, line 33: please, use `Render` class for type `render` argument

packages/ketcher-core/src/chemistry/entities/Fragment.ts
```diff
@@ -13,30 +13,30 @@
...
-function calcStereoFlag(struct: any, stereoAids: Array<number>) {
+export enum StereoFlag {
+  Mixed = 'Mixed',
+  abs = 'ABS (Chiral)',
+  and = 'AND Enantiomer',
+  or = 'OR Enantiomer'
+  // todo: custom in the future
+}
+
+function calcStereoFlag(struct: any, stereoAids: Array<number>): string | null {
```
TomasLight, line 24: and `Struct` (from 'ketcher-core') for `struct` argument

TomasLight (changes requested):

packages/ketcher-core/src/chemistry/entities/Fragment.ts
ensemenova: fixed

packages/ketcher-react/src/script/render/restruct/ReEnhancedFlag.ts
ensemenova: fixed

packages/ketcher-core/src/chemistry/entities/Fragment.ts
```diff
@@ -13,30 +13,35 @@
...
-    and: 'AND Enantiomer',
-    or: 'OR Enantiomer',
-    null: null
-    // todo: custom in the future
-  }
 
+export class Fragment {
   stereoAtoms: Array<number>
-  enhancedStereoFlag?: any
+  enhancedStereoFlag?: boolean | string | undefined
```
AndreiMazol, line 42: enhancedStereoFlag is already marked as undefined ('?' mark) so no needs to use union with 'undefined' explicitly

packages/ketcher-react/src/script/render/restruct/ReObject.ts
```diff
@@ -28,15 +30,15 @@ class ReObject {
     this.visel = new Visel(viselType)
   }
 
-  getVBoxObj(render: any): any {
+  getVBoxObj(render: Render): any {
```
AndreiMazol, line 33: You can return 'Box2Abs' instead of 'any'

AndreiMazol (approved):

TomasLight (approved):

PR #441, opened 2021-04-01
pull

ForBurning:
pull


example/.env
```diff
@@ -1 +1 @@
-REACT_APP_API_PATH=
\ No newline at end of file
+REACT_APP_API_PATH=http://drawengine.ilabpower.develop/api
```
AndreiMazol, line 1: Please use .env.local for defining API PATH.

packages/ketcher-react/src/infrastructure/services/struct/remoteStructService.ts
```diff
@@ -87,22 +87,22 @@ class IndigoService implements StructService {
...
-      isAvailable: boolean = false
-    try {
-      const response = await request('GET', this.baseUrl + 'info')
-      indigoVersion = response['indigo_version']
-      imagoVersions = response['imago_versions']
-      isAvailable = true
-    } catch (e) {
-      isAvailable = false
-    }
+    // let indigoVersion: any,
```
AndreiMazol, line 90: if you  want to  mock structService you need to implement StructService interface, cretae your own StructServiceProvider and update your local version of App.tsx. please look at StandaloneStructService for example: 
```js
import { RemoteStructServiceProvider } from 'ketcher-core'
import 'ketcher-react/dist/index.css'
;(global as any).Miew = Miew

let structServiceProvider: any = new RemoteStructServiceProvider(
  process.env.REACT_APP_API_PATH!
)
if (process.env.MODE === 'standalone') {
  const { StandaloneStructServiceProvider } = require('ketcher-standalone')
  structServiceProvider = new StandaloneStructServiceProvider()
}

const App = () => {
  return (
    <Editor
      staticResourcesUrl={process.env.PUBLIC_URL}
    />
  )
}
```

AndreiMazol:
This PR should not be merged into epam/master branch.
cc: @ForBurning

Issue #107, opened 2020-11-17
Rendering only feature


hadim:
I am wondering whether it's possible to render a molecule given a SMILES without using the editor. In short, I'll need a simple function I can call on the frontend side that input a smiles and output whether an SVG or a PNG. 

Thank you.

AndreiMazol:
Dear @hadim ,

Please double check me, am I right that you would like to have a convertor from SMILES to SVG or PNG files?

hadim:
Yes exactly.

AndreiMazol:
Unfortunately this feature does not supported by Ketcher right now. We will discuss this functionality as a potential enhancement for Ketcher in the future releases.

AndreiMazol:
@StartsevDS  'Save as image" button should be disabled is server is not available

StartsevDS:
To convert SMILES to SVG or PNG, invoke generateImageAsync(smiles, option) method defined on editor instance, where "smiles" - string representation of your SMILES structure, "option" - object with "outputFormat" property ("svg" or "png"). For example:

window.ketcher.generateImageAsync('C1=CC=CC=C1', { outputFormat: 'svg' })

Important: please note that currently the feature is available in "standalone" mode only, in "remote" mode we're receiving 500 Internal Server Error (reported in issue #371)

Svetlana-Asinovskaia:
Retesting will be possible after fixes from the backend side.

Svetlana-Asinovskaia:
Release/2.3 The issue is closed as partly fixed.
It's possible to save the structure as image in Standalone mode (as svg and png), in Remote mode (as svg). It will be possible to save the structure as png in Remote mode after the fix of the issue #371 will be done. Besides the sizes of the view boxes are different in Remote and Standalone modes now. It has been decided that the problems are not blocking, since the main functionality works correctly. Behavior will be rechecked after the fixes will be done.

PR #444, opened 2021-04-02
#107 disable save button if server is down

StartsevDS:


AndreiMazol (changes requested):
Cannot save mol v2000 file

packages/ketcher-react/src/script/ui/component/view/savebutton.jsx
```diff
@@ -63,19 +63,19 @@ const SaveButton = props => {
...
-    onSave()
   }
 
   return (
     <button
       onClick={event => {
         save(event)
       }}
-      className={!data ? `disabled ${className}` : className}
+      className={!server || !data ? `disabled ${className}` : className}
```
AndreiMazol, line 78: Please use 'clsx' to add conditional classes

AndreiMazol (approved):

Issue #443, opened 2021-04-02
'Delete' option works incorrectly 


Svetlana-Asinovskaia:
Affected versions: remote, standalone

Steps to reproduce:

Scenario 1.
1. Launch Ketcher.
2. Add an element, e.g. carbon.
3. Delete it with 'Erase' tool.

Expected result: The element is deleted.
Actual result: The element is present on the canvas.
![CH4](https://user-images.githubusercontent.com/80695128/113422650-bfcafb00-93d5-11eb-8b39-4217f9be4ab3.gif)

Scenario 2.
1. Launch Ketcher.
2. Draw a cycle, e.g. cyclohexane.
3. Delete a part of the structure with 'Erase' tool.
4. Click 'Undo' button.

Expected result: 
3. A part of the structure is deleted.
4. The whole structure is present on the canvas.

Actual result:
3. The whole structure is deleted.
4. CH4 is present on the canvas.
![cyclohexane](https://user-images.githubusercontent.com/80695128/113422787-f56fe400-93d5-11eb-98eb-0d2df5ea97e0.gif)
![image](https://user-images.githubusercontent.com/80695128/113422878-1f290b00-93d6-11eb-9fe1-fb50fe11d275.png)

Svetlana-Asinovskaia:
Issue is closed as fixed.

PR #450, opened 2021-04-07
#443 Fix delete option works incorrectly

irinatelnova:


AndreiMazol (changes requested):
It should be discussed

AndreiMazol:
Fix has been merged into taster in scope of another task. PR is closed

Issue #445, opened 2021-04-05
'Calculated values' - The data in 'Elemental analysis' field is displayed incorrectly 


Svetlana-Asinovskaia:
Affected version: standalone, remote

Steps to reproduce:

1. Launch Ketcher.
2. Draw a structure on the canvas, e.g. chain.
3. Click on the 'Calculated values' button.

Expected result: Numbers in 'Elemental analysis' field are displayed with one decimal place (e.g. C 84.2 H 15.8)
Actual result: Numbers in 'Elemental analysis' field are displayed as integer (e.g. C 84 H 16)
![image](https://user-images.githubusercontent.com/80695128/113579574-2d249900-962d-11eb-8bc8-5747033414c6.png)

The issue was discussed with Alexander Savelyev.

Svetlana-Asinovskaia:
Issue is closed as fixed.

PR #453, opened 2021-04-07
#445 Display Elemental Analysis value with one decimal place

StartsevDS:
OneDecPoint


AndreiMazol (changes requested):
it should be discussed

AndreiMazol (approved):

Issue #456, opened 2021-04-08
It is possible to save a structure in incorrect format when network connection is slow


Svetlana-Asinovskaia:
_Affected versions_: Remote, Standalone

_Steps to reproduce:_
1. Launch Ketcher in Chrome.
2. Open  
[Discodermolide.zip](https://github.com/epam/ketcher/files/6279176/Discodermolide.zip)
3. Choose ‘Slow 3G’ mode in Chrome DevTools.
4. Click on the 'Save structure' button.
5. Choose 'Daylight SMARTS' in Format window.
6. Click on the 'Save To File' button.

_Actual result:_ It's possible to choose a wrong format for saving a structure. It is shown that the molecule is saved in .smarts . If to open the saved file in Ketcher the structure in Mol will be shown. A warning appears after the file has been already saved.

![smarts](https://user-images.githubusercontent.com/80695128/114048668-74ac5e80-9893-11eb-8da1-105a2d510005.gif)

 _Expected result:_ It's not possible to save a structure in a wrong format.

It could be better to disable the 'Save To File' button till the moment when the possibility of saving the certain structure in a certain format will be checked. It is also necessary to show a progress indicator to inform the users about the execution of this operation.

Svetlana-Asinovskaia:
The issue is closed as fixed

PR #459, opened 2021-04-12
 #456 fix saving a structure in incorrect format

MagomedGasanov:


AndreiMazol (changes requested):
You shouldn't update valid flag.

MagomedGasanov:
> You shouldn't update valid flag.

disableContoles flag was added to the state and it is used to control all buttons disabling now

AndreiMazol (approved):

Issue #195, opened 2020-12-21
As soon as user changes the structure the stereoconfiguration should be re-calculated


KseniiaSemenishcheva:
**Steps to reproduce**
### Scenario I
1. Open the attached file stereocenter.mol.
[stereocenter.zip](https://github.com/epam/ketcher/files/5723073/stereocenter.zip)
2. Click the 'Calculate CIP' button. - The R stereo label appears.
3. Change the Single Up Bond with the Single Down Bond.

**Actual result**
- The R stereo label is not changed to S label.
- User should click the 'Calculate CIP' button once again to re-calculate stereoconfiguration.
- 
### Scenario II
4. Delete the stereobond.

**Actual result**
The stereobond is removed. The stereolabel is present still.

**Expected result**
As soon as user changes the structure the stereoconfiguration should be re-calculated. It's a user-friendly behavior.

Svetlana-Asinovskaia:
1) The attached file stereocenter.mol. is opened with wrong symbols.
![image](https://user-images.githubusercontent.com/80695128/113672657-b5a04980-96c0-11eb-96b3-7c456e33d58c.png)
It's also possible to reproduce this issue with the following steps:
1. Draw a structure, e.g. cyclohexane.
2. Add Single Up (Single Down, Single Up/Down) bond.
Or:
1. Choose a Custom template with stereobonds (e.g. one of α-D-Sugars).
2. Change a Single Down bond with a Single Up bond or vice versa. 

2)'Undo' (and 'Redo') option with deleted stereobond works in two steps. 1st click - Stereolabel is returned to the canvas. 2nd click - stereobond is returned to the canvas.

irinatelnova:
When user changes the structure the stereoconfiguration will not be re-calculated automatically in order to avoid making server requests each time on structure changing.
Now **Expected Result**:
As soon as user changes the structure stereoconfiguration marks should be removed.
User should click the 'Calculate CIP' button once again to calculate the stereo configuration.

AndreiMazol:
@Svetlana-Asinovskaia  please transform the following requirement to a separate enhancement:
"As soon as user changes the structure stereoconfiguration marks should be removed"

Svetlana-Asinovskaia:
The enhancement
"As soon as user changes the structure stereoconfiguration marks should be removed" will be done within the task https://github.com/epam/ketcher/issues/539

PR #482, opened 2021-04-15
#195 Remove cip marks after structure changing + revert autorecalc cip

irinatelnova:


AndreiMazol (approved):

packages/ketcher-react/src/script/editor/operations/sgroup/index.ts
```diff
@@ -59,10 +59,12 @@ class SGroupCreate extends BaseOperation {
 
 class SGroupDelete extends BaseOperation {
   data: Data
+  checkEmptiness: boolean
 
-  constructor(sgroupId?: any) {
+  constructor(sgroupId?: any, checkEmptiness = true) {
```
AndreiMazol, line 64: please rename to 'validate' instead

AndreiMazol (changes requested):

packages/ketcher-react/src/script/ui/state/server/index.js
```diff
@@ -147,10 +147,14 @@ export function serverTransform(method, data, struct) {
...
       .then(res =>
         dispatch(
-          load(res.struct, {
-            rescale: method === 'layout',
-            reactionRelayout: method === 'clean'
-          })
+          load(
+            res.struct,
+            {
+              rescale: method === 'layout',
```
AndreiMazol, line 153: look strange. the solution  should be discussed

packages/ketcher-react/src/script/editor/operations/sgroup/index.ts
StartsevDS: Naming booleans starting with "is" considered a good practice. In this case "isValid" or "isEmpty" would be self-descriptive.

AndreiMazol (commented):
please fix comments

MagomedGasanov (approved):

PR #528, opened 2021-05-21
add `DATIVE` to bondTypeMap

hayabusa2009:
when open a molfile which contain a dative bond, ketcher will be crash


MagomedGasanov (approved):

MagomedGasanov:
Thank you for your corrections!

PR #591, opened 2021-06-09
Export ketcher and buildKetcherAsync

waltonseymour:
Exports `Ketcher` and `buildKetcherAsync` to allow for subscribing to editor changes.

Example

```tsx
const ref = useRef(null);
const [smiles, setSmiles] = useState("");

useEffect(() => {
  if (ref.current) {
    buildKetcherAsync({
      element: ref.current,
      staticResourcesUrl: process.env.PUBLIC_URL,
      structServiceProvider: structServiceProvider,
    }).then((k) => {
      k.editor.subscribe("change", () => {
        k.getSmilesAsync(false).then(setSmiles);
      });
    });
  }
}, []);
```

### DIFF
--- a/packages/ketcher-react/package.json
+++ b/packages/ketcher-react/package.json
@@ -31,6 +31,7 @@
     "build": "cross-env NODE_ENV=production rollup -c -m true --silent",
     "start": "cross-env NODE_ENV=development rollup -c -m true -w",
     "test": "run-s test:unit test:lint prettier stylelint test:build",
+    "prepare": "npm run build",
     "test:build": "run-s build",
     "test:lint": "eslint src --ext .ts,.tsx,.js,.jsx",
     "test:unit": "cross-env CI=1 react-scripts test --env=jsdom",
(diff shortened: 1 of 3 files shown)


AndreiMazol:
Hi @waltonseymour ,

could you  explain the scenario when ketcher should be rebuilded by your opinion. According to our design only single instance of ketcher should be connected to <Editor /> react component (1-1 link).

waltonseymour:
Using `buildKetcherAsync` would replace the mounting of the react component. The other alternative is to provide a forwarded ref to the ketcher instance or expose lifecycle methods on the `Editor` component.

waltonseymour:
The idea is for the Editor component to be controlled in terms of the structure/smiles. So it would both have the use case of programmatically setting the smiles and listing to updates in the editor. The easiest way to accomplish this now is by interacting with the underlying ketcher class, which this change facilitates, but ideally that could be exposed through react props (`molecule` `onMoleculeChange`).

AndreiMazol (approved):

Issue #418, opened 2021-03-22
Stereo Flags shouldn't appear for structures without correct stereocenters


ensemenova:
Stereo Flags should't appear for structures without correct stereocenters, for example:
![95749451-4531e600-0ca4-11eb-928a-98b71e16390d](https://user-images.githubusercontent.com/77324533/112000056-a8b32000-8b2e-11eb-9108-939b2c8d4969.jpg)

ensemenova:
@AlexanderSavelyev Please fullfill the details

AlexanderSavelyev:
The following rules should be considered for incorrect stereocenters
1, Stereo bond attached to a "hanging" atom. Pseudo code:
```
   bond.hasStereo & 
   bond.atom1.numNeighbors == 1 & 
   bond.atom2.numNeighbors == 2 &
   bond.atom2.numHydrogens % 2 == 0
```
![image](https://user-images.githubusercontent.com/7427808/122046533-b3a4e600-cde7-11eb-87ab-4417693e072e.png)

2. Stereo bond attached to an atom neighbor to "hanging" atom. Pseudo code:
```
   bond.hasStereo & 
   bond.atom1.numNeighbors == 2 & 
   bond.atom2.numNeighbors == 2 &
   bond.atom2.numHydrogens % 2 == 0
   bond.atom1.neiAtom1.numNeighbors==1
```
![image](https://user-images.githubusercontent.com/7427808/122046490-a7208d80-cde7-11eb-8cbe-a3a1f0b0747c.png)

As user I don't want to see stereo flags for incorrect stereo centers

AleksandrPokrovskii:
**Steps to Reproduce**

1. Launch Ketcher
2. Add two Single bond with common atom
3. Add Single Up bond to common atom
4. Add Single bond to common atom
5. Click "Undo"
6. Click "Redo"


_Expected behavior_
Stereo flag is displayed
![image](https://user-images.githubusercontent.com/85170103/124951214-cf349280-e01b-11eb-8954-dfd5a43ea43e.png)

 
_Actual behavior_
Stereo flag is not displayed
![image](https://user-images.githubusercontent.com/85170103/124951299-e2dff900-e01b-11eb-85a1-bd8922b42ac2.png)

MagomedGasanov:
After copy/paste of structure the implicitH is lost

AleksandrPokrovskii:
**Steps to Reproduce**

1. Launch Ketcher
[...]

PR #597, opened 2021-06-17
#418 stereo flags shouldnt appear for structures without correct stereocenters

MagomedGasanov:


packages/ketcher-react/src/script/editor/actions/bond.ts
```diff
@@ -159,48 +173,173 @@ function fromBondFlipping(restruct: ReStruct, id: number): Action {
...
+      return false
+    }
+
+    return true
+  } else {
+    return false
+  }
+}
+
+function getStereoParity(stereo) {
```
AleksandrPokrovskii, line 340: add types for function

AleksandrPokrovskii (approved):

packages/ketcher-react/src/script/editor/actions/bond.ts
MagomedGasanov: Done

AndreiMazol (changes requested):
ImplicitH should be calculated by Struct object.  Please add an additional operation to trigger recalculation instead of setting implicitH directly.

AndreiMazol (approved):

Issue #231, opened 2020-12-30
Bonds can be overlapped 


ElenaKlepova:
**Affected version:**Release/2.1
**Steps to reproduce:**

1. With any bond (for example Single Bond Tool) click the canvas.
2. With the same bond tool (Single Bond Tool) press any atom of the previously created bond and drag to the second atom of this bond several times.

**Actual result:**
Bonds can be overlapped unlimited number of times.
![2020-12-30_10h44_17](https://user-images.githubusercontent.com/75974843/103337352-1994ef80-4a8c-11eb-81d4-0135cbf067e8.gif)

**Expected result:**
Bonds can not be overlapped.

AleksandrPokrovskii:
In the course of the work, there were problems associated with the architecture. The task is suspended until the architecture is updated

d-mrzv:
Issue is no longer reproducible
Verified on:
Ketcher Version 3.18.0-rc.1 Build at 2026-07-02; 16:09:03
Indigo Toolkit Version 1.46.0-rc.1.0-gf3408986b-x86_64-linux-gnu-11.2.1
Chrome Version 150.0.7871.125 (Official Build) (arm64)
Macos 26.5.2 (25F84)

https://github.com/user-attachments/assets/c491d941-91d8-41aa-b403-895d1d26d435

PR #598, opened 2021-06-17
#231 Bonds can be overlapped

AleksandrPokrovskii:


AndreiMazol (changes requested):
The provided solution is different to  our discussed approach. It is necessary  to  discuss it again.

Issue #543, opened 2021-05-26
Add a warning when saving a reaction equilibrium arrow in a format that does not support it


Svetlana-Asinovskaia:
The only format that supports Reaction equilibrium arrow on save/load is the "ket" format. If user selects another format, then the warning message "The ${Selected_format} does not support drawn elements: the reaction equilibrium arrow will be replaced with the reaction arrow" appears.

AleksandrPokrovskii:
The application crashes when the structure is saved

_Steps to reproduce:_

1. Launch Ketcher.
2. Add Reaction equilibrium arrow
3. Delete Reaction equilibrium arrow
4. Add Reaction equilibrium arrow
5. Click Save button

_Expected result_: It's possible to save with warning

_Actual result_: TypeError: Cannot read property 'mode' of undefined

PR #599, opened 2021-06-18
#543 Add a warning when saving a reaction arrow 

MagomedGasanov:


AndreiMazol (changes requested):
Please use struct.hasRxnArrow() instead of duplicating this logic here again

AndreiMazol (approved):

packages/ketcher-react/src/script/ui/data/convert/structConverter.ts
```diff
@@ -26,6 +26,32 @@ export function couldBeSaved(
 ): string | null {
   let warnings: Array<string> = []
   const formatName: string = getPropertiesByFormat(format).name
+
+  const rxnArrowsSize = struct.rxnArrows.size
+  const hasRxnArrow = struct.hasRxnArrow()
+
+  if (format !== 'graph') {
+    if (hasRxnArrow) {
+      const rxnArrowMode = struct.rxnArrows.values().next().value.mode
```
AndreiMazol, line 35: What do you need it for?

AndreiMazol (changes requested):

packages/ketcher-react/src/script/ui/data/convert/structConverter.ts
MagomedGasanov: Fixed

packages/ketcher-react/src/script/ui/data/convert/structConverter.ts
```diff
@@ -26,6 +27,28 @@ export function couldBeSaved(
 ): string | null {
   let warnings: Array<string> = []
   const formatName: string = getPropertiesByFormat(format).name
+
+  const rxnArrowsSize = struct.rxnArrows.size
+  const hasRxnArrow = struct.hasRxnArrow()
+
+  if (format !== 'graph') {
+    if (hasRxnArrow) {
+      const rxnArrowMode: RxnArrowMode | undefined = struct.rxnArrows.get(0)?.mode
```
AndreiMazol, line 36: You can use 'struct.rxnArrows.get(0)!' instead (you have already checked that  it exists before - struct.hasRxnArrow() function)

AndreiMazol (changes requested):

packages/ketcher-react/src/script/ui/data/convert/structConverter.ts
MagomedGasanov: Fixed

AndreiMazol (approved):

Issue #524, opened 2021-05-19
Vertical/Horizontal flip hot keys don't work correctly


Svetlana-Asinovskaia:
Tested on Ketcher v2.2.1-104-gc7896113

_Steps to reproduce:_

1. Launch Ketcher.
2. Open the structure from file 
[struct.zip](https://github.com/epam/ketcher/files/6509959/struct.zip)
4. Click the 'Horizontal Flip' button.
5. Click 'Alt+H' hot key.

_Expected result:_ It's possible to make the horizontal flip using the 'Alt+H' hot key.

_Actual result:_ The structure is not flipped with the 'Alt+H' hot key.
![2021-05-19_18h32_05](https://user-images.githubusercontent.com/80695128/118841210-c8698900-b8d0-11eb-8ece-d7afea80d780.gif)

The issue is reproduced with 'Vertical Flip' tool.

AleksandrPokrovskii:
In the course of the work, there were problems associated with the architecture. The task is suspended until the architecture is updated

PR #605, opened 2021-06-28
#524 Vertical/Horizontal flip hot keys don't work correctly

AleksandrPokrovskii:


packages/ketcher-react/src/script/ui/views/components/Dialog/Dialog.tsx
```diff
@@ -57,6 +57,9 @@ const Dialog: FC<Props> = props => {
   const dialogRef = useRef<HTMLDivElement>(null)
 
   const exit = mode => {
+    ;(dialogRef.current
+      ?.closest('.smallEditor')
```
AndreiMazol, line 61: It won't work

AndreiMazol (changes requested):

AndreiMazol (approved):

Issue #641, opened 2021-07-15
Structure with NOT[] atom does not open


nseufulina:
**Steps to Reproduce**
1. Launch Ketcher
2. Open 
[29DHIS_prohibitatoms.zip](https://github.com/epam/ketcher/files/6823622/29DHIS_prohibitatoms.zip)

[13-AQuery-NotList.zip](https://github.com/epam/ketcher/files/6830062/13-AQuery-NotList.zip)

**Expected behavior**
Structure opens

**Actual behavior**
Modal window with error appears.

**Screenshots**
![image](https://user-images.githubusercontent.com/87320094/125800898-98d5dbc2-5577-42bb-9411-265ed0dcc562.png)

**Desktop:**
 - OS: Windows 10

vitalgerko:
Release/2.4 Closed as fixed.

PR #686, opened 2021-07-22
#641 Structure with NOT[] atom does not open

AleksandrPokrovskii:


AndreiMazol (changes requested):
I guess Regex would be better instead of replacing strings

Issue #697, opened 2021-07-27
Add additional arrow types


AndreiMazol:
The following arrows should be added in scope of this task: 

- Arrow Open Angle Tool
- Arrow Filled Triangle
- Arrow Filled Bow Tool
- Arrow Dashed Open Angle Tool
- Failed Arrow Tool
- Arrow Both Ends Filled Triangle Tool
- Arrow Equilibrium Filled Triangle Tool
- Arrow Equilibrium Filled Half Bow
- Arrow Equilibrium Open Angle Tool
- Arrow Unbalanced Equilibrium Filled Half Bow Tool
- Arrow Unbalanced Equilibrium Open Half Angle Tool
- Arrow Unbalanced Equilibrium Large Filled Half Bow Tool
- Arrow Unbalanced Equilibrium Filled Half Triangle Tool

For more details please look at Point 2 in #695 .

PR #701, opened 2021-07-30
#697 add additional arrow types

MagomedGasanov:


packages/ketcher-react/src/script/editor/operations/rxn/index.ts
```diff
@@ -1,3 +1,4 @@
+import Restruct, { ReRxnArrow } from '../../../render/restruct'
```
AndreiMazol, line 1: Please revert i: imports should follow after 'copyright' text

MagomedGasanov: Done

AndreiMazol (approved):

Issue #707, opened 2021-08-06
Decomposition of Copy button


aleksandrgordeev1:
As some browsers can only work with 'text' in terms of adding data to clipboard, Copy button should be decomposed into 3:
 - copy as Mol
 - copy as Ket
 - copy as image (already being implemented in the scope of https://github.com/epam/ketcher/issues/691)
in order to insist on choosing one type.

Copy as Mol should be main button. List of options should appear after clicking in the bottom right corner (example: Selection list of options)

This will also help not to waste resources on converting data into 3 types every time Copy button is used.

nseufulina:
No prompt for user in IE when saving Simple Object with Copy (Ctrl+C) that coping does not work.

Steps:
Create Simple Object
Select created object
Click Copy button (Ctrl+C)

Actual result:
No prompt for user in IE when saving Simple Object with Copy (Ctrl+C) that coping does not work.

Expected result:
Some prompt - in Help module or alert window.

PR #714, opened 2021-08-12
#707 Decomposition of Copy button

aleksandrgordeev1:


AndreiMazol (approved):
It won't work in IE11

Issue #691, opened 2021-07-23
Copy drawing to the clipboard as an image


AndreiMazol:
It is necessary to copy structure as png to  clipboard in order to provide possibility to  insert it to Word/PowerPoint. This functionality should be available by pressing a new hotkey (please discuss it with @nseufulina  and @AlexanderSavelyev )

nseufulina:
Cntr+B hotkey is available for this shortcut

nseufulina:
Need to verify if those are bugs to fix:

**1)No alert in Opera**
Opera 77.0.4054.277 Standalone mode

Steps:
1. Draw an object on the canvas.
2. Select object.
3. Click "Copy Image" or click "Ctrl+Shift+F"

Expected result:
Modal window with alert "This feature is not available in your browser" appears.

Actual result:
Modal window does not appear.


**2)No alert in the remote mode**
Opera 77.0.4054.277 Remote mode
FireFox 90.0 Remote mode

Steps:
1. Draw an object on the canvas.
2. Select object.
3. Click "Copy Image" or click "Ctrl+Shift+F"

Expected result:
Modal window with alert "This feature is not available in your browser" appears.

Actual result:
Modal window does not appear.

**3)Copied objects are displayed as black squares in the Clipboard.** 
Steps:
1. Draw objects on the canvas
2. Select object
3. Click "Copy Image" or click "Ctrl+Shift+F"
4. Open Word (or any pplication to paste)
5. Open Clipboard (Win+V)

Expected result:
Objects are displayed as copied structures.

Actual result:
Objects are displayed as black squares.
![image](https://user-images.githubusercontent.com/87320094/128374385-55144b3c-1126-4970-9ec2-c49815ea13bb.png)


**4)Objects from Clipboard are pasted incorrectly from the Clipboard.**
_Scenario1:_
Steps:

1. Create two structures on the canvas.
2. Select first structure and click Copy Image button.
3. Select second structure and click Copy Image button.
4. Open Word/PowerPoint.
5. Open Clipboard (Win+V).
6. Click on the object which is second from the top of the list.

Expected result:
Object is pasted.

Actual result:
Black square is pasted.

_Scenario2:_
Steps:

1. Create two structures on the canvas.
2. Select first structure and click Copy Image button.
3. Select second structure and click Copy Image button.
4. Open Word/PowerPoint.
5. Open Clipboard (Win+V).
6. Click on the object which is second from the top of the list.
7. Click on the object which is first from the top of the list.

[...]

PR #716, opened 2021-08-16
 #691 Can not save structure to clipboard with atom generics

AleksandrRyakhov:
Add alert when user couldn't save the structure


AndreiMazol (approved):

AndreiMazol (changes requested):
Core object  should not depend on browser functions

AndreiMazol (approved):

Issue #723, opened 2021-08-20
Structure with some S-groups cannot be opened


troits:
**Steps to Reproduce**
1. Go to 'https://lifescience.opensource.epam.com/ketcher/demo/ketcher_2_3_remote.html'
2. Click on 'Open...', 'Open From File'
3. Select attached
[UnsupportedSGroup.mol.txt](https://github.com/epam/ketcher/files/7023172/UnsupportedSGroup.mol.txt)
and click Ok

Error: Unsupported S-group type

**Expected behavior**
Support all SGroups described https://discover.3ds.com/sites/default/files/2020-08/biovia_ctfileformats_2020.pdf
or import structure without unsupported SGroup

**Actual behavior**
We cannot open structure at all 

**Screenshots**
![image](https://user-images.githubusercontent.com/992495/130270193-28118c30-2c22-400e-9b21-245cad7eb68a.png)

**Desktop (please complete the following information):**
 - OS: [Windows]
 - Browser [chrome]

AndreiProkofev:
Hi, @troits!

We are implementing a temporary solution that will warn and allow the user to choose whether he wants to import the structure without rendering unsupported s-group or not. Also, we are researching our services to work out a further solution for the implementation of unsupported s-groups.

Thank you for using our product!

PR #727, opened 2021-08-24
#723 Structure with some S-groups cannot be opened

AndreiProkofev:


packages/ketcher-react/src/script/ui/state/shared.js
```diff
@@ -54,60 +54,64 @@ export function load(struct, options) {
...
-            if (stereoProp) {
-              atom.stereoLabel = stereoProp.stereoLabel
-              atom.stereoParity = stereoProp.stereoParity
-            }
+            editor.struct(struct)
+          }
+        },
+        err => {
+          //TODO: add error handler call
+          if (!err.isSGroupError) {
```
AndreiProkofev, line 110: The only change in this file. Other stuff is just code styling from prettier

AndreiMazol (changes requested):
Confirm dialog should be shown once even if file contains a lots of unsupported S-groups

packages/ketcher-react/src/script/ui/state/shared.js
AndreiProkofev: Deprecated

AndreiProkofev:
PR closed due to migration of the solution to release 2.3.1

Issue #602, opened 2021-06-28
Stereo flag is overlapped with a structure


AndreiMazol:
**Steps to Reproduce**
Launch Ketcher.
Create a benzene molecule using the Templates tool.
Add a stereo bond (e.g. Single Up bond) to the following carbon atom:
![image](https://user-images.githubusercontent.com/72735611/123618666-17300a00-d811-11eb-993c-e02df68c2f38.png)
3)Add a single bond to the following carbon atom:
![image](https://user-images.githubusercontent.com/72735611/123618728-23b46280-d811-11eb-9ce1-822dfc409a75.png)
Expected result: Stereo flag is not overlapped with a structure.
Actual result: Stereo flag is overlapped with a structure.

vitalgerko:
Browser: Chrome Version 93.0.4577.63, Standalone version 
Text 'abs' appears inside (outside) the structure (see atach.)

https://user-images.githubusercontent.com/90637536/133435511-bf5414ef-5ee7-4fb0-803d-1ac73bee3045.mp4

ElenaOdnoshivkina:
Text 'abs' appears inside (outside) the structure because of selected settings.
When we set "Label display at stereogenic centres" in "Off" text don't appear (approved by @vitalgerko)

vitalgerko:
The problem was with the cookies. I have previously changed the settings in Сhrome. After clearing the cookies, everything is ok, all default settings appeared

vitalgerko:
Release/2.4 Closed as fixed.

PR #743, opened 2021-09-06
#602 Stereo flag is overlapped with a structure 

ElenaOdnoshivkina:


AndreiMazol (changes requested):
please update imports

AndreiMazol (approved):

Issue #521, opened 2021-05-19
'Settings' - It's not convenient to change the default colors of the stereogenic centers by typing RGB colors


Svetlana-Asinovskaia:
Tested on Ketcher v2.2.1-104-gc7896113

Steps to reproduce:
1. Launch Ketcher.
2. Click the 'Settings' button
3. Change the default settings for any stereogenic center (e.g. "Color of Absolute Center") by typing the parameters of RGB color (e.g. 146, 201, 178)

Expected result: It's easy to specify the parameters of an RGB color value.
Actual result: If the zero value of one of the parameters is cleared before entering a new value, all other parameters are also set to zero.
![2021-05-19_14h55_22](https://user-images.githubusercontent.com/80695128/118808678-6d289e00-b8b2-11eb-9abc-00453535e254.gif)

vitalgerko:
Release/2.4 Closed as fixed.

PR #751, opened 2021-09-09
#521 not convient to change color

r1z3rISGOD:


AndreiMazol (changes requested):
Please look at new dependencies. they should be added to ketcher-react package (not example package.json)

AndreiMazol (approved):

Issue #156, opened 2020-12-09
Atom/Bond property should be applied for all selected objects


AndreiMazol:
As a Ketcher user I want to update atom/bond properties of all selected objects together at once.

PR #753, opened 2021-09-13
#156 Atom/Bond property should be applied for all selected objects

ElenaOdnoshivkina:


AndreiMazol (changes requested):
Please merge operations before execution of them

AndreiMazol (approved):

Issue #734, opened 2021-08-30
Create 'Functional Groups' window


AndreiMazol:
FG window should look like 'Templates' window. The list of features and detailed behavior are described in scope of #692 AC2

PR #754, opened 2021-09-13
#734 Create FG window

AndreiProkofev:


r1z3rISGOD (commented):
Add copyright info to EmptySearchResult and delete duplicate <group> tag in fg.sdf (Bz group)

AndreiMazol (approved):

Issue #713, opened 2021-08-11
Hiding buttons on the left menu


AndreiMazol:
### Discussed in https://github.com/epam/ketcher/discussions/710

<div type='discussions-op-text'>

<sup>Originally posted by **mustafacanp** August  9, 2021</sup>
I want to hide all buttons below the "Rotate Tool" on the left menu. Do you have a plan to add these buttons to the `buttons` property of the `Editor` component?

e.g
```
<Editor
    buttons={{
        sgroup: {hidden: true},
        datasgroup: {hidden: true},
        ...
    }}
/>
```

![Screenshot 2021-08-09 162058](https://user-images.githubusercontent.com/16281650/128720858-f01866fb-25c2-4b09-a75e-9b358a4fd2fd.png)
</div>

vitalgerko:
@AndreiMazol  
hiding buttons works, but such risks appear on the left panel, is this ok?
![Hiding buttons](https://user-images.githubusercontent.com/90637536/133562992-f6aca78b-7b59-4ff8-b0e5-0699321de536.png)

PR #781, opened 2021-09-20
#713 Hiding buttons on the left menu

aleksandrgordeev1:


packages/ketcher-react/src/script/ui/views/toolbars/LeftToolbar/LeftToolbar.tsx
```diff
@@ -26,43 +26,11 @@ import { Bond } from './Bond'
 import { RGroup } from './RGroup'
 import { Shape } from './Shape'
 import { Transform } from './Transform'
+import * as leftToolbarOptions from './leftToolbarOptions'
```
AndreiMazol, line 29: Please do not use alias for star symbol: https://github.com/airbnb/javascript#modules--no-wildcard

AndreiMazol (approved):

Issue #793, opened 2021-09-24
Template Library - Ketcher is broken when clicking the 'Edit' button in different structures card


vitalgerko:
**Steps to Reproduce**

1. Launch Ketcher.
2. Click the 'Custom Templates' button.
3. Click the 'Edit' button on the 'α-D-Altropyranose' structure card ('α-D-Sugars' group).
4. Click the 'Cancel' button.
5. Click the 'Edit' button on the 'Piperidine' structure card ('Heterocyclic Rings' group).
6. Click the 'Cancel' button.
7. Click the 'Edit' button on the 'Ribose monophosphate' structure card ('Nucleobases' group).
8. Click the 'Cancel' button.
9. Click the 'Edit' button on the 'β-D-Fructofuranose' structure card ('β-D-Sugar's group).
10. Click the 'Cancel' button.

**Expected behavior**
Ketcher works correctly.

**Actual behavior**
Ketcher is broken.
Error appears.

https://user-images.githubusercontent.com/90637536/134631766-dfbe4dee-9caa-4331-80d3-117f9d21ab7a.mp4

**Desktop:**

OS: Windows
Browser: Chrome Version 93.0.4577.63, Firefox Version 92.0, Microsoft Edge Version 93.0.961.47

**Additional context**
This bug reproduced when clicking in other different structure cards (try "monkey testing")

vitalgerko:
Release/2.4 Closed as fixed.

vitalgerko:
Reopen.

The bug was fixed, but new regression bug appears after changinging code (not reproduced in the previous version)

**Steps to reproduce**
1. Launch Ketcher.
2. Click the 'Custom Templates' button.
3. Click the 'Edit' button on any structure card.
4. Try to change Template name and Atom/Bond ID.

**Expected behavior**
Possible to change Template name and Atom/Bond ID together.

**Actual behavior**
It is impossible to change Template name and Atom/Bond ID together.

vitalgerko:
Release/2.4 Closed as fixed.

PR #809, opened 2021-09-30
#793 Template Library - Ketcher is broken when clicking the 'Edit' button in different structures card

ElenaOdnoshivkina:


r1z3rISGOD (approved):

packages/ketcher-react/src/script/ui/dialog/template/template-attach.jsx
```diff
@@ -50,9 +50,11 @@ class Attach extends Component {
 
   render() {
     const { name, onNameEdit, onAttachEdit, ...prop } = this.props
-    const { atomid, bondid } =
-      this.tmpl.struct.name !== name ? this.tmpl.props : this.props
     const struct = this.tmpl.struct
+    const { atomid, bondid } =
+      struct.atoms.get(this.props.atomid) && struct.bonds.get(this.props.bondid)
```
aleksandrgordeev1, line 55: This check looks a bit strange. Is this the only way?

Issue #540, opened 2021-05-25
Error message when trying to calculate values of the structures with the hydrogen and dative bonds


Svetlana-Asinovskaia:
Tested on Ketcher v2.2.1-115-gaf97cc42

Steps to reproduce:
1. Launch Ketcher.
2. With the 'Hydrogen Bond' tool draw a bond on the canvas.
3. Click on the 'Calculated Values' button.

Expected result: The fields in the "Calculated Values" window are filled with the correct data.
Actual result: An alert appears 
- "Error: IndigoException: molfile loader: unknown bond type: 10"  (remote mode).
![image](https://user-images.githubusercontent.com/80695128/119528423-d5cbbb00-bd89-11eb-9b3f-88e707c8d0e1.png)
- "Given string could not be loaded as (query or plain) molecule or reaction, see the error messages: 'molfile loader: unknown bond type: 10', 'molfile loader: unknown bond type: 10', 'RXN loader: bad header ', 'RXN loader: bad header " (standalone mode).
![image](https://user-images.githubusercontent.com/80695128/119528540-ef6d0280-bd89-11eb-9cef-d66071074d12.png)

The issue is reproduced with the structures with the dative bonds.

Svetlana-Asinovskaia:
Related issue https://github.com/epam/Indigo/issues/413

vitalgerko:
Reopen.
**Only for Standalone version**

**Steps to reproduce:**
1st case
1. Launch Ketcher.
2. With the 'Hydrogen Bond'/ 'Dative Bond' tools draw a bond on the canvas.
3. Click on the 'Calculated Values' button.

**Expected result:** The fields in the "Calculated Values" window are filled with the correct data.
**Actual result:** The error appears in the console. The fields in the "Calculated Values" window aren't filled.
![Actual result](https://user-images.githubusercontent.com/90637536/135862268-ec471c42-f9e0-4260-8ce9-639939c76faf.png)

2nd case
1. Launch Ketcher.
2. With the 'Any Bond'/ 'Aromatic Bond'/  'Single/ Double Bond' / Single/ Aromatic Bond' / 'Double/ Aromatic Bond' tools draw a bond on the canvas.
3. Click on the 'Calculated Values' button.

**Expected result:** There is no error in console. The fields in the "Calculated Values" window are contain text 'Cannot calculate properties for structures with query features'.
[...]

PR #823, opened 2021-10-05
#540 calculate values inputs do not show errors

r1z3rISGOD:


AndreiMazol (changes requested):
Consumer of StructService is responsible for error handling

Issue #898, opened 2021-10-29
Redesign Error window


AndreiMazol:
This moment ready-to-run app  is used native alert window. It is necessary to  create custom dialog that should be aligned with other modals in our App.

AllaZhbanova:
Tested v.2.4
Error window is redesigned
![error2](https://user-images.githubusercontent.com/93334478/141332443-6a0089e0-a1e2-43f7-b911-55ea76f9a6dd.png)
![error1](https://user-images.githubusercontent.com/93334478/141332452-f23909a7-df47-40fc-8768-2d69c4b0cf7a.png)
:

PR #909, opened 2021-11-02
#898 Redesign Error window 

ElenaOdnoshivkina:


example/src/index.tsx
```diff
@@ -1,6 +1,5 @@
 import 'react-app-polyfill/ie11'
 import 'react-app-polyfill/stable'
-import 'url-search-params-polyfill'
```
AndreiMazol, line 3: Why did you  delete this line?

Issue #961, opened 2021-11-12
Error window is shown when user apply 3D to empty canvas


AllaZhbanova:
**Steps to Reproduce**
1. Clear canvas
2. Select 3D
3. Click on Apply

**Expected behavior**
There is no error message

**Actual behavior**
Error window is shown

**Screenshots**

![empty](https://user-images.githubusercontent.com/93334478/141424664-6c005dae-0f53-4abe-bfe4-fc31863d955f.png)

**Desktop :**
 - OS:Windows 10
 - Browser : Chrome Version 93.0.4577.63, Firefox Version 92.0, Microsoft Edge Version 93.0.961.47

AllaZhbanova:
Fixed,
Error text is shown: Error message
Cannot read properties of null (reading 'sgroups')

PR #968, opened 2021-11-12
#961 Error window is shown when user apply 3D to empty canvas 

ElenaOdnoshivkina:


AndreiMazol (changes requested):
Please update PR based on our discussion

Issue #940, opened 2021-11-10
Redesign the floating windows - Structure check


Alena-Guznischeva:
Change the visual representation of the floating windows appearing for the Structure Check operations. 

GIVEN Ketcher user having something on the canvas WHEN clicking on the Structure Check operation THEN 
Ketcher displays a floating window with settings displayed on the left panel with the ability to mark and unmark the settings AND 
Warnings displayed on the right part of the window AND 
Warnings are displayed in accordance with the Settings marked and the list of warnings is updated immediately when the list of marked settings is updated AND
I can close window with the cross icon or using the Esc hotkey

Use the designs below: 

![image](https://user-images.githubusercontent.com/90251323/141073051-309168fc-ee13-477d-8a3f-5506d2f6081a.png)
![image](https://user-images.githubusercontent.com/90251323/141073077-d90559fa-001e-400c-9155-647edbb5b0c8.png)

AllaZhbanova:
Implemented, tested on v. 2.5

PR #971, opened 2021-11-15
#940 Redesign the floating windows - Structure check

ElenaOdnoshivkina:


packages/ketcher-react/src/script/ui/views/components/Dialog/Dialog.tsx
```diff
@@ -94,25 +94,30 @@ const Dialog: FC<Props> = props => {
...
-              type="button"
-              className={button === 'OK' ? styles.ok : styles.cancel}
-              value={button}
-              disabled={button === 'OK' && !valid()}
-              onClick={() => exit(button)}
-            />
-          )
-        )}
-      </footer>
+      {buttons.length ? (
```
r1z3rISGOD, line 103: Ternary operator can be replaced by &&

AndreiMazol (approved):

Issue #956, opened 2021-11-11
Redesign the floating windows - S-Group


Alena-Guznischeva:
Change the visual representation of the windows for S-Group abilities

GIVEN Ketcher user having something selected with selection tool on the canvas WHEN clicking on the Data S-Group operation THEN
Ketcher displays a floating window with settings displayed on the left panel with the ability to change Context, Field Name, select Field Value, and change the Relative setting AND
I can close confirm the changes for S-Group data saved by clicking on the OK button AND
I can close without saving by clicking on the cross icon or by using Esc hotkey

![image](https://user-images.githubusercontent.com/90251323/141276422-5c248aaf-c17e-4239-adcb-8512f023a097.png)

GIVEN Ketcher user having something selected on the canvas with S-Group tool selected WHEN clicking on the selected fragment THEN Ketcher displays a window with the ability to select S-Group Type and change the specific attributes for the types selected AND
I can close confirm the S-Group settings assigned by clicking on the OK button AND
I can close without saving by clicking on the cross icon or by using Esc hotkey

GIVEN Ketcher user having S-Group on the canvas and S-Group tool or Data S-Group tool selected WHEN clicking on the S-Group on the canvas THEN Ketcher displays a window with the same behavior as for creating new S-Group using S-Group tool

![image](https://user-images.githubusercontent.com/90251323/141276572-899fee22-bf48-481f-84a4-6d8330c8a704.png)

PR #974, opened 2021-11-15
#956 redesign the floating windows SGroup

r1z3rISGOD:


packages/ketcher-react/src/script/ui/views/components/Dialog/Dialog.module.less
```diff
@@ -17,16 +17,17 @@
 @import '../../../../../style/mixins.less';
 
 .ok {
-  background: #3e7bfa;
+  background: #343434;
```
ViktoriiaFedotkina, line 20: According to the design, the button should be blue

AndreiMazol (approved):

Issue #954, opened 2021-11-11
Redesign the floating windows - Attachment Point Tool


Alena-Guznischeva:
Change the visual representation of the window for Attachment Point Tool

GIVEN Ketcher user having something on the canvas and having Attachment Point Tool selected WHEN clicking on the atom on the canvas THEN
Ketcher displays a floating window displaying current settings of the attachment points with the ability to mark and unmark the Primary & Secondary attachment points as visible AND
confirm the new settings by clicking on the OK button AND
close the window without saving by using cross icon or by using Esc hotkey

Use the design:

![image](https://user-images.githubusercontent.com/90251323/141267330-792464b2-3740-499a-8cb3-3bb9c7fe0eeb.png)

AllaZhbanova:
Implemented, tested v.2.5

PR #977, opened 2021-11-16
#954 Redesign the floating windows - Attachment Point Tool

r1z3rISGOD:


dmitrii-kliushentsev (approved):

packages/ketcher-react/src/script/ui/views/modal/components/toolbox/Attach/Attach.tsx
```diff
@@ -31,7 +31,7 @@ const AttachPoints = (props: Props) => {
   const { formState, ...rest } = props
   return (
     <Dialog
-      title="Attachment Points"
+      title="ATTACHMENT POINTS"
```
AndreiMazol, line 34: It would be better to  use text  transform instead

packages/ketcher-react/src/Editor.module.less
```diff
@@ -107,12 +107,18 @@
...
     vertical-align: middle;
     margin: 0 0.3em;
     margin-top: 1px \9; // IE8-9
     line-height: normal;
     // TODO: style focus, active, etc
   }
+
+  input[type='checkbox'] {
+    vertical-align: middle;
+    margin-top: 1px \9; // IE8-9
```
AndreiMazol, line 120: What do you need it for? IE8-9 are OOS

AndreiMazol (changes requested):
see comments

AndreiMazol (approved):

Issue #900, opened 2021-10-29
Settings - Preview images of structures (Templates Library, FG Library) do not respect global display settings


vitalgerko:
**Steps to Reproduce**

1. Launch Ketcher with the following display settings (settings: 'Atoms', 'Show hydrogen labels' - on).
2. Open FG Library.
3. Add FG on canvas and Expanded FG abb.
4. Change settings: 'Atoms', 'Show hydrogen labels' - on).
5. Open Templates Library.
6. Add any Template on canvas.

**Expected behavior**
When Launch Ketcher with the following display settings (settings: 'Atoms', 'Show hydrogen labels' - on)
all structures (in FG Library and in Templates Library) should be displayed according to these settings.

**Actual behavior**
In FG Library and Templates Library, preview images of structures do not respect global display settings.

https://user-images.githubusercontent.com/90637536/139446193-5013bf9e-d71a-40c3-92f2-91fb43cc6f85.mp4

**Desktop:**
OS: Windows 10
Browser: Chrome Version 93.0.4577.63, Firefox Version 92.0, Microsoft Edge Version 93.0.961.47

AllaZhbanova:
The issue is still reproducible on 2.4 and 2.

https://user-images.githubusercontent.com/93334478/144440480-45716de3-5597-4ece-9b79-2a74dcf6340b.mp4

5

AllaZhbanova:
The issue is reproducible if set settings : 'Atoms', 'Show hydrogen labels' - off
Open Templates and click edit

Result: hydrogen label is shown
Expected result: hydrogen label is not shown

AllaZhbanova:
Fixed, tested v.2.5

PR #980, opened 2021-11-17
#900 Preview images of structures (Templates Library, FG Library) do not respect global display settings

dmitrii-kliushentsev:


packages/ketcher-react/src/script/ui/dialog/template/TemplateTable.tsx
```diff
@@ -39,23 +40,22 @@ export interface Template {
...
 
+const getSettingsSelector = state => state.options.settings
+
 function tmplName(tmpl: Template, i: number): string {
   return tmpl.struct.name || `${tmpl.props.group} template ${i + 1}`
 }
 
 const RenderTmpl: FC<{
   tmpl: Template
+  options: any
```
r1z3rISGOD, line 51: We are not using 'any' type

AndreiMazol: It can be fixed later: react components library will be reworked in scope of 2.5

AndreiMazol (approved):

Issue #984, opened 2021-11-18
Cannot read properties of undefined (reading 'struct') error if connect two (or more) Cyclopentadiene(T) and click Aromatize


AllaZhbanova:
**Steps to Reproduce**
1. Open Ketcher
2. Select Cyclopentadiene(T) and place it on the canvas
3. Connect the Cyclopentadiene(T)  with other Cyclopentadiene(T)  (see screenshot(
4. Click Aromatize

**Expected behavior**
Nothing happened.

**Actual behavior**
Cannot read properties of undefined (reading 'struct') error is shown

**Screenshots**
![aromatize](https://user-images.githubusercontent.com/93334478/142374045-ef45d921-9e54-4a07-b7cd-eca4dd269ed3.png)

**Desktop :**
 - OS: Chrome Version 93.0.4577.63, Firefox Version 92.0, Microsoft Edge Version 93.0.961.47

AllaZhbanova:
https://github.com/epam/ketcher/issues/1002

AllaZhbanova:
https://github.com/epam/Indigo/issues/533

AllaZhbanova:
Fixed, tested v.2.4

PR #982, opened 2021-11-17
#961 Error window is shown when user apply 3D to empty canvas 

ElenaOdnoshivkina:

PR #989, opened 2021-11-18
#984 Cannot read properties of undefined (reading 'struct') error if connect two (or more) Cyclopentadiene(T) and click Aromatize

dmitrii-kliushentsev:


packages/ketcher-react/src/script/ui/state/server/index.js
```diff
@@ -157,7 +157,7 @@ export function serverTransform(method, data, struct) {
         )
       })
       .catch(e => {
-        state.editor.errorHandler(e.message)
+        state.editor.errorHandler(e.message || e)
```
AnaMoskvina, line 160: seems, errorHandler expects only strings (not objects)

r1z3rISGOD: It seems like sometimes e can be string

AndreiMazol (changes requested):
I guess this PR can be closed  (Elena made the same changes in scope of https://github.com/epam/ketcher/pull/982

Issue #1020, opened 2021-11-24
Functional Groups - Elements in FG Window should be rendered without SGroup's brackets


r1z3rISGOD:
Need to render images of Functional Groups without SGroup's brackets

AllaZhbanova:
Implemented, tested on v. 2.5
Elements in FG Window are rendered without SGroup's brackets

PR #1026, opened 2021-11-25
#1020 Elements in FG Window should be rendered without SGroup's brackets

r1z3rISGOD:


AnaMoskvina (approved):

AndreiMazol (changes requested):
Please update PR based on our discussion

AndreiMazol (approved):

Issue #1003, opened 2021-11-23
Error when apply 3D to empty canvas


AllaZhbanova:
**Steps to Reproduce**
1. Open Ketcher in any browser
2. Do not place any on the canvas
3. Select 3D
4. Click Apply

**Expected behavior**
 No error message is shown

**Actual behavior**
  Error modal is shown

**Screenshots**

![null](https://user-images.githubusercontent.com/93334478/143014997-e9bca5dc-d49c-49e6-883e-5fe2cb38783e.png)

 **Desktop (please complete the following information):**
 - OS: Windows 10
 - Browsers: Chrome Version 93.0.4577.63, Firefox Version 94.0.1, Microsoft Edge Version 93.0.961.47

Alex-Work-Account:
It's not a Firefox problem. There shouldn't be an error if a user clicked on apply

AllaZhbanova:
Fixed, tested on v. 2.4

PR #1034, opened 2021-11-26
#1003 error when apply 3 d to empty canvas

Alex-Work-Account:


AnaMoskvina (approved):

packages/ketcher-react/src/script/ui/state/shared.js
```diff
@@ -75,6 +75,10 @@ export function load(struct, options) {
     return parseStruct(struct, server, options)
       .then(
         struct => {
+          if (!struct) {
```
AndreiMazol, line 78: It looks like workaround. I guess 'then' expression  should be only executed if struct is not empty. Could you  find the reason of this case?

AndreiMazol (commented):
need to  discuss

AndreiMazol (changes requested):
see comments

karen-sarkisyan (approved):

AndreiMazol (changes requested):
you  should not return hardcoded cml file.

AndreiMazol (approved):

Issue #1029, opened 2021-11-26
Structure shouldn't flipped when a part structure is selected


ElenaOdnoshivkina:
**Steps to Reproduce**
1. Launch Ketcher.
2. Create any structure 
3. Select a part of the structure and choose the 'Horizontal Flip' or 'Vertical Flip' tool.

**Expected behavior**
Structure isn't flipped when a part structure is selected.

**Actual behavior**
Structure is flipped when a part structure is selected.

https://user-images.githubusercontent.com/89905163/143539827-c246966a-2f75-4249-bd46-bea9b4c6cc0c.mp4

AllaZhbanova:
Fixed, tested v. 2.4

PR #1043, opened 2021-11-30
#1029 structure shouldnt flip if part selected

Alex-Work-Account:


packages/ketcher-core/src/application/editor/actions/rotate.ts
```diff
@@ -34,26 +34,37 @@ export function fromFlip(restruct, selection, dir, center) {
 
   const action = new Action()
 
-  if (!selection) selection = structSelection(struct)
+  if (!selection) {
+    selection = structSelection(struct)
+  }
```
r1z3rISGOD, line 39: Why is this needed?

Alex-Work-Account: I think it's much more readable, can't miss the brackets :)

AndreiMazol (approved):

Issue #1044, opened 2021-11-30
Text Tool: Add special symbols


AndreiMazol:
It is necessary to extend  the list of existed controls by additional button that allows to insert special symbols. The list  of special symbols provided below:

'α,β,γ,δ,ε,ζ,η,θ,ι,κ,λ,μ,ν,ξ,ο,π,ρ,σ,τ,υ,φ,χ,ψ,ω,℃,℉,Å,°,ħ,±,‰,√,←,→,↚,↛,↔,⇌,∏,∑,∞,∂,∆,∫,≈,≠,≤,≥'

AllaZhbanova:
Implemented, tested v.2.5

PR #1048, opened 2021-12-01
#1044 Text Tool: Add special symbols

r1z3rISGOD:


packages/ketcher-react/src/icons/index.tsx
```diff
@@ -111,6 +111,7 @@ import UndoIcon from './files/undo.svg'
 import ZoomInIcon from './files/zoom-in.svg'
 import ZoomOutIcon from './files/zoom-out.svg'
 import FunctionalGroupsIcon from './files/functional-groups.svg'
+import Symbols from './files/symbols.svg'
```
AndreiMazol, line 114: Please follow naming convention

AndreiMazol (changes requested):

packages/ketcher-react/src/script/ui/views/modal/components/Text/SpecialSymbols/SpecialSymbolsButton.module.less
```diff
@@ -0,0 +1,63 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ ***************************************************************************/
+
+@import '../../../../../../../style/variables';
+@import '../../../../../../../style/mixins';
+
+.textButton {
```
AnaMoskvina, line 20: Would be nice to re-use styles, as styles in TextButton.module.less are basically identical

AndreiMazol (approved):

Issue #1066, opened 2021-12-08
Floating windows redesign - apply new color scheme for new floating windows 


Alena-Guznischeva:
Apply a new color scheme for all the floating windows changed. Use the color scheme utilizing the EPAM logo color.
The windows in score:
1) R-group, S-Group, Attachment Point tool
2) Save window
3) Settings
4) Structure Check, Calculated Values
5) Error message
6) Functional groups

![image](https://user-images.githubusercontent.com/90251323/145182386-dc2d335a-733d-449c-964c-28a59b35d680.png)

PR #1071, opened 2021-12-09
#1066 Floating windows redesign - apply new color scheme for new floating windows

ViktoriiaFedotkina:


AndreiMazol (changes requested):
variables should be renamed

AndreiMazol (approved):

Issue #1091, opened 2021-12-13
Adding banner into resulting file


ElenaOdnoshivkina:
Now, inside resulting file, we delete the comments and the console logs. It is necessary to expand this functionality and add information about the license and copywriting.

PR #1094, opened 2021-12-14
#1091 Adding banner into resulting file 

ElenaOdnoshivkina:


AndreiMazol (changes requested):
Please use original rollup banner property instead

AndreiMazol (approved):

Issue #1093, opened 2021-12-14
Aromatic structure changes to not aromatic if try to connect with other structure 


AllaZhbanova:
**Steps to Reproduce**
1. Select any aromatic structure
2. Place on canvas
3. Try to connect with any other structure
4. Place other aromatic structure on the canvas
5. Connect with other

**Expected behavior**
3., 5. Aromatic structure is connected with other 
**Actual behavior**
3. Aromatic structure changes to not aromatic
5. Aromatic structure is connected with other

**Screenshots**

https://user-images.githubusercontent.com/93334478/145955501-038b2d02-1a11-4698-8a46-0a89cfd37817.mp4

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser Chrome Version 93.0.4577.63, Firefox Version 94.0.1, Microsoft Edge Version 93.0.961.47

AllaZhbanova:
the issue is fixed, tested v.2.4

PR #1099, opened 2021-12-15
#1093 Fix templates tool aromatic attachment

karen-sarkisyan:
Fixes bug #1093  that I introduced in PR #1064

### DIFF
--- a/packages/ketcher-react/src/script/editor/tool/template.ts
+++ b/packages/ketcher-react/src/script/editor/tool/template.ts
@@ -185,10 +185,10 @@ class TemplateTool {
 
       let loop = molecule.halfBonds.get(bond?.hb1 as number)?.loop
 
-      if (!!loop && loop < 0 && bond?.hb2)
+      if (loop && loop < 0 && bond?.hb2)
         loop = molecule.halfBonds.get(bond.hb2)?.loop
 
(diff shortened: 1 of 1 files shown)


packages/ketcher-react/src/script/editor/tool/template.ts
```diff
@@ -179,20 +179,16 @@ class TemplateTool {
       // calculate fragment center
       const xy0 = new Vec2()
       const bond = molecule.bonds.get(ci.id)
-      if (!bond) return
-      const frid = molecule.atoms.get(bond.begin)?.fragment
-      if (!frid) return
-      const frIds = molecule.getFragmentIds(frid)
+      const frid = molecule.atoms.get(bond?.begin as number)?.fragment
+      const frIds = molecule.getFragmentIds(frid as number)
```
AndreiMazol, line 183: frid can be undefined here

AndreiMazol (changes requested):
need to discuss

AndreiMazol (changes requested):
need to  discuss: this case should be covered by aromaticFusing.ts

karen-sarkisyan:
Closing PR because changes that introduced the bug have been reverted.

Issue #1109, opened 2021-12-16
Some atoms and bonds are not detected as part of FG on doubleclick


karen-sarkisyan:
**Steps to Reproduce**
1. Expanded FG structure is placed on canvas. For example from [this file](https://github.com/epam/ketcher/files/7608676/ketcher.21.zip).
2. With Select tool double click on one atom or bond

**Expected behavior**
None of the atoms or bonds that are doubleclicked on can be edited. _Edit Abbreviation_ modal is shown

**Actual behavior**
Some atoms and bonds can be edited. I found that the atoms and bonds that can be edited are those with id of 0.

**Screencast**
![Запись экрана 2021-12-16 в 13 53 50](https://user-images.githubusercontent.com/18420141/146359939-fa89f782-b905-4d39-9211-17fca7e46cbb.gif)

AllaZhbanova:
Fixed, tested v.2.4

PR #1110, opened 2021-12-16
#1109 FG correctly detected by doubleclick on atom or bond with Select tool

karen-sarkisyan:
Fixes issue #1109 

The problem is with type coercion, atom or bond with id of `0` are coerced to `false` in a logical expression, thus not being considered as non-null.

### DIFF
--- a/packages/ketcher-react/src/script/editor/tool/select.ts
+++ b/packages/ketcher-react/src/script/editor/tool/select.ts
@@ -404,7 +404,7 @@ class SelectTool {
       )
       const atomFromStruct = atomId !== null && struct.atoms.get(atomId)?.a
       if (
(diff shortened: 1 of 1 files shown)


AndreiMazol (changes requested):
need to retest  before merging

karen-sarkisyan:
> need to retest before merging

Retested, the fix is still needed. This error was present before refactoring to TS, `atomId` of 0 would coerce to false: 
https://github.com/epam/ketcher/blob/24ddecff94fea2dddb99eec288ab29c06b7aea2f/packages/ketcher-react/src/script/editor/tool/select.js#L403-L412

AndreiMazol (approved):

Issue #1101, opened 2021-12-15
Redesign the floating windows - Change windows sizes in accordance with standards created


Alena-Guznischeva:
Change the sizes of the redesigned windows in accordance with standards created for the windows.
Consult Figma project & UX/UI designers to confirm windows class for size definition. 

![image](https://user-images.githubusercontent.com/90251323/146159869-987cee29-0fe9-40c7-961e-6c37e6f1d97f.png)

PR #1136, opened 2021-12-21
#1101 Redesign the floating windows - Change windows sizes in accordance with standards created 

ElenaOdnoshivkina:


packages/ketcher-react/src/script/ui/component/cliparea/cliparea.module.less
```diff
@@ -20,7 +20,7 @@
   position: absolute;
   bottom: 0;
   left: -50px;
-  width: 0;
+  width: 0 !important;
```
AndreiMazol, line 23: Please do not use !important

AndreiMazol (changes requested):

AndreiMazol (approved):

Issue #1141, opened 2021-12-23
Use Emotion for styling and theming in peptide editor


karen-sarkisyan:
The goal is to move from CSS modules to CSS-in-JS library `emotion` that we already have installed.

Changes to make:
- Use Emotion Theme provider to build a theme for the editor ([Link to emotion docs on theming](https://emotion.sh/docs/theming#themeprovider-reactcomponenttype))
- Use [emotion styled](https://emotion.sh/docs/styled) and/or [css prop](https://emotion.sh/docs/css-prop) for `Editor` component itself and `monomerLibrary` component.

PR #1147, opened 2021-12-23
#1141 use emotion for styling and theming in peptide editor

ViktoriiaFedotkina:


packages/ketcher-polymer-editor-react/src/components/shared/ui/Tabs/Tabs.jsx
```diff
@@ -31,40 +30,70 @@ function TabPanel({ children, value, index, ...other }) {
...
+  }
+
   return (
     <>
       <Tabs
         value={tabIndex}
         onChange={handleChange}
-        className={clsx(classes.tabs, className)}
-        sx={{ minHeight: 24 }}
+        sx={{ ...styleTabs, ...tabsStyle }}
```
karen-sarkisyan, line 80: sx is `mui`'s built-in styling prop. I believe we should switch to `css` prop or use `styled` component from emotion.

packages/ketcher-polymer-editor-react/src/components/shared/ui/Tabs/Tabs.tsx
```diff
@@ -29,44 +28,79 @@ function TabPanel({ children, value, index, ...other }) {
...
   const tabPanel = tabs[tabIndex]
   const Component = tabPanel?.component
   const componentProps = tabPanel?.props
 
-  const handleChange = (event, newTabIndex) => {
+  const handleChange = (_event, newTabIndex) => {
     setTabIndex(newTabIndex)
   }
 
+  const StyledTabs = styled(Tabs)<TabsProps>({
```
karen-sarkisyan, line 49: We don't need to specify `<TabsProps>` explicitly, `styled()` infers props types from whatever component is passed to it.

We can use this type annotation if we want to change these props types, so it's optional.

packages/ketcher-polymer-editor-react/src/components/shared/ui/Tabs/Tabs.tsx
```diff
@@ -29,44 +28,79 @@ function TabPanel({ children, value, index, ...other }) {
...
+    '&.Mui-selected': {
+      color: theme.colors.text.white,
+      backgroundColor: theme.colors.tab.active
+    },
+
+    ...tabsStyle
+  }))
+
+  const TabListContent = styled('div')({
+    ...contentStyle
```
karen-sarkisyan, line 85: Instead of passing `contentStyle` and `tabsStyle` as props from parent, let's write styles right here, next to component itself.

packages/ketcher-polymer-editor-react/src/components/monomerLibrary/MonomerLibrary.tsx
```diff
@@ -72,14 +78,38 @@ const MonomerLibrary = () => {
...
   return (
-    <div className={classes.library}>
-      <Tabs
-        className={classes.tabs}
-        contentClassName={classes.tabsContent}
-        tabs={tabs}
-      />
-    </div>
+    <MonomerLibraryContainer>
+      <Tabs tabsStyle={styleTabs} contentStyle={styleContent} tabs={tabs} />
```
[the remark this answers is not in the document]
karen-sarkisyan: After refactoring we would also be able to avoid using `useAppTheme` hook above.

packages/ketcher-polymer-editor-react/src/components/shared/ui/Tabs/Tabs.tsx
```diff
@@ -13,13 +13,12 @@
...
-import Tab from '@mui/material/Tab'
-import { useState } from 'react'
-import clsx from 'clsx'
-import classes from './Tabs.module.less'
+
+import Tabs, { TabsProps } from '@mui/material/Tabs'
+import Tab, { TabProps } from '@mui/material/Tab'
+import { ReactElement, useState } from 'react'
 import Box from '@mui/material/Box'
+import { styled } from '@mui/material'
```
karen-sarkisyan, line 21: Though they're quire similar, we need `styled` from `@emotion/styled`, not mui :)

packages/ketcher-polymer-editor-react/src/components/shared/ui/Tabs/Tabs.tsx
karen-sarkisyan: Here's an example. We can even use the same scrollbar mixin, :

```tsx
  const TabListContent = styled('div')(({ theme }) => ({
    ...scrollbarThin(theme),

    display: 'flex',
    flexDirection: 'row',
    flexWrap: 'wrap',
    justifyContent: 'flex-start',
    overflowY: 'auto',
    width: '260px',

    '& > *': {
      margin: '25px 8px'
    }
  }))
```

Or better yet, use template string:

```tsx
  const TabListContent = styled('div')`
    display: flex;
    flex-direction: row;
    flex-wrap: wrap;
    justify-content: flex-start;
    overflow-y: auto;
    width: 260px;

    & > * {
      margin: 25px 8px;
    }
    ${(props) => scrollbarThin(props.theme)}
  `
```
I suggest that we try to use the latter approach with template literals with other components here too.

karen-sarkisyan (changes requested):

packages/ketcher-polymer-editor-react/src/Editor.tsx
```diff
@@ -32,13 +36,23 @@ function Editor(props: EditorProps) {
...
+
   return (
-    <div
-      ref={rootElRef}
-      className={clsx('Ketcher-polymer-editor-root', classes.root)}
-    >
-      <AppContainer />
-    </div>
+    <Provider store={store}>
+      <ThemeProvider theme={theme}>
```
karen-sarkisyan, line 50: There's also an interesting hack @AnaMoskvina used in Miew in [this PR](https://github.com/epam/miew/pull/368/files#diff-b005246af4f160c36f839dc96e255eea1f7424d3c9594e3a54719b7f4b8ea1ee).

If it works well, we can merge theme providers for emotion and MUI. We might need MUI theme to [disable ripple effects globally](https://mui.com/getting-started/faq/#how-can-i-disable-the-ripple-effect-globally), or do other adjustments for MUI's defaults.

ViktoriiaFedotkina: It's already merged into Mui theme via createTheme https://mui.com/customization/theming/#custom-variables

example/src/ErrorModal/ErrorModal.tsx
```diff
@@ -14,13 +14,14 @@
  * limitations under the License.
  ***************************************************************************/
 import './ErrorModal.css'
+import { ReactElement } from 'react'
```
r1z3rISGOD, line 17: For what do we need types from react?

ViktoriiaFedotkina: there is a warning about Missing return type on function

example/src/PolymerToggler/PolymerToggler.tsx
```diff
@@ -15,12 +15,13 @@
...
 
 import './PolymerToggler.css'
+import { ReactElement } from 'react'
 
 interface PolymerTogglerProps {
   toggle: (isEnabled: boolean) => void
 }
 
-const PolymerToggler = ({ toggle }: PolymerTogglerProps) => {
+const PolymerToggler = ({ toggle }: PolymerTogglerProps): ReactElement => {
```
AndreiMazol, line 24: Please use JSX.Element instead

AndreiMazol (changes requested):
Please define Theme https://emotion.sh/docs/typescript

karen-sarkisyan (approved):

AndreiMazol (approved):

Issue #1172, opened 2021-12-31
Extend ketcher API: Add possibility to  run server functions


AndreiMazol:
### Discussed in https://github.com/epam/ketcher/discussions/1116

<div type='discussions-op-text'>

<sup>Originally posted by **horiu** December 17, 2021</sup>
Hi,

thanks for developing Ketcher!

Using the Calculated Values feature of the ketcher editor, I can calculate Chemical Formula and Exact Mass. This is a very nice feature. If I want to run this in Javascript and get the values, how can I do that?
I wish I could do it as a function, like ketcher.getSmilesAsync( ) and ketcher.generateImageAsync( ).

Thanks,</div>

PR #1173, opened 2022-01-10
#1172 extend ketcher api add possibility to run server functions

ViktoriiaFedotkina:


ViktoriiaFedotkina:
The current solution is not suitable, should implement changes according to discussion

AndreiMazol (changes requested):
implemented services should be reworked again.

AndreiMazol (approved):

Issue #1194, opened 2022-01-20
Create Layout component for Peptide editor


ViktoriiaFedotkina:
Create a container for all elements of the editor like Menu, Add to Canvas, Monomer library, etc.

![Property 1=Default Layout](https://user-images.githubusercontent.com/94109373/151505887-b1174f14-38e0-4300-9d1c-ea4f64b6ec11.png)

ViktoriiaFedotkina:
The task is back in progress as some changes should be made:
- remove extra blocks
- update styles according to the updated design

PR #1196, opened 2022-01-20
#1194 Create Layout component for Peptide editor 

ViktoriiaFedotkina:


ViktoriiaFedotkina:
Waiting for merge of https://github.com/epam/ketcher/pull/1154 and https://github.com/epam/ketcher/pull/1152 to add these components in the layout

karen-sarkisyan (approved):

packages/ketcher-polymer-editor-react/src/components/Layout/Layout.tsx
```diff
@@ -0,0 +1,112 @@
...
+  flexDirection: 'column',
+  width: '100%',
+  height: `calc(100vh - 40px - ${PADDING} * 2)`
+})
+
+const Left = styled.div({
+  width: '32px',
+  minWidth: '32px',
+  height: `calc(100vh - 40px - ${PADDING} * 2)`
+})
```
karen-sarkisyan, line 47: Note that this notation allows passing of all the props of a div element, so user is able to do this:
```tsx
...
<Layout.Left style={{position: "absolute", right: 0}}>
  <p>Why am I on the right?</p>
</Layout.Left>
...
```
We would still need to somehow enclose it to a React component to control which props can be passed and to avoid breaking layout.

ViktoriiaFedotkina: I found that we can do it like this ```
styled('div', {
  shouldForwardProp: (prop) => prop === 'children'
})
```

AndreiMazol (approved):

Issue #1200, opened 2022-01-25
Create Shared Modal Window Component


AndreiMazol:
Create shared 'Modal' component that should look like in `ketcher-react`. 
Modal should accept header and display `Modal.Content` and `Modal.Footer` if passed as subcomponents.

PR #1212, opened 2022-01-28
#1200 Create shared Modal window component

Suzie33:


karen-sarkisyan (changes requested):

packages/ketcher-polymer-editor-react/src/components/shared/ui/dialog/Modal.tsx
```diff
@@ -0,0 +1,121 @@
...
+})
+
+type ModalSubcomponent = 'Content' | 'Footer'
+
+export const Modal = ({
+  title,
+  isModalOpen,
+  setIsModalOpen,
+  children
+}: ModalProps) => {
```
karen-sarkisyan: I think Modal should also accept an `onOk` handler, because we will probably use it to confirm some actions.

karen-sarkisyan (changes requested):

karen-sarkisyan (approved):

packages/ketcher-polymer-editor-react/src/components/shared/modal/Modal.tsx
```diff
@@ -0,0 +1,122 @@
...
+export const Modal = ({
+  children,
+  title,
+  isOpen,
+  showCloseButton = true,
+  onClose
+}: ModalProps) => {
+  const theme = useTheme()
+
+  const paperProps = useMemo(
```
ViktoriiaFedotkina, line 61: I don't think you need useMemo here

Suzie33: I used useMemo to memoize this object creation, because I pass this object to the Dialog's props

Issue #1223, opened 2022-02-01
Create RNAContructor component


AndreiMazol:
**Prerequisites**

1. RAP Switcher component has been created
2. Monomer Library Item component is ready  to  be reused

**Scope**

1. Implement RNAContructor component
2. Cover implemented component with unit tests

**Design reference**

![image](https://user-images.githubusercontent.com/72735611/151930808-1a661968-b57a-4a76-9374-8e27dc3f9d39.png)

PR #1230, opened 2022-02-02
#1223 Create RNAContructor component 

ElenaOdnoshivkina:


packages/ketcher-polymer-editor-react/src/components/rna/MonomerSection/MonomerSection.tsx
```diff
@@ -0,0 +1,120 @@
...
+import { useState } from 'react'
+import Button from '@mui/material/Button'
+import { MonomerItemType } from 'components/monomerLibrary/monomerLibraryItem/MonomerItem'
+import { css } from '@emotion/react'
+import { useTheme } from '@mui/material'
+import { Group, MonomerList } from 'components/monomerLibrary'
+
+const ResetButton = ({ onClick }) => {
+  const theme = useTheme()
+  const styles = css`
```
ViktoriiaFedotkina, line 26: why don't you use styled component here?

packages/ketcher-polymer-editor-react/src/components/rna/MonomerSection/MonomerSection.tsx
```diff
@@ -0,0 +1,120 @@
...
+    let updateValue = {}
+    if (activeMonomerType === 'Nucleotide') {
+      updateValue = splitNucleotide(item)
+      setSelectedMonomers(updateValue as selectedMonomersType)
+    } else {
+      updateValue[activeMonomerType] = item.label
+      setSelectedMonomers((prevState) => {
+        return {
+          ...prevState,
+          ...updateValue
```
ViktoriiaFedotkina, line 99: i guess you can just replace `...updateValue` here with `[activeMonomerType]: item.label`

packages/ketcher-polymer-editor-react/src/styles/theme.js
```diff
@@ -24,7 +24,8 @@ const color = {
     primary: '#121212',
     secondary: '#BCBCBC',
     light: '#FFFFFF',
-    dark: '#000000'
+    dark: '#5B6077',
+    black: '#000000'
```
karen-sarkisyan, line 28: I don't think we should extend our theme schema with `black` color, it seems to defeat the purpose of a theme.

If this exact color is needed (and won't change if color palette changes) maybe let's just use it without a theme?

packages/ketcher-polymer-editor-react/src/styles/theme.js
```diff
@@ -52,7 +53,8 @@ const color = {
       primary: '#FFFFFF',
       secondary: '#005662',
       disabled: '#7A7A7A'
-    }
+    },
+    reset: '#FF2727'
```
karen-sarkisyan, line 57: 'reset' seems a bit too specific for a theme. We'll probably have just one button like this, i think we'd better style it on its own.

karen-sarkisyan (changes requested):

packages/ketcher-polymer-editor-react/src/components/monomerLibrary/MonomerLibrary.tsx
```diff
@@ -13,67 +13,169 @@
...
+            onItemClick={onItemClick}
+          />
+        )
+      })}
+    </>
+  )
+}
 
+const MonomerLibrary = () => {
   const tabs = [
```
karen-sarkisyan, line 54: We shouldn't declare both `tabs` and `MonomerLibraryContainer` right within `MonomerLibrary` component, this can have performance issues. You can simply move them up outside the component.

Also, maybe move `tabs` content to a separate file as some sort of a constant?

packages/ketcher-polymer-editor-react/src/components/monomerLibrary/monomerLibraryItem/MonomerItem.tsx
```diff
@@ -13,32 +13,46 @@
...
   onClick: () => void
 }
 
-const MonomerItem = (props: MonomerItemPropTypes) => {
-  const { item, onClick } = props
+interface CardProps extends React.HTMLAttributes<HTMLDivElement> {
+  colorScheme?: string
+}
+
+const Card = styled.div<CardProps>`
```
karen-sarkisyan, line 34: You don't need to create an interface here. I think shorter would be better:
```tsx
const Card = styled.div<{ colorScheme?: string }>`
```

packages/ketcher-polymer-editor-react/src/components/rna/MonomerSection/MonomerSection.tsx
```diff
@@ -0,0 +1,87 @@
...
+        setActiveMonomerType={selectMonomerType}
+      />
+      <MonomerList
+        list={items[activeMonomerType]}
+        onItemClick={selectMonomer}
+      />
+    </>
+  )
+}
+export { MonomerSection as RnaMonomerSection }
```
karen-sarkisyan, line 87: Do we _really_ need to rename as we export? It seems to only lead to confusion.

packages/ketcher-polymer-editor-react/src/components/rna/MonomerSection/MonomerSection.tsx
```diff
@@ -0,0 +1,87 @@
...
+          ...prevState,
+          [activeMonomerType]: item.label
+        }
+      })
+    }
+  }
+
+  const [selectedMonomers, setSelectedMonomers] =
+    useState<selectedMonomersType>(getInitialMonomers())
+  const [activeMonomerType, setActiveMonomerType] = useState('Nucleotide')
```
karen-sarkisyan, line 65: It's a better practice to call `useState` hooks at the very top of the component, so that we see them right away and don't accidentally break [rules of hooks](https://reactjs.org/docs/hooks-rules.html).

karen-sarkisyan (changes requested):

karen-sarkisyan (approved):

Issue #1198, opened 2022-01-25
URL schema 'file' is not supported when loading library.sdf or fg.sdf


zspishere:
Hi, I have a feature request about the baseUrl with 'file:///home/shuzhang' prefix when loading library.sdf or fg.sdf.

**Is your feature request related to a problem? Please describe.**

When I open `index.html` in folder `standalone` which is unpacked from [ketcher-standalone-2.4.1.zip](https://github.com/epam/ketcher/releases/tag/v2.4.1) with chrome browser, I can see the ketcher editor page as below.
However, the button `Custom Templates` and `Functional Groups` are disabled.
![image](https://user-images.githubusercontent.com/19356426/150893231-6df5fbe8-6fb2-4f3b-a3fc-556fb0952eb4.png)

And we could get the error details from Browser Console as below:
![image](https://user-images.githubusercontent.com/19356426/150900293-efa20135-a12c-4100-b1c7-e891f432278e.png)

If I use `ketcher-react@2.4.1` and `ketcher-standalone@1.3.0` in my Electron App, the two buttons `Custom Templates` and `Functional Groups` are also disabled as below.
![image](https://user-images.githubusercontent.com/19356426/150895610-ad27a688-52b7-4298-bd96-52da9a71d0a2.png)

**Describe the solution you'd like**
I think Ketcher could get the sdf files through `__dir` variable with 'file://' schema if they were stored together with `index.html`, or through relative path such as '../templates/library.sdf', instread of 'http://' or 'https://' schema.
```
`file://${join(__dirname, '../renderer/index.html#projects?refresh=0')}`
```

Howover, there is another solution. We could regard the sdf files as user's data which should be stored in remote database.
If so, the default sdf files should be provided by Ketcher's caller, maybe through the Component Method or Ketcher Handler.
And, Ketcher should expose a get method of templates, or an event listener, or a callback function.
```
<Editor
   staticResourcesUrl={""}
   structServiceProvider={structServiceProvider}
   onInit={handleOnInit}
   buttons={hideKetcherBtns}
   templatesDefault={{customTemplates: "Http path or JS Array", functionalGroups: "Http path or JS Array"}}
   templatesHandler={handleTemplateChange}
/>
``` 

Best Regards,
Thanks

AndreiMazol:
Hi @zspishere ,

thank you for your ideas. 
We plane to add an additional abstraction to add possibility to describe how to load templates and FGs. It will be implemented as a part of release 2.5 too.

Regards,
[...]

PR #1237, opened 2022-02-04
#1198 url schema file is not supported when loading library.sdf or fg.sdf

ViktoriiaFedotkina:


AndreiMazol (changes requested):
it has to be fully reworked

Issue #1235, opened 2022-02-04
use iframe embed standalone version ketcher init error


niuzhenjiang:
use iframe embed standalone version ketcher init error
```
var ketcherFrame = document.getElementById('ketcher');
this.ketcher = ketcherFrame.contentWindow.ketcher;
console.log(this.ketcher);
```
and this.ketcher will get  undefined,
i must settimeout and delay 1 second to run the code will get this.ketcher success.

I want to ask how can I listen to the loading event of the ketcher under the iframe tag and execute some initialization code after the loading event is completed

AndreiMazol:
Hi @niuzhenjiang ,

Generally  speaking it is not Ketcher related problem.  'Example' was created for Demo purpose only. For real usage we assume using 'ketcher-core' and 'ketcher-react' packages.

Nevertheless, we will include required changes into Ketcher example in order to allow users to use it as is.

By the way as a workaround you  could try following approach: https://stackoverflow.com/questions/9249680/how-to-check-if-iframe-is-loaded-or-it-has-a-content.

Best Regards,
Andrei

niuzhenjiang:
thanks,I thought the ketcher had a callback event for loading completion, sorry

AndreiMazol:
Ketcher-react provides ability  to make some changes after initialization by passing `onInit` property on `<Editor />` component (please look at https://github.com/epam/ketcher/blob/c9b8109b72448a7de4af20f2f7ad90f07c982a61/example/src/App.tsx#L58).

We plan to extend 'Example' functionality to allow use it OOTB by using post message API (https://developer.mozilla.org/en-US/docs/Web/API/Window/postMessage). 

More likely this functionality will  be included in the next release, 2.5

akolyakov42:
<img width="135" alt="Screenshot 2022-02-16 at 11 57 03" src="https://user-images.githubusercontent.com/97884482/154230691-64135b97-fadd-45ad-af54-92598f5aa38c.png">
В айфрейме всё ещё undefined

PR #1250, opened 2022-02-11
#1235 use iframe embed standalone version ketcher init error

ElenaOdnoshivkina:


AndreiMazol (changes requested):
need to discuss

ElenaOdnoshivkina:
Usage: create an HTML page where an iframe with an example will be embedded and add the following script:

`		window.addEventListener("message", (event) => {
		switch (event.data.eventType){
			case 'init':
				console.log('initialization happens');
				break;
			default:
				break;
			}
		})`
		
Result:
example

Issue #1257, opened 2022-02-15
An error in console when selecting the Bond Tool after closing the 'Enhanced Stereochemistry' window


Svetlana-Asinovskaia:
**Steps to Reproduce**
1. Create a structure with stereobonds (e.g. select any template with stereobonds from the Template library).
2. In the structure select any stereocenter.
3. Click the 'Stereochemistry button'.
4. Close the 'Enhanced Stereochemistry' dialog.
5. Click the canvas.

**Expected behavior**
The stereocenter from step 2 is not selected.

**Actual behavior**
The stereocenter selection remains unchanged.

6. Select any bond tool.

**Expected behavior**
The bond tool is selected. 

**Actual behavior**
An error in the console appears.

**Screenshots**

![2022-02-15_21h49_49](https://user-images.githubusercontent.com/80695128/154129118-9a589491-95c8-495a-a99b-0671e778a15f.gif)
<img width="207" alt="2022-02-15_21h51_25" src="https://user-images.githubusercontent.com/80695128/154129129-4b91d12d-c550-4311-b13d-4d9bcc02129c.png">

**Desktop :**
 - OS: Windows 10
 - Browser: Chrome, Version 98.0.4758.82 
 - Version: v2.4.1-99-g81423737

Mariia-Katkova:
Regression re-checked in build 2.5 - OK

PR #1264, opened 2022-02-17
#1257 An error in console when selecting the Bond Tool after closing the 'Enhanced Stereochemistry' window 

ElenaOdnoshivkina:


packages/ketcher-react/src/script/editor/tool/enhanced-stereo.ts
```diff
@@ -46,6 +46,7 @@ class EnhancedStereoTool {
     this.changeAtomsStereoAction().then(
       (action) => action && editor.update(action)
     )
+    this.isNotActiveTool = true
```
karen-sarkisyan, line 49: That works, but maybe let's make it even more obvious? It should never be an active tool, so let's make it always true:

```typescript
class EnhancedStereoTool {
  editor: Editor
  stereoAtoms: Array<number>
  isNotActiveTool: true

  constructor(editor) {
    const selection = editor.selection()
    this.editor = editor
    this.isNotActiveTool = true
...
```

And we can remove all the other lines with `this.isNotActiveTool = true` after that.

karen-sarkisyan (changes requested):

karen-sarkisyan (approved):

AndreiMazol (approved):

Issue #1292, opened 2022-02-28
Incorrect horizontal scrollbars in Save Structure modal window


akolyakov42:
**Steps to Reproduce**
1. Launch Ketcher
2. Load structure through Open File modal or create it.
3. Press Save Structure button in upper toolbar

**Expected behavior**
Modal layout is nice and clear.

**Actual behavior**
1. Modal has extra horizontal scrollbar in the textarea containing structure code.
2. Errors section also has extra horizontal scrollbar

**Screenshots**
![Screenshot 2022-02-28 at 17 12 15](https://user-images.githubusercontent.com/97884482/155998182-40e82e2b-4519-4d47-98fc-eeb1730cdfbd.png)

**User agent**
Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7)
AppleWebKit/537.36 (KHTML, like Gecko)
Chrome/98.0.4758.109 Safari/537.36

PR #1296, opened 2022-03-02
#1292 Incorrect horizontal scrollbars in save structure modal window

AnaMoskvina:
- get rid of redundant scrollbars
- apply unified font for textarea
- add rule for scrollbars for mozilla

### DIFF
--- a/packages/ketcher-react/src/script/ui/views/modal/components/document/Save/Save.module.less
+++ b/packages/ketcher-react/src/script/ui/views/modal/components/document/Save/Save.module.less
@@ -43,7 +43,7 @@
     height: fit-content;
     border: none;
     border-radius: 4px;
-    margin: 2px;
+    margin: 2px 0;
     padding: 2px 5px;
     overflow: auto;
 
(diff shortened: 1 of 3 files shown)


karen-sarkisyan (changes requested):
I think that besides preventing overflows maybe let's also fix style of a horizontal scrollbar, in case it happens to appear somewhere?

We only need to set its height (same as width when it's vertical) in `mixins.less`:

```diff

  &::-webkit-scrollbar {
    width: 4px;
+   height: 4px;
    background-color: @color-scroll-inactive;
    border-radius: 2px;
    -webkit-border-radius: 2px;
  }

```
So horizontal scrollbars would be thin too:
image

packages/ketcher-react/src/script/ui/views/modal/components/document/Save/Save.module.less
```diff
@@ -43,7 +43,7 @@
     height: fit-content;
```
karen-sarkisyan, line 43: As of now, even 1 line of warnings causes overflow.
image

I suggest that we avoid unnecessary overflow and scroll when we're within `50px` of max-height by using this code or something like it:
```diff
.warnings {
  max-height: 50px;
- height: fit-content;
+ flex-shrink: 0;
  border: none;
  border-radius: 4px;
  margin: 2px 0;
  padding: 2px 5px;
  overflow: auto;
```

karen-sarkisyan (changes requested):

karen-sarkisyan (approved):

AndreiMazol (approved):

PR #1314, opened 2022-03-08
fix focus on label edit popup

fedorseliankin:


AndreiMazol (changes requested):
Please provide more generic solution

AndreiMazol (approved):

Issue #1273, opened 2022-02-18
Full-screen mode support for Ketcher


Alena-Guznischeva:
As a developer of the application using Ketcher as a component, I want to add the ability to use Ketcher in full-screen mode to provide a better user experience for my users with the chance to use big Ketcher canvas even for the application with a small place for Ketcher. 

Acceptance Criteria:
GIVEN the application that uses Ketcher WHEN the option of the full-screen mode is added THEN Ketcher can be opened in full-screen mode in all the browsers supported 

Update on March, 23rd:
Add FullScreen button. Should be placed on the top panel, right top corner. There will be zoom moved near this button to the corner via separate task.

PR #1316, opened 2022-03-09
#1273 full screen mode support for ketcher

AnaMoskvina:
- add full screen support for ketcher
- add full screen support for polymer editor
- design is a draft and most likely be changed later

both buttons maximize only corresponding application (ketcher or polymer editor), so, in example,  ketcher / polymer editor toggler is not available in a full screen mode
for example application, consider single full screen mode button
Ketcher
polymer-editor


demo/src/constants/buttons.ts
```diff
@@ -118,7 +118,7 @@ export const buttonLabelMap = {
   'shape-rectangle': 'Shape Rectangle',
   'shape-line': 'Shape Line',
   text: 'Add Text Tool',
-  'enhanced-stereo': 'Stereochemistry',
+  'enhanced-stereo': 'Stereochemistry'
 }
```
karen-sarkisyan, line 122: I think everything's working great! Maybe let's add an option to hide this button into our demo app right away?

```diff
...
  text: 'Add Text Tool',
  'enhanced-stereo': 'Stereochemistry',
+ fullscreen: 'Full Screen'
}
```

Also need to add 'fullscreen' string into `allButtons` array above.

karen-sarkisyan (changes requested):

demo/src/constants/buttons.ts
AnaMoskvina: added

AndreiMazol (approved):

demo/src/constants/buttons.ts
AnaMoskvina: added

karen-sarkisyan (approved):

Issue #1312, opened 2022-03-04
Extra scrollbar in Data S-Groups modal when Field name contains long string


akolyakov42:
**Version**
Ketcher v2.4.2-118-g6f0fde17

**Steps to Reproduce**
1. Launch Ketcher
2. Load structure or create it on canvas
3. Select structure with Lasso tool
4. Press "Data S-Group" button in left toolbar

**Case 1**
    5. Select "Multifragment" from "Context" dropdown (KETCHER_MULTIPLE_FRAGMENT appears in "Field name" dropdown)

**Case 2**
    5. Select "Fragment" from "Context" dropdown (MDLBG_FRAGMENT_STEREO appears in "Field name" dropdown)
    6. Select MDLBG_FRAGMENT_COEFFICIENT from "Field name" dropdown

**Expected behavior**
Modal window accords it's width to "Field name" dropdown

**Actual behavior**
Extra horizontal scrollbar shows up in the bottom of modal window

**Screenshots**
![Screenshot 2022-03-04 at 16 16 56](https://user-images.githubusercontent.com/97884482/156774500-16af460f-d903-4f38-8dba-22e568e4f63e.png)
![Screenshot 2022-03-04 at 16 27 39](https://user-images.githubusercontent.com/97884482/156772762-7c014f8e-75d0-4d71-beb5-8ccb864f848d.png)

**User agent**
Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) 
AppleWebKit/537.36 (KHTML, like Gecko) 
Chrome/98.0.4758.109 
Safari/537.36

PR #1323, opened 2022-03-11
#1312 extra scrollbar in data s-group modal when field name contains long string

AnaMoskvina:
Scrollbar appears because of fixed standard width of dialog and select width that depends on option string length

Set max width for the select itself, options in dropdown can have bigger width, like this (there's a single edge case):
closed
open

Another option:
- set bigger (not standard) width for data s-group modal (inconsistency with other modals)


AnaMoskvina:
Approved by Alena

AndreiMazol (approved):

Issue #1318, opened 2022-03-10
Add possibility to hide buttons in polymer editor


AnaMoskvina:
Depending on user configuration

PR #1330, opened 2022-03-15
#1318 add possibility to hide buttons in polymer editor

AnaMoskvina:
Use buttons prop for hiding buttons

### DIFF
--- a/example/src/App.tsx
+++ b/example/src/App.tsx
@@ -2,7 +2,10 @@ import 'ketcher-react/dist/index.css'
 
 import { ButtonsConfig, Editor } from 'ketcher-react'
 import { Ketcher, RemoteStructServiceProvider } from 'ketcher-core'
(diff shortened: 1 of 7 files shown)


AndreiMazol (changes requested):
need to discuss

karen-sarkisyan (changes requested):
As discussed, let's try to preserve the pattern of using subcomponent tree as much as possible. Here are the changes I think we should do:
1. Remove functions `renderMenuItem` and `renderSubmenu` in Editor.tsx
2. Instead, use context to expose isHidden prop/function to an item. If it's hidden, it should return `null`.
3. Edge case, when a single item of a submenu is left, will be handled inside of `SubMenu.tsx`. We can filter trough known hidden items received via context and if we know that only 1 item is left, we don't render a dropdown, only a single item.

Issue #1332, opened 2022-03-16
Invisible enhanced flag interferes with user selection


graphite-diamond:
**Steps to Reproduce**
1. Use ketcher v2.4.2 standalone/index.html
2. (optional) Execute ketcher.editor.subscribe('change', (data) => console.log(data));
3. Draw a single oxygen atom in the sketcher, then click the "lasso selection" button
4. Move mouse cursor to above oxygen atom (approximately to within the blue circle as in the below screenshot)
![image](https://user-images.githubusercontent.com/101608359/158643332-ea46d166-47e5-4a00-9331-e2925c88e6f7.png)

5. Hold left click button, move mouse in circle around oxygen atom, release left click button

**Expected behavior**
The lasso should become visible during step 5, and the oxygen atom should be selected

**Actual behavior**
The lasso does not become visible, and the oxygen atom is not selected. Instead, an invisible "enhanced flag" object is selected and moved. The average user may not know that this behavior was caused by moving an invisible "enhanced flag" object, and the user may think that the selection feature is not always working.

Desktop (please complete the following information):

OS: macOS

Browser [chrome or edge]
Version [chrome 99.0.4844.51, edge 98.0.1108.43]

Ketcher version [v2.4.2].
Ketcher standalone v2.4.2, downloaded from https://github.com/epam/ketcher/releases/download/v2.4.2/ketcher-standalone-2.4.2.zip

**Additional context**
Is there a way to disable "enhanced flags" from appearing on the sketcher, using the ketcher API? This could be a solution that a developer could implement right away.

PR #1336, opened 2022-03-18
fix Invisible enhanced flag interferes with user selection (#1332)

IliaVeselov:


AndreiMazol (changes requested):
fix is invalid

packages/ketcher-react/src/script/editor/shared/closest.js
```diff
@@ -240,12 +240,14 @@ function findClosestBond(restruct, pos, skip, minDist, scale) {
   return null
 }
 
-function findClosestEnhancedFlag(restruct, pos) {
+function findClosestEnhancedFlag(restruct, pos, skip, _, scale) {
```
AndreiMazol, line 243: please provide suitable names for arguments: '_' and 'scale'

Issue #1315, opened 2022-03-09
Elements panel & Periodic table redesign


Alena-Guznischeva:
Please, implement changes for the quick periodic elements panel & periodic table floating window:

Change design for the panel with default elements (same list as today) and ability to open the full periodic table. Panel should be aligned on the right part of the canvas and placed vertically. The low part of the panel should provide ability to open Periodic Table floating window. Right now it should contain 2 tabs - Periodic Table (should be 1st and opened by default) and Extended table. 
![image](https://user-images.githubusercontent.com/90251323/157402787-59298089-1dd7-4ae5-8441-b4e6c46e24c6.png)

Linked issue: https://github.com/epam/ketcher/issues/1340

Svetlana-Asinovskaia:
Elements panel mockup in the task is obsolete. Implemented according to the redesign.  
<img width="28" alt="2022-04-18_22h08_29" src="https://user-images.githubusercontent.com/80695128/163862036-d8f81cad-5e12-4897-a3d5-f866b55906d3.png">
discussed with @Alena-Guznischeva
"Panel should be aligned on the right part of the canvas" will be implemented in https://github.com/epam/ketcher/issues/1340.

PR #1338, opened 2022-03-21
#1315 right toolbar redesign

Mariia-Katkova:
Changed toolbar view for desktop, moved button classes to atom component, added some styles according to the design


karen-sarkisyan (changes requested):

packages/ketcher-react/src/script/ui/views/toolbars/RightToolbar/RightToolbar.module.less
```diff
@@ -19,6 +19,12 @@
 .root {
   .toolbar();
   .vertical-toolbar();
+  position: relative;
+  top: 0;
+  right: 60px;
+  margin-top: 20px;
+  margin-right: 20px;
+  z-index: 10;
```
karen-sarkisyan, line 28: Note that this toolbar is nested within a rightmost column of a grid defined in `App.module.less`. Having a `margin-right` causes grid column to bulge, but `right` makes the toolbar render outside of its grid-area:
image

As long as we don't need that column anymore, maybe it'd be better to remove right column from root grid at all, and just position the Elements panel `absolute`ly?

karen-sarkisyan (changes requested):

packages/ketcher-react/src/script/ui/views/toolbars/RightToolbar/RightToolbar.tsx
```diff
@@ -73,25 +77,20 @@ const RightToolbar = (props: Props) => {
             onAction={onAction}
           />
           <AtomsList atoms={freqAtoms} active={active} onAction={onAction} />
-        </Group>
-
-        <Group>
-          <div ref={sizeRef}>
+          <div className={classes.toolbarButton} ref={sizeRef}>
             <ToolbarGroupItem id="period-table" {...rest} />
```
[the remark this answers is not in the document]
Mariia-Katkova: changed the icon

packages/ketcher-react/src/script/ui/views/toolbars/RightToolbar/RightToolbar.module.less
Mariia-Katkova: removed the right column

packages/ketcher-react/src/script/ui/component/view/Atom/Atom.jsx
```diff
@@ -14,11 +14,16 @@
  * limitations under the License.
  ***************************************************************************/
 
-function Atom({ el, shortcut, className, ...props }) {
+import classes from '../../../views/toolbars/ToolbarGroupItem/ActionButton/ActionButton.module.less'
```
AndreiMazol, line 17: All styles related to Atom component should de defined inside Atom.module.less. Please do  not reuse styles of ActionButton component

AndreiMazol (changes requested):
see comments

Issue #1324, opened 2022-03-14
Floating windows - Change About window design


Alena-Guznischeva:
Implement changes for the floating window About: provide new design for the window. 
AC:
GIVEN Ketcher user WHEN clicking on the ? icon on the top panel WHEN About window is opened presenting the Ketcher logo, Ketcher and Indigo version information, links to the Ketcher and Indigo pages on the EPAM LIfe Science portal, link to the main page of EPAM LIfe Science portal and Feedback link to the suggestion of the Ketcher feedback AND I links are opening in the new browser tab when I click on them AND I can close the form by Esc OR clicking on the cross icon.

Use the design below. Ignore the Close button, the bottom panel should be small and empty. 

![image](https://user-images.githubusercontent.com/90251323/158136036-85457232-36a4-440f-94fe-4b1fc3527417.png)

PR #1339, opened 2022-03-21
#1324 Floating window change About window design

AnaMoskvina:
- change About window design, not dialog design (delimiter and buttons belong to dialog component)
- add possibility to forward HeaderContent component
- other dialog windows are not affected
About_redesign


fedorseliankin:
2FE4F2E6-4958-4AEB-AA1C-C7CAC740DF4C

Issue #1306, opened 2022-03-03
UX-UI Improvements: Hand tool


Alena-Guznischeva:
Implement the new tool to allow an easy view of content on the canvas: Hand Tool. The tool should be available on the left panel and should be enabled when the icon is clicked: 
![image](https://user-images.githubusercontent.com/90251323/156566569-1e7af22a-7b41-4106-87c3-053e6a496da9.png)

AC1: GIVEN Ketcher user working with canvas content WHEN clicking the Hand tool on the panel THEN the tool icon should be presented as selected and cursor presentation changes to Hand icon when hovering over the canvas AND content of the canvas is not changing and selections, highlights, and other canvas objects are not changing

AC2: GIVEN Ketcher user working with canvas content with Hand tool enabled WHEN clicking on the left mouse button over the canvas THEN cursor icon should be changed to Hand compressed icon: 
![image](https://user-images.githubusercontent.com/90251323/156566211-5ebc52e8-54f8-4406-ad83-d2129846ff6f.png)

AC3: GIVEN Ketcher user working with the canvas content with scrolls presented with Hand tool enabled WHEN clicking on the left mouse button over the canvas AND moving the mouse THEN Ketcher canvas content should be moved like the canvas was dragged by the hand to the direction of the mouse move AND canvas content should be presented as if scrolls were moved AND scrolls position should be updated and scrolls should hide if the move has resulted in displaying all canvas content in the window

AC4: GIVEN Ketcher user working with the canvas content without scrolls presented with Hand tool enabled WHEN clicking on the left mouse button over the canvas AND moving the mouse THEN Ketcher canvas content should be moved like the canvas was dragged by the hand to the direction of the mouse move AND srolls may appear if the move results in hiding part of the canvas content 

AC5: GIVEN Ketcher user working with the canvas content with Hand tools WHEN clicking on any object on the canvas THEN this object should not be changed, selected or moved

Alena-Guznischeva:
Open question: can we do the animation for the Hand -> Compressed Hand pointer transition.

Alena-Guznischeva:
Hotkey SPACE added for the new tool

Alena-Guznischeva:
Hotkey Ctrl + H replacing SPACE. Space will be reserved for using Hand tool with hold  - will be implemented via separate issue.

PR #1357, opened 2022-03-25
#1306 hand tool

fedorseliankin:


AnaMoskvina:
after merging, styles or left toolbar buttons are affected (expand icon positioning), need to be fixed

PR #1387, opened 2022-03-31
#1315 atomslist styles fix

Mariia-Katkova:
added hover and selected styles for atoms list

### DIFF
--- a/packages/ketcher-react/src/script/ui/component/view/Atom/Atom.jsx
+++ b/packages/ketcher-react/src/script/ui/component/view/Atom/Atom.jsx
@@ -22,8 +22,9 @@ import styled from '@emotion/styled'
 import α from 'color-alpha'
 
 const atomStyle = (props) => {
-  if(!props.withBorder) return '';
-  const atomColor = props.el && props.el.label ? ElementColor[props.el.label] : '#000';
+  if (!props.withBorder) return ''
+  const atomColor =
+    props.el && props.el.label ? ElementColor[props.el.label] : '#000'
   return css`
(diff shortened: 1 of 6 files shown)


packages/ketcher-react/src/script/ui/component/view/Atom/Atom.module.less
```diff
@@ -15,45 +15,32 @@
...
- }
- 
- .selected {
-    background-color: @color-button-primary;
-    color: white;
- }
 
- 
+   &:active {
+    color: #fff;
```
AnaMoskvina, line 42: we can use variable here

packages/ketcher-react/src/script/ui/component/view/Atom/Atom.jsx
```diff
@@ -16,19 +16,42 @@
...
+    }
+    &.selected {
+      color: #fff;
+      background-color: ${α(atomColor, 0.8)};
+
+      &:hover {
+        background-color: ${atomColor};
+      }
+    }
+  `
```
AnaMoskvina, line 42: In order to prevent recreating AtomButton function, we can move it outside the Atom component (e.g., create getAtomButton function, which accepts atomColor ant returns styled component)

AnaMoskvina (changes requested):

AndreiMazol (changes requested):
'withBorder' prop should be deleted

AnaMoskvina (approved):

AndreiMazol (approved):

PR #1400, opened 2022-04-05
correct double click behavior

fedorseliankin:


packages/ketcher-react/src/script/ui/dialog/template/TemplateDialog.tsx
```diff
@@ -165,9 +165,9 @@ const TemplateDialog: FC<Props> = (props) => {
       ? sdfSerializer.serialize(templateLib)
       : sdfSerializer.serialize(functionalGroups)
 
-  const select = (tmpl: Template): void => {
+  const select = (tmpl: Template, activateEmmidiatly = false): void => {
```
AndreiMazol, line 168: please fix the variable name

Issue #1350, opened 2022-03-24
Open, Save and Reset buttons in Settings window looks unclickable (grey)


akolyakov42:
**Steps to Reproduce**
1. Launch Ketcher
2. Click Settings button in top right corner of window
3. Overview Settings modal window header
3.1. Save settings button
3.2. Open settings button
3.3. Reset settings button

**Expected behavior**
Save and Open buttons are always look clickables. 
Reset settings buttons is clickable if there are settings to reset.

**Actual behavior**
Buttons are gray and user may understand that they are turned off and unclickable.

**Screenshots**
<img width="122" alt="Screenshot 2022-03-24 at 20 14 10" src="https://user-images.githubusercontent.com/97884482/159972842-b8b6c57e-8a47-4799-b0f3-b3bb3c98f7c4.png">

**User agent**
Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) 
AppleWebKit/537.36 (KHTML, like Gecko) 
Chrome/99.0.4844.83 
Safari/537.36

**Ketcher version**
2.4.2-135-gcdead093

**Additional info**
All buttons are working properly. Problem is only in appearance

ViktorMakarov-EPAM:
Additional requirements: 

- [ ] Buttons must be located on left part of header
- [ ] Add behaviour on hover: If icon is active it changes color

Svetlana-Asinovskaia:
Verified in Release/2.5.0.
<img width="168" alt="2022-04-13_21h06_17" src="https://user-images.githubusercontent.com/80695128/163242991-5c539ea9-3d98-4189-8530-cc514e2fce2d.png">
 The correct location of the buttons is near the 'Settings' header.

Mariia-Katkova:
Regression re-check, in build 2.5 - OK

PR #1403, opened 2022-04-06
#1350 buttons in settings window looks unclickable

ViktorMakarov-EPAM:
Снимок экрана 2022-04-07 в 09 58 33


packages/ketcher-react/src/script/ui/views/modal/components/meta/Settings/Settings.tsx
```diff
@@ -73,6 +74,25 @@ const SettingsDialog = (props: Props) => {
...
+
+  let changes = {}
+
+  const onChangeForm = (name: string, value: string | number | boolean) => {
+    changes[name] = value
+    if (initState[name] === value) {
+      delete changes[name]
+    }
+    const changedFields = Object.keys(changes)
+    setHasChanges(Boolean(changedFields.length))
```
AndreiMazol, line 87: please use `length > 0` instead

AndreiMazol (approved):

Issue #1392, opened 2022-04-01
Floating windows - Open Window Redesign iteration #2


Alena-Guznischeva:
Update design of the Open Menu window. The floating window should be opened when Open operation on the top menu is clicked on. 
The window should contain areas for 3 options: Paste from Clipboard, Open from file, Import from Image. 
Updates in this iteration:
1. Change look & feel, add icons, change wording in accordance with design
2. Provide ability to initiate operations while clicking on the full rectangle area for the option (drug & drop for 1st and 3rd should also work for the full rectangle area): 
3. (Out of scope) Ctrl + V text should be excluded from design. Ability to use this Hotkey will be implemented in the separate task in 2.6. 

![image](https://user-images.githubusercontent.com/90251323/161289159-28e87644-9afc-4a58-8ba1-e001444c294f.png)

Alena-Guznischeva:
3 will be done via this issue: https://github.com/epam/ketcher/issues/1393

Alena-Guznischeva:
Use the following message for the situation when Recognize is not available:
_Image Recognition service is not available_

PR #1412, opened 2022-04-08
#1392 Floating windows - Open Window Redesign iteration #2 

ViktorMakarov-EPAM:
Снимок экрана 2022-04-11 в 15 49 40


AnaMoskvina (approved):

packages/ketcher-react/src/icons/index.tsx
```diff
@@ -63,6 +63,8 @@ import ClearIcon from './files/clear.svg'
 import NotFoundIcon from './files/not-found.svg'
 import OpenIcon from './files/open.svg'
 import PasteIcon from './files/paste.svg'
+import PasteIcon1 from './files/paste1.svg'
```
AndreiMazol, line 66: could you provide more  meaningful name

packages/ketcher-react/src/script/ui/views/modal/components/document/Open/Open.module.less
```diff
@@ -18,6 +18,21 @@
 @import '../../../../../../../style/mixins.less';
 
 .open {
+  -moz-box-shadow: 0 5px 20px #67688440;
```
AndreiMazol, line 21: it is not necessary to  add vendor prefixes.

packages/ketcher-react/src/script/ui/views/modal/components/document/Open/Open.module.less
```diff
@@ -18,6 +18,21 @@
...
+  -webkit-box-shadow: 0 5px 20px #67688440;
```
AndreiMazol, line 22: please use variables instead of 'magic' color stirngs

AndreiMazol (changes requested):
see comments

Issue #1396, opened 2022-04-04
Floating windows - S-Group windows redesign iteration #2


Alena-Guznischeva:
Update the design of the 2 S-Group Properties windows (S-Group Tool, S-Group Properties Tool). 
Use the designs below. The changes should be the following:

1. Change color scheme & look and feel
2. Add Cancel button to forms
3. Change order of the Context & Field Name for the S-Group Propertries tool form. 
4. Validation and Default values should stay the same. 
![image](https://user-images.githubusercontent.com/90251323/161521957-58253d96-8ebf-4380-bf15-c8508da99e82.png)
![image](https://user-images.githubusercontent.com/90251323/161521905-6af71dbf-4baf-457e-9fc7-27002c5e64d6.png)
![image](https://user-images.githubusercontent.com/90251323/161521996-6fbd52de-672a-429e-9d2c-418f468ff18e.png)
![image](https://user-images.githubusercontent.com/90251323/161522021-a2dac62d-9467-437d-bc08-0b796c152bb9.png)
![image](https://user-images.githubusercontent.com/90251323/161522058-aea99c72-3e51-45e8-9593-6709c6fd798e.png)

ZhannaNovikova:
Doesn't fit the requirements, should be reopened:

- Overlaping the text of the "Field name" on the arrow button to open the "Field name" drop down after selecting the MDLBG_FRAGMENT_COEFFICIENT field name;
- Incorrect radio buttons clickable area;
- Horisontal scroll when the "Field value" winwow is extended.

fedorseliankin:
Regression re-check, in build 2.5 - OK

fedorseliankin:
Regression re-check, in build 2.5 - OK

PR #1420, opened 2022-04-12
#1396 s groups windows redesign

Mariia-Katkova:
Floating windows - S-Group windows redesign


packages/ketcher-react/src/script/ui/views/components/Dialog/Dialog.tsx
```diff
@@ -81,7 +81,7 @@ const Dialog: FC<Props> = (props) => {
   }, [])
 
   const isButtonOk = (button) => {
-    return button === 'OK' || button === 'Save'
+    return button === 'OK' || button === 'Save' || button === 'Apply'
```
AndreiMazol, line 84: 'Apply' option should be deleted

packages/ketcher-react/src/script/ui/dialog/toolbox/sgroup.module.less
```diff
@@ -14,48 +14,73 @@
...
-
+@import "src/style/mixins";
+// stylelint-disable selector-pseudo-class-no-unknown
+// stylelint-disable no-descending-specificity
 .sgroup {
   width: @width-small;
+  border-radius: 6px;
+  background-color: @color-background-primary;
+
+  header {
```
AndreiMazol, line 25: 'header' styles should be moved to Dialog styles

AndreiMazol (approved):
see comments

AndreiMazol (changes requested):
update PR regarding to  comments. NB: global Dialog styles will be added to  Dialog component so they should be deleted from 's-group' component

packages/ketcher-react/src/script/ui/views/components/Dialog/Dialog.tsx
ViktorMakarov-EPAM: We set Apply button instead of OK button. What should we do instead of this decision?

packages/ketcher-react/src/script/ui/component/form/Select/Select.module.less
```diff
@@ -28,7 +28,7 @@
     color: @color-text-primary;
     font-family: @Inter;
     font-size: @font-size-regular;
-    padding: 0 8px;
+    padding: 4px 8px;
```
AnaMoskvina, line 31: seems, it's a shared component; changes didn't affect other components?

packages/ketcher-react/src/script/ui/dialog/toolbox/sgroup.module.less
```diff
@@ -14,48 +14,73 @@
...
+    border-bottom: 1px solid @color-grey-3;
+    margin: 0;
+    padding: 0 12px;
+    height: 36px;
+    font-weight: 500;
+  }
+
+  & > div[class^='Dialog-module_dialog_body'], footer {
```
AnaMoskvina, line 33: Maybe, dialog body styles can be redefined in Dialog module styles (if are same in all dialogs);
it's safer to not use 'class^='Dialog-module_dialog_body'', 'div' selector would be enough here

AnaMoskvina:
And few comments on markup:
- think, it's better to consult the designer on select dropdown styles (maybe, it should have rounded corners and shadow, like on other mockups)
- text on buttons looks not centered (line height property can be changed, e.g.)
- we decided to stick current input styles (radio buttons vs checkboxes)?

dropdown

packages/ketcher-react/src/script/ui/views/components/Dialog/Dialog.tsx
Mariia-Katkova: done

packages/ketcher-react/src/script/ui/dialog/toolbox/sgroup.module.less
Mariia-Katkova: moved header to dialog

packages/ketcher-react/src/script/ui/component/form/Select/Select.module.less
Mariia-Katkova: removed this padding change, not needed anymore

packages/ketcher-react/src/script/ui/dialog/toolbox/sgroup.module.less
Mariia-Katkova: removed this style

AndreiMazol (approved):

Issue #1402, opened 2022-04-06
Left panel - Change how the menus are open + Fix the shadows under menus


Alena-Guznischeva:
Change the behaviour & design of the left tool panel:

1. Change how the menus are opened for the items that have menus (Lasso, Bonds, Rotation, Arrows, Reaction, R-Groups, Simple Objects): if the tool in the group is selected the menu should be opened just by clicking on this selected tool, not only by clicking on the menu triangle. 
2. Change how the shadows look for the menu opening. Use the design below:

![image](https://user-images.githubusercontent.com/90251323/161964707-749c286f-cde3-4aa2-8d46-09cb19c41368.png)

![image](https://user-images.githubusercontent.com/90251323/161964770-b8998e96-f96b-4ff2-97f4-1719ed632f93.png)

Svetlana-Asinovskaia:
The menus are opened in the following way:
if the tool in the group is selected
- with single click on the menu triangle;
- with single click on the selected tool.
if the tool in the group is not selected
- with single click on the menu triangle;
- by double clicking on the selected tool. 

@Alena-Guznischeva 

Verified in Release/2.5.0

PR #1421, opened 2022-04-12
#1402 fix how left loobar menus and shadows

fedorseliankin:


ZaimBeshich1:
Are you sure we are merging into the master branch?

AndreiMazol (approved):

Issue #1424, opened 2022-04-12
Floating windows - Change design for Label Edit form


Alena-Guznischeva:
Change the design of the Label Edit form. 
The changes should be the following:
1. Use color scheme, header, buttons & other components from the ones used in other forms
2. Apply the alignment & pixel amount rules for the components on the form using the rules for the standard forms. 
3. Change the button OK label to Apply
4. Validation and logic should stay the same. 

Old form: 
<img width="129" alt="image" src="https://user-images.githubusercontent.com/90251323/162981303-f77e3397-3b9b-43de-9e7c-08917f65d6f0.png">

Mariia-Katkova:
Verified in Release 2.5

PR #1460, opened 2022-04-15
#1424 redesign label edit form

ZaimBeshich1:


packages/ketcher-react/src/script/ui/dialog/toolbox/labeledit.jsx
```diff
@@ -66,14 +67,28 @@ function deserialize(value) {
...
+  }
   return (
     <Dialog
       title="Label Edit"
-      className="labeledit"
       valid={() => valid}
+      withDivider={true}
       result={() => deserialize(result.label)}
+      className={styles.labelEdit}
+      buttons={getButton()}
```
AnaMoskvina, line 91: basically, Apply button is OK button
we can add following props:
 buttons: ['OK'],
 buttonsNameMap={{ OK: 'Apply' }}

in this case, no other changes are needed
You can check out TemplateDialog as a reference

packages/ketcher-react/src/script/ui/dialog/toolbox/labelEdit.module.less
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+    input {
+      margin-left: 8px;
+      min-width: 140px;
+    }
+  }
+
+  footer {
+    padding: 16px 12px 12px;
```
AnaMoskvina, line 42: I will add these changes to global styles, so you can remove them

packages/ketcher-react/src/script/ui/dialog/toolbox/labelEdit.module.less
```diff
@@ -0,0 +1,64 @@
...
+
+    .okButton {
```
AnaMoskvina, line 44: with buttonsNameMap prop we won't need it
also, you can use corresponding mixins when needed

AnaMoskvina (changes requested):

packages/ketcher-react/src/script/ui/dialog/toolbox/labeledit.jsx
ZaimBeshich1: Ok thank u, now I understand how to use it.

packages/ketcher-react/src/script/ui/dialog/toolbox/labelEdit.module.less
ZaimBeshich1: deleted

packages/ketcher-react/src/script/ui/dialog/toolbox/labelEdit.module.less
ZaimBeshich1: deleted

packages/ketcher-react/src/script/ui/dialog/toolbox/labeledit.jsx
```diff
@@ -66,14 +67,28 @@ function deserialize(value) {
 
 function LabelEdit(props) {
   const init = { label: props.letter || serialize(props) }
-  const { formState, ...prop } = props
+  const { formState, onOk, ...prop } = props
```
AnaMoskvina, line 70: as we extract onOk prop here, it won't get into params prop in Dialog
it should affect functionality

ZaimBeshich1: Deleted unused prop

AnaMoskvina (approved):

AndreiMazol (approved):

Issue #1426, opened 2022-04-12
Floating windows - Change Atom properties window design


Alena-Guznischeva:
Implement changes for the design of Atom Properties window. 
What to change:
1. Use the color scheme, header, buttons, other standard components from the other forms
2. Apply alignment & pixel rules for the elements on the form. 
3. Buttons: add button Cancel, change label for OK button to Apply. Use the standard approach to place the buttons on the left in the buttons section. 
4. Validation and logic should stay the same. 
Current version of the form:

<img width="307" alt="image" src="https://user-images.githubusercontent.com/90251323/162982881-1f00158b-2784-4697-99ef-4266bc2f3639.png">

Alena-Guznischeva:
Update:
1. Redesign of the form is needed. Accordeon solution to implement with the General section expanded by default when the form is opened. Use the designs below
2. Tab General should be opened when the form is opened. If other tabs were expanded on the form and it was closed and opened repeatedly - the General tab should be opened, not the previously expanded tab. 

![image](https://user-images.githubusercontent.com/90251323/163825618-fbeb4d4d-e393-4e44-bc8c-085396c04a13.png)

4. Scroll should be applied for the all content inside the form, just like it is on the Template library form.

ansivgit:
Not only one tab can be opened at any point. Scroll appear
![image](https://user-images.githubusercontent.com/56927722/164047698-663ab343-ef08-402d-baee-41390da57972.png)

PR #1469, opened 2022-04-18
#1426 Floating windows - Change Atom properties window design

ViktorMakarov-EPAM:
Снимок экрана 2022-04-19 в 15 18 28

### DIFF
--- a/packages/ketcher-react/src/Editor.module.less
+++ b/packages/ketcher-react/src/Editor.module.less
@@ -119,6 +119,9 @@
     &[readonly],
     fieldset[disabled] & {
       cursor: not-allowed;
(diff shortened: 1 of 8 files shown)


AndreiMazol (approved):

Alena-Guznischeva:
Tab behaviour: General tab should be opened when the form is opened. Don't save the tab state when the form is closed and opened again.

ViktorMakarov-EPAM:
> Tab behaviour: General tab should be opened when the form is opened. Don't save the tab state when the form is closed and opened again.

it is.

AndreiMazol (approved):

PR #1522, opened 2022-04-26
correct map enum and values

fedorseliankin:


packages/ketcher-react/src/script/ui/utils/index.ts
```diff
@@ -71,7 +71,10 @@ export const getSelectOptionsFromSchema = (schema): Array<Option> => {
   return schema.enum.reduce((options, value, index) => {
     options.push({
       value,
-      label: schema?.enumNames?.[index] || value
+      label:
+        schema?.enumNames?.[index] !== undefined
+          ? schema?.enumNames?.[index]
+          : value
```
ViktorMakarov-EPAM, line 77: Can we change it to  `label: schema?.enumNames?.[index] ?? value`

ViktorMakarov-EPAM (approved):

Issue #1515, opened 2022-04-22
Text Editor design issues


ZhannaNovikova:
Case 1
**Steps to Reproduce**
1. Launch Ketcher
2. Open the Text Editor window clicking on the Text Tool icon on the left menu
3. Observe the format icons style and alignment

**Expected behavior**
The format icons (The B/I/ X<sup>2</sup> / X<sub>2</sub>) much the mockup 

![image](https://user-images.githubusercontent.com/103422815/164693833-8119215d-7545-4da4-b95e-7d9bc24e351e.png)

**Actual behavior**
The format icons (The B/I/ X<sup>2</sup> / X<sub>2</sub>) don't much the mockup 
![image](https://user-images.githubusercontent.com/103422815/164696598-b2810e9e-8d96-4bca-9902-8bfe48141ddf.png)

Case 2
**Steps to Reproduce**
1. Launch Ketcher
2. Open the Text Editor window clicking on the Text Tool icon on the left menu
3. Observe the arrow to open the Font Size drop down

**Expected behavior**
The arrow to open the Font Size drop down much the mockup 
![image](https://user-images.githubusercontent.com/103422815/164699467-ebe02657-5892-445a-8f3e-c547667e6dcd.png)

**Actual behavior**
The arrow to open the Font Size drop down doesn't much the mockup 
![image](https://user-images.githubusercontent.com/103422815/164699539-76f69a7a-ac63-4510-8f9a-cbbf5815bf28.png)

Case 3
**Steps to Reproduce**
1. Launch Ketcher
2. Open the Text Editor window clicking on the Text Tool icon on the left menu
3. Extend tte Symbols drop down

**Expected behavior**
The Symbols much the mockup 
![image](https://user-images.githubusercontent.com/103422815/164700891-e58a70c0-6ef9-47d1-af45-14153c764288.png)

**Actual behavior**
The Symbols don't much the mockup 
![image](https://user-images.githubusercontent.com/103422815/164700528-a6c0b193-2c21-46dd-ab21-3c824e678173.png)

Case 4
**Steps to Reproduce**
1. Launch Ketcher
2. Open the Text Editor window clicking on the Text Tool icon on the left menu
3. Enter text to make a scroll appear in the field

**Expected behavior**
The width between the text and the bottom border of the field matches the mockup -> 12px

**Actual behavior**
The width between the text and the bottom border of the field doesn't much the mockup 
[...]

PR #1529, opened 2022-04-26
#1515 Text editor design issues

ZaimBeshich1:
Problems in cases 4 and 5 will be solved in a separate tasks in 2.6


packages/ketcher-react/src/script/ui/views/modal/components/Text/SpecialSymbols/SpecialSymbolsButton.module.less
```diff
@@ -38,21 +39,21 @@
...
+    fill: @main-color;
     transition: 0.2s;
   }
 
   &:hover {
     background-image: none;
 
     & > svg {
-      fill: #000;
+      fill: @active-color;
```
AnaMoskvina, line 56: think, it should be #00838f here (which is @color-button-primary-hover), same as for buttons in toolbars

packages/ketcher-react/src/script/ui/views/modal/components/Text/TextButton/TextButton.module.less
```diff
@@ -14,28 +14,29 @@
...
+    fill: @main-color;
     transition: 0.2s;
   }
 
   &:hover {
     background-image: none;
 
     & > svg {
-      fill: @main-color;
+      fill: @active-color;
```
AnaMoskvina, line 39: same here

packages/ketcher-react/src/icons/index.tsx
```diff
@@ -149,8 +149,102 @@ import DropdownIndicatorIcon from './files/dropdown-indicator.svg'
...
+import π from './files/π.svg'
+import ρ from './files/ρ.svg'
+import σ from './files/σ.svg'
+import τ from './files/τ.svg'
+import υ from './files/υ.svg'
+import φ from './files/φ.svg'
+import χ from './files/χ.svg'
+import ψ from './files/ψ.svg'
+import ω from './files/ω.svg'
+import DegreeCelseus from './files/degree-celseus.svg'
```
AnaMoskvina, line 176: let's name icons according to convention, e.g. DegreeCelseusIcon
for symbols (not named icons) it's ok

AnaMoskvina (changes requested):

AnaMoskvina (approved):

Issue #1512, opened 2022-04-21
Text Tool issues


Svetlana-Asinovskaia:
1. Selected formatting is reset after clicking the text editing field.
![2022-04-21_22h57_11](https://user-images.githubusercontent.com/80695128/164542552-a062702d-96b9-4126-93a1-af74769b0800.gif)

2. Previous formatting is activated after changing the font size.
Steps to reproduce:
1) In the 'Text Editor' window select 'Bold'
2) Type some symbols in the text editing field.
3) Unselect 'Bold'
4) Change the font size
![2022-04-21_23h03_46](https://user-images.githubusercontent.com/80695128/164543711-3b204474-93dd-4d43-8711-1988b8e15964.gif)

3. Not selected text changes it's position when 'Superscript' is applied to the selected part of the text.
![2022-04-21_23h11_53](https://user-images.githubusercontent.com/80695128/164544729-bc2d4e8f-dde8-4456-8559-45c4ec3cdf6a.gif)

4. Drop-down lists in the 'Font Size' and 'symbols' fields do not close when clicking on the text edit field.
5. There is a space between characters when the subscript/superscript is applied to the beginning of the text.

<img width="201" alt="2022-04-22_13h01_43" src="https://user-images.githubusercontent.com/80695128/164684844-8722db4a-e940-4e45-8745-799174e5f9fa.png">

**Ketcher version** 
Ketcher v2.4.2-212-g4dee48be

ZhannaNovikova:
The cases: 1, 3, 5 - verified as fixed on Release 2.5.0

PR #1532, opened 2022-04-27
#1512 text tool issues

Mariia-Katkova:
fixed: 
-formatting reset after clicking the text editing field;
-not selected text changes it's position when 'Superscript'  is applied;
-space between characters when the subscript/superscript is applied;


packages/ketcher-react/src/script/ui/views/modal/components/Text/Text.tsx
```diff
@@ -136,17 +142,26 @@ const Text = (props: TextProps) => {
...
-      transformOrigin: 'left'
+      transformOrigin: 'left',
+      lineHeight: 0
     }
   }
 
+  const refEditor = useRef() as MutableRefObject<Editor>
+  const setFocusInEditor = refEditor.current
+    ? refEditor.current.focus
+    : () => console.log('refEditor not found')
```
ViktorMakarov-EPAM, line 158: Do we really want to show the message?

packages/ketcher-react/src/script/ui/views/modal/components/Text/Text.tsx
```diff
@@ -136,17 +142,26 @@ const Text = (props: TextProps) => {
...
+
+  // set focut on component mount
+  useEffect(() => {
+    setFocusInEditor()
+  })
```
ViktorMakarov-EPAM, line 163: As I can see we set focus on any updates. If you want to set focus on mount, you should pass empty array as second argument.

ZaimBeshich1 (approved):

packages/ketcher-react/src/script/ui/views/modal/components/Text/Text.tsx
Mariia-Katkova: removed console log

Issue #1525, opened 2022-04-26
Design issues of the Templates window


ZhannaNovikova:
Case 1
**Steps to Reproduce**
1. Steps to Reproduce
2. Launch Ketcher
3. Open Templates window
4. Hover on the "Search by element..." field

**Expected behavior**
- The "Search by element..." field border/border rounding, and color matches the mockup

**Actual behavior**
- The "Search by element..." field border/border rounding, and color doesn't match the mockup

![image](https://user-images.githubusercontent.com/103422815/165270372-3d150f3c-ebf0-4224-a6fe-9f096aa4820a.png)
![image](https://user-images.githubusercontent.com/103422815/165270513-257b79bb-8325-492f-b41f-2d93cf6eee61.png)
https://user-images.githubusercontent.com/103422815/165270629-07d8b168-bdf0-459f-8092-08376c240d43.mov

Case 2
**Steps to Reproduce**

1. Launch Ketcher
2. Open Templates window
3. Open alpha-D-Sugars tab of Templates window
4. Click on Edit button in right down corner of template (e.g. alpha-D-Arabinofuranose)
5. Hover on the "Molecule name" field

**Expected behavior**
- Template edit window size should not be changed after hovering on the "Molecule name" field
- The left down border of the Template edit window should have rounding
- Text area paddings should be 12px
- Margins: should be 12px for all labels

**Actual behavior**
- Template edit window doesn't match the mockup

https://user-images.githubusercontent.com/103422815/165274533-eaf73122-15e8-4171-bed4-bba9ef2618b8.mp4

![image](https://user-images.githubusercontent.com/103422815/165284703-2df8bc3a-c48e-4a2d-8edc-bb8f29b6acee.png)

Case 3
**Steps to Reproduce**

1. Launch Ketcher
2. Create structure on the canvas
3. Click on the Atom with the Fragment Selection Tool on the left menu
4. Click on the Atom with the S-Group Tool on the left menu
5. Select the Atom in the Contex drop down
6. Extend the "Field value" drop down

**Expected behavior**
The "Field value" drop down has the list values with scroll 

**Actual behavior**
The "Field value" drop down has the long list values (it's inconvenient).
![image](https://user-images.githubusercontent.com/103422815/165514330-c2984408-f8ed-464a-b9f2-83269960f6f2.png)

OS: [Windows NT 10.0 Win64; x64]
Browsers: Google Chrome: Version 100.0.4896.88/FireFox: Version 99.0.1 (64-bit)
[...]

PR #1534, opened 2022-04-27
#1525 fixed design issues in templates window

Mariia-Katkova:
- fixed search hover style and template window styles;
- added max-height for long lists in select dropdowns.


packages/ketcher-react/src/script/ui/component/form/Select/Select.module.less
```diff
@@ -20,6 +20,7 @@
 // stylelint-disable no-descending-specificity
 .dropdownList {
   :global(.MuiPaper-root) {
+    max-height: 200px;
```
ViktorMakarov-EPAM, line 23: I think we should use the same scroll which we have in Templates and Settings windows

AnaMoskvina (approved):

Issue #1519, opened 2022-04-25
Floating windows - Extended table window design


AnaMoskvina:
![Extended table](https://user-images.githubusercontent.com/77681522/165036049-69c10169-9e3e-4f5a-a70b-9cb12e600eba.PNG)

PR #1537, opened 2022-04-27
#1519 floating windows extended table window design

AnaMoskvina:
fixed_extended
periodic


packages/ketcher-react/src/script/ui/views/modal/components/ExtendedTable/ExtendedTable.jsx
```diff
@@ -0,0 +1,95 @@
...
+}
+
+function mapSelectionToProps(editor) {
+  const selection = editor.selection()
+
+  if (
+    selection &&
+    Object.keys(selection).length === 1 &&
+    selection.atoms &&
+    Object.keys(selection.atoms).length === 1
```
ViktorMakarov-EPAM, line 67: Can we replace it to if (selection?.atoms?.length === 1)

ZaimBeshich1 (approved):

ViktorMakarov-EPAM (approved):

Issue #1551, opened 2022-05-02
Blank page after applying the 'Shift + f' hotkey


Svetlana-Asinovskaia:
**Steps to Reproduce**
1. Launch Ketcher.
2. Apply the 'Shift + f' hotkey

**Expected behavior**
The 'Functional Groups' library is opened.

**Actual behavior**
A blank page is displayed. Errors appear in the console.

**Screenshots**
<img width="959" alt="2022-05-02_15h05_23" src="https://user-images.githubusercontent.com/80695128/166230851-391de5cf-6d42-4817-936c-fb09c92cb9c8.png">

**Desktop:**
 - OS: Windows 10
 - Browser: Chrome 
 - Version: 100.0.4896.127

**Ketcher version** 
- v2.4.2-219-g11bc5160

ZhannaNovikova:
Verified as fixed on Release 2.5.0.

PR #1565, opened 2022-05-05
#1551 Blank page after applying the 'Shift + f' hotkey 

ViktorMakarov-EPAM:


packages/ketcher-react/src/script/ui/dialog/template/TemplateDialog.tsx
```diff
@@ -145,7 +145,10 @@ const TemplateDialog: FC<Props> = (props) => {
     ...rest
   } = props
 
-  const [tab, setTab] = useState(TemplateTabs.TemplateLibrary)
+  const initialTab = useSelector(
+    (state: RootStateOrAny) => state.modal?.prop?.tab
```
AnaMoskvina, line 149: don't like RootStateOrAny type, can we not use it?

PR #1751, opened 2022-10-04
fix: remove addFragment()

porcelain11:
remove addFragment() (not implemented method) from ketcher.ts and docks


porcelain11:
NOTE:
Instead of **addFragment** one should use **setMolecule** method.

PR #1758, opened 2022-10-11
feat: implement addFragment method

porcelain11:
implement addFragment method


porcelain11:
Create separate PRs for addFragment implementation and refactoring

PR #1760, opened 2022-10-11
#0000 refactor variable names

porcelain11:
Refactor: variable names, change 'vars' for 'let' and 'const' where possible


porcelain11:
Don't remember, is it relevant or not, addFragment method is working now, the rest is not urgent, closing this PR

Issue #1762, opened 2022-10-12
File with text saved in .ket format won't open


Zhirnoff:
**Steps to Reproduce**
1. Open ketcher v.2.5.2
2. Write some text
3. Save in .ket format
4. Open saved file

**Expected behavior**
Text appeared on canvas

**Actual behavior**
Error message when open saved file

**Screenshots**
![2022-10-10_19h35_41](https://user-images.githubusercontent.com/62464241/195287777-55cb76bc-3a96-4936-bffe-b96b8b085275.png)
![2022-10-10_19h45_06](https://user-images.githubusercontent.com/62464241/195287787-0d71be40-817a-4d5e-8f08-2a7dc6932e5d.png)

**Desktop (please complete the following information):**
OS: Windows 10
Browser Chrome
Version 106.0.5249.103 (Official Build) (64-bit)
 

**Ketcher version** 
[Version 2.5.2](https://lifescience.opensource.epam.com/ketcher/index.html)
Build at 2022-10-07; 10:29:05

PR #1763, opened 2022-10-12
fix ket format not open issue

LadaYudovina:


packages/ketcher-core/src/domain/entities/text.ts
```diff
@@ -28,8 +28,16 @@ export enum TextCommand {
...
   // TODO: add Interface for content type
   content: string
-  position: Vec2
-  pos: Array<Vec2>
+  position: Point
+  pos: Array<Point>
+}
+
+function preparePositions(positions) {
+  if (!positions || positions === []) {
```
porcelain11, line 36: if (!positions || !positions.length) { ...

Screenshot 2022-10-13 at 13 59 11

packages/ketcher-core/src/domain/entities/text.ts
```diff
@@ -28,8 +28,16 @@ export enum TextCommand {
 export interface TextAttributes {
   // TODO: add Interface for content type
   content: string
-  position: Vec2
-  pos: Array<Vec2>
+  position: Point
+  pos: Array<Point>
+}
+
+function preparePositions(positions) {
```
porcelain11, line 35: Probably, it would be good to add types:

function preparePositions(positions:Array<Point>) {

porcelain11 (commented):
Please avoid using ' in the branch name :)
#1762-File_with_text_saved_in_.ket_format_**won't_**open

packages/ketcher-core/src/domain/entities/text.ts
LadaYudovina: Fix, thank you!

packages/ketcher-core/src/domain/entities/text.ts
LadaYudovina: Fix, thank you!

evgeniiFrolov1 (approved):

Issue #1694, opened 2022-08-16
Elements styles are overwriting other styles


MariaDolotova:
Hi! My name is Maria and I'm working on integrating Ketcher into cheminformatcs package used on Datagrok analytical platform.
And the problem is the following: in Ketcher some styles are set on elements like 'input'. 
And when Ketcher is built in Datagrok UI those styles affect all input elements.
![image](https://user-images.githubusercontent.com/44060410/184816710-0264e9cb-e12f-4656-85ff-3509388c3607.png)
Is it possible to use styles not on the elements them selfs but using classes?

skalkin:
@evgeniiFrolov1 , thanks for looking at it! Just FYI, this is the only issue that prevents us from using Ketcher as one of the supported sketchers for Datagrok

https://www.youtube.com/watch?v=k1NVdTRpYOM&ab_channel=Datagrok

PR #1767, opened 2022-10-13
Fix local `input` styles overriding other styles

yuleicul:
There were `input` selectors lying in the outer-most layer of a `.less` file, which led to overriding other styles when importing the `less` file.

After going through all `.less` files, I found just one file having that problem. I changed the tag selector to its corresponding class selector. I've checked styles that may be affected by my changes to make sure everything works fine.

* Update selector `input` to `.input`

Resolves: #1694

### DIFF
--- a/packages/ketcher-react/src/script/ui/component/form/input.jsx
+++ b/packages/ketcher-react/src/script/ui/component/form/input.jsx
@@ -34,7 +34,7 @@ export function GenericInput({
         value={value}
         onInput={onChange}
         onChange={onChange}
-        className={classes.input}
+        className={clsx(classes.input, classes.genericInput)}
         {...props}
       />
       {type === 'checkbox' && <span className={classes.checkbox} />}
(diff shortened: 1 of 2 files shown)


Nitvex:
I propose removing className here https://github.com/epam/ketcher/blob/master/packages/ketcher-react/src/script/ui/component/form/input.jsx#L129 as it is not used, but may have an effect if other application will use "radio" class. @yuleicul what do you think?

yuleicul:
> I propose removing className here https://github.com/epam/ketcher/blob/master/packages/ketcher-react/src/script/ui/component/form/input.jsx#L129 as it is not used, but may have an effect if other application will use "radio" class. @yuleicul what do you think?

Hi @Nitvex, thank you for bringing it up. I've checked `feildset` whose classname is `radio` in dialog Structure Check. Nothing happens after I remove `class="radio"`, so I think you're right, we can remove it.

Nitvex (approved):

Issue #1801, opened 2022-10-28
Refactor standalone struct service. Get rid of unnecessary Indigo-worker creation.


evgeniiFrolov1:
**Is your feature request related to a problem? Please describe.**
Indigo worker is terminated and recreated every single API call. 
This is the reason why API call to wasm module take extremely long time.

**Describe the solution you'd like**
The main idea of the issue is to create indigo worker one to avoid unnecessary wasm module sending and re-compilation.

PR #1803, opened 2022-10-31
#1801 Refactor standalone struct service. Get rid of unnecessary Indigo-worker creation.

Nitvex:


packages/ketcher-standalone/src/infrastructure/services/struct/standaloneStructService.ts
```diff
@@ -148,20 +150,45 @@ function mapWarningGroup(property: string) {
...
+  private readonly EE: EventEmitter = new EventEmitter()
 
   constructor(defaultOptions: StructServiceOptions) {
     this.defaultOptions = defaultOptions
+    this.worker = new IndigoWorker()
+    this.worker.onmessage = (e: MessageEvent<OutputMessage<string>>) => {
+      const message: OutputMessage<string> = e.data
+      if (message.type !== undefined) {
+        const event = messageTypeToEventMapping[message.type]
+        console.log(`emitted event ${event}`)
```
evgeniiFrolov1, line 182: Could you remove `console.log` pls?

packages/ketcher-standalone/src/infrastructure/services/struct/standaloneStructService.ts
```diff
@@ -460,7 +490,10 @@ class IndigoService implements StructService {
         data: commandData
       }
 
-      worker.postMessage(inputMessage)
+      this.EE.removeListener('calculateCip', action)
```
evgeniiFrolov1, line 493: What do you think about replacing string literal with `enum` 
for example :
```js
 this.EE.removeListener(ListenerType.CalculateCip, action)
```

evgeniiFrolov1 (changes requested):

packages/ketcher-standalone/src/infrastructure/services/struct/standaloneStructService.ts
Nitvex: Good proposal :) Will do

packages/ketcher-standalone/src/infrastructure/services/struct/standaloneStructService.ts
Nitvex: Okay, I thought that was intentional.

evgeniiFrolov1 (approved):

Issue #1822, opened 2022-11-09
When open template window search text should be selected as default


AlexanderSavelyev:
**Is your feature request related to a problem? Please describe.**
RIght now if one opens templates window, need to click to search text area to start search

**Describe the solution you'd like**
Focus automatically moved to search text area when open template window

PR #1851, opened 2022-11-16
#1822 – fixing Template-window input focus

ansivgit:
It works but we have errors in console
MicrosoftTeams-image

### DIFF
--- a/packages/ketcher-react/src/script/ui/component/form/input.jsx
+++ b/packages/ketcher-react/src/script/ui/component/form/input.jsx
@@ -102,7 +102,8 @@ function Select({
       value={value}
       name={name}
       multiple={multiple}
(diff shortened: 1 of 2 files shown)


packages/ketcher-react/src/script/ui/component/form/input.jsx
```diff
@@ -25,8 +25,14 @@ export function GenericInput({
   value = '',
   onChange,
   type = 'text',
+  isFocus = false,
```
KonstantinEpam23, line 28: Maybe the name should be "isFocused"? since its describing a particular state.

ansivgit: Thank you, changed

Nitvex (approved):

Issue #1852, opened 2022-11-17
Automatic selection of mol v2000 or mol v3000 encoding


Nitvex:
**Is your feature request related to a problem? Please describe.**
Introduce a single method to get a molfile from a structure. For now, calling `ketcher.getMolfile('v2000')` can not be applied for certain features such as enhanced stereochemistry. So you need to call the `ketcher.getMolfile('v3000')`. With this approach you need to know, which version of molfile is supposed to be used.
It is required to simplify determining versions of molfile and make this process automatic.

**Describe the solution you'd like**
It is proposed to make version of Molfile parameter optional. Calling `ketcher.getMolfile()` should return the result for v2000 if possible and if not for v3000.

Note: automatic version selection is already implemented on indigo side. It is required to set `molfile-saving-mode: 'auto'`.
For more information, refer https://lifescience.opensource.epam.com/indigo/options/input-output-options.html#molfile-saving-mode

Nitvex:
Files for testing:
[Archive.zip](https://github.com/epam/ketcher/files/10089815/Archive.zip)

There are two files: v2000 and v3000. You can open one of these files, then call `ketcher.getMolfile()` and structure is supposed to be saved to corresponding format.

Nitvex:
For reference. Situations, when v3000 format must be used:
* Structure highlighting – The V3000 format is required for molecule or collection highlighting.
* Enhanced stereochemistry features - The V3000 format is required when using the enhanced stereochemical representations.
* Long fields - If any of the fields with fixed widths for any connection table properties are exceeded, the V3000 format is used.For example, if the number of atoms (or bonds) exceeds 999. This is because the number of atoms (or bonds)on the V2000 counts line can not exceed 3 columns.
* Template block - useful for representing large structures, such as biological molecules.
* New properties supported only in V3000 format - such as atom CLASS and SEQID

PR #1870, opened 2022-11-28
#1852 Implements automatic selection of mol encoding

porcelain11:
Implements automatic selection of mol encoding: add "molAuto" to SupportedFormat type and to formatProperties - with { 'molfile-saving-mode': 'auto' } option.

### DIFF
--- a/packages/ketcher-core/src/application/ketcher.ts
+++ b/packages/ketcher-core/src/application/ketcher.ts
@@ -29,22 +29,6 @@ import assert from 'assert'
 import { EventEmitter } from 'events'
 import { runAsyncAction } from 'utilities'
 
-const Format = {
-  mol: 'mol',
-  molV3000: 'molV3000',
-  molAuto: 'molAuto',
-  rxn: 'rxn',
-  rxnV3000: 'rxnV3000',
-  smiles: 'smiles',
-  smilesExt: 'smilesExt',
-  smarts: 'smarts',
-  inChI: 'inChI',
-  inChIAuxInfo: 'inChIAuxInfo',
-  cml: 'cml',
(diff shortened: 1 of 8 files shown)


packages/ketcher-core/src/application/ketcher.ts
```diff
@@ -29,6 +29,22 @@ import assert from 'assert'
 import { EventEmitter } from 'events'
 import { runAsyncAction } from 'utilities'
 
+const Format = {
```
Nitvex, line 32: I propose to add type for this object. It can be something like:
```
const Format: {
  [key in SupportedFormat]: SupportedFormat
}
```
This way you don't need to write 
```
Format.SomeFormat as SupportedFormat
```

porcelain11: Fixed, pls check, if it's a good variant

packages/ketcher-react/src/script/ui/views/modal/components/process/Miew/Miew.tsx
```diff
@@ -117,7 +122,7 @@ const MiewDialog = ({
     (miew: MiewAsType) => {
       miewRef.current = miew
       const factory = new FormatterFactory(server)
-      const service = factory.create('cml')
+      const service = factory.create('cml' as SupportedFormat)
```
Nitvex, line 125: Could you please also change to `SupportedFormat.cml`?

porcelain11: Fixed

Nitvex (approved):

Issue #1826, opened 2022-11-09
Keyboard shortcut should not change current tool if cursor is over an atom


vanoprenko:
Currently when user uses a keyboard shortcut, this always changes current/selected tool, so that all further clicks on canvas will apply this tool (e.g. if 'F' is pressed, current tool will be set to Fluorine atom and applied on mouse clicks until another tool is selected via a button or a shortcut). 

This logic should be changed in case **mouse cursor is hovering over an atom**. If keyboard shortcut is used in such a case, it **should not change the current/selected tool**. Instead corresponding tool should be applied once while keeping current tool unchanged. E.g. if user has single bond tool selected, puts mouse cursor over an atom and presses 'I', Iodine tool should be applied once but single bond tool to be kept selected and applied on further mouse clicks.

PR #1877, opened 2022-11-30
#1826 dont select tools on hover

KonstantinEpam23:


packages/ketcher-react/src/script/ui/state/hotkeys.js
```diff
@@ -78,6 +97,13 @@ function keyHandle(dispatch, state, hotKeys, event) {
   }
 }
 
+function getHoveredAtomId(atoms) {
```
Nitvex, line 100: Propose to add types
```
function getHoveredAtomId(atoms: Atom[]): number | null {
...
}
```

KonstantinEpam23: its a JS file :(

porcelain11: > Maybe, you can rename it for .ts?

Nitvex (approved):

Issue #1827, opened 2022-11-09
Highlight currently selected tool with mouse cursor and toolbox icons


vanoprenko:
Currently when user selects a tool from left-side toolbox (e.g. double bond) by clicking toolbox button or using a keyboard shortcut, corresponding tool icon in the toolbox is highlighted. 

Tool highlighting behavior should be extended, so that if an atom tool is selected
1. Atom tool icon should be highlighted in the right-hand panel -- similarly to left-hand one.
2. Atom symbol should be displayed under mouse cursor while mouse hovers over the canvas (e.g. display arrow with 'F' if Fluorine is selected). 

![image](https://user-images.githubusercontent.com/18439849/201414904-a62c738b-768e-4652-8b07-9d1c99d1cff5.png)

PR #1883, opened 2022-12-02
#1827 currently selected tool with mouse cursor

ansivgit:
Add Atom hoverIcon when atom clicked on the right toolbar. You can see chosen atom icon when you put mouse on the canvas


packages/ketcher-react/src/script/editor/Editor.ts
```diff
@@ -200,6 +206,11 @@ class Editor implements KetcherEditor {
 
     const tool = new toolMap[name](this, opts)
 
+    // hide icon if not AtomToll chosen
```
Nitvex, line 209: Please, fix a typo `Toll` :)

packages/ketcher-react/src/script/editor/Editor.ts
```diff
@@ -109,6 +110,7 @@ class Editor implements KetcherEditor {
   historyPtr: any
   errorHandler: ((message: string) => void) | null
   highlights: Highlighter
+  hoverIcon: any
```
Nitvex, line 113: Can't we use more specific type?

packages/ketcher-react/src/script/editor/tool/atom.ts
```diff
@@ -129,7 +139,11 @@ class AtomTool {
 
   mousemove(event) {
     const rnd = this.editor.render
+    const { layerX, layerY } = event
+
     if (!this.dragCtx || !this.dragCtx.item) {
+      // magic numbers to adjust the position of the icon relative to the mouse pointer
+      this.editor.hoverIcon.attr({ x: layerX - 9, y: layerY - 7 })
```
Nitvex, line 146: Can't we measure it? :) 
May be functions like `getBBox()` or `getBoundingClientRect()`?

Nitvex (changes requested):

packages/ketcher-react/src/script/editor/Editor.ts
Nitvex: But this kind of comments we can replace with:
```
const isAtomToolChosen = name === 'atom'
if (!isAtomToolChosen) {
  ...
}
```
I think this makes the same sense.

packages/ketcher-react/src/script/editor/tool/atom.ts
KonstantinEpam23: or at least put these int constants?

Nitvex (approved):

KonstantinEpam23 (approved):

Issue #1823, opened 2022-11-09
When select a structure from template window it automatically starts to be added to canvas


AlexanderSavelyev:
**Is your feature request related to a problem? Please describe.**
User selects a structure from template window and after need to press button "Add to Canvas". 

**Describe the solution you'd like**
When user selects a structure, it automatically makes the add to canvas action

PR #1890, opened 2022-12-06
#1823 Add a template to the clipboard when selected

epam-nikolai-iakushchenko:
When select a structure from template window it automatically starts to be added to canvas


epam-nikolai-iakushchenko:
branch need to be renamed

PR #1892, opened 2022-12-07
#1823 Add a template to the clipboard when selected

epam-nikolai-iakushchenko:
When select a structure from template window it automatically starts to be added to canvas

### DIFF
--- a/packages/ketcher-react/src/script/ui/dialog/template/TemplateDialog.tsx
+++ b/packages/ketcher-react/src/script/ui/dialog/template/TemplateDialog.tsx
@@ -293,12 +293,12 @@ export default connect(
       return { ...template, modifiedStruct: struct }
     })
   }),
-  (dispatch: Dispatch<any>, props) => ({
+  (dispatch: Dispatch<any>, props: Props) => ({
     onFilter: (filter) => dispatch(changeFilter(filter)),
     onSelect: (tmpl) => {
       dispatch(selectTmpl(tmpl))
       dispatch(onAction({ tool: 'template', opts: tmpl }))
-      ;(props as any).onOk(tmpl)
+      props.onOk(tmpl)
     },
     onChangeGroup: (group) => dispatch(changeGroup(group)),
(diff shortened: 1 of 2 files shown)


packages/ketcher-react/src/script/ui/dialog/template/TemplateDialog.tsx
```diff
@@ -176,28 +176,15 @@ const TemplateDialog: FC<Props> = (props) => {
     )
   }
 
-  const result = (): Result | null => {
```
Nitvex, line 179: Can't we remove `Result` interface?

packages/ketcher-react/src/script/ui/dialog/template/TemplateDialog.tsx
```diff
@@ -318,13 +302,13 @@ export default connect(
   }),
   (dispatch: Dispatch<any>, props) => ({
```
Nitvex, line 303: propose to add a type for props
```
(dispatch: Dispatch<any>, props: Props) => ({
```
then you don't need to cast it to any 
```
props.onOk(tmpl)
```

Nitvex (changes requested):

packages/ketcher-react/src/script/ui/dialog/template/TemplateDialog.tsx
epam-nikolai-iakushchenko: 👍 [33de419](https://github.com/epam/ketcher/pull/1892/commits/33de419369cf7716a3b04f3155a58292cedce379)

packages/ketcher-react/src/script/ui/dialog/template/TemplateDialog.tsx
epam-nikolai-iakushchenko: sure, thanks for the catch
[61ba157](https://github.com/epam/ketcher/pull/1892/commits/61ba157cc3b9b426ed8171bd35fcd3b7a543a8f3)

Issue #1864, opened 2022-11-22
If the coordinates have a centroid that is far from zero, the molecule can be placed far offscreen.


Zhirnoff:
**Steps to Reproduce**
1. Open Ketcher
2. Open browser console
3. Paste into console: 
await ketcher.setMolecule(`                
  Mrv2118 11202219232D  
  
  0  0  0     0  0            999 V3000
M  V30 BEGIN CTAB
M  V30 COUNTS 8 8 0 0 0
M  V30 BEGIN ATOM
M  V30 1 C 51.8392 -54.8062 0 0
M  V30 2 C 51.8392 -53.2662 0 0
M  V30 3 N 53.1738 -55.5762 0 0
M  V30 4 C 54.5074 -54.8062 0 0
M  V30 5 C 54.5074 -53.2662 0 0 CFG=1
M  V30 6 O 53.1738 -52.4962 0 0
M  V30 7 N 55.841 -55.5762 0 0
M  V30 8 C 55.8411 -52.4962 0 0
M  V30 END ATOM
M  V30 BEGIN BOND
M  V30 1 1 1 2
M  V30 2 1 1 3
M  V30 3 1 6 2
M  V30 4 2 3 4
M  V30 5 1 4 5
M  V30 6 1 4 7
M  V30 7 1 5 6
M  V30 8 1 5 8 CFG=1
M  V30 END BOND
M  V30 BEGIN COLLECTION
M  V30 MDLV30/STEREL1 ATOMS=(1 5)
M  V30 END COLLECTION
M  V30 END CTAB
M  END
`)
4. Look at canvas

**Actual behavior**
The molecule placed far offscreen. It is necessary to scroll a long way in order to see it or make 'Zoom out'

**Expected behavior**
After 'setMolecule' is called, the coordinates are normalised and the molecule is displayed in the middle of the canvas.

**Screenshots**
![2022-11-22_17h01_45](https://user-images.githubusercontent.com/62464241/203335018-2d756408-7041-46cc-a5ab-2dace4771e26.png)

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser Chrome
 - Version 107.0.5304.107 (Official Build) (64-bit)

**Ketcher version**
[Version 2.6.2]
Build at 2022-11-17; 13:01:16

PR #1893, opened 2022-12-07
#1864 molecule placed offscreen issue

porcelain11:
Try to fix recoordinate and Render.setOffset - works, but there questions about other parts:

1. recoordinate (Editor.ts) didn't do anything with passed structCenter (probably, was not implemented) => updated with passing structCenter coords to setScrollOffset.

2. setScrollOffset operates with confused types of coords - it seems, that coords passed from structCenter ketcher understands as scaled (i.e., values > 50 - as in the issue sample - interpreted not as 50px), but setScrollOffset adds them up with clientArea.clientWidth & clientArea.clientHeight, which are in px - it's not clear, if is it done intentionally or not, seems not valid.

3. then setScrollOffset changes clientArea scroll positions to those passed to function - is case of issue sample, it doesn't help, as it sets scroll positions to ~50px, structure renders in scaled coords (=> changed: coords multiplied by options.scale).

4. Subsequent this.update removed from setScrollOffset, as it seems redundant - and, in the scope of the issue - it resets the scroll position (so the structure is left offscreen).


packages/ketcher-core/src/application/render/restruct/restruct.ts
```diff
@@ -296,9 +297,10 @@ class ReStruct {
   }
 
   clearVisel(visel: Visel): void {
-    visel.paths.forEach((path) => {
-      path.remove()
-    })
+    // TODO: do we need this? visel.clear() contains this.paths = []
```
KonstantinEpam23, line 300: Does just setting it to empty array remove it from canvas though?

porcelain11: Right, it doesn't, moved back

porcelain11:
Too many changes decided to make them gradually

Issue #1872, opened 2022-11-28
Add right-click menu for bonds


Nitvex:
It is required to add right-click menu for bonds to simplify user interaction. 
It shall contain the following items:
- Edit. Opens dialog which is currently opens on double click
- Bond type. Should contain vertical submenu with icons and names of bonds. Click on corresponding type changes bond type.
- Delete. Deletes bond

---
Nov 29, 2022 updated

And regarding the second point, it can look like: (green square is icon)
![MicrosoftTeams-image (3)](https://user-images.githubusercontent.com/27288153/204439765-bd23e376-5751-4a90-ac2b-9552423f82ec.png)

---
Dec 8, 2022 updated

Q: In what situation the context menu needs to show when you right click?
A: Right-click menu should always appear without any dependency on selected tool. A selected tool shouldn't reset, when right-click.

PR #1829, opened 2022-11-10
#1681 Upgrade to React 18 and replace react-contextmenu with react-contexify

TatianaGrigorovich:
Resolves: #1681 

(by @yuleicul) **NB:** Because this PR includes lots of code and it's not friendly for reviewers, I'll highlight the code which needs more attention in the following to ease the review work. Thanks for your review in advance❤. (Btw, I have reviewed React upgrade part)

---

### Updates _by @TatianaGrigorovich_ :
- Upgrade to React 18.2.0, includes
  * ketcher-react
  * ketcher-polymer-editor-react
  * demo
  * example
- Upgrade Jest and React Testing Library

----
### New updates _by @yuleicul_:

- Replace `react-contextmenu` with [react-contexify](https://github.com/fkhadra/react-contexify)
- Update functional group menu items
  - if all selected functional groups are unexpanded
    image
  - if all selected functional groups are expanded
    image
  - if some are expanded, others are unexpanded
    image

PR #1896, opened 2022-12-08
#1872 Add right-click menu for bonds

yuleicul:
Resolves: #1872

## 1. Feature description

#1872 

## 2. UI styles

Just follow styles of the existing context menu:

image

#### Scenario 1

Given: a molecule and any active tool
When: the user right-clicks on the bond
Then: the context menu shows

#### Scenario 2

Given: a molecule and any active tool
When: the user right-clicks on the atom(or any other non-bond things)
Then: the context menu doesn't show 

GIF of the general usage:
Ketcher-general-usage

#### Scenario 1

Given: the user has selected a bunch of things
When: the user right-clicks any selected bond
Then: the context menu shows

#### Scenario 2

Given: Scenario 1
When: the user clicks "Delete"
Then: All selected bonds are deleted (note that if left atoms become detached, they will be deleted too)

#### Scenario 3
Given: the user has selected a bunch of things
When: the user right-clicks a non-selected bond
Then: the context menu doesn't show

GIF of batch updates:
Ketcher-batch-updates

### 3.3 Handling conflict with the existing functional-group context menu

Note: the following scenarios are compatible with the existing logic process, but maybe will change after refactoring the functional-group context menu.

#### Scenario 1

Given: an expanded functional group
When: the user right-clicks a bond in the f-group
Then: only the functional-group context menu shows

#### Scenario 2
Given: the user has selected a bunch of things including functional groups
When: the user right-clicks a selected or non-selected bond
Then: only the functional-group context menu shows 

Ketcher-two-context-menus

## 4. My mind map before coding

https://excalidraw.com/#json=PtVEKcmFEDYpJ7MGCU7Ye,W21nPx7QUdIeCAAs_apUQw

## 5. Next steps
- [ ] #1899 
- [ ] Add context menu document to internal KB (Ketcher first step - Code structure) and `help.md`
- [ ] Add icons for top-level items (❓ what do icons look like)
- [ ] Cut/Copy + Paste for structures (Paste should work, when clicking anywhere)
- [ ] If you selected both atoms and bonds (for batch update), then there should be two menu items: Atom Properties and Bond Properties (🤔 depends on which one is right-clicked)
- [ ] Add a shortcut to the right of each menu item (🤔 some bonds don't have shortcuts)
- [ ] Add types to variable `closestItem` (I found the source file, packages/ketcher-react/src/script/editor/shared/closest.js, which generates ci, is written by JavaScript. In my experience, there are two ways to add types for it. One is to refactor the whole file with TypeScript, the other is to create a declaration file. The former will affect all functions in the file, the latter may only affect the function which we want to add types for)

After merging this PR and #1829:
- Reduce two context-menu triggers to one if possible, and make sure two context menus will not appear at the same time.
- Put all things related to the context menu into the folder where `StructEditor` is


packages/ketcher-react/src/script/ui/views/components/ContextMenu/ContextMenu.tsx
```diff
@@ -0,0 +1,182 @@
...
+import type { ItemParams, PredicateParams } from 'react-contexify'
+import { Item, Menu, Submenu } from 'react-contexify'
+import 'react-contexify/ReactContexify.css'
+import { useAppContext } from 'src/hooks'
+import Editor from 'src/script/editor'
+import tools from 'src/script/ui/action/tools'
+import Icon from 'src/script/ui/component/view/icon'
+import styles from './ContextMenu.module.less'
+
+export const CONTEXT_MENU_ID = 'KeTcHeR-CoNtExT-MeNu'
```
Nitvex, line 28: Just interesting, why are we using id with such case? :) 
can't it be `ketcher-context-menu`
or
`ketcherContextMenu`?

packages/ketcher-react/src/script/ui/views/components/ContextMenu/ContextMenu.tsx
```diff
@@ -0,0 +1,182 @@
...
+  const { getKetcherInstance } = useAppContext()
+
+  const handleEdit = useCallback(
+    async ({ props }: ItemParams<ContextMenuItemProps, ItemData>) => {
+      const editor = getKetcherInstance().editor as Editor
+      const bondId = props?.ci.id
+      const bond = editor.render.ctab.bonds.get(bondId)?.b
+
+      try {
+        const newBond = await editor.event.bondEdit.dispatch(bond)
```
Nitvex, line 47: What kind of error may occur?

packages/ketcher-react/src/script/ui/views/components/ContextMenu/ContextMenu.tsx
```diff
@@ -0,0 +1,182 @@
...
+  const isHidden = ({
+    props
+  }: PredicateParams<ContextMenuItemProps, ItemData>) => !!props?.selected
+  const isBatchOpHidden = ({
+    props
+  }: PredicateParams<ContextMenuItemProps, ItemData>) => !props?.selected
+
+  return (
+    <Menu id={CONTEXT_MENU_ID} animation={false} className={styles.contextMenu}>
+      {/** ****** Menu items for single update: ********/}
```
Nitvex, line 139: I propose to create two components for single/batch updates.
Won't it work if
```
<Menu id={CONTEXT_MENU_ID} animation={false} className={styles.contextMenu}>
  {isHidden && <SingleBondOperations />}
  {isBatchOpHidden && <BatchBondOperations />}
</Menu>
```

packages/ketcher-react/src/script/ui/views/components/ContextMenu/ContextMenu.tsx
```diff
@@ -0,0 +1,182 @@
...
+
+      editor.update(action)
+    },
+    [getKetcherInstance]
+  )
+
+  const isHidden = ({
+    props
+  }: PredicateParams<ContextMenuItemProps, ItemData>) => !!props?.selected
+  const isBatchOpHidden = ({
```
Nitvex, line 133: Could you use longer names? 
`isBatchOperationsHidden`?

packages/ketcher-react/src/script/ui/views/components/ContextMenu/ContextMenuTrigger.tsx
```diff
@@ -0,0 +1,111 @@
...
+    (event) => {
+      const editor = getKetcherInstance().editor as Editor
+      const ci = editor.findItem(event, ['bonds'])
+
+      if (!ci) {
+        hideAll()
+        return
+      }
+
+      // Resolve conflict with existing functional group context menu
```
Nitvex, line 40: Could you, please, describe in a few words what a conflict occurs?
Showing two menus at the same time?
Then I propose to move this part of code at separate function.

packages/ketcher-react/src/script/ui/views/components/StructEditor/StructEditor.jsx
```diff
@@ -200,33 +204,36 @@ class StructEditor extends Component {
         onMouseDown={(event) => event.preventDefault()}
         {...props}
       >
-        <ContextMenuTrigger
+        <FunctionalGroupTrigger
```
Nitvex, line 207: Hmm, I think the appropriate nesting would be:
```
<ContextMenuTrigger>
  <FunctionalGroupTrigger>
  </FunctionalGroupTrigger>
</ContextMenuTrigger>
```
Or will it break something?

packages/ketcher-react/src/script/ui/views/components/ContextMenu/ContextMenu.tsx
```diff
@@ -0,0 +1,182 @@
+/****************************************************************************
```
Nitvex, line 1: I propose to move this file to the same folder, where existing context menu is located (or vice versa):
image

Nitvex (changes requested):

packages/ketcher-react/src/script/ui/views/components/ContextMenu/ContextMenu.tsx
yuleicul: It can be `ketcher-context-menu` or `ketcherContextMenu`.
Just because I wanted it unique, in case there would be other context menus. But now I'm thinking if there are more context menus we'd better use enums to manage the IDs.

packages/ketcher-react/src/script/ui/views/components/ContextMenu/ContextMenu.tsx
yuleicul: image

If clicking the Cancel button, will throw an error `undefined`. What I did is catch it and `console.error` it.

I searched `bondEdit.dispatch(bond)` in the codebase to see how they handled it. I got this:
```js
catch(() => null) // w/o changes
``` 
So maybe it's better to ignore it as they did? What's your take on that

packages/ketcher-react/src/script/ui/views/components/ContextMenu/ContextMenuTrigger.tsx
yuleicul: Oh sorry, I described it in the PR description but I didn't put it in the code. I'll put the PR link into the code and move this part to a function as you suggested. Thank you for pointing it out.

packages/ketcher-react/src/script/ui/views/components/StructEditor/StructEditor.jsx
yuleicul: Unfortunately, it doesn't work if nesting like that. The `FunctionalGroupTrigger` is directly imported from the library `react-contextmenu`. I guess it captures the event `onContextMenu` and stops it from propagating.
Who nests who doesn't matter, because we are going to refactor `FunctionalGroupTrigger`, like combining it with `ContextMenuTrigger`, right?

packages/ketcher-react/src/script/ui/views/components/ContextMenu/ContextMenu.tsx
yuleicul: You are right I've noticed that. I recommend putting them in the folder where `StructEditor` is, because only `StructEditor` needs them. And I think it will be better to do this when refactoring the functional group context menu.

yuleicul:
Thanks for your review @Nitvex, I am going to create a new issue on refactoring the existing functional-group context menu. And list all the things I plan to do:

- [ ] Reduce two context-menu triggers to one
- [ ] Put all things related to the context menu into the folder where `StructEditor` is
- [ ] Refactor the existing functional-group context menu by `react-contexify`, and remove the library `react-contextmenu` from dependencies.

What's your take on this?

packages/ketcher-react/src/script/ui/views/components/ContextMenu/ContextMenu.tsx
Nitvex: Yes, you can do the same way. Or you can make it more explicit, for example:
```
const noOperation = () => {}
catch(noOperation)
```

Nitvex:
> Thanks for your review @Nitvex, I am going to create a new issue on refactoring the existing functional-group context menu. And list all the things I plan to do:
> 
> * [ ]  Reduce two context-menu triggers to one
> * [ ]  Put all things related to the context menu into the folder where `StructEditor` is
> * [ ]  Refactor the existing functional-group context menu by `react-contexify`, and uninstall the library `react-contextmenu` from dependencies (preparation for upgrading to React 18)
> 
> What's your take on this?

Yes, that's a good point. I agree with all. Small notes:
– Please, describe in a refactoring task a conflict between two menus and how to avoid a conflict.
– Actually, we already have a PR (https://github.com/epam/ketcher/pull/1829) for React 18 – and here react-contextmenu is replaced by MUI menu.

yuleicul:
> > Thanks for your review @Nitvex, I am going to create a new issue on refactoring the existing functional-group context menu. And list all the things I plan to do:
> > 
> > * [ ]   Reduce two context-menu triggers to one
> > * [ ]   Put all things related to the context menu into the folder where `StructEditor` is
> > * [ ]   Refactor the existing functional-group context menu by `react-contexify`, and uninstall the library `react-contextmenu` from dependencies (preparation for upgrading to React 18)
> > 
> > What's your take on this?
> 
> Yes, that's a good point. I agree with all. Small notes: – Please, describe in a refactoring task a conflict between two menus and how to avoid a conflict. – Actually, we already have a PR (#1829) for React 18 – and here react-contextmenu is replaced by MUI menu.

Oh, 😮 I didn't notice that we already have a PR (#1829) for React 18. Actually, all these refactoring things are based on replacing `react-contextmenu` with `react-contexify`.

As `react-contextmenu` has been replaced by MUI menu, I recommend all these refactoring things happen after merging the two PRs, otherwise, it's really hard to resolve the code conflicts.

packages/ketcher-react/src/script/ui/views/components/ContextMenu/ContextMenuTrigger.tsx
yuleicul: Here are the code comments: https://github.com/epam/ketcher/pull/1896/files#diff-65d4d0aa8707ab4ab7d93b7f42d0b60f211dbcfd6e5d7b06398fd43a8b1b68f3R31-R48

And I added a GIF to the PR description to show more details.

packages/ketcher-react/src/script/ui/views/components/ContextMenu/ContextMenu.tsx
Nitvex: Let's use enum 😄 
And in this case, I guess, we can name like `ketcherBondContextMenu`. So we use `ketcher` in order to avoid conflicts with other libraries/tools and `bond` to specify its purpose.

packages/ketcher-react/src/script/ui/views/components/ContextMenu/ContextMenuTrigger.tsx
```diff
@@ -0,0 +1,140 @@
...
+   *
+   * Scenario 2
+   * Given: the user has selected a bunch of things including functional groups
+   * When: the user right-clicks a selected or non-selected bond
+   * Then: only the functional-group context menu shows
+   *
+   * More details: https://github.com/epam/ketcher/pull/1896
+   */
+  const hasConflictWithFunctionalGroupMenu = useCallback(
+    (ci: any) => {
```
Nitvex, line 49: If that's possible, could you, please, add types?
My bet, you have found it somewhere and there were no types as well. So, I think, it would be great to add types there.
In addition, I propose to use more clear naming for `ci` => `closestItem`. It's clear, when you know it, but for newcomers it may be unclear.

packages/ketcher-react/src/script/ui/views/components/ContextMenu/ContextMenuTrigger.tsx
```diff
@@ -0,0 +1,140 @@
...
+        (item) => item.relatedSGroupId === functionalGroupId
+      )
+
+      if (functionalGroupId !== null && hasRelatedSGroup) {
+        return true
+      }
+
+      const selection = editor.selection()
+
+      if (selection && selection.atoms) {
```
Nitvex, line 68: Minor: `selection?.atoms`

packages/ketcher-react/src/script/ui/views/components/ContextMenu/ContextMenuTrigger.tsx
```diff
@@ -0,0 +1,140 @@
...
+          const functionalGroupId = FunctionalGroup.findFunctionalGroupByAtom(
+            struct.functionalGroups,
+            atomId
+          )
+          const hasRelatedSGroupId = struct.functionalGroups.some(
+            (item) => item.relatedSGroupId === functionalGroupId
+          )
+          return functionalGroupId !== null && hasRelatedSGroupId
+        })
+        if (hasSelectedFunctionalGroup) return true
```
Nitvex, line 79: I propose to use brackets even for single-line `if`.
(For me, for example, it is simpler to debug, because it is clear, when debugger steps into `if` body)

packages/ketcher-react/src/script/ui/views/components/ContextMenu/items/BondBatchOperations.tsx
```diff
@@ -0,0 +1,125 @@
...
+      <Submenu
+        label="Bond type"
+        hidden={isHidden}
+        className={styles.subMenu}
+        {...props}
+      >
+        {bondTools.map((name) => (
+          <Item id={name} onClick={handleBatchTypeChange} key={name}>
+            <Icon name={name} className={styles.icon} />
+            <span>{tools[name].title.slice(0, -5)}</span>
```
Nitvex, line 113: Thanks for adding this, but I think, it will be unclear, if we just remove `bond` part this way. I propose to create a separate function.

Nitvex (changes requested):

Nitvex (approved):

PR #1917, opened 2022-12-13
#1915 – improved search performance in template dialog

### LINKED ISSUE #1915: Improve search performance in Templates Dialog
Current implementation of search in Templates Dialog is slow, what negatively impacts user experience. This is noticeable, when opening a tab with a lot of structures, for example, 'Salts and Solvents'.
It is required to speed up search and remove significant delays.

Nitvex:
Improved search performance in template dialog by caching rendering result.
resolves #1915

### DIFF
--- a/packages/ketcher-react/src/script/ui/component/structrender.tsx
+++ b/packages/ketcher-react/src/script/ui/component/structrender.tsx
@@ -30,12 +30,22 @@ function prepareStruct(struct: Struct) {
   return struct
 }
 
+/**
+ * Is used to improve search and opening tab performance in Template Dialog
+ * Rendering a lot of structures causes great delay
+ */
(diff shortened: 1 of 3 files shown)


packages/ketcher-react/src/script/ui/component/structrender.tsx
```diff
@@ -65,7 +72,7 @@ class StructRender extends Component<StructRenderProps> {
   }
 
   shouldComponentUpdate(previousProps) {
-    return this.props.id && this.props.id !== previousProps.id
+    return !!(this.props.id && this.props.id !== previousProps.id)
```
KonstantinEpam23, line 75: maybe instead of `!!(...)` we can use `Boolean(...)`?

Issue #1963, opened 2022-12-20
Custom Templates: When switching between tabs, the focus is not active on the search bar


Zhirnoff:
**Steps to Reproduce**
1. Launch Ketcher
2. Open 'Custom Templates'
3. Switch between tabs 

**Actual behavior**
When switching between tabs, the focus is not active on the search bar

**Expected behavior**
When switching between tabs, the focus is active on the search bar

**Screenshots**

https://user-images.githubusercontent.com/62464241/208676933-de22ea96-43cd-41e0-ab9b-b76934a66ae2.mp4

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser Chrome
 - Version 108.0.5359.125 (Official Build) (64-bit)

**Ketcher version** 
[Version 2.7.0-rc.3](https://lifescience.opensource.epam.com/ketcher/index.html)

**Additional context**

Zhirnoff:
Bug is reproduced in [Version 2.8.0-rc.2](https://lifescience.opensource.epam.com/ketcher/index.html)
Build at 2023-01-31; 09:01:03

https://user-images.githubusercontent.com/62464241/216364382-be64de28-7752-4c07-9bb4-cf61299c8743.mp4

PR #1980, opened 2022-12-22
#1963 Custom Templates: When switching between tabs, the focus is not active on the search bar

porcelain11:
Create an additional state member in TemplateDialog, which is changing after tab changed, and pass it as a param to input, and then - as a dependency to useEffect, setting focus.


packages/ketcher-react/src/script/ui/component/form/input.jsx
```diff
@@ -26,12 +26,15 @@ export function GenericInput({
...
   type = 'text',
   isFocused = false,
+  searchFocusBack = false,
   ...props
 }) {
   const inputRef = useRef(null)
   useEffect(() => {
-    if (inputRef.current && isFocused) inputRef.current.focus()
-  }, [inputRef, isFocused])
+    if (inputRef.current && isFocused) {
```
Nitvex, line 34: Can we do the same logic in tabs, not in GenericInput? It belongs to tabs changing, but not for input itself.
I see the solution the following way in tabs component:
```
const inputRef = useRef(null)
...
const onTabChange = () => {
  ... 
  inputRef.current.focus();
}
...
return (
 ...
 <GenericInput ref={inputRef} />
 ...
)
```

porcelain11: Fixed

KonstantinEpam23:
@porcelain11 hmm, this should be against master, not release branch, the corresponding task is on 2.8.0 milestone

packages/ketcher-react/src/script/ui/dialog/template/TemplateDialog.tsx
```diff
@@ -172,6 +174,10 @@ const TemplateDialog: FC<Props> = (props) => {
     setFilteredFG(filterFGLib(functionalGroups, filter)[FUNCTIONAL_GROUPS])
   }, [functionalGroups, filter])
 
+  useEffect(() => {
+    searchInputRef.current?.focus()
```
Nitvex, line 178: May be we can simply move this to `handleTabChange`?

Nitvex (changes requested):

packages/ketcher-core/src/application/formatters/identifyStructFormat.ts
```diff
@@ -77,7 +77,10 @@ export function identifyStructFormat(
     return SupportedFormat.inChI
   }
 
-  if (sanitizedString.indexOf('\n') === -1) {
+  if (
+    sanitizedString.indexOf('\n') === -1 &&
+    sanitizedString === sanitizedString.toUpperCase()
```
Nitvex, line 82: It shouldn't be here. This check was removed

packages/ketcher-core/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "ketcher-core",
-  "version": "2.6.4",
+  "version": "2.7.0-rc.4",
```
Nitvex, line 3: Are you sure, we need this? 
In master we still have 2.6.4, in release there is a 2.7.0-rc.8

Nitvex (changes requested):

PR #2010, opened 2022-12-27
#1934 hot key hot key 1 turns sulfur into carbon when cursor over the atom

### LINKED ISSUE #1934: Hot key: Hot key "1" turns sulfur into carbon when cursor over the atom
Bond hotkeys ("1", "2", etc.) are applied incorrectly when mouse cursor is over an atom. It is expected that corresponding type of bond is attached to the atom under cursor while previously selected tool is kept intact. Instead pressing bond hotkey leads to atom replacement with carbon.

_Example:_
Create a single isolated "S" atom, move mouse cursor over it and hit "1". It is expected that single bond is attached to the sulfur atom. Instead sulfur (H2S) turns into carbon (CH4).
 
Similar problems appear for hotkeys associated with other tools from the left toolbar.

**NOTE:** Hotkey "1" is currently linked to a few bond types and allows switching between them. When "1" is pressed over an atom, single bond should always be applied.

Nitvex:
Added shortcuts handling for all tools. Previously it was implemented only for atoms.
Moved `sgroupDialog` as a static member of `SGroup` class.
Changed shortcut for 'hand' tool because it concedes with MacOS 'Cmd+H', which just hides the application window.

Commented types produce a lot of TS error, which are required to be fixed in future.

closes #1934


packages/ketcher-react/src/script/ui/state/handleHotkeysOverAtom.ts
```diff
@@ -0,0 +1,214 @@
...
+    },
+    hand: () =>
+      dispatch(
+        onAction({
+          tool: 'hand'
+        })
+      )
+  }
+  const toolHandler = toolsMapping[newAction.tool]
+  const changeStructureTool =
```
KonstantinEpam23, line 52: isChangeStructureTool ? (just to have them consistent)

packages/ketcher-react/src/script/editor/tool/sgroup.ts
```diff
@@ -143,7 +143,11 @@ class SGroupTool {
         return
       }
 
-      sgroupDialog(this.editor, id !== undefined ? id : null, this.type)
+      SGroupTool.sgroupDialog(
+        this.editor,
+        id !== undefined ? id : null,
```
KonstantinEpam23, line 148: id: id ?? null

KonstantinEpam23 (approved):

PR #2034, opened 2022-12-29
#1963 Custom Templates  when switching between tabs focus is not active on search bar

porcelain11:
Forward ref from TemplateDialog  - trough Input component - to GenericInput, use it to preserve focus on search input when tab changed (rename input.jsx to .tsx, add types)

### DIFF
--- a/packages/ketcher-react/src/script/ui/component/form/input.tsx
+++ b/packages/ketcher-react/src/script/ui/component/form/input.tsx
@@ -197,7 +197,7 @@ FieldSet.val = function (ev, schema) {
   const inputCollection = fieldset?.querySelectorAll('input')
   let result
 
-  if (inputCollection && inputCollection.length) {
+  if (inputCollection?.length) {
     result = Array.from(inputCollection).reduce(
       (res, inp: HTMLInputElement, i) =>
         !inp.checked ? res : [enumSchema(schema, i), ...res],
--- a/packages/ketcher-react/src/script/ui/component/form/input.tsx
+++ b/packages/ketcher-react/src/script/ui/component/form/input.tsx
@@ -14,23 +14,46 @@
  * limitations under the License.
  ***************************************************************************/
 
-import { Component, useRef, useEffect } from 'react'
+import React, { PureComponent, ComponentType, useRef, useEffect } from 'react'
 
-import { omit } from 'lodash/fp'
 import classes from './input.module.less'
 import clsx from 'clsx'
+
+type Props = {
+  component?: ComponentType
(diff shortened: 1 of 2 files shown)


packages/ketcher-react/src/script/ui/component/form/input.tsx
```diff
@@ -275,47 +311,62 @@ function ctrlMap(component, { schema, multiple, onChange }) {
...
-function shallowCompare(a, b) {
-  for (const key in a) {
-    if (!(key in b)) return true
-  }
-  for (const key in b) {
-    if (a[key] !== b[key]) return true
+class Input extends PureComponent<
+  Props & { innerRef: React.Ref<HTMLInputElement> }
+> {
+  component: any
```
epam-nikolai-iakushchenko, line 337:
```suggestion
  component: React.Component
```

porcelain11: Unfortunately, it's not that simple =( I'll try once again to set proper type

Nitvex (approved):

packages/ketcher-react/src/script/ui/component/form/input.tsx
```diff
@@ -158,19 +183,29 @@ function FieldSet({
...
+  const fieldset = input?.parentNode?.parentNode?.parentNode
+  const inputCollection = fieldset?.querySelectorAll('input')
+  let result
+
+  if (inputCollection && inputCollection.length) {
+    result = Array.from(inputCollection).reduce(
+      (res, inp: HTMLInputElement, i) =>
+        !inp.checked ? res : [enumSchema(schema, i), ...res],
+
+      [] as HTMLInputElement[]
```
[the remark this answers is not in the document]
porcelain11: Not sure it is a good idea - inside this function, we will need to check if it is an array of inputs or options - so that using  'checked' or 'selected' param, and cast types somehow, I think it will look a bit ugly and wouldn't make things clearer.

packages/ketcher-react/src/script/ui/component/form/input.tsx
porcelain11: There is an issue with **schema** here - it probably, needs some refactoring - type for component should look smth like this:
ComponentType<{}> | typeof TextArea | typeof GenericInput | typeof Slider | typeof FieldSet | typeof Select
but for some of these ^^ components schema is required, but for class Input itself it is optional (so if we set type like this, we get an error  - for inner component with ref - that schema is missing
Propose to leave it like this for subsequent refactoring

Issue #1999, opened 2022-12-26
S-Group: Hover over the S-Group displays always acetaldehyde


Zhirnoff:
**Steps to Reproduce**
1. Launch Ketcher
2. Draw a structure
3. Add an atom data S-Group
4. Add a label and value
5. Press 'Apply'
6. Hover over the S-Group

**Actual behavior**
Hover over the S-Group displays always acetaldehyde

**Expected behavior**
Hover over the S-Group displays a preview related to the represented atom

**Screenshots**

https://user-images.githubusercontent.com/62464241/209544644-f3411f6c-df1b-4edb-a668-7a93c1355c37.mp4

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser Chrome
 - Version 108.0.5359.125 (Official Build) (64-bit)

**Ketcher version** 
[Version 2.7.0-rc.6](https://lifescience.opensource.epam.com/ketcher/index.html)
Build at 2022-12-23; 13:33:39

**Additional context**

PR #2037, opened 2022-12-29
#1999 S-group: hover over the s group displays always acetaldehyde

porcelain11:
Fix getFunctionalGroupByName - so that it doesn't return first functionalGroup for empty searchName;

Show name=value pair on hover for 'DAT' sGroup (and nothing - for other sGroup types):
inside InfoPanel - check if groupStruct is null and there is sGroup and sGroup.type is 'DAT' - render div with sGroup name and value.

### DIFF
--- a/packages/ketcher-react/src/script/ui/views/components/StructEditor/InfoPanel.tsx
+++ b/packages/ketcher-react/src/script/ui/views/components/StructEditor/InfoPanel.tsx
@@ -102,8 +102,17 @@ const InfoPanel: FC<InfoPanelProps> = (props) => {
     sGroup
   } = props
   const [molecule, setMolecule] = useState<Struct | null>(null)
+  const [sGroupData, setSGroupData] = useState<string | null>(null)
   const childRef = useRef(null)
   const groupName = sGroup?.data?.name
+
+  useEffect(() => {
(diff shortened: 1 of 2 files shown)


packages/ketcher-react/src/script/ui/views/components/StructEditor/InfoPanel.tsx
```diff
@@ -102,9 +102,18 @@ const InfoPanel: FC<InfoPanelProps> = (props) => {
     sGroup
   } = props
   const [molecule, setMolecule] = useState<Struct | null>(null)
+  const [sGroupData, setSGroupData] = useState<string | null>(null)
   const childRef = useRef(null)
   const groupName = sGroup?.data?.name
 
+  useEffect(() => {
+    if (!groupStruct && sGroup && sGroup.type === 'DAT') {
```
KonstantinEpam23, line 110: `sGroup && sGroup.type === 'DAT'` -> `sGroup?.type === 'DAT'`

porcelain11: Fixed

KonstantinEpam23 (approved):

Issue #2035, opened 2022-12-29
Changing the logic of interaction with the 'Selection tool' hotkey


Zhirnoff:
**Is your feature request related to a problem? Please describe.**
The current hotkey for the 'Selection tool' is an escape button (ESC), which can mislead the user. Logically, by pressing escape the user will try to reset the used tool, but instead selections are made between: Rectangle, Lasso, Fragment. Also if the user's mouse is over an atom of the structure, then when pressing the hotkey (ESC) for the 'SelectionTool', the structure atom highlighted and selected.

https://user-images.githubusercontent.com/62464241/209958776-a6ec6a11-e203-49c0-ac01-320aff97fc05.mp4

**Describe the solution you'd like**
Change the hotkey (ESC) for selection between: Rectangle, Lasso, Fragment to another one that is not misleading and leave the structure selections only on mouse click. If the user's mouse is over an atom or a bond of the structure, then when pressing the hot key for the 'SelectionTool', the structure should not be highlighted and selected. Hotkey should only change the select options between: Rectangle, Lasso, Fragment

**Describe alternatives you've considered**

**Additional context**

PR #2056, opened 2023-01-05
#2035 Changing the logic of interaction with the 'Selection tool' hotkey

St-Permiakov:
Changed behaviour of "Select" tool hotkey: now it ignores hovering status and acts consistently in all cases.
Solves #2035


packages/ketcher-react/src/script/ui/state/hotkeys.ts
```diff
@@ -76,7 +76,7 @@ function keyHandle(dispatch, state, hotKeys, event) {
       const newAction = actions[actName].action
       const hoveredItemId = getHoveredAtomId(render.ctab.atoms)
       const isHoveringOverAtom = hoveredItemId !== null
-      if (isHoveringOverAtom) {
+      if (isHoveringOverAtom && newAction.tool !== 'select') {
```
Nitvex, line 79: Could you, please, also remove logic for select tool from `handleHotkeyOverAtom`?

Nitvex (approved):

Issue #2043, opened 2022-12-30
Thumbnail images are not readable in Template Dialog


Nitvex:
**Steps to Reproduce**
1. Open template dialog
2. Switch to "Functional Groups" tab
3. Observe some large structure

**Actual behavior**
Some structures are not fully readable due to small font size

**Expected behavior**
It is required to increase font size for elements to make structures readable

**Screenshots**
Example of unreadable structure:
![image](https://user-images.githubusercontent.com/41272128/210068069-302a353f-5a05-4c29-bb37-02967cb58b08.png)

porcelain11:
Screenshots for different font sizes and bond thicknesses are in the attachment
[screenshots.zip](https://github.com/epam/ketcher/files/10360107/screenshots.zip)

PR #2063, opened 2023-01-06
#2043 thumbnail images are not readable in template dialog

porcelain11:
Change font size, bondThickness and doubleBondWidth separately for Salts&Salvents and rest templates (**main changes are  in the last but one commit** )
Rename some files to .tsx and add types (tried to fix the issue with ForwardedRef type )


packages/ketcher-react/src/script/ui/component/form/MeasureInput/measure-input.tsx
```diff
@@ -92,8 +92,12 @@ function convertValue(value, measureFrom, measureTo) {
...
 
   return measureTo === 'px' || measureTo === 'pt'
-    ? ((value * measureMap[measureFrom]) / measureMap[measureTo]).toFixed() - 0
-    : ((value * measureMap[measureFrom]) / measureMap[measureTo]).toFixed(3) - 0
+    ? Number(
+        ((value * measureMap[measureFrom]) / measureMap[measureTo]).toFixed()
+      ) - 0
+    : Number(
+        ((value * measureMap[measureFrom]) / measureMap[measureTo]).toFixed(3)
+      ) - 0
```
KonstantinEpam23, line 100: Looks like these two lines only differ in that the second one uses "toFixed(3)" - maybe we could put this into a separate function? e.g. `getFormatedValue` or something like that?

packages/ketcher-react/src/script/ui/dialog/template/TemplateTable.tsx
```diff
@@ -71,7 +71,14 @@ const RenderTmpl: FC<{
   return (
     <StructRender
       struct={tmpl.struct}
-      options={{ ...options, autoScaleMargin: 15 }}
+      options={{
+        ...options,
+        autoScaleMargin: 15,
+        fontsz: 55,
```
KonstantinEpam23, line 77: could we put these into constants, since it looks like these are re-used in several files?

packages/ketcher-react/src/script/ui/component/form/MeasureInput/measure-input.tsx
porcelain11: Done

packages/ketcher-react/src/script/ui/dialog/template/TemplateTable.tsx
porcelain11: Done

KonstantinEpam23:
@porcelain11 I went in a slightly different direction with this: https://github.com/epam/ketcher/pull/2098
perhaps we will close this PR (@porcelain11 lets discuss when you are available)

PR #2071, opened 2023-01-09
#2060: Don't break the selection if the user's cursor goes beyond the canvas

### LINKED ISSUE #2060: Don't break the selection if the user's cursor goes beyond the canvas
Select the 'Selection Tool', then try to select a structure, which is near to the edge of the canvas. At the current moment, if a users cursor will appear out of canvas, then 'Selection Tool' is stopped and selection is interrupted.
It is required not to interrupt selection tool, when user moved cursor beyond the canvas.

St-Permiakov:
Now selection tool does not break when cursor leaves canvas.
Fixes #2060

### DIFF
--- a/packages/ketcher-react/src/script/editor/Editor.ts
+++ b/packages/ketcher-react/src/script/editor/Editor.ts
@@ -616,17 +616,18 @@ function updateLastCursorPosition(editor: Editor, event) {
 function domEventSetup(editor: Editor, clientArea) {
   // TODO: addEventListener('resize', ...);
   ;[
-    'click',
-    'dblclick',
-    'mousedown',
-    'mousemove',
-    'mouseup',
-    'mouseleave',
-    'mouseover'
-  ].forEach((eventName) => {
+    { target: clientArea, eventName: 'click' },
+    { target: clientArea, eventName: 'dblclick' },
+    { target: clientArea, eventName: 'mousedown' },
+    { target: document, eventName: 'mousemove' },
+    { target: document, eventName: 'mouseup' },
+    { target: document, eventName: 'mouseleave' },
+    { target: clientArea, eventName: 'mouseover' }
+  ].forEach(({ target, eventName }) => {
     editor.event[eventName] = new DOMSubscription()
     const subs = editor.event[eventName]
-    clientArea.addEventListener(eventName, subs.dispatch.bind(subs))
+
+    target.addEventListener(eventName, subs.dispatch.bind(subs))
 
     subs.add((event) => {
       updateLastCursorPosition(editor, event)
--- a/packages/ketcher-react/src/script/editor/Editor.ts
+++ b/packages/ketcher-react/src/script/editor/Editor.ts
@@ -598,7 +598,7 @@ function isMouseRight(event) {
 }
 
 function resetSelectionOnCanvasClick(editor: Editor, eventName: string) {
-  if (eventName === 'mouseup') {
+  if (eventName === 'mouseup' && editor.selection()) {
     editor.selection(null)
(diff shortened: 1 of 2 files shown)


packages/ketcher-react/src/script/editor/Editor.ts
```diff
@@ -577,17 +577,18 @@ function isMouseRight(event) {
...
   ;[
-    'click',
-    'dblclick',
-    'mousedown',
-    'mousemove',
-    'mouseup',
-    'mouseleave',
-    'mouseover'
-  ].forEach((eventName) => {
+    { target: clientArea, eventName: 'click' },
```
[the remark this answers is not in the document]
St-Permiakov: Thanks for the idea! If it is to be extracted as a standalone variable, that definitely yes. Otherwise I don't see a benefit: it is not reused, it is not a prop, and the type is easily inferred.

What I would do is move all eventNames to a enum, but it should involve refactoring multiple locations and I think deserve a separate PR.

KonstantinEpam23 (approved):

Nitvex (changes requested):
I've checked this PR locally and found that tools, which have multiple options (for example, selections and bonds) don't work from first time, if click on left panel.
You need to perform at least, two clicks in order to open, for example, panel for selection tool:
image
Could you, please, have a look?

Nitvex (approved):

packages/ketcher-react/src/script/editor/Editor.ts
```diff
@@ -616,17 +616,18 @@ function updateLastCursorPosition(editor: Editor, event) {
...
+    { target: clientArea, eventName: 'dblclick' },
+    { target: clientArea, eventName: 'mousedown' },
+    { target: document, eventName: 'mousemove' },
+    { target: document, eventName: 'mouseup' },
+    { target: document, eventName: 'mouseleave' },
```
yuleicul, line 624: Hi 😃 seems it causes a little bug: pasted structure is here to stay when the mouse is moving outside the canvas.

**Steps to Reproduce**
1. Cut or copy a structure
2. `Ctrl+v` to paste it on the canvas
3. Move the mouse to the toolbar

**Actual behavior**
The pasted structure is still here
ketcher-paste-bug-07


**Expected behavior**
The pasted structure disappears. The following GIF is created on v2.7.0 RC test remote version.
ketcher-paste-bug-08

St-Permiakov: Do we have explicit requirements to remove it when over a toolbar?
Why should we use controversial logic for different tools? We do not break selection when going over toolbars, why should we break copy-paste action?

yuleicul: You are right. Actually, sometimes I'm a bit confused about some controversial logic. Thanks for your fast reply :)

yuleicul: And please check out this one. The hand tool is still following the cursor when moving outside the canvas.

ketcher-paste-bug-09



I'm not sure if it's caused by this line of code, but I guess it's because the `mouseleave` event is not triggered. The following code in `ketcher-react/src/script/ui/views/components/StructEditor/StructEditor.jsx` may help.

```jsx
case 'leave':
  this.editorRef.current.classList.remove(classes.enableCursor)
  this.setState({
    enableCursor: false
  })
  break
```

St-Permiakov: Yeah, that does look weird. Thanks for noticing.
I think it deserves its own issue on the board. I'll look into it later.

PR #2077, opened 2023-01-10
#1906 drawing bonds in one direction does not change the bond created in the other direction

### LINKED ISSUE #1906: Drawing bonds in one direction does not change the bond created in the other direction
**Steps to Reproduce**
1. Create two atoms
2. Link the first atom with the second with a single bond
3. Link the second atom with the first with a double bond

**Actual behavior**
Ketcher displays a triple bond

**Expected behavior**
The second bond replaces the first bond, so result must be double bond (in provided example)

**Screenshots**
Two atoms attached with a single bond:
image
Then, linking from NH2 to H3C results in:
image

2.6.4

Nitvex:
closes #1906

### DIFF
--- a/packages/ketcher-react/src/script/editor/tool/bond.ts
+++ b/packages/ketcher-react/src/script/editor/tool/bond.ts
@@ -23,12 +23,10 @@ import {
   FunctionalGroup,
   SGroup,
   fromOneBondDeletion
-  // BondDelete
 } from 'ketcher-core'
 
 import utils from '../shared/utils'
 import Editor from '../Editor'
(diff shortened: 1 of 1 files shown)


packages/ketcher-react/src/script/editor/tool/bond.ts
```diff
@@ -246,9 +252,16 @@ class BondTool {
             beginPos,
             endPos
           )[0]
+          if (existingBondId) {
+            this.dragCtx.existedBond = bond
```
KonstantinEpam23, line 256: existedBond -> existingBond ?

KonstantinEpam23 (approved):

PR #2087, opened 2023-01-12
#2085: Selection is reset when using context menu

### LINKED ISSUE #2085: Selection is reset when using context menu
**Steps to Reproduce**
1. Draw any structure
2. Select multiple atoms and bonds
3. Right-click on any selected atom

**Actual behavior**
Selection is reset. There is no possibility to perform context-menu action.

**Expected behavior**
Selection is not reset. User can use right-click menu in order to perform actions.

St-Permiakov:
Improved tracking of out-of-canvas mouse events
Resolves #2085


packages/ketcher-react/src/script/editor/Editor.ts
```diff
@@ -613,7 +622,30 @@ function updateLastCursorPosition(editor: Editor, event) {
   }
 }
 
-function domEventSetup(editor: Editor, clientArea) {
+function toolIsUsed(
```
Nitvex, line 625: One small proposal – may be it is worth to rename a function to `useToolIfNeeded`?
And then
```
const isToolUsed = useToolIfNeeded(...)
if (isToolUsed) {
  return true
}
```

Not 100% sure it is better, but at the moment the function `toolIsUsed` doesn't only check the condition (as the name implies), but also has side effect – it applies the tool

Nitvex (approved):

Nitvex (approved):

PR #2089, opened 2023-01-12
#2047: Templates: When the structure is pasted or dragged on each other, carbon atoms are highlighted

### LINKED ISSUE #2047: Templates: When the structure is pasted or dragged on each other, carbon atoms are highlighted
**Steps to Reproduce**
1. Launch Ketcher
2. Select 'Benzene' from Templates
3. Paste it on canvas
4. Select 'Benzene' from Templates and paste over 'Benzene'
5. Place two 'Benzene' on the canvas and drag one onto the other

**Actual behavior**
When the structure is pasted or dragged on each other, carbon atoms are wrongly highlighted as "bad valence".

**Expected behavior**
Fully overlapping identical templates using Paste should result in merging without change.
Fully overlapping identical templates using using Dragging-drop should result in merging without change.

**Screenshots**

https://user-images.githubusercontent.com/62464241/210354700-5c1443f5-ffc3-40a7-87f3-cf53cdfa0e21.mp4

 - OS: Windows 10
 - Browser Chrome
 - Version 108.0.5359.125 (Official Build) (64-bit)

**Ketcher version** 
[Version 2.7.0-rc.8](https://lifescience.opensource.epam.com/ketcher/index.html)

**Additional context**

St-Permiakov:
Fixes a big issue with valence validation upon merging structures.
Previously validation was performed atom by atom alongside other modifications, which resulted in miscalculations when taking into account atoms that were queued for deletion.
Now it is performed separately, after all merges complete, and takes into account all connected atoms in case structure is changed.

Resolves #2047

### DIFF
--- a/packages/ketcher-core/src/application/editor/actions/closelyFusing.ts
+++ b/packages/ketcher-core/src/application/editor/actions/closelyFusing.ts
@@ -15,7 +15,7 @@
  ***************************************************************************/
 
 import { Action } from './action'
-import { fromAtomMerge } from './atom'
+import { checkAtomValence, fromAtomMerge } from './atom'
 import { fromBondsMerge } from './bond'
 import utils from '../shared/utils'
 
@@ -25,6 +25,12 @@ export function fromItemsFuse(restruct, items) {
   if (!items) return action
 
   const usedAtoms = new Set()
+
+  const connectedAtomIds = getAllConnectedAtomsIds(
+    restruct,
+    mergeMapOfAtomsToSet(items.atoms),
+    mergeMapOfAtomsToSet(items.bonds)
+  )
 
   // merge single atoms
   items.atoms.forEach((dst, src) => {
@@ -36,6 +42,8 @@ export function fromItemsFuse(restruct, items) {
 
   // merge bonds
   action = fromBondsMerge(restruct, items.bonds).mergeWith(action)
+
+  action = valenceCheck(restruct, connectedAtomIds).mergeWith(action)
 
   return action
 }
@@ -97,4 +105,53 @@ function closestToMerge(struct, closestMap) {
   if (mergeMap.atoms.size === 0 && mergeMap.bonds.size === 0) return null
 
   return mergeMap
+}
+
+function mergeMapOfAtomsToSet(items: Map<number, number>): Set<number> {
+  const itemsSet = new Set<number>()
+  items.forEach((value, key) => {
+    itemsSet.add(value).add(key)
+  })
+  return itemsSet
+}
+
+function getAllConnectedAtomsIds(restruct, atomsIds, bondsIds) {
+  const initialAtoms = new Set(atomsIds)
+  const connectedAtoms = new Set()
+
+  for (const bondId of bondsIds) {
+    const bond = restruct.bonds.get(bondId)
+    if (bond) {
+      const { begin, end } = bond.b
+      initialAtoms.add(begin).add(end)
+    }
+  }
+
+  for (const initialAtom of initialAtoms) {
+    if (connectedAtoms.has(initialAtom)) continue
+    const relevantConnectedComponent = [
+      ...restruct.connectedComponents.values()
+    ].find((component) => component.has(initialAtom))
+    if (relevantConnectedComponent)
+      relevantConnectedComponent.forEach((id) => connectedAtoms.add(id))
+  }
+
+  return connectedAtoms
+}
+
+function valenceCheck(restruct, atomIds) {
+  let action = new Action()
+
+  if (!atomIds) return action
+  const usedAtoms = new Set()
+
+  // merge single atoms
+  atomIds.forEach((atomId) => {
+    if (usedAtoms.has(atomId)) return
+
+    action = checkAtomValence(restruct, atomId).mergeWith(action)
+    usedAtoms.add(atomId)
+  })
+
+  return action
 }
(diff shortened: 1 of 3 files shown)


packages/ketcher-core/src/application/editor/actions/closelyFusing.ts
```diff
@@ -98,3 +106,52 @@ function closestToMerge(struct, closestMap) {
 
   return mergeMap
 }
+
+function mergeItemsFuseMap(items: Map<number, number>): Set<number> {
```
epam-nikolai-iakushchenko, line 110: This function name is not clear to me.
Does this function filter out duplicated items?
Then maybe filterOutDuplicatedItems?
Thoughts?

packages/ketcher-core/src/application/editor/actions/closelyFusing.ts
```diff
@@ -98,3 +106,52 @@ function closestToMerge(struct, closestMap) {
...
+  const itemsSet = new Set<number>()
+  items.forEach((value, key) => {
+    itemsSet.add(value).add(key)
+  })
+  return itemsSet
+}
+
+function getAllConnectedAtomsIds(restruct, atomsIds, bondsIds) {
+  const srcAtoms = new Set(atomsIds)
```
epam-nikolai-iakushchenko, line 119: srcAtoms? atoms from the source? 🤔

St-Permiakov: Source atoms. The atoms we are beginning with. I couldn't come up with a better naming. Any ideas?

packages/ketcher-core/src/application/editor/actions/closelyFusing.ts
St-Permiakov: Well, it merges items into one array. Thus — mergeItems.
I can name it "mergeItemsIntoOneArrayFuseMap" if it helps.
Can't get rid of "FuseMap" part because it's highly niсhe and not supposed to be used anywhere else.

packages/ketcher-core/src/application/editor/actions/closelyFusing.ts
epam-nikolai-iakushchenko: why not just 'atoms'?

packages/ketcher-core/src/application/editor/actions/closelyFusing.ts
epam-nikolai-iakushchenko: can't say 'mergeItemsIntoOneSetFuseMap' looks easier to understand from my point of view 🤔 
Can you please explain what this function does if it doesn't filter out duplicated items?
And why is this 'fuseMap' thing needed? What does it mean?

St-Permiakov: > Can you please explain what this function does if it doesn't filter out duplicated items?
->
> ...it merges items into one Set.

It sure does remove any duplicated items, but it's not the point of this function.

> And why is this 'fuseMap' thing needed?
->
> ...because it's highly niсhe and not supposed to be used anywhere else.

The wording of this part can be tweaked, just not removed entirely, so the whole sense would not be lost.

packages/ketcher-core/src/application/editor/actions/closelyFusing.ts
St-Permiakov: Because the result of the function is also "atoms". So basically we can name result "atoms" as well and we'll get function making atoms from atoms.
In order to differentiate them in some way we divide them into "source atoms" — the ones we are using for the process — and the resulting "connected atoms", which are the goal of the process.

packages/ketcher-core/src/application/editor/actions/closelyFusing.ts
epam-nikolai-iakushchenko: > It sure does remove any duplicated items, but it's not the point of this function.
> The wording of this part can be tweaked, just not removed entirely, so the whole sense would not be lost.

What is the point of this function then? Sorry, I struggle to grasp the whole sense of this function 🤷

packages/ketcher-core/src/application/editor/actions/closelyFusing.ts
epam-nikolai-iakushchenko: Isn't it 

```
() => {
 const atoms;
 ?
 return  ̶p̶r̶o̶f̶i̶t̶ connectedAtoms
}
```

?

St-Permiakov: This way we have no clue about the purpose of this variable.
Are those target atoms? Are they going to be modified? Returned? What are we going to do with them?
Defining them as "source atoms" clearly states that it's an input to some process and they are not going to be used in any other way.

TimSPb89: maybe `initialAtoms` would be more straightforward then, what do you think?

St-Permiakov: Do not see the difference, but ok, will rename it if this helps.

packages/ketcher-core/src/application/editor/actions/closelyFusing.ts
St-Permiakov: Clarified in direct messages. 
And renamed a little to make the process more transparent.

Nitvex (approved):

KonstantinEpam23 (approved):

PR #2102, opened 2023-01-17
#2091 Using paste tool (or new template tool) - clearing the canvas causes console errors

### LINKED ISSUE #2091: Using paste tool (or new template tool) - clearing the canvas causes console errors
**Steps to Reproduce**
1. Open browser console
2. Paste a molecule
3. Click on "clear canvas"

**Actual behavior**
Multiple type errors appear in the console

**Expected behavior**
Errors should not appear in the console

**Screenshots**
212008181-5f3116c8-be19-468e-907a-cc6cd95f8acd

**Ketcher version** 
v2.8.0

**Additional context**

ansivgit:
fixed #2091

### DIFF
--- a/packages/ketcher-core/src/application/editor/operations/sgroup/SGroupAttr.ts
+++ b/packages/ketcher-core/src/application/editor/operations/sgroup/SGroupAttr.ts
@@ -36,10 +36,12 @@ export class SGroupAttr extends BaseOperation {
 
   execute(restruct: ReStruct) {
     const struct = restruct.molecule
-    const sgroupId = this.data.sgid
-    const sgroup = struct.sgroups.get(sgroupId)
+    const sgroupId = this.data.sgid!
+    const sgroup = struct.sgroups.get(sgroupId)!
 
-    if (!sgroup) return
+    if (!sgroup) {
+      return
+    }
 
     const sgroupData = restruct.sgroupData.get(sgroupId)
     if (sgroup.type === 'DAT' && sgroupData) {
--- a/packages/ketcher-core/src/application/editor/operations/sgroup/SGroupAttr.ts
+++ b/packages/ketcher-core/src/application/editor/operations/sgroup/SGroupAttr.ts
@@ -37,7 +37,9 @@ export class SGroupAttr extends BaseOperation {
   execute(restruct: ReStruct) {
     const struct = restruct.molecule
     const sgroupId = this.data.sgid
-    const sgroup = struct.sgroups.get(sgroupId)!
+    const sgroup = struct.sgroups.get(sgroupId)
+
+    if (!sgroup) return
 
     const sgroupData = restruct.sgroupData.get(sgroupId)
     if (sgroup.type === 'DAT' && sgroupData) {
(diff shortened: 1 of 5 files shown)


packages/ketcher-core/src/domain/entities/sgroupForest.ts
```diff
@@ -142,11 +142,18 @@ export class SGroupForest {
   }
 
   remove(id) {
-    assert(this.parent.has(id), 'sgid is not in the forest')
-    assert(this.children.has(id), 'sgid is not in the forest')
+    // assert(this.parent.has(id), 'sgid is not in the forest')
```
ansivgit, line 145: Do we really need assert module here or we can change it with `if(...) return`?

packages/ketcher-core/src/application/editor/operations/sgroup/SGroupAttr.ts
```diff
@@ -37,7 +37,9 @@ export class SGroupAttr extends BaseOperation {
   execute(restruct: ReStruct) {
     const struct = restruct.molecule
     const sgroupId = this.data.sgid
-    const sgroup = struct.sgroups.get(sgroupId)!
+    const sgroup = struct.sgroups.get(sgroupId)
+
+    if (!sgroup) return
```
Nitvex, line 42: I propose don't use single line `if`s

packages/ketcher-core/src/domain/entities/sgroupForest.ts
```diff
@@ -142,11 +142,18 @@ export class SGroupForest {
   }
 
   remove(id) {
-    assert(this.parent.has(id), 'sgid is not in the forest')
-    assert(this.children.has(id), 'sgid is not in the forest')
+    try {
+      assert(this.parent.has(id), 'sgid is not in the forest')
```
Nitvex, line 146: Are you sure `assert` throws an error? I thought it just prints error into console.
Won't it be simpler
```
if (!this.parent.has(id) && !this.children.has(id)) {
  console.info('error: sgid is not in the forest')
}
```

packages/ketcher-core/src/domain/entities/sgroup.ts
```diff
@@ -296,6 +296,8 @@ export class SGroup {
   }
 
   static removeAtom(sgroup: SGroup, aid: number): void {
+    if (!sgroup) return
```
Nitvex, line 299: Hmmm, not sure if it is a good place to do that. I mean a function is quite simple and I think, if it is not provided first argument, then higher function in stack didn't provide it – and that is the real problem.

KonstantinEpam23 (approved):

packages/ketcher-react/src/script/editor/Editor.ts
```diff
@@ -629,6 +629,8 @@ function useToolIfNeeded(
   event
 ) {
   const EditorTool = editor.tool()
+  // console.log(EditorTool)
```
KonstantinEpam23, line 632: lets remove this

packages/ketcher-core/src/application/editor/operations/sgroup/SGroupAttr.ts
KonstantinEpam23: I agree, there is a bigger issue, but it looks like it may relate to the way paste tool implements following the cursor. Since we will be overhauling that behaviour - I believe we can get back to this. I will add a separate task for that and link it to this one.

PR #2118, opened 2023-01-20
#2112: The hand tool is still following the cursor when moving outside the canvas

### LINKED ISSUE #2112: The hand tool is still following the cursor when moving outside the canvas
The hand tool is still following the cursor when moving outside the canvas.

ketcher-paste-bug-09

I'm not sure if it's caused by that line of code, but I guess it's because the `mouseleave` event is not triggered. The following code in `ketcher-react/src/script/ui/views/components/StructEditor/StructEditor.jsx` may help.

```jsx
case 'leave':
  this.editorRef.current.classList.remove(classes.enableCursor)
  this.setState({
    enableCursor: false
  })
  break
```

_Originally posted by @yuleicul in https://github.com/epam/ketcher/pull/2071#discussion_r1082176201_

St-Permiakov:
Resolves #2112

### DIFF
--- a/packages/ketcher-react/src/script/editor/Editor.ts
+++ b/packages/ketcher-react/src/script/editor/Editor.ts
@@ -654,8 +654,13 @@ function domEventSetup(editor: Editor, clientArea: HTMLElement) {
     { target: document, eventName: 'mousemove' },
     { target: document, eventName: 'mouseup' },
     { target: document, eventName: 'mouseleave' },
+    {
+      target: clientArea,
+      eventName: 'mouseleave',
+      toolEventName: 'mouseLeaveClientArea'
+    },
     { target: clientArea, eventName: 'mouseover' }
-  ].forEach(({ target, eventName }) => {
+  ].forEach(({ target, eventName, toolEventName }) => {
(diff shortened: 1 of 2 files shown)


Nitvex (approved):

packages/ketcher-react/src/script/editor/Editor.ts
```diff
@@ -654,8 +654,13 @@ function domEventSetup(editor: Editor, clientArea: HTMLElement) {
     { target: document, eventName: 'mousemove' },
     { target: document, eventName: 'mouseup' },
     { target: document, eventName: 'mouseleave' },
+    {
+      target: clientArea,
+      eventName: 'mouseleave',
+      toolEventName: 'mouseLeaveClientArea'
+    },
```
KonstantinEpam23, line 661: Maybe we should add a type for these after all? :) Or at least an enum?

St-Permiakov: We do. But I still stand for the need of a separate task for that. Especially now, when it got even more complicated.

KonstantinEpam23 (approved):

PR #2120, opened 2023-01-20
#1889 add ability to flip a part of structure

### LINKED ISSUE #1889: Add ability to flip a part of structure
It is required to have an ability to flip a part of structure horizontally and vertically. The flip function should be available for selected fragment of a structure only if the fragment is connected with the rest of the structure via one and only one bond.

# Connecting bond is not selected
If the bond connecting selected fragment with the rest of the structure is not included into selection, the fragment should be flipped around its horizontal or vertical center axis and shifted on the canvas, so that connecting atom in the selected fragment keeps its position: 

image

# Connecting bond is selected
If the connecting bond is included into selection, the entire selection (including bond) should be flipped around horizontal or vertical axis crossing the point when the bond connects to the rest of the structure.

image

Nitvex:
closes #1889

### DIFF
--- a/packages/ketcher-core/src/application/editor/actions/rotate.ts
+++ b/packages/ketcher-core/src/application/editor/actions/rotate.ts
@@ -63,9 +63,114 @@ export function fromFlip(restruct, selection, dir, center) {
   })
 
   if (typeof isFragFound === 'number') {
-    return action // empty action
+    return flipPartOfStructure({
+      fids,
+      struct,
+      restruct,
+      dir,
+      action,
+      selection
+    })
   }
 
+  return flipStandaloneStructure({
+    fids,
+    struct,
+    restruct,
+    center,
+    dir,
+    action,
+    selection
+  })
+}
+
+function getRotationPoint(struct, selection) {
+  const { bonds } = struct
+  const isAttachmentBond = ({ begin, end }) =>
+    (selection.atoms.includes(begin) && !selection.atoms.includes(end)) ||
+    (selection.atoms.includes(end) && !selection.atoms.includes(begin))
+  const isSelectedAtom = (atomId) => selection.atoms.includes(atomId)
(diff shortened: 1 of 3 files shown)


packages/ketcher-core/src/application/editor/actions/rotate.ts
```diff
@@ -63,30 +63,59 @@ export function fromFlip(restruct, selection, dir, center) {
...
-    fragment.forEach((aid) => {
-      const atom = struct.atoms.get(aid)
-      const d = flipItemByCenter(atom, calcCenter, dir)
-      action.addOp(new AtomMove(aid, d))
-    })
+function getRotationPoint(struct, selection) {
+  const { bonds } = struct
+  const isAttachmentBond = ({ begin, end }) =>
+    (selection.atoms.includes(begin) && !selection.atoms.includes(end)) ||
+    (selection.atoms.includes(end) && !selection.atoms.includes(begin))
```
KonstantinEpam23, line 91: Can we put these into separate variables?
e.g. `const atomsHaveBeginningBond = selection.atoms.includes(begin)`

packages/ketcher-core/src/application/editor/actions/rotate.ts
```diff
@@ -63,30 +63,59 @@ export function fromFlip(restruct, selection, dir, center) {
...
+  const isSelectedAtom = (atomId) => selection.atoms.includes(atomId)
```
KonstantinEpam23, line 92: Maybe lets pass the "selection" here as an argument?

Nitvex: Yes, we can, but it doesn't mean too much outside this function. The purpose of this function is just code clarity and reduction of code repetition.

KonstantinEpam23 (approved):

Issue #2055, opened 2023-01-05
Hand tool: Unable to scroll canvas down and to the right


Zhirnoff:
**Steps to Reproduce**
1. Launch Ketcher
2. Create few structures on canvas
3. Try to scroll canvas down and to the right using Hand tool

**Actual behavior**
Unable to scroll canvas down and to the right

**Expected behavior**
Able to scroll canvas down and to the right

**Screenshots**

https://user-images.githubusercontent.com/62464241/210779961-23dc5979-872d-4ef8-9d7a-60b714ac31d2.mp4

![2023-01-05_15h28_57](https://user-images.githubusercontent.com/62464241/210780710-051ad5c6-3696-4144-8e08-77e762b46921.png)

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser Chrome
 - Version 108.0.5359.125 (Official Build) (64-bit)

**Ketcher version** 
[Version 2.7.0-rc.8](https://lifescience.opensource.epam.com/ketcher/index.html)

**Additional context**
Note: It is necessary to restrict the user from endless scrolling the canvas. The restriction must take effect when the last pixel of the structure disappears from the visibility area. Now it is implemented so that you can scroll endlessly and then the user can hardly find his structures on the canvas.

Zhirnoff:
Bug is not reproduced in 2.17.0-rc.3

PR #2121, opened 2023-01-23
#2055 Hand tool: Unable to scroll canvas down and to the right

epam-nikolai-iakushchenko:
closes: #2055 

WIP

TODO:

- [ ] Fix 'Unable to scroll canvas down and to the right'

- [ ] restrict the user from endless scrolling the canvas. The restriction must take effect when the last pixel of the structure disappears from the visibility area. Now it is implemented so that you can scroll endlessly and then the user can hardly find his structures on the canvas.


St-Permiakov:
Just a note: it still happens only with a second try.

Issue #2031, opened 2022-12-29
Update ci/cd and other infrastructure-related scripts


KonstantinEpam23:
**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**

Issue #2030, opened 2022-12-29
Update all documentation throughout the project


KonstantinEpam23:
**Describe the solution you'd like**
Replace all documentation that mentions yarn with correct npm alternatives.
This includes all "readme" and .md files, wiki and any other places where yarn may be mentioned.
Make sure the documented npm commands work correctly.

Also, if applicable - update lifescience portal documentation: https://github.com/epam/lifescience

epam-nikolai-iakushchenko:
Looks like this page needs to be updated
https://kb.epam.com/display/EPMLSOP/Ketcher+Release+Preparation#

Issue #2008, opened 2022-12-27
Move from yarn to npm


KonstantinEpam23:
**Is your feature request related to a problem? Please describe.**
At the moment the project uses yarn, which seems to be problematic sometimes , for example we had problems with compiling different versions: https://github.com/epam/ketcher/issues/1970 Yarn seems to be too aggressive with its caching and .lock file handling when working with the workspaces feature.
npm is the standard tool for node, has larger userbase and more stable releases. Many developers are more familiar with it. It also supports some features out of the box that yarn requires plugins for.

We would  like to switch to using npm as it an official tool for node and JS and .

**Describe the solution you'd like**
Replace yarn with npm

Steps that need to be taken:
-   completely replacing yarn with npm commands throughout the project (this includes package.json scripts, husky integrations, etc)
-   update documentation: update all documentation throughout the project, including readme files
-   announce changes to repo
-   update ci/cd and other infrastructure-related scripts that may contain yarn-specific commans
-   update node to at least verion 14 to be able to use npm version 6 workspace features
-   updateupdate github ci

**Additional context**
It is also important to note that such changes will need to be announced to the community, since many peoples workflows may depend on yarn. Probably this should be done within a major version release.

**Related issues**
#2031 #2030 #2029

KonstantinEpam23:
Related ticket: https://github.com/epam/ketcher/pull/1740
It needs to be re-evaluated after this task is complete

PR #2126, opened 2023-01-25
#2029: Completely replace yarn with npm commands throughout the project

### LINKED ISSUE #2029: Completely replace yarn with npm commands throughout the project
**Describe the solution you'd like**
Completely replace yarn with npm commands throughout the project package.json files and any other script files that may rely on yarn.
This includes package.json scripts in all ./packages/* directories and in example apps:
```
./package.json
./packages/ketcher-react/package.json
./packages/ketcher-core/package.json
./packages/ketcher-standalone/package.json
./example/package.json
```
Each package and example need to be able to be built separately with npm and also together at the same time when build script is run from the project root.

Also, at the moment there are also some commands that are yarn-specific, e.g. the project is currently built via yarn foreach plugin: `yarn workspaces foreach -ptvR --from ketcher-react-app run build`
For such commands either npm alternatives or replacements need to be found.

In addition, yarn.lock file should be replaced with package-lock.json and husky scripts should be updated for local development.

As a result - the project should build correctly using only npm for both local development and deployment.

KonstantinEpam23:
closes #2029 #2030 #2031 #2008


demo/package.json
```diff
@@ -8,9 +8,9 @@
     "@testing-library/user-event": "^13.5.0",
     "@types/jest": "^27.4.1",
     "@types/node": "^16.11.26",
-    "ketcher-core": "workspace:*",
-    "ketcher-react": "workspace:*",
-    "ketcher-standalone": "workspace:*",
```
KonstantinEpam23, line 13: This is yarn-specific syntax, that is not supported by npm cli

packages/ketcher-react/src/constants.ts
```diff
@@ -17,3 +17,5 @@
 export const KETCHER_INIT_EVENT_NAME = 'ketcher-init'
 
 export const KETCHER_SAVED_SETTINGS_KEY = 'ketcher_editor_saved_settings'
+
+export const TEST = 1
```
Nitvex, line 21: Guess, we need to remove it?

DEVNOTES.md
```diff
@@ -1,9 +1,41 @@
...
+npm install
+```
+*NOTE!* this command should only be run from root directory
+
+Build the project:
+```
+npm run build
+```
+
 ## Prerequisites
```
Nitvex, line 35: I propose to move `Prerequisites` section to the top of the file

AnastasiiaPlyako:
I think we also can update read.me files in packages/ketcher-core etc

packages/ketcher-react/src/constants.ts
KonstantinEpam23: Ah, good catch! :) Will remove

KonstantinEpam23:
> I think we also can update read.me files in packages/ketcher-core etc

Well, for people who use our packages as npm modules - we don't really care if they use npm or yarn, they should work for both. These changes are mainly for people who work with the repository directly (like us 😄 )

KonstantinEpam23:
Hello @papillot , thank you checking this PR and pointing this out! I believe you are correct about the reason that the error appears b/c the library is not part of the non-dev builds.
Currently ketcher is mostly oriented on using its separate packages ([ketcher-core](https://www.npmjs.com/package/ketcher-core), [ketcher-react](https://www.npmjs.com/package/ketcher-react), etc) and building your own app (e.g. using the /example or /demo directories from the repository)

In the future we may be looking to avoiding using husky for non-dev builds.
Hopefully this clarifies things a bit, thank you for your input again!

PR #2140, opened 2023-01-30
#2093 combine s group menus and remove generic s group

### LINKED ISSUE #2093: Combine s-group menus and remove generic s-group
**Background**
The terms and navigation are not intuitive if you're not intimately familiar with all of the terms. Consider ways to either update the terms or provide more detailed help menus.

**Solution**
* Keep only one single s-group menu. Use the icon from "S-group" on the left
* Remove S-Group type Generic
* Add Data S-Group type
* The following types should be available with the order: [Data, Multiple group, SRU polymer, Superatom]
* If user selects "Data" type then data s-group menu should be added with additional fields (similar to the current "Data s-group"): Context, Field name, Field value, Absolute, Relative, Attached. For all other types fields should be the same as currently

yuleicul:
Resolves #2093 

### Before:

image

### After:

image

More details: https://excalidraw.com/#json=akG_vHaNa4J29lFtRzbSY,eFq4Q1LGavcpuj25tG3Mtw


packages/ketcher-react/src/script/ui/dialog/toolbox/SDataFieldset.jsx
```diff
@@ -0,0 +1,74 @@
...
+  )
+}
+
+const content = (schema, context, fieldName, fieldValue, checked) =>
+  Object.keys(schema.properties)
+    .filter(
+      (prop) => prop !== 'type' && prop !== 'context' && prop !== 'fieldName'
+    )
+    .map((prop) => {
+      return prop === 'radiobuttons' ? (
```
Nitvex, line 51: I propose to don't use two nested ternary operators.

packages/ketcher-react/src/script/ui/dialog/toolbox/SGroupFieldset.jsx
```diff
@@ -0,0 +1,58 @@
...
+      {content(type)}
+    </fieldset>
+  )
+}
+
+const content = (type) =>
+  Object.keys(schemes[type].properties)
+    .filter((prop) => prop !== 'type')
+    .map((prop) => {
+      const props = {}
```
Nitvex, line 39: Small proposal, may be the following can be better:
```
if (sGroupProperty === 'connectivity') {
  <Field
    name={prop}
    key={`${type}-${prop}`}
    {...props}
    component={Select}
    options={getSelectOptionsFromSchema(schemes[type].properties[prop])}
  />
}
const propsMapping = {
  name: { maxLength: 15 },
  fieldName: { maxLength: 30 },
  fieldValue: { type: 'textarea' },
  radiobuttons: { type: 'radio' }
}
const props = propsMapping[sGroupProperty] || {}
return <Field name={prop} key={`${type}-${prop}`} {...props} />
```

packages/ketcher-react/src/script/ui/dialog/toolbox/SDataFieldset.jsx
yuleicul: Done. Thank you.

packages/ketcher-react/src/script/ui/dialog/toolbox/SGroupFieldset.jsx
yuleicul: Done. Really good suggestions, thank you so much.

Nitvex (approved):

PR #2150, opened 2023-01-31
#2123: Hover effect doesn't appear after clicking on bonds or atoms

### LINKED ISSUE #2123: Hover effect doesn't appear after clicking on bonds or atoms
**Steps to Reproduce**
Scenario 1:
1. Select bond tool
2. Add bond to canvas
3. Don't move the cursor

Scenario 2:
1. Select bond tool and add a bond to canvas
2. Move the cursor slightly so the hover effect appears
3. Click on the bond again

**Actual behavior**
There is no hover effect on the bond, same applies to clicking on atoms

**Expected behavior**
Bond has hover effect without moving the cursor
image

TimSPb89:
This PR resolves #2123 
Fixed by removing hover reset call from click handlers

### DIFF
--- a/packages/ketcher-react/src/script/editor/tool/atom.ts
+++ b/packages/ketcher-react/src/script/editor/tool/atom.ts
@@ -241,6 +241,11 @@ class AtomTool {
     this.editor.event.message.dispatch({
       info: false
     })
+    this.editor.hover(
+      this.editor.findItem(event, ['atoms', 'functionalGroups']),
+      null,
+      event
+    )
   }
 }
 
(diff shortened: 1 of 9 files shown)


St-Permiakov:
I really not sure this is a good idea to remove hover resets from all tools.
They are there for some reason, I guess.
Is it expected that hover will never disappear for all other tools (like template, for example)?

packages/ketcher-react/src/script/editor/tool/template.ts
```diff
@@ -506,7 +504,6 @@ function addSaltsAndSolventsOnCanvasWithoutMerge(
   const [action] = fromTemplateOnCanvas(restruct, template, dragCtx.xy0, 0)
   editor.update(action)
   editor.selection(null)
-  editor.hover(null)
```
St-Permiakov, line 509: I'm really not sure this is a good idea to remove hover resets from all tools.
They are there for some reason, I guess.
Is it expected that hover will never disappear for all other tools (like here, for example)?

TimSPb89: reworked

KonstantinEpam23 (approved):

PR #2176, opened 2023-02-06
#2170: Connected contracted functional groups become non-interactive

### LINKED ISSUE #2170: Connected contracted functional groups become non-interactive
**Steps to Reproduce**
1. Open template window functional groups tab and select some functional group (e.g. Cbz)
2. Add two groups on canvas and expand them
3. Add a single bond and using select tool connect two atoms of the separate groups together with this bond
4. Contract both functional groups

**Actual behavior**
Functional groups loose interactivity, they no longer can be selected.

**Expected behavior**
Both functional groups are interactive

**Screenshots**

https://user-images.githubusercontent.com/101649391/216348711-efbc2046-6bc2-4ede-8d80-84555a4a79e1.mov

**Ketcher version** 
v2.8.0

KonstantinEpam23:
Closes #2170

### DIFF
--- a/packages/ketcher-core/src/application/render/restruct/resgroup.js
+++ b/packages/ketcher-core/src/application/render/restruct/resgroup.js
@@ -129,16 +129,27 @@ class ReSGroup extends ReObject {
     let width = 0
     let height = 0
     const sGroupItem = this.item
-    const [firstAtomId] = sGroupItem.atoms
-    const sGroupAtom = this.render.ctab.atoms.get(firstAtomId)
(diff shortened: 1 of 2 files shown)


KonstantinEpam23:
The original issue was caused b/c when an s-group is contracted - it looses data about its boundingBox. This data is calculated based on the actual svg element paths and stored in the `visel` property of ReSgroup, but is lost when the group is contracted. This PR stores the boundingBox data into oldBoundingBox property on visel and calculates x,y position based on struct and not the rendered svg element like it used to previously.

Issue #1990, opened 2022-12-23
Functional group does not connect with another functional group on click&drag


Zhirnoff:
**Steps to Reproduce**
1. Launch Ketcher
2. Select any 'Functional Group' abbreviation (e.g. Bn)
3. Place on canvas (do not expand)
4. Select any 'Functional Group' abbreviation (e.g. Bn) with Template or Paste tools
5. Hover over 'Bn' on canvas, click and drag

**Actual behavior**
At the moment, a notification window appears:
EDIT ABBREVIATION
A change was detected for the abbreviation. Do you want to remove the abbreviation information from the structure and continue work with separate atoms and bonds?

**Expected behavior**
If existing functional group is not yet connected to another structure, the new one should be added and connected with existing one with single bond.
![2023-02-09_15h21_26](https://user-images.githubusercontent.com/62464241/217811824-5823834f-31ed-431e-9774-6c1f854e5863.png)

If existing functional group is already connected with another structure, 'Edit Abbreviation' dialog should be opened (current behavior).

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser Chrome
 - Version 108.0.5359.99 (Official Build) (64-bit)

**Ketcher version**
[Version 2.7.0-rc.5](https://lifescience.opensource.epam.com/ketcher/index.html)
Build at 2022-12-22; 10:43:36

**Additional context**

Nitvex:
Issue is reproduced in v2.7.0-rc.9, moving to v2.9.0-rc.1

yuleicul:
> **Expected behavior**
>
> ......
>
> If existing functional group is already connected with another structure, 'Edit Abbreviation' dialog should be opened (current behavior).

Hi @St-Permiakov , could you please check if the expected behavior above has been implemented? If I understand correctly, it hasn't 😔. The 'Edit Abbreviation' dialog doesn't pop up as expected.

PR #2208, opened 2023-02-13
#1990: Functional group does not connect with another functional group on click&drag

St-Permiakov:
**Fixed**
- Attaching salt or solvent with mousemove to atoms

**Added**
- Create bond between groups when adding a group with Template or Paste tool
- Get attachment atom id method in sgroup entity (needs fixed MOL files to work properly)

**Refactored**
- `getGroupsIdsFromBondId` — since bond can belong to more than one group, it should return array and not single number. Previously added `getGroupIdFromBondId" left for compatibility purposes and should be removed later
- `findClosestFG` extended with possibility to skip items, but for now not used (replaced with easier solution). Potentially required to get target group for hover when pasting/inserting another group (to avoid editor picking the same group you're pasting as "closest")


packages/ketcher-react/src/script/editor/tool/template.ts
```diff
@@ -403,7 +354,7 @@ class TemplateTool {
     let action
     let pasteItems = null
 
-    if (SGroup.isSaltOrSolvent(this.template.molecule.name)) {
+    if (this.isSaltOrSolvant) {
```
Nitvex, line 357:
```suggestion
    if (this.isSaltOrSolvent) {
```

packages/ketcher-react/src/script/editor/tool/paste.ts
```diff
@@ -112,4 +227,31 @@ class PasteTool {
   }
 }
 
+type Template = {
+  aid?: number
+  molecule?: Struct
+  xy0?: Vec2
+  angle0?: number
+}
+function prepareTemplateFromSingleGroup(molecule: Struct): Template | null {
```
Nitvex, line 236: May be we need to add some comment that this function calculates the position of FG and its rotation angle?

packages/ketcher-react/src/script/editor/tool/paste.ts
```diff
@@ -14,23 +14,40 @@
...
   mergeItems: any
+  isSingleContractedGroup: boolean
 
   constructor(editor, struct) {
     this.editor = editor
     this.editor.selection(null)
     this.struct = struct
 
+    this.isSingleContractedGroup =
+      struct.isSingleGroup() && !struct.functionalGroups.get(0).isExpanded
```
Nitvex, line 49: Hmm, are you sure we can rely on `0` ?
I guess, in this case we need to get first FG by something like `struct.functionaGroups.values().next()`

St-Permiakov: I am sure.
If id is something else that `0`, it can be considered a bug and fixed.
This "struct" is considered to be a brand-new structure with ids different from actual canvas.
And if first (and only, as we verify beforehand) group has non-0 id, it's a clear marker of a bug.

packages/ketcher-react/src/script/editor/tool/paste.ts
St-Permiakov: Added.

Issue #1953, opened 2022-12-19
The bond between two Functional Group abbreviations dissapears


Zhirnoff:
**Steps to Reproduce**
1. Launch Ketcher
2. Paste this structure:
```

Ketcher 12162215312D 1   1.00000     0.00000     0

  8  8  0  0  0  0  0  0  0  0999 V2000
   10.3670   -6.7750    0.0000 C   0  0  0  0  0  0  0  0  0  0  0  0
   11.2330   -6.2750    0.0000 C   0  0  0  0  0  0  0  0  0  0  0  0
    9.5008   -7.2755    0.0000 C   0  0  0  0  0  0  0  0  0  0  0  0
    8.6340   -6.7753    0.0000 C   0  0  0  0  0  0  0  0  0  0  0  0
    7.7639   -7.2781    0.0000 C   0  0  0  0  0  0  0  0  0  0  0  0
    7.7703   -8.2780    0.0000 C   0  0  0  0  0  0  0  0  0  0  0  0
    8.6336   -8.7739    0.0000 C   0  0  0  0  0  0  0  0  0  0  0  0
    9.5001   -8.2732    0.0000 C   0  0  0  0  0  0  0  0  0  0  0  0
  1  2  1  0     0  0
  3  4  2  0  0  0  0
  8  3  1  0  0  0  0
  7  8  2  0  0  0  0
  6  7  1  0  0  0  0
  5  6  2  0  0  0  0
  4  5  1  0  0  0  0
  3  1  1  0  0  0  0
M  STY  1   1 SUP
M  SLB  1   1   1
M  SAL   1  7   1   3   4   5   6   7   8
M  SBL   1  1   1
M  SMT   1 Bn
M  END
```

3. Select 'Bn' from the 'Custom Templates' -> 'Functional Groups' list, and click on the methyl group
4. Look to the bond between two Functional Group abbreviations
Note: If you expand the abbreviations, then the bond is in place. The bond disappears only when the abbreviations are collapsed. Visual problem.

**Actual behavior**
The bond between two Functional Group abbreviations dissapears

**Expected behavior**
The bond between two Functional Group abbreviations does not disappear

**Screenshots**

https://user-images.githubusercontent.com/62464241/208429817-390248ce-59f0-425a-bb7e-007efdb8910f.mp4

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser Chrome
 - Version 108.0.5359.125 (Official Build) (64-bit)

**Ketcher version** 
[...]

PR #2286, opened 2023-02-28
Backmerge: #1953 The bond between two Functional Group abbreviations disappears

yuleicul:
Resolves: #1953

### DIFF
--- a/packages/ketcher-core/src/application/render/restruct/rebond.ts
+++ b/packages/ketcher-core/src/application/render/restruct/rebond.ts
@@ -67,8 +67,7 @@ class ReBond extends ReObject {
       FunctionalGroup.isBondInContractedFunctionalGroup(
         bond,
         sgroups,
-        functionalGroups,
-        true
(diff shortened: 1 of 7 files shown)


packages/ketcher-core/src/application/render/restruct/rebond.ts
```diff
@@ -109,15 +107,13 @@ class ReBond extends ReObject {
...
-    const sgroupsIds = struct.getGroupsIdsFromBondId(bid)
     if (
+      bond &&
       FunctionalGroup.isBondInContractedFunctionalGroup(
         bond,
         sgroups,
-        functionalGroups,
-        false
-      ) &&
-      sgroupsIds.length < 2
```
yuleicul, line 120: Have refactored `isBondInContractedFunctionalGroup` as:
image

so don't need `sgroupsIds.length < 2`

KonstantinEpam23 (approved):

Issue #2216, opened 2023-02-14
S-Group pop-up tool tip is positioned so that it overlaps structure


Zhirnoff:
**Steps to Reproduce**
1. Open Ketcher
2. Create structure and make Data S-Group with 'Field name' and 'Field value'
3. With 'Selection tool' hover over created S-Group

**Actual behavior**
S-Group pop-up tool tip is positioned so that it overlaps structure and makes it difficult to interact with structure. 
**Expected behavior**
S-Group pop-up tool tip is positioned so that it doesn't overlap a structure.
Should make a pop-up tooltip away from highlighted S-Group structure.
**Screenshots**

https://user-images.githubusercontent.com/62464241/218767336-6e48befa-d671-4579-b783-9bbfe2e34d5f.mp4

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser Chrome
 - Version 109.0.5414.120 (Official Build) (64-bit)

**Ketcher version** 
[Version 2.8.0-rc.2](https://lifescience.opensource.epam.com/ketcher/index.html)

**Additional context**

PR #2289, opened 2023-03-01
#2216 S-Group pop-up tool tip is positioned so that it overlaps structure

Konstantin1996:
**What was done?**

For now it calculates according to the green rectangle that appears when user hovering over the element. Function calculateMiddleCoordsForRect calculate middle coord for each side of the rect. After that panel appears relativly to the middle point of side.

Panel info is always in the bottom if it is center of the page.

image
image

Close to borders
image
image
image

- created WrapperPositionedRelativeToRectangle component.
- moved calculateScrollOffsetX , calculateScrollOffsetY into helpers.ts to reduce code duplication
- created calculateMiddleCoordsForRect accepts array of arrays [[1,2], [3,4], [3,4]] => [{ x: 2, y: 3 }, { x: 3, y: 4}]
- added wrapperRef on div to get dinamically changed width / height of the element

Please, review and add to the q

### DIFF
--- a/packages/ketcher-react/src/script/ui/views/components/StructEditor/WrapperPositionedRelativeToRectangle.tsx
+++ b/packages/ketcher-react/src/script/ui/views/components/StructEditor/WrapperPositionedRelativeToRectangle.tsx
@@ -1,0 +1,110 @@
+import { useEffect, useRef, useState, FC } from 'react'
+import { Vec2, Render, SGroup, Struct } from 'ketcher-core'
+import clsx from 'clsx'
+import classes from './InfoPanel.module.less'
+import {
+  calculateMiddleCoordsForRect,
+  calculateScrollOffsetX,
+  calculateScrollOffsetY
+} from './helpers'
+
+/* eslint-disable-line camelcase */
+const BAR_PANEl_SIZE = 32
+const LEFT_PADDING_MULTIPLIER = 3
+
+function getPanelPositionRelativeToRect(
+  clientX: number,
+  clientY: number,
+  sGroup: SGroup,
+  render: Render,
+  width: number,
+  height: number
+): Array<Vec2> {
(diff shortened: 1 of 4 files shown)


packages/ketcher-react/src/script/ui/views/components/StructEditor/WrapperPositionedRelativeToRectangle.tsx
```diff
@@ -0,0 +1,117 @@
...
+    render?.clientArea?.clientHeight - BAR_PANEL_SIZE - height
+  const viewportRightLimit =
+    render?.clientArea?.clientWidth - BAR_PANEL_SIZE - width
+
+  // We need to remove first element of the path. Example of the path => ['M', 23, 43]
+  const rectCoords: Array<Array<number>> = sGroup.hovering.attrs?.path?.map(
+    (line) => line.slice(1)
+  )
+
+  const [mLeftSide, mBottomSide, mRightSide, mTopSide] =
```
KonstantinEpam23, line 33: What does the 'm' prefix mean in variable names? :)

packages/ketcher-react/src/script/ui/views/components/StructEditor/WrapperPositionedRelativeToRectangle.tsx
```diff
@@ -0,0 +1,117 @@
...
+  width: number,
+  height: number
+): Vec2 | null {
+  const viewportLeftLimit = BAR_PANEL_SIZE * LEFT_PADDING_MULTIPLIER + width
+  const viewportBottomLimit =
+    render?.clientArea?.clientHeight - BAR_PANEL_SIZE - height
+  const viewportRightLimit =
+    render?.clientArea?.clientWidth - BAR_PANEL_SIZE - width
+
+  // We need to remove first element of the path. Example of the path => ['M', 23, 43]
```
KonstantinEpam23, line 28: Why do we need to remove it?

KonstantinEpam23 (approved):

Issue #2160, opened 2023-01-31
Always display enhanced stereo information for atoms


AlexanderSavelyev:
**Background**
If a user designates a stereocenter as "AND" or "OR", it is not always clear on canvas since the stereo label (e.g. "AND enantiomer") does not show which atom has enhanced stereo

**Solution**
if "ignore-chiral-flag"=true (see #2161)
Always display enhanced stereo information for atoms (&1, or1, etc).

**Examples**

* Paste structure ```CC[C@@H](C)CCCC```
* Change Enhanced stereochemistry for stereo atom to AND
* Ketcher should display the enhanced stereo &1

![image](https://user-images.githubusercontent.com/7427808/215844870-1f212cb9-74a8-4699-9be6-f55cd6ea9c83.png)

* Change Enhanced stereochemistry for stereo atom to OR
* Ketcher should display the enhanced stereo or1

![image](https://user-images.githubusercontent.com/7427808/215845111-676641c8-f2cb-4aa0-ace6-0d832bdf089a.png)

ansivgit:
If "ignore-chiral-flag"=false (by default) we leave the current behavior

PR #2296, opened 2023-03-03
#2161 Add ignore-chiral-flag parameter to ketcher settings

### LINKED ISSUE #2161: Add ignore-chiral-flag parameter to ketcher settings
**Background**
Chiral flag is a legacy property from MDL molfiles. It should not be used

**Solution**
* Add a parameter ```Ignore the chiral flag``` inside Settings/Stereochemistry. Check box, default off. Location after "Text of Mixed flag"
* Add help text for mouse hover "Ignore chiral flag while loading from molfiles. By default all the stereo will be ABS"
* Add API parameter ```ignoreChiralFlag``` (Boolean; default=false) which can be changed during loading Ketcher editor
* Update help documentation for the flag behavior using this document
[ignore_chiral_flag.pdf](https://github.com/epam/ketcher/files/10743177/ignore_chiral_flag.pdf)

[ignore_chiral_flag.docx](https://github.com/epam/ketcher/files/10743178/ignore_chiral_flag.docx)

* if   ```ignoreChiralFlag``` = true 
  * input structures are treated as ABS until any enhanced stereo is defined.
  * do not display any labels ( "ABS" or "AND Enantiomer" etc) 
  * all enhanced stereo should be displayed always (related to #2160)
  * saving into molfile v2000 should set chiral flag = 1 (see doc file)

ansivgit:
Closes #2161 & #2160


packages/ketcher-core/src/application/ketcher.ts
```diff
@@ -111,41 +115,36 @@ export class Ketcher {
...
+    )
 
-    if ('dearomatize-on-load' in options) {
-      return {
-        'general.dearomatize-on-load': options['dearomatize-on-load']
-      }
-    }
-    if ('ignoreChiralFlag' in options) {
-      return {
-        'general.ignore-no-chiral-flag': options.ignoreChiralFlag
```
KonstantinEpam23: @ansivgit I dont think this needs to be in server-style, I changed "general.ignore-no-chiral-flag" to "ignoreChiralFlag" here - do you think this is OK?

packages/ketcher-core/src/application/formatters/formatterFactory.ts
```diff
@@ -56,6 +60,11 @@ export class FormatterFactory {
       molfileParseOptions.badHeaderRecover = badHeaderRecover
     }
 
+    if (typeof ignoreChiralFlag === 'boolean') {
+      molfileParseOptions.ignoreChiralFlag = ignoreChiralFlag
+      structServiceOptions['ignore-no-chiral-flag'] = ignoreChiralFlag
```
KonstantinEpam23, line 65: Just curios about the flag name "ignore-no-chiral-flag" - is it already supported by Indigo?

packages/ketcher-core/src/application/ketcher.ts
ansivgit: Ok

packages/ketcher-core/src/application/formatters/formatterFactory.ts
```diff
@@ -70,6 +79,7 @@ export class FormatterFactory {
     switch (format) {
       case SupportedFormat.ket:
         formatter = new KetFormatter(new KetSerializer())
+        // formatter = new KetFormatter(new KetSerializer(molSerializerOptions))
```
Konstantin1996, line 82: Do you need it? Please, remove

packages/ketcher-core/src/domain/serializers/mol/molSerializer.ts
```diff
@@ -37,27 +37,34 @@ export class MolSerializer implements Serializer<Struct> {
...
+          })
         } catch (ex1) {
           //
         }
         try {
           // check for a missing first line
           // this sometimes happens when pasting
-          return molfile.parseCTFile(
-            [''].concat(lines),
-            this.options.reactionRelayout
```
Konstantin1996, line 59: this.options.reactionRelayout is not used anymore in molSerializer.ts.

I think we can delete it from class itself and from other methods like:

parseCTFile, parseRxn...

KonstantinEpam23 (approved):

Issue #2273, opened 2023-02-24
When user selects and moves atom on Functional Group or Salt it doesn't replaces atom in the structure 


d-mrzv:
**Steps to Reproduce**
1. Open Ketcher
2. Add Functional Group or Salt to the canvas 
3. Add an atom to the canvas
4. Using selection tool select the atom and move it over FG or Salt

**Actual behavior**
Atom doesn't replace FG or Salt, it displayed on the canvas over the FG or Salt

**Expected behavior**
Atom replaces a FG or Salt

**Screenshots**

https://user-images.githubusercontent.com/124911723/221199332-941b5c81-9d18-4241-99ae-30e545c840d2.mov

**Desktop (please complete the following information):**
 - OS: MacOS 13.2 (22D49)
 - Browser: Chrome
 - Version: Version 110.0.5481.100 (Official Build) (x86_64)

**Ketcher version** [e.g. v2.4.2]. 
[Version 2.8.0-rc.4]

**Additional context**
Related discussion: https://github.com/epam/ketcher/discussions/2315

Zhirnoff:
Bug reproduces in 2.10.0-rc.1
https://user-images.githubusercontent.com/62464241/234938521-6d59158a-9273-4c41-8781-aa8bcc6a8a0e.mp4

PR #2304, opened 2023-03-07
#2273 - When user selects and moves atom on Functional Group or Salt it doesn't replaces atom in the structure

yuleicul:
Resolves: #2273

### DIFF
--- a/packages/ketcher-core/src/application/editor/actions/closelyFusing.ts
+++ b/packages/ketcher-core/src/application/editor/actions/closelyFusing.ts
@@ -70,6 +70,9 @@ export function getHoverToFuse(items) {
     bonds: Array.from(items.bonds.values()),
     ...(items.functionalGroups && {
       functionalGroups: Array.from(items.functionalGroups.values())
+    }),
+    ...(items.atomToFunctionalGroup && {
+      sgroups: Array.from(items.atomToFunctionalGroup.values())
     })
   }
 
@@ -88,17 +91,20 @@ export function mergeMapOfItemsToSet(items: Map<number, number>): Set<number> {
  * @param struct
  * @param closestMap {{
  * 		atoms: Map<number, number>,
- * 		bonds: Map<number, number>
+ * 		bonds: Map<number, number>,
+ *    atomToFunctionalGroup: Map<number, number>
  * }}
  * @return {{
  * 		atoms: Map<number, number>,
- * 		bonds: Map<number, number>
+ * 		bonds: Map<number, number>,
+ *    atomToFunctionalGroup: Map<number, number>
  * }}
  */
 function closestToMerge(struct, closestMap) {
   const mergeMap = {
     atoms: new Map(closestMap.atoms),
-    bonds: new Map(closestMap.bonds)
+    bonds: new Map(closestMap.bonds),
+    atomToFunctionalGroup: new Map(closestMap.atomToFunctionalGroup)
   }
 
   closestMap.bonds.forEach((dstId, srcId) => {
@@ -113,7 +119,12 @@ function closestToMerge(struct, closestMap) {
     }
   })
 
-  if (mergeMap.atoms.size === 0 && mergeMap.bonds.size === 0) return null
+  if (
+    mergeMap.atoms.size === 0 &&
+    mergeMap.bonds.size === 0 &&
+    mergeMap.atomToFunctionalGroup.size === 0
+  )
+    return null
 
   return mergeMap
 }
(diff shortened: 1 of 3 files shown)


packages/ketcher-core/src/application/editor/actions/closelyFusing.ts
```diff
@@ -113,7 +119,12 @@ function closestToMerge(struct, closestMap) {
     }
   })
 
-  if (mergeMap.atoms.size === 0 && mergeMap.bonds.size === 0) return null
+  if (
+    mergeMap.atoms.size === 0 &&
+    mergeMap.bonds.size === 0 &&
+    mergeMap.atomToFunctionalGroup.size === 0
+  )
+    return null
```
Konstantin1996, line 127: Better to create curly brackets:

```
if (...) {

}
```

packages/ketcher-react/src/script/editor/shared/closest.js
```diff
@@ -505,21 +506,63 @@ function findCloseMerge(restruct, selected, maps = ['atoms', 'bonds'], scale) {
...
+function mergeAtomToFunctionalGroup(atomId, restruct, atomPosition, result) {
+  if (FunctionalGroup.isAttachmentPointAtom(atomId)) return false
+
+  const closestFunctionalGroup = findClosestFG(restruct, atomPosition, null)
+  if (closestFunctionalGroup) {
+    result.atomToFunctionalGroup.set(atomId, closestFunctionalGroup.id)
+    return true
+  }
+
+  return false
```
Konstantin1996, line 563: Can we somehow join these conditions into single condition?

``` 
if (FunctionalGroup.isAttachmentPointAtom(atomId)) return false

return false
```

Konstantin1996: FunctionalGroup.isAttachmentPointAtom(atomId) has already return you boolean

Maybe just like this 

`
  const closestFunctionalGroup = findClosestFG(restruct, atomPosition, null)
  if (closestFunctionalGroup) {
    result.atomToFunctionalGroup.set(atomId, closestFunctionalGroup.id)
    return true
  }

 return !FunctionalGroup.isAttachmentPointAtom(atomId)
`

Konstantin1996:
We need to create rule for auto-formatting curly brackets

packages/ketcher-react/src/script/editor/tool/helper/dropAndMerge.ts
```diff
@@ -1,17 +1,28 @@
...
+  atomToFunctionalGroup?: Map<number, number>
 }
 
 export function dropAndMerge(
   editor: Editor,
-  mergeItems: MergeItems,
+  mergeItems: MergeItems | null,
   action?: Action
 ): void {
+  if (!mergeItems) return
```
Konstantin1996, line 24: Brackets

packages/ketcher-react/src/script/editor/shared/closest.js
```diff
@@ -505,21 +506,63 @@ function findCloseMerge(restruct, selected, maps = ['atoms', 'bonds'], scale) {
...
-      if (item && !selected[mp].includes(item.id)) res.set(srcId, item.id)
+  const result = {
+    atoms: new Map(),
+    atomToFunctionalGroup: new Map()
+  }
 
-      return res
-    }, new Map())
+  maps.forEach((map) => {
+    if (map === 'atoms') {
```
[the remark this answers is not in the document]
yuleicul: If I understand correctly, what you mean is something like
```
const atom = 'atoms'
...
maps.forEach((map) => {
    if (map === atom ) {
...
```
I think it's a bit redundant. The better solution might be to add a type to `maps`, but unfortunately, this file is now still only written by JavaScript. I suggest keeping the way I wrote it and refactoring it in TypeScript in the future.

packages/ketcher-react/src/script/editor/shared/closest.js
yuleicul: Thanks for your review again, but I can't think of a better way to join them into a single condition, and unfortunately, your suggestion is not gonna work.

The logic is like
ketcher-review-1-2023-03-09-1343

and your proposal is like
ketcher-review-2-2023-03-09-1343

They're two different logic. Anyway, you have brought up a good point, if you have any other ideas please don't hesitate to tell me! 😃

yuleicul:
> We need to create rule for auto-formatting curly brackets

Good idea. I suggest creating a new ticket for that.

Konstantin1996:
Create ticket for curly braces https://github.com/epam/ketcher/issues/2317

KonstantinEpam23 (approved):

Issue #2293, opened 2023-03-02
Templates Library tab shows extra structures


lorken24:
**Steps to Reproduce**
1. Open Ketcher
2. Open the 'Templates' window
3. In 'Functional Groups' or 'Salts and Solvents' select any structure and place it on the canvas
4. Open the 'Templates' window again, on the 'Templates Library' tab scroll down

**Actual behavior**
After the 'User Templates' folder structures from the previously selected tab are displayed

**Expected behavior**
The 'Templates Library' tab has only 
-α-D-Sugars;
-Aromatics;
-β-D-Sugars;
-Bicycles;
-Bridged Polycyclics;
-Crown Ethers;
-D-Amino Acids;
-D-Sugars;
-Heterocyclic Rings;
-L-Amino Acids;
-Nucleobases;
-Rings;
-Sugars;
-3D Templates;
-User Templates (if a user created a custom template).

**Screenshots**

https://user-images.githubusercontent.com/65543532/222448661-109fbaf3-44c2-40e2-a39b-c53be576a496.mov

**Desktop (please complete the following information):**
 - OS: macOS
 - Browser Chrome
 - Version 110.0.5481.177 (Official Build) (arm64)

**Ketcher version** 
[Version 2.8.0-rc.5](https://lifescience.opensource.epam.com/ketcher/index.html)
Build at 2023-03-02; 12:56:58

**Test case**
EPMLSOPKET-1697 for information

**Additional context**
Possibly related to issue #2264

PR #2305, opened 2023-03-07
Backmerge: #2293 Templates Library tab shows extra structures

KonstantinEpam23:


KonstantinEpam23:
Backmerge for: https://github.com/epam/ketcher/pull/2303

PR #2392, opened 2023-03-20
#2385 Migrate to Indigo v1.10.0-rc.5 in-browser module

KonstantinEpam23:


KonstantinEpam23:
We will be integratingin Indigo-v1.10-rc.6 instead

PR #2430, opened 2023-03-30
#2411 [Investigation] Move to Vite from Rollup for project build

### LINKED ISSUE #2411: Move to vite from react-app-rewired for example project development build
**Background**

**Solution**

**Alternatives**

**Additional context**

yuleicul:
Redirect to #2432


yuleicul:
Too many conflicts after setting master as the target branch. Redirect to #2432

amcraig:
Is there an ETA for the next release, and or possibility for a interem patched-release?
My work is blocked by the `Unsupported URL Type "workspace:": workspace:*` bug.

yuleicul:
@amcraig Could you tell me how to reproduce?

Issue #2420, opened 2023-03-27
Hotkey (Del) can't delete Functional Groups and Salts abbreviation


Zhirnoff:
**Steps to Reproduce**
1. Open Ketcher
2. Add Functional Groups and Salts abbreviations on canvas
3. Hover mouse over Functional Groups and Salts abbreviations and press hotkey (Del)

**Actual behavior**
Functional Groups and Salts abbreviations are not deleted

**Expected behavior**
Functional Groups and Salts abbreviations are deleted

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser Chrome
 - Version 109.0.5414.120 (Official Build) (64-bit)

**Ketcher version** 
[Version 2.9.0-rc.5](https://lifescience.opensource.epam.com/ketcher/index.html)

Zhirnoff:
A contracted function group abbreviation is removed with the DEL hotkey. But now it became possible to delete parts of the expanded functional group or salt with the DEL hotkey. When trying to change the functional group or salt, a warning window should appear (EDIT ABBREVIATION)

https://user-images.githubusercontent.com/62464241/233986425-50d481fe-042f-401c-8d34-f9a807f2bfe5.mp4

PR #2478, opened 2023-04-12
#2420 Hotkey (Del) can't delete Functional Groups and Salts abbreviation

Ziprion:
closes [#2420](https://github.com/epam/ketcher/issues/2420)

### DIFF
--- a/packages/ketcher-core/src/application/editor/actions/erase.ts
+++ b/packages/ketcher-core/src/application/editor/actions/erase.ts
@@ -28,7 +28,7 @@ import { removeAtomFromSgroupIfNeeded, removeSgroupIfNeeded } from './sgroup'
 
 import { Action } from './action'
 import assert from 'assert'
-import { atomGetDegree } from './utils'
+import { atomGetDegree, formatSelection } from './utils'
 import { fromBondStereoUpdate } from '../actions/bond'
 import { fromFragmentSplit } from './fragment'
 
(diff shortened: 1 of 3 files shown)


packages/ketcher-core/src/application/editor/actions/erase.ts
```diff
@@ -84,23 +84,30 @@ export function fromOneBondDeletion(restruct, id) {
...
-  }
+  const selection = formatSelection(rawSelection)
+
+  selection.sgroups.forEach((sgroupId) => {
+    const sgroup = restruct.sgroups.get(sgroupId)
+    const sgroupAtoms = sgroup.item.atoms
+
+    selection.atoms = selection.atoms.concat(sgroupAtoms) as any[]
+
+    Object.values(restruct.molecule.bonds).forEach((bond: any, bid) => {
```
KonstantinEpam23, line 102: Looks like you are using iteration index here for `bid`? This may not be a correct bond id.
Also, struct.bonds is a Pool object, are you sure passing it to Object.values() produces a correct result?

Ziprion: You're right. Fixed it

packages/ketcher-core/src/application/editor/actions/erase.ts
```diff
@@ -84,23 +84,30 @@ export function fromOneBondDeletion(restruct, id) {
...
-  }
+  const selection = formatSelection(rawSelection)
+
+  selection.sgroups.forEach((sgroupId) => {
+    const sgroup = restruct.sgroups.get(sgroupId)
+    const sgroupAtoms = sgroup.item.atoms
+
+    selection.atoms = selection.atoms.concat(sgroupAtoms)
+
+    restruct.molecule.bonds.forEach((bond, bid) => {
```
Nitvex, line 102: Could you, please, use `bondId` instead?

Nitvex:
@Ziprion, could you, please, fix a conflict and provide a description for Pull Request. It can be like "Selection didn't include some bonds/atoms that's why ..."

packages/ketcher-core/src/application/editor/actions/erase.ts
Ziprion: Done

Nitvex (approved):

PR #2487, opened 2023-04-18
#2411 Move to vite from react-app-rewired for example project development build

### LINKED ISSUE #2411: Move to vite from react-app-rewired for example project development build

yuleicul:
Closes #2411 

## Getting Started

At root

```shell
npm install
```

then go to example

```shell
# standalone
npm run dev:standalone
# or remote
npm run dev:remote
```

The steps above have been added to [DEVNOTES.md](https://github.com/epam/ketcher/pull/2487/files#diff-e6fecc1d2ecdd042c611b8f62f78af878293785f983f410a11453b2ff6dcdeafR82-R96)

### TypeScript IntelliSense and `tsconfig`

The problem: every time we changed types in kecher-core, we had to rebuild ketcher-core to get correct TypeScript IntelliSense where importing ketcher-core, such as ketcher-react.

Solution: I rename the previous `tsconfig.json` of in ketcher-react to `tsconfig.build.json`, and use it to build. Then I create a new `tsconfig.json` which contains `paths`. Through `paths`, VSCode can give us TypeScript IntelliSense without rebuilding ketcher-core.

~~### Missing source file warning~~

~~The problem: Vite can find imported CSS files, which means during development we don't need to `import 'ketcher-react/dist/index.css'` in example/src/App.tsx.~~

~~Solution: resolve 'ketcher-react/dist/index.css' by a custom resolver which returns `null`, and please **ignore** this warning when starting dev server.~~

~~[warningforcsssourcemap](https://user-images.githubusercontent.com/27288153/232698280-0085c4b1-e16e-4996-85c9-2a889bedfc53.png)~~


example/package.json
```diff
@@ -3,6 +3,8 @@
   "version": "1.0.0",
   "private": true,
   "scripts": {
+    "dev:standalone": "cross-env MODE=standalone vite",
```
KonstantinEpam23, line 6: I think lets replace the "start:standalone"/"start:remote" commands with these ones, not sure why we would want to keep the old versions.

packages/ketcher-react/src/script/ui/views/components/StructEditor/StructEditor.jsx
```diff
@@ -112,12 +112,17 @@ class StructEditor extends Component {
     })
 
     this.editor.event.cursor.add((csr) => {
+      let clientX, clientY
```
Nitvex, line 115: Is there a specific point for doing this?

yuleicul: I would say this may not be the best solution, but it's somehow a not bad workaround. 🤔 

The problem is: Vite seems to have bugs about hoisting decelerations in case block.
Untitled-2023-04-20-1601

 😵‍💫It doesn't occur when I set debuggers. It seems that Vite has enough time to decelare(hoist) `clientX` when there is a debugger.

https://user-images.githubusercontent.com/27288153/233313212-c1f8f726-f066-4b61-b55b-d3668a8f57ae.mp4

Why i said that workaround is not bad is because eslint gave warning _Unexpected lexical decelaration in case block_, as you can see in the snapshot. That workaround somehow avoids Vite bug and fixes eslint warning.

KonstantinEpam23 (approved):

PR #2492, opened 2023-04-18
Add a function to detect when opened file is loaded

Konstantin1996:
**What was done?**

- emitted event for remote and standalone version


KonstantinEpam23:
This functionality is required for autotests

Nitvex (approved):

PR #2519, opened 2023-04-22
Move render struct to ketcher core

KonstantinEpam23:


KonstantinEpam23:
Resolved in: https://github.com/epam/ketcher/pull/2651

Issue #2463, opened 2023-04-07
Rotation Tool: display predefined rotation angles and current angle


vanoprenko:
Enhance the rotation tool introduced in scope of #2090, so that user can see predefined rotation angles' marks as well as current rotation angle and can easier rotate the selection on needed angle. The rotation tool should work as follows:

1. Once user clicks the rotation handle and rotation mode is activated, the rounded rectangle should be replaced with a circle. Center of the circle should be in the center of rotation, circle diameter equal to the height of replaced rectangle. The rotation handle should not change its position while switching from selection to rotation mode and back.
2. The following predefined angles should be marked on the rotation circle: -90, -60, -45, -30, 0, 30, 45, 60, 90
3. Once user starts rotating the structure, the set of angles shown should always cover [-90 - +90] range relatively to the handle position, e.g. if the structure is rotated for 60 degrees, user should see -30, 0, 30, 45, 60, 90, 120, 135, 150
4. Current rotation angle should be displayed at starting position (0 degree). 
5. Once user releases left mouse button, the rotation mode is deactivated and selected fragment stays in its new position (rotated). Rotation circle should be replaced with rounded rectangle (handle on top). 

<p align="center">
<img width="300" alt="hover-handle" src="https://user-images.githubusercontent.com/18439849/230630688-c7ad3b27-71f6-4398-b7b8-4ae67b1fb72a.png"> 
</p>

<p align="center">
<img  width="500" alt="rotation-start" src="https://user-images.githubusercontent.com/18439849/230630912-91472746-fc34-4078-ab16-9763c70d7205.png"> 
</p>

<p align="center">
<img width="250" alt="rotation in progress" src="https://user-images.githubusercontent.com/18439849/230634356-4f2f350b-a6f6-40df-86a0-61fc19e93e5b.png">
</p>

#184121704

yuleicul:
> 3. Once user starts rotating the structure, the set of angles shown should always cover [-90 - +90] range relatively to the handle position, e.g. if the structure is rotated for 60 degrees, user should see -30, 0, 30, 45, 60, 90, 120, 135, 150
> 
> <p align="center">
> <img width="250" alt="rotation in progress" src="https://user-images.githubusercontent.com/18439849/230634356-4f2f350b-a6f6-40df-86a0-61fc19e93e5b.png">
> </p>

One question here, according to ''_the set of angles shown should always cover [-90 - +90] range relatively to the handle position_", angles of the image above should be like

[...]

PR #2526, opened 2023-04-25
#2090 Selection Tool: support rotation using in-place handle

### LINKED ISSUE #2090: Selection Tool: support rotation using in-place handle
Improve structure rotation functionality, so that user could rotate selected structure or fragment using in-place tools instead of toolbar buttons.

For this purpose the following additional elements should be shown once structure/fragment eligible for rotation is selected: 

1. Dash-lined rounded rectangle surrounding the selection
2. Rotation handle on top of selection area
3. A small cross indicating the center of rotation 

selected-structure

Once user presses left mouse button on the rotation handle, the rotation mode is activated and the following changes should be applied (see below sketch):

1. Rotation handle changes its view (color/icon)
2. Center of rotation highlighted with blue color
3. Dashed line connecting the rotation handle with the center of rotation appears

hover-handle

User can further rotate the structure around the center using handle. All elements of the rotation tool rotate along with selected fragment. 

rotation

Once user releases left mouse button, the rotation mode is deactivated, selected fragment stays in its new position (rotated) while elements of rotation tool (surrounding rectangle with handle) are redrawn at their initial position (zero rotation degree, handle on top). Note: all rotated atom labels should be redrawn so that symbols are aligned with horizontal axis. 

#184121704

yuleicul:
Closes #2090


KonstantinEpam23:
We have the following feedback on this:
- when non-symetric molecule is selected - the center of rotation should be just in the center of the selection, we dont need any special calculations for this:
Screenshot 2023-04-26 at 22 11 41
-  Undo/Redo (ctrl+z) do not seem to work for rotations
- We don't need to show the rotation box when only one atom is selected:
Screenshot 2023-04-26 at 22 15 51
- The rotation handle may appear ouside of the top of viewport and it is impossible to scroll up to it, perhaps we should check if there is enough space to display it, and if not - display it on the bottom?
Screenshot 2023-04-26 at 22 16 48

yuleicul:
Thanks for you detailed feeback! @KonstantinEpam23 I have some questions for some of them:

> 1. When non-symetric molecule is selected - the center of rotation should be just in the center of the selection, we dont need any special calculations for this:
> Screenshot 2023-04-26 at 22 11 41
> 
> 2. When a molecule that is connected to a structure with only one attachment point is selected - the origin of rotation should be the attachment point atom:
> Screenshot 2023-04-26 at 22 23 15

I'm wondering the two feedbacks are just for rotation handle(the new feature) or both Rotate Tool. if you are using Rotate Tool, you'll get the same rotation center as using handle, because handle is implemented following current logic of Rotate Tool. For example, for the two structures you mentioned:


https://user-images.githubusercontent.com/27288153/234740402-24d8c8c0-37a9-48c1-b693-f4272e917501.mp4


> 3. Undo/Redo do not seem to work for rotations
>
> 4. We don't need to show the rotation box when only one atom is selected:
> Screenshot 2023-04-26 at 22 15 51

These two are bugs, thanks for catching. I'll try to fix them in the scope of this PR.


> 5. The rotation handle may appear ouside of the top of viewport and it is impossible to scroll up to it, perhaps we should check if there is enough space to display it, and if not - display it on the bottom?
> Screenshot 2023-04-26 at 22 16 48

I usually drag the structure a little bit down in this case 🤣. If putting it on the bottom, we need to consider the next step #2463  -- how to show the predefined degrees. Maybe like this: 🤔 
 
image

KonstantinEpam23:
@yuleicul 
> I'm wondering the two feedbacks are just for rotation handle(the new feature) or both Rotate Tool. if you are using Rotate Tool, you'll get the same rotation center as using handle, because handle is implemented following current logic of Rotate Tool. 

I think both tools should behave the same, so the behaviour needs to be changed for both.

KonstantinEpam23:
> I usually drag the structure a little bit down in this case. If putting it on the bottom, we need to consider the next step https://github.com/epam/ketcher/issues/2463 -- how to show the predefined degrees. Maybe like this? 🤔

Based on feedback - lets ignore this case entirely for now and keep the current behaviour :)

PR #2534, opened 2023-04-26
#2420 Hotkey (Del) can't delete Functional Groups and Salts abbreviation

Ziprion:
Closes [#2420](https://github.com/epam/ketcher/issues/2420)

Added warning dialog when user is deleting atom/bond in expanded group

### DIFF
--- a/packages/ketcher-react/src/script/ui/state/handleHotkeysOverItem.ts
+++ b/packages/ketcher-react/src/script/ui/state/handleHotkeysOverItem.ts
@@ -6,7 +6,6 @@ import {
   fromBondsAttrs,
   fromFragmentDeletion,
   FunctionalGroup,
-  SGroup,
   Atom
 } from 'ketcher-core'
 import { openDialog } from './modal'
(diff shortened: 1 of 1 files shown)


packages/ketcher-react/src/script/ui/state/handleHotkeysOverItem.ts
```diff
@@ -56,15 +55,42 @@ function handleOtherActions({
...
+function handleEraser({
+  editor,
+  hoveredItem,
+  newAction,
+  dispatch
+}: HandleHotkeyOverItemProps) {
+  const item = mapItemsToArrays(hoveredItem)
+  const itemType = Object.keys(hoveredItem)[0]
+
+  if (['atoms', 'bonds'].includes(itemType)) {
```
VandaPonikarova, line 80: I would suggest to create const for 'atoms', 'bonds', etc, as we use them in different places.

Nitvex:
All the comments were fixed in https://github.com/epam/ketcher/pull/2562, closing this as obsolete.

Issue #2213, opened 2023-02-14
Data S-Group: 'Field value' text when placed on a structure becomes hard to access


Zhirnoff:
**Steps to Reproduce**
1. Open Ketcher
2. Create structure and make Data S-Group with 'Field name' and 'Field value'
3. With 'Selection tool' place 'Field value' text to structure
4. Try select 'Field value' text and move to new position

**Actual behavior**
'Field value' text when placed on a structure becomes hard to access

**Expected behavior**
User can easily select 'Field value' text and move to desired location

**Screenshots**

https://user-images.githubusercontent.com/62464241/218750066-a78604a0-df5c-450e-8382-0f530700f9c3.mp4

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser Chrome
 - Version 109.0.5414.120 (Official Build) (64-bit)

**Ketcher version** 
[Version 2.8.0-rc.2](https://lifescience.opensource.epam.com/ketcher/index.html)

PR #2537, opened 2023-04-27
#2213 Data S-Group: 'Field value' text when placed on a structure becomes hard to access

VandaPonikarova:
Added high priority for S-Group data when we make actions with elements (hover/click).


AnastasiiaPlyako (approved):

packages/ketcher-react/src/script/editor/shared/closest.js
```diff
@@ -454,22 +454,29 @@ function findClosestItem(restruct, pos, maps, skip, scale) {
   // eslint-disable-line max-params
```
gairon, line 454: I found the same function in in Core - packages/ketcher-core/src/application/editor/shared/closest.js:415
Do we need to update this one also?

Issue #2113, opened 2023-01-20
Hovered structure gets into the saved molecule


yuleicul:
**Steps to Reproduce**
1. Click the tool _Single Bond_
2. Click on the canvas to draw a structure
3. Click the tool _Selection_, and select the structure drawn on step 2
4. Press Ctrl+x to cut it
5. Press Ctrl+v, but do not click on the canvas (means you haven't yet decided to paste it)
6. Press Ctrl+s

**Actual behavior**
You can see the data of the structure(which you haven't pasted on the canvas) shows on the preview text

**Expected behavior**
The data of the structure(which you haven't pasted on the canvas) doesn't show on the preview text

**Screenshots**
![ketcher-paste-bug-10](https://user-images.githubusercontent.com/27288153/216329204-cf8ede4f-1760-4fc2-9b88-372aaa2944d5.gif)

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser Chrome
 - Version 109.0.5414.120 (Official Build) (64-bit)

**Ketcher version** [e.g. v2.4.2]. 
v2.7.0 rc.10

**Additional context**
_Originally posted by @KonstantinEpam23 in https://github.com/epam/ketcher/issues/2079#issuecomment-1378765574_

Zhirnoff:
Templates, Template Library structures, Functional Groups, Salts and Solvents, if they are under cursor at time of saving file, then get into the saved file.
Steps for Reproduce:
1. Select any Template (e.g. Benzene)
2. Do not put on canvas
3. Click 'Save' button or CTRL+S
4. Look at preview for saved file
5. Repeat steps 2-4 for Library structures, Functional Group and Salts and Solvents
6. 
https://user-images.githubusercontent.com/62464241/225099971-4f5adf40-a8d8-4132-8369-ed6eed636495.mp4

**Ketcher version** 
[Version 2.9.0-rc.3](https://lifescience.opensource.epam.com/ketcher/index.html)

Zhirnoff:
Bug reproduces in 2.10.0-rc.1

https://user-images.githubusercontent.com/62464241/234937401-ab3ef678-f04a-4bcb-a897-b514e954aa3d.mp4


https://user-images.githubusercontent.com/62464241/234937715-80432d72-fce4-4280-a925-3aaf787cbe6b.mp4

PR #2555, opened 2023-04-28
#2113 Hovered structure gets into the saved molecule

VandaPonikarova:
Added removal not rendered struct on `save` action.

### DIFF
--- a/packages/ketcher-react/src/script/ui/action/index.js
+++ b/packages/ketcher-react/src/script/ui/action/index.js
@@ -222,6 +222,20 @@ const config = {
     },
     hidden: (options) => isHidden(options, 'any-atom')
   },
+  // Removing from what should be saved - structure, which was added to paste tool,
+  // but not yet rendered on canvas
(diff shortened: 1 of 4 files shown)


packages/ketcher-react/src/script/ui/action/index.js
```diff
@@ -222,6 +222,20 @@ const config = {
     },
     hidden: (options) => isHidden(options, 'any-atom')
   },
+  // Removing from what should be saved - structure, which was added to paste tool,
+  // but not yet rendered on canvas
+  'remove-not-rendered-struct': {
+    action: {
+      thunk: (dispatch) => {
+        const savedSelectedTool = SettingsManager.selectionTool
```
Nitvex, line 230: I propose to move this into separate function as this logic is already used in `src/script/ui/state/hotkeys.ts` in `removeNotRenderedStruct`.
May be it is worth to remove this logic from before mentioned file and put it here.

Nitvex (approved):

PR #2574, opened 2023-05-04
#2463 Rotation Tool: display predefined rotation angles and current angle

### LINKED ISSUE #2463: Rotation Tool: display predefined rotation angles and current angle

yuleicul:
Closes #2463 

Note: 
1. According to UX design, the predefined degrees will be hidden if the radius is less than 65px.

rotation-2463-01

~~2. One **existing** implementation that needs attention is **15° is the smallest rotation degree**, which means the user cannot rotate like 13° or 34° etc.~~

~~[rotation-2463-02](https://user-images.githubusercontent.com/27288153/236608057-12083049-3bf8-421d-b8ff-3770e8cd00ea.gif)~~

### DIFF
--- a/packages/ketcher-react/src/script/editor/Editor.ts
+++ b/packages/ketcher-react/src/script/editor/Editor.ts
@@ -228,7 +228,7 @@ class Editor implements KetcherEditor {
 
     const isSelectToolChosen = name === 'select'
     if (!isSelectToolChosen) {
-      this.rotateController.hide()
+      this.rotateController.clean()
     }
 
     if (!tool || tool.isNotActiveTool) {
@@ -363,7 +363,7 @@ class Editor implements KetcherEditor {
 
     this.render.ctab.setSelection(this._selection) // eslint-disable-line
     this.event.selectionChange.dispatch(this._selection) // eslint-disable-line
-    this._selection === null && this.rotateController.hide()
+    this._selection === null && this.rotateController.clean()
 
     this.render.update(false, null, { resizeCanvas: false })
     return this._selection // eslint-disable-line
--- a/packages/ketcher-react/src/script/editor/Editor.ts
+++ b/packages/ketcher-react/src/script/editor/Editor.ts
@@ -321,9 +321,10 @@ class Editor implements KetcherEditor {
     }
 
     let ReStruct = this.render.ctab
-
+    let selectAll = false
     this._selection = null // eslint-disable-line
     if (ci === 'all') {
+      selectAll = true
       // TODO: better way will be this.struct()
       ci = structObjects.reduce((res, key) => {
         res[key] = Array.from(ReStruct[key].keys())
@@ -363,7 +364,12 @@ class Editor implements KetcherEditor {
 
     this.render.ctab.setSelection(this._selection) // eslint-disable-line
     this.event.selectionChange.dispatch(this._selection) // eslint-disable-line
-    this._selection === null && this.rotateController.clean()
+
+    if (selectAll) {
+      this.rotateController.rerender()
+    } else if (this._selection === null) {
+      this.rotateController.clean()
+    }
 
     this.render.update(false, null, { resizeCanvas: false })
     return this._selection // eslint-disable-line
--- a/packages/ketcher-react/src/script/editor/tool/rotate-controller.test.ts
+++ b/packages/ketcher-react/src/script/editor/tool/rotate-controller.test.ts
@@ -74,7 +74,7 @@ describe('Rotate controller', () => {
     }
 
     // @ts-ignore
-    controller.mouseDown({
+    controller.dragStart({
       buttons: 2, // Right button
       stopPropagation: () => null
     })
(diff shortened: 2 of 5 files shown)


KonstantinEpam23:
Thanks for your PR @yuleicul, looks great!
Just some feedback we got on the changes:

1. When the rotation wheel goes near 180 degrees - only half of the angle marks are shown:
Screenshot 2023-05-09 at 00 58 19-upd

2. If a page is scrolled - the circle is drawn in the incorrect position. Steps to reproduce: 
 - have a canvas big enough to have a scroll bar
 - select a structure and rotate it
 - scroll the canvas and try to rotate the structure again
Screenshot 2023-05-09 at 01 08 41-upd

3. Can we please highlight the angle mark text for the one nearest to the cursor angle position.
Like so:
Screenshot 2023-05-09 at 01 22 40

4. This is more of just a nitpick, but sometimes the handle line and the circle line go a bit out of sync:
Screenshot 2023-05-09 at 01 33 10

5. This may only be reproducable on WIndows, but the rotation circle seems to stay on the canvas if Print Screen is pressed. Possible steps to reproduce (not tested):
 - select a structure and start rotating it
 - while the rotation circle is visible - press Print Screen to take a screenshot of part of the screen
 - after taking the screenshot -  continue rotating the figure
MicrosoftTeams-image (1)

yuleicul:
@KonstantinEpam23 Thx for your feedback!
 
> 1. When the rotation wheel goes near 180 degrees - only half of the angle marks are shown:
> 
> Screenshot 2023-05-09 at 00 58 19-upd

I'm not sure if this is a bug because according to the requirement "_the set of angles shown should always cover [-90,+90] range relative to the handle position_",  positive 180 or 150 is not in the range of [-240, -60] when the current rotation degree is -150.

KonstantinEpam23:
> I'm not sure if this is a bug because according to the requirement "_the set of angles shown should always cover [-90,+90] range relative to the handle position_", positive 180 or 150 is not in the range of [-150-90, -150+60] when the current rotation degree is -150.

I think in this case for example for 180 degrees it should treat the [-1, +1] range to result in [179, -179]
Basically always showing +/-90 degrees on both sides, something like this:
Screenshot 2023-05-09 at 09 43 39

yuleicul:
> 3. Can we please highlight the angle mark text for the one nearest to the cursor angle position.
Like so:
> Screenshot 2023-05-09 at 01 22 40

I want to make sure how it works in the following cases, where the rotation degree doesn't overlap with the marks:
(note that the rotation step is 18)

image

I want to know your thoughts on which one is better, or maybe you have other ideas

KonstantinEpam23:
> I want to make sure how it works in the following cases, where the rotation degree doesn't overlap with the marks: (note that the rotation step is 18)
> Let me know your thoughts @KonstantinEpam23 on which one is better, or maybe you have other ideas

We decided that for now we should only highlight the mark if it matches the current angle exactly. This should work well for the default situation when the rotation step is 15 degrees

KonstantinEpam23:
@yuleicul found one more issue - when selecting all elements with Ctrl+A - the rotation wheel does not appear.

yuleicul:
> 5. This may only be reproducable on WIndows, but the rotation circle seems to stay on the canvas if Print Screen is pressed. Possible steps to reproduce (not tested):
> 
> * select a structure and start rotating it
> * while the rotation circle is visible - press Print Screen to take a screenshot of part of the screen
> * after taking the screenshot -  continue rotating the figure
>   MicrosoftTeams-image (1)

My solution to this one is to make cursor 'not-allowed' after finishing screenshot. If you continue dragging, it will go back to its initial state.

rotation-2463-03

KonstantinEpam23:
@yuleicul I think this solution is fine.

Also, one more small thing in addition to the ctrl+a issue - the box also does not appear when selecting while holding the shift key and clicking.

yuleicul:
> @yuleicul I think this solution is fine.
> 
> Also, one more small thing in addition to the ctrl+a issue - the box also does not appear when selecting while holding the shift key and clicking.

The bounding box shows while holding and clicking from my side:
rotation-2463-shift

As you can see, if you are only clicking **bond** the box will not appear because stuff eligible for rotation must contain atoms.

yuleicul:
One small mistake I made is that the **cross** rotated with structures. According to UX design, it shouldn't be rotated. Fixed.

yuleicul:
@KonstantinEpam23 Thanks again for your detailed review! ❤️ I've fixed all above bugs you mentioned, please review again

KonstantinEpam23 (approved):

Issue #2196, opened 2023-02-09
The Save button is active for empty canvas


lorken24:
**Steps to Reproduce**
1. Open Ketcher
2. With the empty canvas click on the 'Save as.." button

**Actual behavior**
Save structure displaying active 'Save" button

**Expected behavior**
When the canvas is empty, the "Save" button is disabled

**Screenshots**

https://user-images.githubusercontent.com/65543532/217822717-fbae35d9-ef4a-4e5c-b347-16455402846e.mov

**Desktop:**
 - OS: MacOS 13.2
 - Browser Chrome
 - Version 109.0.5414.119 (Official Build) (arm64)

**Ketcher version** 
v2.8.0

PR #2579, opened 2023-05-08
#2196 - the Save button is active for empty canvas

gairon:
* Fix `disabled` prop for `SaveButton` component
* Add tests for `disabled` prop


yuleicul (approved):
Learned `setupFilesAfterEnv` from your code. Thank you!

Issue #2423, opened 2023-03-28
Hotkeys for atoms don't work on Functional Groups and Salts abbreviations


Zhirnoff:
**Steps to Reproduce**
1. Open Ketcher
2. Add Functional Groups and Salts and Solvents abbreviations on canvas
3. Press 'O' or 'N' hotkey on Functional Groups and Salts and Solvents abbreviations on canvas

**Actual behavior**
Nothing happens

**Expected behavior**
Oxygen or Nitrogen atoms replace a Functional Groups and Salts and Solvents abbreviations on canvas

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser Chrome
 - Version 109.0.5414.120 (Official Build) (64-bit)

**Ketcher version** 
[Version 2.9.0-rc.5](https://lifescience.opensource.epam.com/ketcher/index.html)

PR #2580, opened 2023-05-08
#2423 Hotkeys for atoms don't work on Functional Groups and Salts abbreviations

AnastasiiaPlyako:
What was done?
When you select functional group and tap hotkey (o / n), it changes to the appropriate atom.

Firstly we take selected sgroup. If it is contracted, we delete all atoms (except the attachment point) and all bonds. After that, we add atoms from atomProps.

I made logic for sgroups in the atom tool because it is impossible to realize this in the sgroup tool. The reason is that the atom tool makes editor.selection null and selected elements in the sgroup tool will be empty.  

UPD: 
Discussed with @Nitvex : It will be better to delete logic with sgroup from atoms. His suggestion was to add "sgroups" in "getTool Handler" and invoke the function "handleHotkeyOverItem". 
It's hard to implement because "handleHotkeyOverItem" work with one item and in our case we need an array of items.  If I started changing code for supporting array, it started to affect a lot of places. 
In this case, I decided to create a new separate function for handling logic with selection sgroup.  

UPD 2:
- Add one more condition `newAction.tool === 'atom'` because this function should invoke only for type 'atoms'
- Change checking from to `isContractedFunctionalGroup` because this block should invoke for functionalGroup, Salt and Solvents, unknown superatom (not SRU type)


VasilevDO (approved):

kaluginserg:
Am I correct that the required behavior is two cases:
1) when we select the collapsed group and click on Atom we expect to replace the group with Atom. This behavior works fine for replacing atoms, but not for groups. 
2) by a hotkey trigger for the same scenario (just without clicking on the toolbar).
@AnastasiiaPlyako ? 

If so - I think that your first idea of placing this inside the Atom tool handler was correct, it is Atom tool's responsibility to replace any other elements with the selected atom.

PR #2586, opened 2023-05-09
#2585 Change behaviour of "Show hydrogen labels" setting so the "on" value works the same way as "terminal and hetero"

### LINKED ISSUE #2585: Change behaviour of "Show hydrogen labels" setting so the "on" value works the same way as "terminal and hetero"
**Background**
In previous versions the "ON" setting for "Show hydrogen labels" did not work correctly (worked the same way "Terminal and Hetero" should work) As part of #2245 the behaviour for "ON" option was fixed, but it was previously selected as default behaviour, so all users who have updated any settings in the previous versions will have this setting set to "ON", while the default should be "Terminal and Hetero"

**Actual behavior**
When the setting is "ON" all atoms show labels and hydrogens

**Expected behavior**
When setting is "ON" it should show labels and hydrogens only for terminal and hetero atoms (same as "Terminal and Hetero" option works now)
Option "ALL" is added that makes it show labels and hydrogens for all atoms.

**Proposed solution**
1. All current options for "Show hydrogen labels" work the same way as they do now, except for "On" option
2. "On" option is saved as "all" when settings are updated
3. If the user has "on" value in the saved settings - interpret it as "Terminal and Hetero" option

**Ketcher version**
v2.10.0-rc.2

KonstantinEpam23:
Closes #2585

### DIFF
--- a/packages/ketcher-core/src/application/render/restruct/reatom.ts
+++ b/packages/ketcher-core/src/application/render/restruct/reatom.ts
@@ -51,7 +51,8 @@ export enum ShowHydrogenLabels {
   Hetero = 'Hetero',
   Terminal = 'Terminal',
   TerminalAndHetero = 'Terminal and Hetero',
-  On = 'on'
+  On = 'on',
+  All = 'all'
 }
 
 class ReAtom extends ReObject {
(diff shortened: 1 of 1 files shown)


yuleicul (approved):
LGTM

PR #2588, opened 2023-05-09
#2585 Change behaviour of "Show hydrogen labels" setting so the "on" value works the same way as "terminal and hetero"

KonstantinEpam23:


KonstantinEpam23:
Had to re-open the pull request, code was updated for new requirements

KonstantinEpam23:
Currently the schema will reset to default value if the provided one is not in the enum. So users who have "on" in the saved settings will have the value set to default from the schema, which is "Terminal and Hetero" at the moment.

gairon:
Looks good

PR #2604, opened 2023-05-15
#2127 Change esc behaviour logic for selection tool

### LINKED ISSUE #2127: Change esc behaviour logic for selection tool
At the moment pressing `esc` on the keyboard allows to change currently active mode of selection tool.

**Describe the solution you'd like**
What needs to be changed:
* It is only allowed to choose selection tool mode (rectangle, lasso, structural) by mouse
* If user closes Ketcher and then opens again, the last chosen selected tool must be selected (done in #1818)
* If user presses `esc`, then last chosen selected tool must be selected and pressing `esc` doesn't choose another mode of selection tool

**Additional notes**
Other tools with dropdown share the same logic. New behaviour must only be applied for selection tool

VasilevDO:
Closes #2127

### DIFF
--- a/packages/ketcher-react/src/script/ui/state/hotkeys.ts
+++ b/packages/ketcher-react/src/script/ui/state/hotkeys.ts
@@ -121,8 +121,7 @@ function keyHandle(dispatch, state, hotKeys, event) {
         })
       } else {
         if (newAction.tool === 'select') {
-          const savedSelectionTool = SettingsManager.getSettings().selectionTool
-          newAction = savedSelectionTool
+          newAction = SettingsManager.getSettings().selectionTool
         }
 
         dispatch(onAction(newAction))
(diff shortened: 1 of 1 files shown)


packages/ketcher-react/src/script/ui/state/hotkeys.ts
```diff
@@ -119,6 +120,11 @@ function keyHandle(dispatch, state, hotKeys, event) {
           dispatch
         })
       } else {
+        if (newAction.tool === 'select') {
+          const savedSelectionTool = SettingsManager.getSettings().selectionTool
```
AnastasiiaPlyako, line 124: Could we avoid creating an extra variable and write just:
`newAction = SettingsManager.getSettings().selectionTool`?

AnastasiiaPlyako (approved):

PR #2607, opened 2023-05-15
#2512 Toolbar arrow is active after the Ketcher with iFrame is loaded and disappears when you click on it

### LINKED ISSUE #2512: Toolbar arrow is active after the Ketcher with iFrame is loaded and disappears when you click on it
**Steps to Reproduce**
1. Open Ketcher with iFrame
2. Set Height to 632
3. Click 'Open modal'
4. Look at left toolbar navigation arrow
5. Click on arrow

**Actual behavior**
Toolbar arrow is active even though all buttons are placed in the iFrame and there's no need to scroll.
When you click on the arrow, it disappears and there is no way to scroll

**Expected behavior**
Toolbar arrows not activates because toolbar buttons fit into the iFrame so there's no need to scroll. 

**Screenshots**

https://user-images.githubusercontent.com/62464241/233634115-1c6deee1-20a2-47a1-9c22-a6f2f00385d9.mp4

 - OS: Windows 10
 - Browser Chrome
 - Version 112.0.5615.138 (Official Build) (64-bit)

**Ketcher version** 
[Version 2.10.0-rc.1](https://lifescience.opensource.epam.com/ketcher/index.html)

#184993424

VasilevDO:
Closes #2512

Details:
1) I have set the visibility threshold for the "scroll up" and "scroll down" reference elements to 1 (100%). This means that we consider the reference element as visible only when the entire element is within the view area. 
2) To address potential issues with the right toolbar, I have extracted the first basic atom into a separate element. This allows us to link the "scroll up" reference to a single button. If we were to leave the reference on the entire "atoms list" group, it could cause problems when the height values are low. In such cases, the height of the "atoms list" group would be greater than the actual view area height, resulting in the "scroll up" toolbar arrow persisting permanently.
3) The "arrow scroll" component appears only when the height of the toolbar content is greater than the height of the view area.

Corner iFrame heights:
Left toolbar: 609-610px
Right toolbar: 499-500px

### DIFF
--- a/packages/ketcher-react/src/script/ui/views/toolbars/LeftToolbar/LeftToolbar.module.less
+++ b/packages/ketcher-react/src/script/ui/views/toolbars/LeftToolbar/LeftToolbar.module.less
@@ -43,8 +43,6 @@
 }
 
 .groupItem {
-  margin: 2px;
-
   // TODO: move to action button styles, when all buttons will have same style
   svg {
     right: 2px;
--- a/packages/ketcher-react/src/script/ui/views/toolbars/LeftToolbar/LeftToolbar.module.less
+++ b/packages/ketcher-react/src/script/ui/views/toolbars/LeftToolbar/LeftToolbar.module.less
@@ -43,6 +43,8 @@
 }
 
 .groupItem {
+  margin: 2px 0;
+
   // TODO: move to action button styles, when all buttons will have same style
   svg {
     right: 2px;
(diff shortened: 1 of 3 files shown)


yuleicul (commented):
Hi @VasilevDO , thanks for your job, and it was great that you even considered the potential issues with the right toolbar! 👍 
My only concern is, according to the expected behavior in the issue description, the Toolbar arrows would be hidden when the toolbar buttons just right fit into the iFrame (for example, when the height is 632). 

> Expected behavior
> Toolbar arrows not activates because toolbar buttons fit into the iFrame so there's no need to scroll.

So please take a look at this. If I get it wrong, let me know. 😃

VasilevDO:
> Hi @VasilevDO , thanks for your job, and it was great that you even considered the potential issues with the right toolbar! 👍 My only concern is, according to the expected behavior in the issue description, the Toolbar arrows would be hidden when the toolbar buttons just right fit into the iFrame (for example, when the height is 632).
> 
> > Expected behavior
> > Toolbar arrows not activates because toolbar buttons fit into the iFrame so there's no need to scroll.
> 
> So please take a look at this. If I get it wrong, let me know. 👀

Hi, @yuleicul 
I have found a better approach to address this issue, so I have made changes to the code and updated the details of this PR :)

yuleicul (approved):

Issue #2576, opened 2023-05-05
Right click menu shown for highlighted Functional Group instead of clicked


Zhirnoff:
**Steps to Reproduce**

**Scenario1**
1. Open Ketcher
2. Add two Functional Groups on canvas ( one expand )
3. Select with Selection tool expanded Functional Group
4. Make right click on contracted Functional Group abbreviation

**Actual behavior**
The right click menu is shown for highlighted Functional Group instead of the clicked Functional Group.

**Expected behavior**
When right clicking on a Functional Group that is not currently highlighted, the right click menu should be shown for the clicked Functional Group.

**Screenshots**

https://user-images.githubusercontent.com/62464241/236462235-1cd49b85-6bc5-4275-abd4-3e1256a91d53.mp4

**Scenario2**
1. Open Ketcher
2. Add two atoms on canvas
3. Select with Selection tool one atom
4. Make right click on another atom

**Actual behavior**
Context menu is shown as if both atoms were selected and from right side menu for bonds
![2023-05-05_16h11_35](https://user-images.githubusercontent.com/62464241/236469239-a361d8e8-4cd5-4084-84d9-f259a46ed82d.png)

**Expected behavior**
When right clicking on an atom that is not currently highlighted, the right click menu should be shown for the clicked atom.
![2023-05-05_16h11_54](https://user-images.githubusercontent.com/62464241/236469421-e0a72793-c437-4360-b080-99cabffa1c35.png)

**Scenario3**
1. Open Ketcher
2. Add structure on canvas (e.g. Benzene)
3. Select with Selection tool all structure
4. Make right click on any place on canvas

**Actual behavior**
Context menu is shown and structure is highlight

**Expected behavior**
Context menu is not shown and structure is unhighlight

**Scenario4**
1. Open Ketcher
2. Add two structures on canvas (e.g. Benzene)
3. Select with Selection tool one of structure
4. Make right click on atom of not highlighted structure

**Actual behavior**
Context menu is shown for selected structure and structure is highlight

**Expected behavior**
Context menu is shown for clicked atom and selected structure is unhighlight

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser Chrome
 - Version 112.0.5615.138 (Official Build) (64-bit)

**Ketcher version** 
[Version 2.10.0-rc.2](https://lifescience.opensource.epam.com/ketcher/index.html)
Note: work in v2.8.0

yuleicul:
[...]

PR #2613, opened 2023-05-17
#2576 - Right click menu shown for highlighted functional group instead of clicked

gairon:
* Show context menu for the closest item only if there is no selected item(s)
* Show context menu for a selection only if a user clicked on an item included in the selection
* Otherwise reset the current selection and show the context menu for the closest item
* Clear selection if a user clicks RMB outside of selection and other items
* Remove parentheses from menu items for General selection context menu
* Show "Edit selected atom..." menu item if atoms and something except bonds and Functional Groups are selected
* Show "Edit selected bonds..." menu item if bonds and something except atoms and Functional Groups are selected

### DIFF
--- a/packages/ketcher-react/src/script/ui/views/components/ContextMenu/ContextMenuTrigger.tsx
+++ b/packages/ketcher-react/src/script/ui/views/components/ContextMenu/ContextMenuTrigger.tsx
@@ -92,7 +92,11 @@ const ContextMenuTrigger: React.FC<PropsWithChildren> = ({ children }) => {
           selectedSGroupsIds
         })
       ) {
-        triggerType = ContextMenuTriggerType.Selection
+        if (!selection.bonds && !selection.atoms) {
+          triggerType = ContextMenuTriggerType.None
+        } else {
+          triggerType = ContextMenuTriggerType.Selection
+        }
       } else {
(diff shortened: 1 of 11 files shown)


packages/ketcher-react/src/script/ui/views/components/ContextMenu/ContextMenuTrigger.tsx
```diff
@@ -45,9 +53,18 @@ const ContextMenuTrigger: React.FC<PropsWithChildren> = ({ children }) => {
...
+
+      const sGroupId = struct.sgroups.find((_, sGroup) =>
+        sGroup.atoms.includes(atomId)
+      )
+
+      sGroupId !== null && selectedSGroupsIds.add(sGroupId)
     })
 
-    return selectedFunctionalGroups
+    return {
```
AnastasiiaPlyako, line 64: if the method now returns functional and sgroup, maybe rename the function? For example, just "getSelectedGroups"

gairon: Renamed to `getSelectedGroupsInfo`

AnastasiiaPlyako (approved):

Issue #2632, opened 2023-05-19
Refactoring request: improve Editor tools in 'ketcher-react'


kaluginserg:
Improve code quality for editor tools.
Place: `/packages/ketcher-react/src/script/editor/tool`

What to do:
1) create interface `AbstractTool` with optional methods like: 'mousedown', 'mousemove', 'cancel' ...

2) for each tool in the folder implement the interface

3) in the file Editor `/Users/sergei_kalugin/Developer/ketcher/packages/ketcher-react/src/script/editor/Editor.ts`
- add a new type for the '_tool' property.
- update related interfaces and properties

Affected scope: folder `/packages/ketcher-react/src/script/editor`

PR #2635, opened 2023-05-19
#2632 Refactoring request: improve Editor tools in 'ketcher-react'

kaluginserg:
- in the folder `/packages/ketcher-react/src/script/editor/tool` 
  - create interface AbstractTool with optional methods like: 'mousedown', 'mousemove', 'cancel' ...
  - for each tool in the folder implement the interface
  - for each tool make the strictest modifiers for the private properties and methods
  - remove unused properties
- in the file Editor `/packages/ketcher-react/src/script/editor/Editor.ts`
  - add a new type for the '_tool' property.
  - update related interfaces and properties


packages/ketcher-react/src/script/editor/tool/AbstractTool.ts
```diff
@@ -0,0 +1,35 @@
...
+
+export interface AbstractTool extends ToolEventHandler {
+  close?(): void
+
+  cancel?(): void
+
+  isSelectionRunning?(): boolean
+
+  isNotActiveTool?: boolean
+}
```
yuleicul, line 29: I'm wondering why `close()`, `isSelectingRunning()`, `isNotActiveTool()` should be in AbstractTool. I thought that not every tool needed them, right?

packages/ketcher-react/src/script/editor/tool/AbstractTool.ts
```diff
@@ -0,0 +1,35 @@
...
+  mouseup?(event: Event): void
+
+  mouseleave?(event: Event): void
+
+  mouseLeaveClientArea?(event: Event): void
+
+  mouseover?(event: Event): void
+}
+
+export interface AbstractTool extends ToolEventHandler {
```
yuleicul, line 21: I thought TypeScript had a keyword [abstract](https://www.typescriptlang.org/docs/handbook/2/classes.html#abstract-classes-and-members) which was specific for abstract class. 

I'm not sure if we can:
```suggestion
abstract class ToolEventHandler {}
abstract class Tool extends ToolEventHandler {}
```

I thought the advantage of `interface` was there would be no extra JavaScript output, but one good thing about `abstract` was if we forget to implement the base class’s abstract members, we’ll get an error.

What's your take on this?

packages/ketcher-react/src/script/editor/tool/AbstractTool.ts
kaluginserg: actually, it `isSelectingRunning()` and `isNotActiveTool` is something that is needed for every tool, as it is an interface used from the outside to control abstract tool.

Example `Editor.ts:633`:
```
  const EditorTool = editor.tool()
  ....
    clientArea.contains(event.target) || EditorTool.isSelectionRunning?.(),
  ]
```

Example `Editor.ts:234`:
```
if (!tool || tool.isNotActiveTool) {
```


Even I think, we should mark those methods as required and implement them in each tool. Consistency leads to simplicity.

Not sure about `.close()` - can't find the case... Looks like it is my mistake.

@yuleicul what do you think about it?

packages/ketcher-react/src/script/editor/tool/AbstractTool.ts
kaluginserg: I doubt using the base class as it often tends to use real inheritance with class extends and common logic, which is often tough to manage.
I agree that the 'Abstract' word could lead to misunderstanding. 
So my proposal is to rename `AbstractTool` to `Tool` and use it as is as a common interface.

@yuleicul let me know if you have other ideas

packages/ketcher-react/src/script/editor/tool/AbstractTool.ts
yuleicul: Thanks for your examples, make sense

yuleicul (approved):

Issue #2471, opened 2023-04-10
When you try to add a structure to a canvas with an expanded functional group, an EDIT ABBREVIATION window appears


Zhirnoff:
**Steps to Reproduce**
1. Launch Ketcher
2. Open file [ketcher.zip](https://github.com/epam/ketcher/files/11190114/ketcher.37.zip)
3. Try put it on canvas

**Actual behavior**
The structure is under mouse cursor and when you try to put it on canvas, EDIT ABBREVIATION window appears.
Note: Perhaps this is due to the fact that the Oxygen atom is located close to the Carbon atom. The Oxygen atom is highlighted and, apparently, when clicked on canvas, editing of the abbreviation is triggered.
**Expected behavior**
A structure with an expanded Functional Group is added to the canvas

**Screenshots**
![2023-04-10_14h25_04](https://user-images.githubusercontent.com/62464241/230892822-98b99b74-17d4-42f7-acc1-1f3f2c8e1af6.png)
![2023-04-10_14h23_58](https://user-images.githubusercontent.com/62464241/230892851-a4ded299-c591-47b4-9cad-4df8c10bfa87.png)

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser Chrome
 - Version 109.0.5414.120 (Official Build) (64-bit)

**Ketcher version** 
[Version 2.9.0-rc.5](https://lifescience.opensource.epam.com/ketcher/index.html)

kaluginserg:
this bug is affected by changes https://github.com/epam/ketcher/issues/2110. Continue investigating.

kaluginserg:
Found the details. 
I checked how Ketcher works now, for a couple of cases. 
Screenshot:
![Screenshot 2023-05-22 at 16 29 15](https://github.com/epam/ketcher/assets/6398891/e27cb4a2-979c-4438-815a-61ab014d5ec7)

A) Whole structure is from the clipboard (PASTE tool triggered). Oxygen and Carbon are not considered to be merged, as there is a space between them. 
So the 'Paste' functionality works fine.

B) Whole structure is from the clipboard (PASTE tool triggered). [Incorrect current implementation] Oxygen and Carbon are considered to be merged (there is a highlight for AttachmentPoint).
So the 'Paste' functionality doesn't work well for this case.

The decision, TODO:
[...]

Issue #2110, opened 2023-01-19
Incorrect merging of Functional Groups


yuleicul:
**Steps to Reproduce**
1. Use the select tool to select a functional group or paste a functional group from clipboard
2. Drag it close to other structure (such as a benzene)
3. Click on the canvas to insert the functional group

**Actual behavior**
After step 2, there may be multiple highlighted green circles that indicate atoms to be merged, and the circles correspond to the atoms **inside** the **expanded** functional group.
After step 3, there may be multiple functional groups being merged.
Additionaly, the group may get merged with other structures that are in close proximity.

**Expected behavior**
Contracted functional group should ignore all atoms apart form attachment point when dragging (and/or pasting) 

**Additional info**
It looks like a contracted functional group is being treated as an expanded one while it is being dragged along the canvas, hence it is showing multiple connection points wherever atoms of an expanded group are located.
Based on the reproduction steps:
After step 2, there will be only one highlighted green circle that indicates the unique attachment atom of the functional group.
After step 3, there will be only one functional group being merged and it will be merged only to the structure that is highlighted

*Example of correct merge of functional group:*
![image](https://user-images.githubusercontent.com/41272128/213690068-dc98535d-d763-44f2-b8e1-9653216db1f8.png)

*Salts or solvents are special case of this behaviour and they are not supposed to be merged at all. Example of adding salt or solvent to canvas over benzene ring:*
![image](https://user-images.githubusercontent.com/41272128/213690241-25779312-685e-4a81-a09a-4f7b320a8a0e.png)

**Screenshots**
![ketcher-paste-bug-06](https://user-images.githubusercontent.com/27288153/213422674-f4946c29-739b-4321-af7f-8cfb228cf7e7.gif)

**Ketcher version**
v2.7.0 (remote rc version)

**Additional context**
Two related bugs:
- https://github.com/epam/ketcher/pull/2079#issuecomment-1387044291
- https://github.com/epam/ketcher/pull/2079#issuecomment-1396169423

PR #2643, opened 2023-05-22
#2471 – When you try to add a structure to a canvas with an expanded functional group, an EDIT ABBREVIATION window appears

kaluginserg:
Removed implementation of function [getMergeItems.ts](https://github.com/epam/ketcher/pull/2643/files#diff-072d72f1eeecad47a3ecc8cbeeda6333a78820982c35c7ad7b7181781b49c443) and changed to another approach.

Why?
The getMergeItems - function was incorrect, it filtered past items that finally result in incorrect behavior. We already have the commonly used approach with [getItemsToFuse](https://github.com/epam/ketcher/blob/c8f9479518996d448bfe4a6b231ad67056cceab1/packages/ketcher-core/src/application/editor/actions/closelyFusing.ts#L51) function.
It works correctly.

Detailed explanation in the issue thread: [comment1](https://github.com/epam/ketcher/issues/2471#issuecomment-1557157122), [comment2](https://github.com/epam/ketcher/issues/2471#issuecomment-1557270993)


yuleicul (changes requested):
@kaluginserg Thanks for your detailed comments! Unfortunately, removing `getMergeItems.ts` causes issue #2110 again. I recorded different behaviors between your branch and rc, please take a look


https://github.com/epam/ketcher/assets/27288153/44359786-804e-47f7-8fa2-03bf6f895239

yuleicul (approved):

KonstantinEpam23 (approved):

PR #2666, opened 2023-05-24
#2584 #2750 - Selection tool: add flip and delete buttons & Rotation Tool: non-selected end of the selected bond should be the rotation center

### LINKED ISSUE #2584: Selection tool: add flip and delete buttons
Add 'flip horizontally', 'flip vertically' and 'delete' buttons to the selection tool. 

Flip and delete buttons

### LINKED ISSUE #2750: Rotation Tool: non-selected end of the selected bond should be the rotation center
One end of selected bond is selected, while the other end is non-selected, in this situation, the rotation center should be the non-selected end.

image

yuleicul:
Closes #2584 
Closes #2750 

# What's new
## Floating buttons 

image

- add new floating buttons for flipping and deleting structures.
- remove rotate tool, horizontal flip tool, and vertical flip tool from the left toolbar
- remove rotation's shortcut _alt+r_
- delete rotation icon from codebase

## New flip support

- Support flipping multiple structures as a whole
- Support flipping reaction arrows, pluses, and texts

more details:

Frame 1ketcher-flip-axis (7)


packages/ketcher-react/src/script/ui/views/toolbars/FloatingTools/FloatingTools.container.ts
```diff
@@ -0,0 +1,35 @@
...
+} from './FloatingTools'
+
+import { connect } from 'react-redux'
+import { Dispatch } from 'redux'
+import type RotateController from 'src/script/editor/tool/rotate-controller'
+import { onAction } from '../../../state'
+
+const mapStateToProps = (state: any): FloatingToolsProps => {
+  const rotateController: RotateController | undefined =
+    state.editor?.rotateController
```
kaluginserg, line 14: I'm not sure that this is a good practice to use a mutable object (`state.editor`) as a source of reactions.
The `state.editor` is probably added to the redux store by mistake or for debugging. 

Could we try to set up the new state like 'state.floatingToolsVisible' or something like that, and subscribe to events from the editor to update this state? So in this case Redux will handle changes in the proper way.

packages/ketcher-react/src/script/ui/views/toolbars/ToolbarGroupItem/ToolbarGroupItem.tsx
```diff
@@ -33,7 +33,7 @@ interface ToolbarGroupItemProps extends ToolbarItem {
 
 interface ToolbarGroupItemCallProps {
   onAction: (action: UiActionAction) => void
-  onOpen: (menuName: string, isSelected: boolean) => void
+  onOpen?: (menuName: string, isSelected: boolean) => void
```
kaluginserg, line 36: why did we change the common interface and make it less strict?
it will affect all other toolbars that depend on this interface.

packages/ketcher-react/src/script/ui/views/toolbars/FloatingTools/FloatingTools.container.ts
```diff
@@ -0,0 +1,32 @@
...
+import {
+  FloatingTools,
+  type FloatingToolsCallProps,
+  type FloatingToolsProps
+} from './FloatingTools'
+
+const mapStateToProps = (state: any): FloatingToolsProps => {
+  return {
+    visible: getFloatingToolsVisible(state),
+    HandlePos: getFloatingToolsHandlePos(state),
```
kaluginserg, line 17: curious why it is capitalised? Looks like it is not a common convention, any reason?

packages/ketcher-core/src/application/render/raphaelRender.js
```diff
@@ -96,6 +96,17 @@ Render.prototype.page2obj = function (event) {
...
 }
 
+/**
+ * @param {Vec2} coord Vector in Raphael coordinate system,
+ *                     scaled atom's position with offset
+ * @returns {Vec2} Vector in View coordinate system,
+ *                 coordinate relative to viewport (excluding toolbar)
+ */
+Render.prototype.raphael2View = function (coord) {
+  const obj = coord.sub(this.options.offset).scaled(1 / this.options.scale)
```
kaluginserg, line 106: propose to rename 'obj' to 'vec' or 'vector', as it could be a more specific name.

packages/ketcher-core/src/application/render/raphaelRender.js
```diff
@@ -96,6 +96,17 @@ Render.prototype.page2obj = function (event) {
   return this.view2obj(pp)
 }
 
+/**
+ * @param {Vec2} coord Vector in Raphael coordinate system,
+ *                     scaled atom's position with offset
+ * @returns {Vec2} Vector in View coordinate system,
+ *                 coordinate relative to viewport (excluding toolbar)
+ */
+Render.prototype.raphael2View = function (coord) {
```
kaluginserg, line 105: naming for 'raphael2View' is not clear, from my side it looks like it does not reflect the reason and the meaning. Also for me, it is hard to recognize from implementation - what is the difference with the .obj2view(vec)? 
Could you clarify?

packages/ketcher-react/src/script/ui/state/floatingTools/index.ts
```diff
@@ -0,0 +1,63 @@
+import { Reducer } from 'react'
+import { Vec2 } from 'ketcher-core'
+
+export interface FloatingToolsState {
+  visible: boolean
+  handlePos: Vec2
```
kaluginserg, line 6: 1) Could we define a more specific name for this state `handlePos`, from the usage - as I can see - it is more like: `toolBarPosition`? 

2) I would suggest making the interface stricter, like that: {x: number, y: number}, because it would be better to not allow to access 'Vec2' properties (that is related to renderer) from React level.

packages/ketcher-react/src/script/ui/views/components/StructEditor/StructEditor.jsx
```diff
@@ -212,6 +215,17 @@ class StructEditor extends Component {
...
+            onScroll={throttle(40, () => {
+              if (!this.editor.rotateController.handle) {
+                return
+              }
+              const handleCenter = this.editor.rotateController.handleCenter
+              const handleCenterInViewport =
+                this.editor.render.raphael2View(handleCenter)
+              this.editor.event.updateFloatingTools.dispatch(
+                updateFloatingToolsPos(handleCenterInViewport)
+              )
```
kaluginserg, line 227: Could I propose moving the implementation of handling the scrolling inside the Editor.ts to an additional public method, so this logic will be encapsulated close to rotateController and rendering?

packages/ketcher-react/src/script/editor/tool/rotate-controller.ts
```diff
@@ -27,15 +31,15 @@ const RIGHT_ARROW_PATH =
...
   private editor: Editor
   private rotateTool: RotateTool
   private originalCenter!: Vec2
   private normalizedCenterInitialHandleVec!: Vec2
-  private handleCenter!: Vec2
   private initialRadius!: number
   private isMovingCenter!: boolean
 
-  private handle?: RaphaelElement
+  handle?: RaphaelElement
```
kaluginserg, line 42: it looks like making this property public is not very required, so we can make a new public method with a stricter interface, let's say you can implement scrolling logic inside this method even without exposing internal properties.

packages/ketcher-react/src/script/ui/views/toolbars/FloatingTools/FloatingTools.tsx
```diff
@@ -0,0 +1,60 @@
...
+  onAction
+}) => {
+  if (!visible) return null
+
+  return (
+    <div
+      className={classes.wrapper}
+      style={{ left: HandlePos.x, top: HandlePos.y }}
+    >
+      {FLOATING_TOOLS.map((id) => (
```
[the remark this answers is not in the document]
yuleicul: My bad, this is a typo

packages/ketcher-core/src/application/render/raphaelRender.js
yuleicul: The reason why I name it `obj` is the `obj` here has the same meaning as `obj` in `obj2View`. What `obj` means is not clear to me as well. I'll take a closer look into this in issue [#2697](https://github.com/epam/ketcher/issues/2697)

packages/ketcher-core/src/application/render/raphaelRender.js
yuleicul: Maybe `raphael2View` is not a good name, I'll rename it in [#2697](https://github.com/epam/ketcher/issues/2697) if needed. As for something like `obj2View`, `page2Obj`, etc. I don't have an accurate idea of what they are, but I guess they are different coordinate systems, like

packages/ketcher-react/src/script/ui/views/toolbars/FloatingTools/FloatingTools.container.ts
yuleicul: My bad, this is a typo. Thank you so much!

packages/ketcher-react/src/script/ui/state/floatingTools/index.ts
yuleicul: `handlePos` is the center position of the rotation handle. 


I'd rename it to `rotateHandlePosition`.

packages/ketcher-react/src/script/editor/tool/rotate-controller.ts
yuleicul: Such a great idea, thank you!

packages/ketcher-core/src/application/render/raphaelRender.js
kaluginserg: I discovered - it is a Vec2 instance, Vec2 - itself is a vector class with some Vector calculus abilities.

packages/ketcher-react/src/script/editor/tool/rotate-controller.ts
```diff
@@ -27,11 +32,11 @@ const RIGHT_ARROW_PATH =
 
 class RotateController {
   isRotating!: boolean
+  handleCenter!: Vec2
```
kaluginserg, line 35: is it still required to be public? or it can be 'private'

packages/ketcher-core/src/application/render/raphaelRender.js
kaluginserg: @yuleicul thanks for the thorough analysis. 
But my point is - if we merge it and release it, our users will be able to use this method ()it is a public API of ketcher-core functionality, isn't it? And later if we will try to rename it we will have to do a deprecation process. 
Probably for now, it would be better to hide this ability somehow from the public to be able to refactor without issues.

packages/ketcher-react/src/script/ui/state/floatingTools/index.ts
kaluginserg: Second idea was about changing the interface, 
from the `Vec2` to `{x: number, y: number}`

why?
Because we don't need all internal logic from the Vec2 in our UI - correct?
I see only the plase where it is implemented:
`style={{ left: rotateHandlePosition.x, top: rotateHandlePosition.y }}` 

So we can do it, and in this case, late we can more easily refactor the implementation because we will have simpler interface.

packages/ketcher-react/src/script/editor/tool/rotate-controller.ts
yuleicul: My bad, ignored it

yuleicul:
After merging #2680 into this branch, another issue occurs:

At this moment, arrows are not allowed to flip

    rotate-arrow-floating
    
    I propose to:
    support flipping arrows in the scope of this pull request
               OR 
    disable flipping buttons for arrows (and maybe create a new issue to support flipping)

@KonstantinEpam23 @Nitvex What do you think? 💭

packages/ketcher-react/src/script/editor/tool/rotate-controller.ts
```diff
@@ -1,5 +1,10 @@
 import { Action, Scale, Vec2 } from 'ketcher-core'
 import { throttle } from 'lodash'
+import {
+  hideFloatingTools,
+  showFloatingTools,
+  updateFloatingToolsPosition
+} from 'src/script/ui/state/floatingTools'
```
kaluginserg, line 7: just noticed, if possible we should avoid coupling the `Editor` and the UI redux event.

So we have mapped in the `packages/ketcher-react/src/script/ui/state/editor/index.js` and in the current implementation you have: 
```
onUpdateFloatingTools: memoizedDebounce((action) => {
  dispatch(action)
})
```
but probably we should do smth like that so, real mapping:
```
onUpdateFloatingTools: memoizedDebounce((params: UpdateFloatingToolsData) => {
if(params ....)
  dispatch(hideFloatingTools())
else ...
   dispatch(showFloatingTools(...))
else ...
 dispatch(updateFloatingToolsPosition(...))
})
```

In this case, UI will know about the interface of the Event of `Editor` (`UpdateFloatingToolsInterface`), but `Editor` will not know about UI logic.

KonstantinEpam23:
> After merging #2680 into this branch, another issue occurs:
> At this moment, arrows are not allowed to flip
> I propose to: support flipping arrows in the scope of this pull request OR disable flipping buttons for arrows (and maybe create a new issue to support flipping)

Hi @yuleicul! I believe we do need to support flipping arrows for reactions (especially since we already support rotating them)
Flipping arrow by itself is probably a rare usecase, but if there are other structures in the selection - we cant disable the flip buttons and it might be confusing for the user if the arrows dont flip.

I would propose to support flipping arrows in the scope of this pull request, unless it is hard to implement.

packages/ketcher-react/src/script/editor/Editor.ts
```diff
@@ -37,6 +37,7 @@ import { toolsMap } from './tool'
 import { Highlighter } from './highlighter'
 import { setFunctionalGroupsTooltip } from './utils/functionalGroupsTooltip'
 import { contextMenuInfo } from '../ui/views/components/ContextMenu/contextMenu.types'
+import { FloatingToolsPayload } from '../ui/state/floatingTools'
```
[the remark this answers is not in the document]
yuleicul: I think it'd be better to put `FloatingToolsPayload` in `ui/state`, because `FloatingToolsPayload` is defined along with function `updateFloatingTools` which exists in `ui/state`. I'm curious what your understanding of producer and consumer is, because I'm confused about them

kaluginserg (approved):

kaluginserg:
@yuleicul thanks for your work, it is a good solution.
I saw you've started working on additional functionality for choosing the rotation center.... 
Could we please do it in a separate PR?  
It is not a good practice to have long-live PRs and also it is not a good practice to have PR with various changes.
Why? Using small pull requests can lead to more efficient code reviews, faster feedback, fewer merge conflicts, and a more stable and maintainable codebase.

yuleicul:
@kaluginserg Thanks for your review! Actually, the reason why I close two issues in this PR is because the two issues have overlapped code, if I open separate PRs, it may be hard to resolve conflicts.

But anyway, thanks for your kind reminder, I'll pay more attention next time.

PR #2713, opened 2023-05-30
#2660 Can't attach copied or cutted Functional Group to atoms of structure

### LINKED ISSUE #2660: Can't attach copied or cutted Functional Group to atoms of structure
**Steps to Reproduce**
1. Open Ketcher
2. Add benzene ring to canvas
3. Add contracted Functional Group to canvas
4. Select Functional Group and copy (CTRL+C) then press (CTRL+V) or cut it (CTRL+X then CTRL+V)
5. Click on any atom of Benzene ring

**Actual behavior**
Can't attach copied or cutted Functional Group to atoms of structure

**Expected behavior**
Can attach copied or cutted Functional Group to atoms of structure

**Screenshots**

https://github.com/epam/ketcher/assets/62464241/11294ae0-96c9-4187-a328-a0a24196aff6

 - OS: Windows 10
 - Browser Chrome
 - Version 112.0.5615.138 (Official Build) (64-bit)

**Ketcher version** 
[Version 2.10.0-rc.5](https://lifescience.opensource.epam.com/ketcher/index.html)

kaluginserg:
Resolves #2660 

What was done:
- Tool Paste 
    - removed all logic related to drag functionality (it didn't work, so without proper tests and naming it's just trash)
    - removed flag `isSingleContractedGroup` as it is not working as expected, finally in the current implementation it works only when we past group on group, but if we try to past Atom on group or structure on structure, or group on atom - it's not working. So the decision is to have it as is, and later fix all issues with Past tool at once with the new implmementation.
- add the helper function `filterSlatAndSolventFromMerge()` in the tools folder. It's the short-term solution as in the better way it must be inside the ketcher-core `closest.merge` (the same as `findCloseMerge`), and removed from the tools level.
- small refactoring for related files (adding typescript types and interfaces)

How it works now (On the video: begining from Template, then using select tool, and after that using Paste tool):

https://github.com/epam/ketcher/assets/6398891/d80df0be-a16c-4fbd-9098-9c78da6259a6

It is still a difference in the behavior of the tools:
- Template Tool - the group is merged as collapsed
- Select tool - the group is merged as expanded
- Paste tool - the group is merged as expanded


packages/ketcher-core/src/domain/entities/sgroup.ts
```diff
@@ -240,7 +240,7 @@ export class SGroup {
     return Vec2.diff(sgroup.pp, sgroup.bracketBox.p1)
   }
 
-  static isSaltOrSolvent(moleculeName: string): boolean {
+  static isSaltOrSolvent(moleculeName?: string): boolean {
```
gairon, line 243: Do we need to make it optional?

kaluginserg: actually yes we need in the current implementation at least here:
```js
const isDraggingSaltOrSolventOnStructure = SGroup.isSaltOrSolvent(
      possibleSaltOrSolvent?.item.data.name
    )
```

And also it looks like it could be more convenient, because I see several places like that:
```js
!SGroup.isSaltOrSolvent(functionalGroup?.name || '')
```
That is not necessary to use || '', because implementation of the function works well with undefined value.

AlexeyGirin:

AlexeyGirin:

Issue #2717, opened 2023-05-30
Tooltip for S-Group appears in upper left area of canvas


Zhirnoff:
**Steps to Reproduce**
1. Open Ketcher
2. Create structure and make Data S-Group with 'Field name' and 'Field value'
3. With 'Selection tool' hover over created S-Group

**Actual behavior**
Tooltip for S-Group appears in upper left area of canvas
![2023-05-30_16h57_13](https://github.com/epam/ketcher/assets/62464241/86553d31-731c-4779-8913-65d31743a5f9)

**Expected behavior**
Tooltip for S-Group appears under S-Group structure

![2023-05-30_17h00_43](https://github.com/epam/ketcher/assets/62464241/dfd38006-28d5-4837-9258-db956c13ed2a)

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser Chrome
 - Version 112.0.5615.138 (Official Build) (64-bit)

**Ketcher version** 
[Version 2.11.0-rc.1](https://lifescience.opensource.epam.com/ketcher/index.html)

PR #2723, opened 2023-05-31
#2717 Tooltip for S-Group appears in upper left area of canvas

KonstantinEpam23:


packages/ketcher-core/src/domain/entities/sgroup.ts
```diff
@@ -632,7 +632,14 @@ export class SGroup {
     if (!sGroup) {
       return false
     }
-    return sGroup.type === 'SUP'
+    return sGroup.type === SGroup.TYPES.SUP
+  }
+
+  static isDataSGroup(sGroup) {
```
Nitvex, line 638: I propose to add types here. 
And do we need this `if` condition?

KonstantinEpam23: In a way this function replaces sgroup?.type === 'DAT', changed this to conditional chaining.

Nitvex (approved):

PR #2739, opened 2023-06-05
#2165 - Change attachment point visualization to IUPAC style

### LINKED ISSUE #2165: Change attachment point visualization to IUPAC style
**Background**
Current attachment point visualization is different from IUPAC (e,g, from rendering to PNG) 

**Solution**
* Make visualization of attachment points the same as in rendering to PNG/SVG
* The visual representation is for the R-group attachment points (APO) see the screenshots below
* if there is 1 bond (single or double) attached then angle should be 120
* if there is 1 triple bond attached, then angle should be 180
* there are 2+ bonds attached then make bisector between bonds
* The following rules for the show/not show label
if all primary attachment points => do not display
if at least one attachment point is secondary => show labels 1,2 near the curve line (see screenshots)

**Additional context**
All examples can be found in [examples.zip](https://github.com/epam/ketcher/files/10558047/examples.zip)

* example1 (1 attachment point)
image
* example2 (1 attachment point connected to single bond)
imagedd01e0244dd2.png)
* example3 (2 attachment points)
image
* example4 (1 triple bond attached)
image

#184325922

kaluginserg:
resolves #2165


yuleicul:
https://github.com/epam/ketcher/assets/27288153/e7925166-e3b9-4ff1-aa35-0752c534946e

Could you take a look at the three issues in the above recording:
1. After clicking _Save To Templates_, the attachment point touches the bounding (I think there would be some paddings)
2. When opening the Templates Modal to check out the saved template, the attachment point doesn't display in the preview
3. After choosing my saved template and clicking on the canvas, the number displays

I'm not sure the first and the second ones are somehow related to the ability to select attachment point which is out of scope of this PR according to your https://github.com/epam/ketcher/issues/2165#issuecomment-1577933560 .

yuleicul (commented):
https://github.com/epam/ketcher/assets/27288153/e7925166-e3b9-4ff1-aa35-0752c534946e

Could you please take a look at the three issues in the above recording:
1. After clicking _Save To Templates_, the attachment point touches the bounding. (I think there would be some paddings)
2. When opening the Templates Modal to check out the saved template, the attachment point doesn't display in the preview
3. After choosing my saved template and clicking on the canvas, the number displays.

I'm not sure if the first and the second ones are somehow related to the ability to select attachment point which is out of scope of this PR according to your https://github.com/epam/ketcher/issues/2165#issuecomment-1577933560 . 

And in terms of the third one, maybe it can be summarized as a situation where there are multiple attachment points on the **canvas**, but every structure has just one attachment point, like:

https://github.com/epam/ketcher/assets/27288153/60115f44-0c98-4651-85c3-1ddbce211155

kaluginserg:
@yuleicul thanks for checking this behavior.

1,2 - yes, you're right, they are connected to Attachment points to be as selectable and manipulable canvas entities along as atoms. In this case, we will be able to define better borders for them.

3 - interesting, it's really missed out. Looks like we don't have a calculation for each atom each time, we have incremental calculations only for changed atoms and bonds, so that's why it is not rerendered. BTW if we will try to drag an atom without a label - the attachment point will be rendered. So, in the current rendering mechanism, we can't do this requirement without bugs. As on the canvas level, we don't have reactivity (like in react) or signals.

packages/ketcher-core/src/application/render/restruct/reatom.ts
```diff
@@ -1109,6 +1095,36 @@ function pathAndRBoxTranslate(path, rbb, x, y) {
...
+}
+
+function newVectorFromAngle(angle: number): Vec2 {
+  return new Vec2(Math.cos(angle), Math.sin(angle))
+}
+
+function getLargestSectorFromNeighbors(
+  atom: ReAtom,
+  struct: Struct
+): { neighborAngle; largestAngle } {
```
yuleicul, line 1127: Maybe remove the return type here? This line makes the reture type `{ neighborAngle: any, largestAngle: any}`, but it has been implicitly infered as `{ neighborAngle: number, largestAngle: number}`

packages/ketcher-core/src/application/editor/actions/atom.ts
```diff
@@ -99,13 +100,15 @@ export function fromAtomsAttrs(restruct, ids, attrs, reset) {
...
     const atomNeighbors = restruct.molecule.atomGetNeighbors(atomId)
-    const bond = restruct.molecule.bonds.get(atomNeighbors[0]?.bid)
+    const bond = restruct.molecule.bonds.get(atomNeighbors?.[0]?.bid as number)
     if (bond) {
       action.mergeWith(fromBondStereoUpdate(restruct, bond))
     }
     // when a heteroatom connects to an aromatic ring it's necessary to add a ImplicitHCount
     // property to this atom to specify the number of hydrogens on it.
     const atom = restruct.molecule.atoms.get(atomId)
+    assert(atom != null)
```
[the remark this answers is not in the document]
kaluginserg: @yuleicul thanks for your suggestion.

I'm thinking... AFAIK '!=' condition is a commonly used practice in JS wolds, due to it checks not only undefined cases but also null cases. And it is very useful for such nullish checks. So I did it to be on the same page with commonly used practices. 

What do you think?

packages/ketcher-core/src/application/render/restruct/reatom.ts
kaluginserg: fixed

yuleicul (dismissed):
Please fix ci

yuleicul (approved):

PR #2741, opened 2023-06-06
#2689 – Make cyclopentadiene ring systems fusion without valence errors

### LINKED ISSUE #2689: Make cyclopentadiene ring systems fusion without valence errors
**Background**
Double cyclopentadiene ring systems create valence errors

**Solution**
Following rules to be applied when adding cyclopentadiene to the existing structure and **fusion is done using bond**

1. if connect to a single bond with two atoms, where each atom is connected to any atom with a double bond, then rotate ring and replace double bond by single bond for the fusion
image
image

2. If connect to a double bond with two atom, where each atom is connected to any atom with a single bond, then rotate ring and use double bond for the fusion
image
image

3. if connect to a single bond with two atoms, but one atom is connected with a single bond and another with a double bond then rotate and fuse by single bond in cyclopentadiene 
image

image

4. if all bonds are single then rotate and fuse by single bond in cyclopentadiene 

image
image

5. In other cases use existing behavior

**Additional context**
See the attached ket file with all the examples above
[cyclopentadiene_examples.zip](https://github.com/epam/ketcher/files/11585447/cyclopentadiene_examples.zip)

#185151979

Nitvex:
closes #2689


packages/ketcher-core/src/application/editor/actions/template.ts
```diff
@@ -207,6 +208,36 @@ export function fromTemplateOnBondAction(
   )
 }
 
+function getConnectingBond(
+  tmpl: Struct,
```
AnastasiiaPlyako, line 212: Could we rename it to "template"?

packages/ketcher-core/src/application/editor/actions/template.ts
```diff
@@ -308,22 +323,45 @@ function fromTemplateOnBond(restruct, template, bid, flip) {
...
+          action.addOp(
+            new BondAttr(newBondId, 'type', newBondType).perform(restruct)
+          )
+        }
+        if (isCyclopentadieneTemplate) {
+          const { beginBondIds, endBondIds } = Bond.getBondNeighbourIds(
+            struct,
+            bid
+          )
+          const bondBegin = struct.bonds.get(beginBondIds[0])!
```
AnastasiiaPlyako, line 342: We have an eslint rule which disables using non null assertion. Could we fix the eslint error? 
image

packages/ketcher-core/src/application/editor/actions/template.ts
```diff
@@ -207,6 +208,37 @@ export function fromTemplateOnBondAction(
   )
 }
 
+function getConnectingBond(
+  template: Struct,
+  struct: Struct,
+  bid: number,
```
AnastasiiaPlyako, line 214: I think it would be great to call it bondId :)

PR #2758, opened 2023-06-12
#2745 move selected structures using arrow keys

### LINKED ISSUE #2745: Move selected structures using arrow keys
**Background**
Once a structure is selected, allow users to move it using the keyboard arrow keys.

Acceptance criteria:
– It should support two "steps": 1px move, when pressing arrow button and 10px move, when also shift is pressed
– canvas should be automatically expanded, when structure is moved outside the canvas
– canvas should be automatically scrolled, when structure is moved outside the canvas

#184086826

Nitvex:
closes #2745


packages/ketcher-react/src/script/ui/state/moveSelectedItems.ts
```diff
@@ -0,0 +1,115 @@
...
+    editor.render.ctab,
+    selectedItems,
+    distinationVector.scaled(isShiftPressed ? fasterStepFactor : stepFactor)
+  )
+  editor.update(action, false, { resizeCanvas: true })
+  const isClose = isCloseToTheEdgeOfCanvas(selectedItems, editor, key)
+  if (isClose) {
+    scrollByVector(distinationVector, editor.render)
+  }
+  editor.rotateController.rerender()
```
kaluginserg, line 46: Just curious why we need to use the rotate controller here?

packages/ketcher-react/src/script/ui/state/moveSelectedItems.ts
```diff
@@ -0,0 +1,115 @@
...
+  const distinationVector = distinationVectorMapping[key]
+  const action = fromMultipleMove(
+    editor.render.ctab,
+    selectedItems,
+    distinationVector.scaled(isShiftPressed ? fasterStepFactor : stepFactor)
+  )
+  editor.update(action, false, { resizeCanvas: true })
+  const isClose = isCloseToTheEdgeOfCanvas(selectedItems, editor, key)
+  if (isClose) {
+    scrollByVector(distinationVector, editor.render)
```
kaluginserg, line 44: have concerns, point to discuss

packages/ketcher-react/src/script/ui/state/moveSelectedItems.ts
Nitvex: Because when we select structure "rotateController" appears. And then, we press some arrow button, structure is moved and rotateController doesn't move with the selected items

kaluginserg (approved):
Good job

PR #2767, opened 2023-06-14
#2600 Preview peptide monomers in Library

### LINKED ISSUE #2600: Preview peptide monomers in Library
**Requirements**
Once hover on a peptide in the Library, a dialog window appears with a monomer preview.

image

VasilevDO:
Closes #2600


packages/ketcher-polymer-editor-react/src/components/shared/StructRender/index.tsx
```diff
@@ -0,0 +1,41 @@
+/****************************************************************************
```
Nitvex, line 1: Can't we use this component directly from `ketcher-core`?

packages/ketcher-polymer-editor-react/src/components/monomerLibrary/monomerLibraryGroup/usePreview.ts
```diff
@@ -0,0 +1,61 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ ***************************************************************************/
+import { RefObject, useState } from 'react'
+import { MonomerItemType } from '../monomerLibraryItem/types'
+import { preview } from './consts'
+import { PreviewStyle } from './types'
+
+export const usePreview = (
```
Nitvex, line 21: I guess, it is worth to move this hook to `hooks` folder. I bet, this won't be the only case of using it.

packages/ketcher-polymer-editor-react/src/components/monomerLibrary/monomerLibraryGroup/usePreview.ts
```diff
@@ -0,0 +1,61 @@
...
+] => {
+  const [previewMonomer, setPreviewMonomer] = useState<
+    MonomerItemType | undefined
+  >()
+  const [previewStyle, setPreviewStyle] = useState<PreviewStyle | undefined>()
+
+  const setPreview = (monomer?: MonomerItemType, rect?: DOMRect) => {
+    if (monomer && rect && ref.current) {
+      const editorRect = document
+        .querySelector('.Ketcher-polymer-editor-root')
```
Nitvex, line 36: We have a constant for this class name

packages/ketcher-polymer-editor-react/src/components/shared/StructRender/index.tsx
VasilevDO: This is simply a container for ketcher-core render

packages/ketcher-react/src/script/ui/views/components/StructEditor/InfoPanel.tsx
```diff
@@ -135,16 +134,13 @@ const InfoPanel: FC<InfoPanelProps> = (props) => {
     >
       <StructRender
         struct={molecule}
-        id={groupName}
```
Nitvex, line 138: Could you clarify, why have you removed these props?

VasilevDO: Since I replaced the class StructRender component with a functional component and updated all occurrences I also removed the redundant props: childRef was just set without any actual usage, the same for id, viewSz prop does not exists within the RenderOptions type. So I just cut them and tested it locally - struct renders works the same as before

rrodionov91 (approved):

PR #2796, opened 2023-06-26
#2111 - Position newly added bond so that mouse cursor is over its atom

### LINKED ISSUE #2111: Position newly added bond so that mouse cursor is over its atom
When user selects bond tool and clicks on empty space of canvas, the newly added bond is positioned so that its center appears under mouse cursor. 
image
If user doesn't move mouse cursor and clicks again, the bond switches from single to double and from double to triple. 
image

Instead the newly added bond should be positioned so that one of its ends (atoms) appears under mouse cursor.
image
In such a case additional clicks will lead to adding new bonds to the same atom.
image
 

#184005652

rrodionov91:
Closes #2111 

### Single/double/tripple bond addition, zoom out and single bond addition

https://github.com/epam/ketcher/assets/14859362/771d856b-d6f0-45f9-924d-5a48cc1303cc

### DIFF
--- a/packages/ketcher-react/src/script/editor/tool/bond.ts
+++ b/packages/ketcher-react/src/script/editor/tool/bond.ts
@@ -293,18 +293,24 @@ class BondTool implements Tool {
   mouseup(event) {
     if ('dragCtx' in this) {
       const dragCtx = this.dragCtx;
-      const rnd = this.editor.render;
-      const struct = rnd.ctab.molecule;
+      const render = this.editor.render;
+      const struct = render.ctab.molecule;
       if ('action' in dragCtx) {
         this.restoreBondWhenHoveringOnCanvas(event);
         this.editor.update(dragCtx.action);
       } else if (!('item' in dragCtx)) {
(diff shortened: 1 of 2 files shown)


packages/ketcher-react/src/script/editor/tool/bond.ts
```diff
@@ -303,18 +303,24 @@ class BondTool implements Tool {
...
-      const struct = rnd.ctab.molecule
+      const render = this.editor.render
+      const struct = render.ctab.molecule
       if ('action' in dragCtx) {
         this.restoreBondWhenHoveringOnCanvas(event)
         this.editor.update(dragCtx.action)
       } else if (!('item' in dragCtx)) {
-        const xy = rnd.page2obj(event)
+        const editorOptions = this.editor.options()
+        const QUARTER_OF_BOND_WIDTH = 20 * editorOptions.zoom
```
Nitvex, line 313: Would be better if we had some constant instead of `20` magic number. Or I think, it is supposed to be calculated somehow

Nitvex (approved):

PR #2802, opened 2023-06-27
#2128 Automatically adjust zoom when opening a structure

### LINKED ISSUE #2128: Automatically adjust zoom when opening a structure
If a user works with large structures, they are not always fit the screen and canvas size. 
It is required to automatically change zoom level to show whole opened structure. In addition, zoom level needs to be adjusted automatically if user pastes a structure to an empty canvas.

**Additional requirements**
* Zoom level step is 1%
* Minimal zoom level is 10%
* Maximum zoom level is 100%

#184326364

rrodionov91:
Closes #2128 

## Changes
- added method in editor to zoom according content
- added this method execution after drawing opened structure

### Open big structure, delete, open small structure, delete, open big structure

https://github.com/epam/ketcher/assets/14859362/a5b3fa6e-99eb-4d2a-be0f-39b2b80401b0

### DIFF
--- a/packages/ketcher-react/src/script/editor/Editor.ts
+++ b/packages/ketcher-react/src/script/editor/Editor.ts
@@ -330,6 +330,27 @@ class Editor implements KetcherEditor {
     return this.render.options.zoom
   }
 
+  zoomAccordingContent() {
+    this.zoom(1)
+    const clientAreaBoundingBox = this.render.clientArea.getBoundingClientRect()
+    const paper = this.render.paper
+    const MIN_ZOOM_VALUE = 0.1
+    const MAX_ZOOM_VALUE = 1
+    const newZoomValue =
+      paper.height - clientAreaBoundingBox.height >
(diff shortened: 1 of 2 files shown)


packages/ketcher-react/src/script/editor/Editor.ts
```diff
@@ -330,6 +330,27 @@ class Editor implements KetcherEditor {
     return this.render.options.zoom
   }
 
+  zoomAccordingContent() {
```
kaluginserg, line 333: Just curious, there is "autoScale" property in the render options, that do something similar, can't use use it?

rrodionov91: Looks like autoscale is about scaling the size of structures after its addition on canvas:
https://github.com/epam/ketcher/assets/14859362/cd6c1f72-ab56-4caf-b032-921fcc89e7fd

Nitvex (approved):

Issue #197, opened 2020-12-21
Attachment points should be within R-Group brackets


ElenaKlepova:
**Description:** 
When attachment points are attached to a structure that is a R-Group member,  the attachment points are beyond the R-Group brackets. 

**Steps to reproduce:**
1. Create a structure.
2. Make the structure a R-Group member using _R-Group Fragment Tool_.
3. Attach attachment points to the structure using _Attachment Point Tool_. 

**Actual result:**
Attachment points are beyond R-Group brackets.
![2020-12-21_17h17_11](https://user-images.githubusercontent.com/75974843/102787231-cc40c000-43b1-11eb-9211-7dd238ab4281.png)

**Expected result:**
Attachment points are within R-Group brackets.

Zhirnoff:
Made only for R-Group Fragment, not for S-Group
![2023-07-07_11h38_06](https://github.com/epam/ketcher/assets/62464241/42fe3f47-35be-45f9-ab70-ead453cda584)
![2023-07-07_11h38_39](https://github.com/epam/ketcher/assets/62464241/3285b363-8095-46f3-8adc-d0dd0f9ed8ef)

PR #2888, opened 2023-07-07
#197 - Attachment points should be within R-Group brackets

StarlaStarla:
Closing #197 
including the attachment points bounding box as accumulating the r-group box

## Check list
- [ ] unit-tests written
- [ ] e2e-tests written
- [ ] documentation updated


packages/ketcher-core/src/application/render/restruct/rergroup.js
```diff
@@ -63,19 +63,34 @@ class ReRGroup extends ReObject {
 
     if (ret) ret = ret.extend(BORDER_EXT, BORDER_EXT)
 
+    let attachmentPointsVBox = render.ctab.getAttachmentsPointsVBox(
```
Nitvex, line 66: If the variable doesn't change, please use `const` instead

packages/ketcher-core/src/application/render/restruct/rergroup.js
```diff
@@ -63,19 +63,34 @@ class ReRGroup extends ReObject {
...
+      this.getAtoms(render)
+    )
+    if (attachmentPointsVBox)
```
Nitvex, line 69: Could you please add `{ }` for even single-line `if`?

packages/ketcher-core/src/application/render/restruct/rergroup.js
```diff
@@ -63,19 +63,34 @@ class ReRGroup extends ReObject {
...
+      attachmentPointsVBox = attachmentPointsVBox.extend(BORDER_EXT, BORDER_EXT)
+
+    ret =
```
Nitvex, line 72: If you understood what `ret` means, I would like to ask you to rename it

packages/ketcher-core/src/application/render/restruct/rergroup.js
```diff
@@ -63,19 +63,34 @@ class ReRGroup extends ReObject {
...
+    let bb = this.calcBBox(render)
 
     if (!bb) {
       console.error(
         'Abnormal situation, empty fragments must be destroyed by tools'
       )
       return {}
+    } else {
+      // add a little space between the attachment points and brackets
+      const vext = new Vec2(0.2, 0.4)
```
Nitvex, line 92: And here :) 
Unfortunately, there are a lot of variables with meaningless name. Let's not add new ones.

packages/ketcher-core/src/application/render/restruct/rergroup.js
StarlaStarla: but I did change the value at line 70

packages/ketcher-core/src/application/render/restruct/rergroup.js
StarlaStarla: sure, will do

packages/ketcher-core/src/application/render/restruct/rergroup.js
StarlaStarla: sure, will do

packages/ketcher-core/src/application/render/restruct/rergroup.js
StarlaStarla: sure, will do

Zhirnoff:
Made only for R-Group Fragment but not for S-Group
2023-07-07_13h21_47
2023-07-07_13h21_56

rrodionov91 (approved):

StarlaStarla:
The S-Group fix please refer to this bug  https://github.com/epam/ketcher/issues/2949

Issue #2958, opened 2023-07-25
Atom lable disappears after updating R-Group attachment point to none


yuleicul:
**Steps to Reproduce**
1. Activate tool 'R-Group Attachment Point'
2. Click on an atom to set attachment points
3. Click the same atom again to remove attachment points

**Actual behavior**
The label of the atom disappears

**Expected behavior**
The label of the atom appears

**Screenshots**

![rgroup-ap-label-bug](https://github.com/epam/ketcher/assets/27288153/d4a020a0-e212-43a0-a714-b823f16d9583)

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser: Chrome 
 - Version: 114.0.5735.199 (Official Build) (64-bit)

**Ketcher version** [e.g. v2.4.2]. 
Version 2.13.0-rc.1
Build at 2023-07-19; 13:44:23

MartaWilliams:
Test result: passed
Check on branch: 2958-atom-lable-disappears-after-updating-r-group-attachment-point-to-none
![2023-08-02_14h19_38](https://github.com/epam/ketcher/assets/82752700/59e17d5c-7adb-475c-b3aa-ecbc8727fa8c)

Issue #2742, opened 2023-06-06
Attachment point selection edition and deletion


kaluginserg:
### Background
Based on inputs from @AlexanderSavelyev.
Currently, we have the tool 'Attachment point', and each time when we need to remove an attachment point or edit it we have to use this tool. The attachment point on Canvas is not hoverable and is not selectable.

### Proposal
We have to introduce new behavior for editing and removing attachment points.
1) Attachment point should be hoverable
2) There should be the ability to select an attachment point (with a selection tool) and see the selected attachment point.
3) Should be the ability to remove attachment points (select and `Del` hotkey or with Erase tool)
4) Should be the ability to edit an attachment point by clicking on the Context menu item 'Edit' -> dialog with attachment point type should be displayed (Choosing Primary/Secondary or both).

### Visualisation details
> Selection and visualization of the attachment point: ![Attachment point selection](https://user-images.githubusercontent.com/125649799/219609843-8e8e55b3-d459-48bd-8d54-98dde51787bd.png) Attachment point in SVG ![Attachment point](https://user-images.githubusercontent.com/125649799/219609871-b83237a8-79e7-46d9-b667-b94dc2961bf5.svg) ![Attachment point selected](https://user-images.githubusercontent.com/125649799/219610632-9f0b8109-f2fe-4a32-b377-601472568e34.svg) ![Attachment point hover](https://user-images.githubusercontent.com/125649799/219610644-91684891-606a-4b12-a745-763f9f0ee4b7.svg)

_Originally posted by @PetrKurkin in https://github.com/epam/ketcher/issues/2165#issuecomment-1434393564_

### Additional acceptance criteria (in the scope of the task)
We have to fix these issues (it is supposed to be fixed as we will have Attachment points as interaction elements on Canvas):
> https://github.com/epam/ketcher/assets/27288153/e7925166-e3b9-4ff1-aa35-0752c534946e 
> 1. After clicking _Save To Templates_, the attachment point touches the bounding. (wrong paddings)
> 2. When opening the Templates Modal to check out the saved template, the attachment point doesn't display in the preview

_Originally posted by @yuleicul_

### Additional context
Connected #2165

yuleicul:
> Should be the ability to edit an attachment point by clicking on the Context menu item 'Edit' -> dialog with attachment point type should be displayed (Choosing Primary/Secondary or both).

A question: what's the trigger object for the context menu? 
- If the trigger object is the R-Group attachment point
[...]

PR #2961, opened 2023-07-25
#2958 - Atom lable disappears after updating R-Group attachment point to none

yuleicul:
## Check list
- [ ] unit-tests written
- [ ] e2e-tests written
- [ ] documentation updated

### DIFF
--- a/packages/ketcher-core/src/application/render/restruct/reatom.ts
+++ b/packages/ketcher-core/src/application/render/restruct/reatom.ts
@@ -648,7 +648,7 @@ function shouldDisplayStereoLabel(
 }
 
 function isLabelVisible(restruct, options, atom) {
-  const isAttachmentPointAtom = atom.a.attpnt;
+  const isAttachmentPointAtom = Boolean(atom.a.attpnt);
   const isCarbon = atom.a.label.toLowerCase() === 'c';
   const visibleTerminal =
     options.showHydrogenLabels !== ShowHydrogenLabels.Off &&
--- a/packages/ketcher-core/src/application/render/restruct/reatom.ts
+++ b/packages/ketcher-core/src/application/render/restruct/reatom.ts
@@ -648,7 +648,7 @@ function shouldDisplayStereoLabel(
 }
 
 function isLabelVisible(restruct, options, atom) {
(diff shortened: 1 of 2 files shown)


yuleicul:
Proposal to add type to `attpnt` :https://github.com/epam/ketcher/issues/2962

packages/ketcher-core/src/application/render/restruct/reatom.ts
```diff
@@ -648,7 +648,7 @@ function shouldDisplayStereoLabel(
 }
 
 function isLabelVisible(restruct, options, atom) {
-  const isAttachmentPointAtom = atom.a.attpnt != null;
+  const isAttachmentPointAtom = atom.a.attpnt;
```
kaluginserg, line 651: Could I propose to use this approach: 

```javascript
= Boolean(atom.a.attpnt)
``` 

(in this case the variable `isAttachmentPointAtom` will store the boolean type but not one of three: null, undefined, 0).

packages/ketcher-core/src/domain/entities/atom.ts
```diff
@@ -46,6 +46,13 @@ export interface AtomAttributes {
...
   ringBondCount?: number;
   explicitValence?: number;
+  /**
+   *  Values can be `0 | 1 | 2 | 3 | null`.
+   * `1` - has a `primary` R-Group Attachment Point
+   * `2` - has a `secondary` R-Group Attachment Point
+   * `3` - has a `primary` and a `secondary` R-Group Attachment Points
+   * `null` and `0` both mean the atom has no R-Group Attachment Points
+   */
   attpnt?: any;
```
kaluginserg, line 56: Curious, can we also remove the 'any' type and make it like this:

```typescript
attpnt?: 0 | 1 | 2 | 3 | null;
```

yuleicul: Yes this is what I wanted, but considering my current task #2742 is using a lot `attpnt`. I'm afraid it would take efforts to resolve conflicts so I created a refactoring request https://github.com/epam/ketcher/issues/2962

SashaGraves (approved):

Issue #1859, opened 2022-11-18
Save structure window: formats in drop-down list is not ordered


Zhirnoff:
**Steps to Reproduce**
1. Open Ketcher
2. Create any structure
3. Press 'Save' button
4. Click to drop-down 'File format'
5. Check order of file formats

**Actual behavior**
File formats ordered not as in design
![2022-11-18_17h04_47](https://user-images.githubusercontent.com/62464241/202777085-303512f5-603a-44d8-adb4-344f252a490c.png)

**Expected behavior**
File formats ordered as in design

**Screenshots**
![2022-11-18_21h17_04](https://user-images.githubusercontent.com/62464241/202777122-124ab2b3-a0ae-4bd5-b6f0-0faeda83d0bc.png)

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser Chrome
 - Version 107.0.5304.107 (Official Build) (64-bit)

Related issue: https://github.com/epam/ketcher/issues/1376
**Ketcher version**
 [Version 2.6.2]
Build at 2022-11-17; 13:01:16

PR #2976, opened 2023-07-27
#1859 – Save structure window: formats in drop-down list is not ordered

nanoblit:

Zrzut ekranu 2023-07-27 081726

## Check list
- [x] unit-tests written
- [ ] e2e-tests written
- [x] documentation updated


yuleicul:
This part is not the same as expected in the issue description. 
image
image


I think it's because the issue is outdated, could you please contact @Zhirnoff and confirm the latest design and update the issue description?

SashaGraves (approved):

nanoblit:
@yuleicul  I renamed that mockStore file as you suggested and also renamed another one in other place for consistency. I left that one as js though, because typing it would cause an error because of the way its reducer is defined in another js file. I decided not to make a generic helper, because there were enough differences between just these two uses that it would add more unnecessary code on every use.

yuleicul (approved):
Thank you! Great job 👍

Ustianskii:
Fixed. Tested on SA.
image
image

Issue #2948, opened 2023-07-21
Bond thickness is not changed via Ketcher API


Nitvex:
**Steps to Reproduce**
1. Draw any structure
2. Change bond thickness via API
```
ketcher.setSettings({
    bondThickness: 5
})
```

**Actual behavior**
Bond thickness is not changed:
![image](https://github.com/epam/ketcher/assets/41272128/2e85f3e4-5bf0-4052-9d65-fa1e53ad9e78)

The same behaviour can be observed, if user wants to get an image by calling `generateImage` method:
![image](https://github.com/epam/ketcher/assets/41272128/2df792ed-591a-437b-a8fc-1dcb881ba450)

**Expected behaviuor**
– Bond thickness is changed
– Structure is redrawn without any additional actions
– Bond with changed thickness can be received using `generateImage` method
![image](https://github.com/epam/ketcher/assets/41272128/ef06644a-da12-4b90-854f-3f013db29db0)

**Additional context**
It is possible to change bond thickness manually via Ketcher options

PR #3062, opened 2023-08-08
#2948 - Bond thickness is not changed via Ketcher API

AKZhuk:
Resolves [#2948](https://github.com/epam/ketcher/issues/2948)
## How the feature works?

https://github.com/epam/ketcher/assets/16281184/e7908690-24bc-462e-91be-d6a6902ca920

Fix for:
- Change bond thickness via API
-  Bond thickness for for "save as"  method
-  Bond thickness for for "save as" preview  
-  Bond thickness for "copyToCLipBoard" method

## Check list
- [ ] unit-tests written
- [x] e2e-tests written
- [x] documentation updated


packages/ketcher-core/src/application/indigo.ts
```diff
@@ -200,12 +201,13 @@ export class Indigo {
   ): Promise<string> {
     const outputFormat = options?.outputFormat || 'png';
     const backgroundColor = options?.backgroundColor || '';
-
+    const bondThickness = options?.bondThickness || 1;
```
Nitvex, line 204: I propose to create a constant for default value, which will be used across app :) 
And additional question. Here you have `1` default value, but somewhere below you have `2`. Why? :)

AKZhuk: propose to create a constant for default value - _Done_.
Regarding value `2` it is used just for testing

packages/ketcher-core/src/application/editor/shared/constants.ts
```diff
@@ -34,3 +34,5 @@ export const selectionKeys = [
   'simpleObjects',
   'texts',
 ] as const;
+
+export const defaultBondThickness = 1;
```
KonstantinEpam23, line 38: Can we please set this to `2` and refer to it in here: https://github.com/epam/ketcher/blob/master/packages/ketcher-react/src/script/ui/data/schema/options-schema.ts#L224

KonstantinEpam23: or probably even create a separate constant for this and refer to it in both places

AKZhuk: done

KonstantinEpam23 (approved):

PR #3071, opened 2023-08-09
2334 bond peptides center to center

### LINKED ISSUE #2334: Create non-cyclic peptide chain (horizontal bonds only) - draft
**Background**

Chain - at least 2 monomers
This story covers creation of a one chain / separate chains 

**Requirements**

1)	A bond can be created if both connecting the monomers have at least 1 connection point available (each)
2)	The bond is represented in the following way: starting from the place of the first click (one monomer) to the place of the actual cursor position while dragging to the other monomer -? 
3)	The monomer that was placed more on the left on the canvas is represented earlier in the chain and the monomer that was placed more on the right is represented later
4)	The bond is created using Carboxyl end (R1) of the one monomer and Amino end (R2) of the other monomer
5)	The bond is represented as a two-colored line: R1 is colored in blue, R2 is colored in red
6)	Chain enumeration is represented in the following way: n1, 2, 3, …, N+1 – from first left monomer to the last right monomer

If one chain  is already built:
1)	When connecting a new monomer / monomer chain to the first or the last monomer of the already existing chain – new monomer chain is “REenumerated” as follows: n1, 2, 3, …, N+1

rrodionov91:
Closes #2334 

## Check list
- [ ] unit-tests written
- [ ] e2e-tests written
- [ ] documentation updated


packages/ketcher-core/src/application/editor/actions/polymerBond.ts
```diff
@@ -0,0 +1,46 @@
...
+  startPosition;
+  endPosition;
+};
+
+export function fromPolymerBondAddition(
+  renderersContainer: ReStruct,
+  data: Data,
+): Action {
+  const action = new Action();
+  action.addOp(new PolymerBondAdd(data)).perform(renderersContainer);
```
vitaepam, line 33: Could you please rename this function to 'addOperation'?

vitaepam (approved):

packages/ketcher-core/src/application/editor/actions/polymerBond.ts
rrodionov91: This method is used in 140+ places in code. Maybe it should be done in separate PR

Issue #2541, opened 2023-04-27
Missing "Tool" word for Arrow Filled Triangle


AleksandraSmolianinova:
1. Navigate to ketcher
2. Navigate mouse to Arrow Filled Triangle

AR: it is "Arrow Filled Triangle"
ER: it should be "Arrow Filled Triangle Tool"

[Version 2.10.0-rc.1](https://lifescience.opensource.epam.com/ketcher/index.html)
Build at 2023-04-20; 18:30:36

Google Chrome - Version 111.0.5563.149

AleksandraSmolianinova:
Test cases are being created.

PR #3086, opened 2023-08-10
#2541 - missing tool word for arrow filled triangle

chgayane:
Navigate to ketcher
Navigate mouse to Arrow Filled Triangle
AR: it is "Arrow Filled Triangle"
ER: it should be "Arrow Filled Triangle Tool"

## Check list
- [x] unit-tests written
- [ ] e2e-tests written
- [ ] documentation updated

### DIFF
--- a/packages/ketcher-react/src/script/ui/component/view/ToolActions.test.tsx
+++ b/packages/ketcher-react/src/script/ui/component/view/ToolActions.test.tsx
@@ -1,4 +1,4 @@
-import toolActions from '../../../../../../ketcher-react/src/script/ui/action/tools';
+import toolActions from '../../action/tools';
 
 describe('ToolActions', () => {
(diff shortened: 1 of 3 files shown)


packages/ketcher-react/src/script/ui/component/view/ToolActions.test.tsx
```diff
@@ -0,0 +1,13 @@
+import toolActions from '../../action/tools';
+
+describe('ToolActions', () => {
+  const toolsWithTitles = Object.values(toolActions).filter(
```
Nitvex, line 4: Actually, my proposal is to remove `filter`. That's because it doesn't allow to check the `title` attribute. 
If some button doesn't have title, then it will be filtered out and test will pass.

Issue #2909, opened 2023-07-10
Movement of structure ceases when it reaches the boundaries of canvas, requiring the consistent movement of mouse to continue its motion


Zhirnoff:
**Steps to Reproduce**
1. Open Ketcher
2. Add two Benzene rings on canvas
3. Select one of them and move to the edges of canvas

**Actual behavior**
The movement of the structure occurs in shaking when the user brings the structure close to the edge of the canvas, and in order for the structure to continuously move beyond the canvas boundaries, continuous mouse movements are required.

**Expected behavior**
When the user brings the structure close to the edge of the canvas, smooth and continuous movement should be ensured instead of shaking motion. Additionally, the structure should be able to move beyond the canvas boundaries without the need for continuous mouse movements. 

**Screenshots**

https://github.com/epam/ketcher/assets/62464241/289f3432-1baa-4293-a9aa-9222fbfe79e6

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser Chrome
 - Version 112.0.5615.138 (Official Build) (64-bit)

**Ketcher version** 
[Version 2.12.0-rc.3](https://lifescience.opensource.epam.com/ketcher/index.html)

Zhirnoff:
Test cases are written by me.

Zhirnoff:
Pre-tested. Bug is fixed.

https://github.com/epam/ketcher/assets/62464241/6c2b087d-c7da-4df1-8223-9b65141f1d8a

PR #3342, opened 2023-09-19
#2909 - movement of structure ceases when it reaches the boundaries of canvas requiring the consistent movement of mouse to continue its motion

nanoblit:
Made smooth scroll work using requestAnimationFrame to update the scroll every frame (it only happens when the user is holding a structure close to a screen border.). Also made sure to make scrolling framerate independent, among other things.

## Check list
- [ ] unit-tests written
- [ ] e2e-tests written
- [x] documentation updated

### DIFF
--- a/packages/ketcher-react/src/script/editor/tool/select.ts
+++ b/packages/ketcher-react/src/script/editor/tool/select.ts
@@ -75,7 +75,7 @@ class SelectTool implements Tool {
   readonly #lassoHelper: LassoHelper;
   private readonly editor: Editor;
   private dragCtx: any;
-  private previousMouseMoveEvent: any;
+  private previousMouseMoveEvent?: MouseEvent;
   isMouseDown = false;
   readonly isMoving = false;
 


nanoblit:
Most of the issues below are also reproducible in live version.
Issues:
- When the view is moved up or left and trying to move a structure up or left further, there is an area on the screen where the view will scroll backwards.
- Scroll speed is inconsistent.
- When moving a structure down or right and scrolling then leaving the structure, the canvas will be resized to the position of the structure. This happens only if the structure is close enough to the border of the canvas. This doesn't happen when moving the structure up or left.
- When moving the structure down and then left or right / moving the structure right then up or down, the canvas position will reset. This doesn't happen for up and left.
- When moving a structure far in any direction and returning, the canvas' size will reset. It won't reset if leaving the structure anywhere in between the farthest position and the original position, leaving a lot of empty space.

nanoblit:
all-bonds.spec.ts tests passing locally:
all bonds spec ts PASS

Other failing tests should be fixed by Nitvex soon. They are unrelated to this task.

y-holik (approved):

packages/ketcher-react/src/script/editor/tool/select.ts
```diff
@@ -54,12 +53,30 @@ import { handleMovingPosibilityCursor } from '../utils';
...
+  [Direction.DOWN]: new Vec2(0, canvasOffsetPerSecond),
+};
+
 class SelectTool implements Tool {
   readonly #mode: SelectMode;
   readonly #lassoHelper: LassoHelper;
   private readonly editor: Editor;
   private dragCtx: any;
-  isMousedDown = false;
+  private previousMouseMoveEvent: any;
```
Nitvex, line 78: I propose to use proper type `MouseEvent`

chgayane (approved):

Issue #3040, opened 2023-08-04
Provide additional logging


Nitvex:
**Background**
At the moment, Ketcher has a lot of try/catch blocks which catch errors (for example, when parsing files) and show appearing errors in modal. This errors sometimes neither informative for end-user nor developers.

**Solution**
1. It is required to add `console.error` to catch block in order to easy find the places, where error appeared
2. It is required to prevent non-informative errors to be shown to users (Example error: "cannot read property null of undefined")

Zhirnoff:
Not tested by QA.

PR #3375, opened 2023-09-29
#3040 - Provide additional logging

nanoblit:

Adds a KetcherLogger class, which can be used to log in dev and prod with specific flags in window.ketcher enabled.
Adds logs to throws so we can see where they happened in prod.

## Check list
- [ ] unit-tests written
- [ ] e2e-tests written
- [x] documentation updated


packages/ketcher-core/src/utilities/KetcherLogger.ts
```diff
@@ -0,0 +1,77 @@
...
+
+  static warn(warning: unknown): void {
+    if (!this.isMinimumLogLevel(LogLevel.WARN)) {
+      return;
+    }
+
+    window.console.warn(warning);
+  }
+
+  static error(error: unknown): void {
```
Nitvex, line 52: Could you, please, clarify the difference between `error` and `showExceptionLocation`? What is the difference between error and exception?  
My proposal is to have only one method `error`, which can work like this:
`KetcherLogger.error('someFile::someFunction', error);` – please, note that error can be an object

packages/ketcher-core/src/application/formatters/identifyStructFormat.ts
```diff
@@ -26,7 +27,11 @@ export function identifyStructFormat(
     if (JSON.parse(sanitizedString)) {
       return SupportedFormat.ket;
     }
-  } catch (er) {} // eslint-disable-line
+  } catch (er) {
+    KetcherLogger.showExceptionLocation(
```
Nitvex, line 31: General comment for everywhere – to pass exception to logger. If someone has any error, then you'll be able to find the place, where error has occurred, but won't be able to understand the error it self.

example/src/App.tsx
```diff
@@ -71,6 +73,13 @@ const App = () => {
         structServiceProvider={structServiceProvider}
         onInit={(ketcher: Ketcher) => {
           window.ketcher = ketcher;
+
+          KetcherLogger.settings = {
```
Nitvex, line 77: My proposal is to remove this from here. Actually, for 99% of users logger won't be required. It is needed for cases, when we will directly ask customer like "Open the console, write this command, and send logs to us". 
So, can you additionally add to description, how logger can be enabled via Dev Tools console?

packages/ketcher-core/src/utilities/KetcherLogger.ts
```diff
@@ -0,0 +1,77 @@
+export enum LogLevel {
+  ERROR = 0,
+  WARN = 1,
+  LOG = 2,
```
Nitvex, line 4: Can we add a method `info` here? It can be called this way `console.info()`. This is just additional method for the same logging, but it will allow you to reduce number of logs by setting proper log level

Nitvex (changes requested):

OlgaMazurina (approved):

example/src/App.tsx
nanoblit: To enable logger type in the console:
`window.ketcher.logger.enabler = true`

The default log level is "ERROR" or 0. 
To change log level type in the console:
`window.ketcher.logger.level = 1` to enable warnings
2 to enable infos
3 to enable logs

To show traces on logs and infos, type in the console:
`window.ketcher.logger.showTrace = true`

packages/ketcher-core/src/utilities/KetcherLogger.ts
nanoblit: Done

packages/ketcher-core/src/application/formatters/identifyStructFormat.ts
nanoblit: Done

example/src/App.tsx
nanoblit: Done

packages/ketcher-core/src/utilities/KetcherLogger.ts
nanoblit: Done

Nitvex (approved):

PR #3396, opened 2023-10-04
Macro: #3373 - Rotate Attachment Point Unused

### LINKED ISSUE #3373: Macro: Rotate attachment point unused
## Background
Monomer connection point (CP) rotates around the center of the monomer if CP is connected to a bond and the bond changes its position relatively to the monomer. In certain cases rotated CP may partially or fully overlap with another CP. In order to minimize the amount of such cases free connection points should be rearranged around the monomer once such an overlap occurs.

Example of current behaviour:
image

Example of expected behaviour:
image

## Solution:

The rules for rearrangement are as follows:

1. Only free CPs should be rearranged, connected ones should keep their positions.
2. Free CPs should be rearranged only when CP rotation is finished, no rearrangements while user keeps dragging connected monomer. 
3. There are 8 predefined sectors with center axis at 0 (right), 45, 90 (top), 135, 180 (left), 225, 270 (down) and 315 degrees. Two CPs are considered overlapping if they appear in the same sector, i.e. they overlap or are close enough. 
4. Once rearrangement starts, first priority is to move each free CP to the center of its default sector: R1 - left, R2-right, R3-top, R4-bottom, R5-top-left, R6-top-right, R7-bottom-left, R8-bottom-right. 
5. Free CP can be moved to its default sector if there is no connected CP in this sector, other free CPs are not taken into consideration. If CP cannot be moved to its default connection, it is assigned to a sector on further steps (see below).
6. Once CPs assignment on default positions is finished, second priority is to keep free CPs at their old positions if no conflicts occurred on previous steps. 
7. Remaining free CPs are distributed across available sectors, where available sector is the one having neither connected nor free CP. The algorithm should search through sectors starting from the one at 0 degrees and moving clockwise.

image

SashaGraves:

closes #3373
## Check list
- [ ] unit-tests written
- [ ] e2e-tests written
- [ ] documentation updated


Nitvex:
Points to fix:
1) Labels overlap, when attachment points are placed near each other
image

packages/ketcher-core/src/domain/AttachmentPoint.ts
```diff
@@ -102,6 +46,9 @@ export class AttachmentPoint {
     bodyHeight,
     canvas,
     RN,
+    isUsed,
```
Nitvex, line 49: Can we use more meaningful names instead of `RN`? In addition, I highly encourage you to add more types for variables – there a lot of them with `any` type.

packages/ketcher-core/src/domain/AttachmentPoint.ts
```diff
@@ -158,15 +118,21 @@ export class AttachmentPoint {
...
+    } else {
+      const angleRadians = this.rotateToAngle(
+        this.monomer.attachmentPointsToBonds[this.attachmentPointName],
+        this.monomer.id ===
+          this.monomer.attachmentPointsToBonds[this.attachmentPointName]
+            ?.firstMonomer?.id,
+      );
 
-    const angleDegrees = Vec2.radians_to_degrees(angleRadians);
+      angleDegrees = Vec2.radians_to_degrees(angleRadians);
```
Nitvex, line 134: I propose to rename it to `radiansToDegrees`

packages/ketcher-core/src/domain/entities/BaseMonomer.ts
```diff
@@ -150,6 +150,30 @@ export class BaseMonomer extends DrawingEntity {
     return this.firstFreeAttachmentPoint;
   }
 
+  public get usedAttachmentPointsNamesList() {
+    const list: string[] = [];
```
Nitvex, line 154: I believe it is not a string – it is `AttachmentPointName` type

Nitvex (changes requested):

rrodionov91 (approved):

Nitvex (approved):

MartaWilliams:
Tested - functionality works correctly, only attachment points are sometimes on different hights
2023-10-24_09h30_35

Issue #2925, opened 2023-07-17
Macro: Support monomers connection dragging AP


OlZhur:
**Background**
An attachment point (AP) should have a visual representation to be able to select a specific point.

**Requirements**

Monomers should be connected by a bond for the following cases of monomers connections:
* User clicks on the center of the first monomer and drag a bond to the specific AP of the second monomer. A modal window for selection of free AP should be displayed if the first monomer has more than one free AP
* User clicks on the specific AP of the first monomer, clicks on it and drags a bond to the center of the second monomer. A modal window for selection of free AP should be displayed if the second monomer has more than one free AP
* User clicks on the specific AP of the first monomer and drag a bond to the specific AP of the second monomer

There should be a different visual style for free AP compared to non-free AP.
While monomers are chaining by a bond it needs to rotate both appropriate attachment points to the bond direction. 
When users hover over a bond, both AP should be displayed.

_Warning messages:_
1. Warning for Peptides only: flow R1 to R1 or R2 to R2. 'You have connected monomers with attachment points of the same group'
2. No free attachment points. 'Monomers do not have any attachment points available'

Position of connection points on the monomer - according to mock ups

nightnick:
![Frame 2197](https://github.com/epam/ketcher/assets/27892118/d0ebeee3-4e09-4db6-b108-4d2341beb8e4)
![Frame 2198](https://github.com/epam/ketcher/assets/27892118/d4c3476e-ff6d-42e9-8905-93e1d0ed0061)

nightnick:
Warning and error messages are WIP. Need some correction.

nightnick:
Updated screens
![Frame 2197](https://github.com/epam/ketcher/assets/27892118/e0c6d85b-9739-406e-bc67-f116ec236b57)
![Frame 2198](https://github.com/epam/ketcher/assets/27892118/e3547392-bf33-469c-8286-1eb0e2d19942)

nightnick:
New updated designs:
![Bonds Scenarios](https://github.com/epam/ketcher/assets/27892118/0639c173-491d-485e-a626-6b84a1cebcad)

nightnick:
New designs again:
![Bonds Scenarios_ Improvement (in discussion)](https://github.com/epam/ketcher/assets/27892118/828844ed-118e-409b-af6e-f9854f161e70)

OlZhur:
[...]

PR #3524, opened 2023-10-30
Macro: #2925 - Support monomers connection dragging AP

nanoblit:

## Check list
- [ ] unit-tests written
- [ ] e2e-tests written
- [ ] documentation updated


OlgaMazurina:
I ran this test 5 times and once it was with error. So it is flaky now.
Снимок экрана 2023-11-15 в 14 45 19

packages/ketcher-core/src/domain/AttachmentPoint.ts
```diff
@@ -124,6 +139,79 @@ export class AttachmentPoint {
...
+      },
+      {
+        x: -halfWidth,
+        y: -areaHeight + 10,
+      },
+      { x: -AttachmentPoint.radius, y: AttachmentPoint.radius },
+    ];
+
+    const lineFunction = line<Coordinates>()
+      .x(function (point) {
```
Nitvex, line 174: Can we just
```
.x(({ x }) => x)
```

packages/ketcher-core/src/application/editor/operations/monomer/index.ts
```diff
@@ -48,6 +48,19 @@ export class MonomerHoverOperation implements Operation {
     );
   }
 }
+export class APHoverOperation implements Operation {
```
Nitvex, line 51: I propose to rename `AP` to attachmentPoint everywhere.

Nitvex, line 51: Actually, I don't understand the usage of "operation" and "command" regarding visual effect. Main purpose of commands is History, so that you can perform "undo" and "redo". Let's take an example: you hovered over a monomer, then you will create a command, so when you click "undo" the selection will be reset? 
The answer is – no. Commands should affect only model (chemical) changes, not hovers, selections, modals and any other visual effects.

Nitvex (commented):
I propose to move the requested changes to the next milestone

packages/ketcher-core/src/domain/AttachmentPoint.ts
OlgaMazurina: Done

packages/ketcher-core/src/application/editor/tools/Bond.ts
```diff
@@ -345,6 +479,101 @@ class PolymerBond implements BaseTool {
     }
     return false;
   }
+
+  // private shouldInvokeModal(
```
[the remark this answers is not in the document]
OlgaMazurina: Of course, good catch!

packages/ketcher-core/src/application/editor/operations/monomer/index.ts
OlgaMazurina: Replaced

OlgaMazurina: It is not my idea to use operation in such cases. If you look at this file, you will find similar approach there (MonomerHoverOperation). If we want to replace all this places, I think it is better to create a task further.

Nitvex: Okay, I guess, we need to discuss it with @rrodionov91

Issue #3525, opened 2023-10-31
Macro: Wrong usage of AP when establishing a bond between monomers (R1 and R2) -> (R2)


OlZhur:
**Steps to Reproduce**
1. Add to canvas 2 monomers. One of them has R1 and R2, the second one has only R2
2. Drug the bond from the first one (with R1 and R2) to the second one.

**Actual behavior**
A bond is established between R2 and R2

**Expected behavior**
A bond is established between R1 and R2

**Screenshots**
![image](https://github.com/epam/ketcher/assets/128700628/f439526c-b0ce-4be7-b1cd-8e4b4fa7833e)

MartaWilliams:
Tested 
![image](https://github.com/epam/ketcher/assets/82752700/3e3a9dc0-cb62-4457-bb19-ab3d3c708bbd)

PR #3536, opened 2023-11-03
Macro: #3525 - Wrong usage of AP when establishing a bond between monomers r1 and r2 r2

ivan-arkhipov-epam:
This fix prevents forming invalid R2-R2 polymer bond in cases when there is an R1 point in the first monomer

## Check list
- [x] unit-tests written
- [ ] e2e-tests written
- [ ] documentation updated

### DIFF
--- a/packages/ketcher-core/src/domain/entities/DrawingEntitiesManager.ts
+++ b/packages/ketcher-core/src/domain/entities/DrawingEntitiesManager.ts
@@ -375,6 +375,18 @@ export class DrawingEntitiesManager {
       command.addOperation(operation);
     }
 
+    if (
+      availableAttachmentPointForBondEnd === 'R2' &&
+      !monomer.hasAttachmentPoint('R1') &&
+      bond.firstMonomer.hasAttachmentPoint('R1')
+    ) {
+      // Prevents forming invalid R2-R2 bonds when second monomer does not have R1 point
+      bond.firstMonomer.removePotentialBonds();
+      bond.firstMonomer.setPotentialBond('R1', bond);
+      const operation = new MonomerHoverOperation(bond.firstMonomer, true);
+      command.addOperation(operation);
+    }
+
     const operation = new MonomerHoverOperation(monomer, true);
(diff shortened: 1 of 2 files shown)


AnastasiyaPiatrovaKlu (approved):

packages/ketcher-core/src/domain/entities/DrawingEntitiesManager.ts
```diff
@@ -375,6 +375,18 @@ export class DrawingEntitiesManager {
...
+      !monomer.hasAttachmentPoint('R1') &&
+      bond.firstMonomer.hasAttachmentPoint('R1')
+    ) {
+      // Prevents forming invalid R2-R2 bonds when second monomer does not have R1 point
+      bond.firstMonomer.removePotentialBonds();
+      bond.firstMonomer.setPotentialBond('R1', bond);
+      const operation = new MonomerHoverOperation(bond.firstMonomer, true);
+      command.addOperation(operation);
+    }
+
```
SashaGraves, line 389: Hi! Not bad, it works! But you also should check for bond.firstMonomer.isAttachmentPointUsed('R1'). Currently you will get infinite number of bonds:
image

Also i would suggest to refactor this whole function and make every 'if' return command. This way they will not stack one on another. But it can make additional effort, so maybe we can leave this for future tasks.

SashaGraves (changes requested):

nanoblit (approved):

packages/ketcher-core/src/domain/entities/DrawingEntitiesManager.ts
rrodionov91: JFYI each 'if' will be replaced to specific monomer class soon in another PR.

ivan-arkhipov-epam: (hopefully) fixed.

SashaGraves (approved):

Issue #3509, opened 2023-10-25
Macro: RNA tab view while opening the RNA tab


OlZhur:
Once RNA tab is opened by user for the first time, the RNA builder should be presented in a collapsed empty state and Presets shouldn't be blocked from utilizing 
![image](https://github.com/epam/ketcher/assets/128700628/27d78f5d-2060-4df8-9c36-b258ad63a77b)

Zhirnoff:
Autotest written

AnastasiyaPiatrovaKlu:
Tested.

https://github.com/epam/ketcher/assets/72072840/ebe2367e-0b1d-441f-bfda-f87ac8605587

PR #3562, opened 2023-11-09
Macro:  #3509 - RNA tab view while opening the RNA tab

rrodionov91:
- fixed presets deletion
- fixed monomer selection

## Check list
- [ ] unit-tests written
- [ ] e2e-tests written
- [ ] documentation updated

### DIFF
--- a/packages/ketcher-polymer-editor-react/src/components/monomerLibrary/RnaBuilder/RnaBuilder.tsx
+++ b/packages/ketcher-polymer-editor-react/src/components/monomerLibrary/RnaBuilder/RnaBuilder.tsx
@@ -74,7 +74,7 @@ export const RnaBuilder = ({ libraryName }) => {
   return (
     <RnaBuilderContainer>
       <RnaEditor
-        duplicatePreset={duplicatePreset}
+        duplicatePreset={() => duplicatePreset()}
         activateEditMode={activateEditMode}
       />
       <RnaAccordion
--- a/packages/ketcher-polymer-editor-react/src/state/rna-builder/rnaBuilderSlice.ts
+++ b/packages/ketcher-polymer-editor-react/src/state/rna-builder/rnaBuilderSlice.ts
@@ -110,11 +110,13 @@ export const rnaBuilderSlice = createSlice({
     deletePreset: (state, action: PayloadAction<IRnaPreset>) => {
       const preset = action.payload;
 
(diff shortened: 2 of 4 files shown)


packages/ketcher-polymer-editor-react/src/components/monomerLibrary/RnaBuilder/RnaBuilder.tsx
```diff
@@ -74,7 +74,7 @@ export const RnaBuilder = ({ libraryName }) => {
   return (
     <RnaBuilderContainer>
       <RnaEditor
-        duplicatePreset={duplicatePreset}
+        duplicatePreset={() => duplicatePreset()}
```
StarlaStarla, line 77: I am just curious, is`duplicatePreset` and `() => duplicatePreset()` has the same effect here? why isn't `duplicatePreset` better because this funtion is always the same so it would not ben redefined, while `() => duplicatePreset()` is just an anonymous function that would get redefined everytime during render?

packages/ketcher-polymer-editor-react/src/state/rna-builder/rnaBuilderSlice.ts
```diff
@@ -110,11 +110,13 @@ export const rnaBuilderSlice = createSlice({
...
 
+      const presetIndexInList = state.presets.findIndex(
+        (presetInList) => presetInList.name === preset.name,
+      );
+      state.presets.splice(presetIndexInList, 1);
+
       if (preset.presetInList) {
-        const presetIndexInList = state.presets.findIndex(
-          (presetInList) => presetInList.name === preset.presetInList?.name,
-        );
```
StarlaStarla, line 116: Do we need to change the logic in savePreset too? because a new created preset will be saved in `presetInList` of activePreset and it seems could not be deleted if we dont use `presetInList?.name` to match here

packages/ketcher-polymer-editor-react/src/components/monomerLibrary/RnaBuilder/RnaBuilder.tsx
rrodionov91: It's done because to exclude providing "event" as a first parameter of duplicatePreset function, because it waits another optional parameter

StarlaStarla (approved):

Issue #3530, opened 2023-11-02
Macro: Labels in attachment points selection popup are small for large molecules


rrodionov91:
**Steps to Reproduce**
1. Add 2 Chems with name "Test-6-AP-Chem"
2. Drag a bond between chems
3. Popup with monomer previews appeared

**Actual behavior**
R1..Rn labels scaled out and became small.
<img width="956" alt="image" src="https://github.com/epam/ketcher/assets/14859362/8947f27c-4891-4c2c-9647-79802e098c20">

**Expected behavior**
R1..Rn labels have fixed size.
<img width="447" alt="image" src="https://github.com/epam/ketcher/assets/14859362/fa3b013d-9fd1-42a8-99b4-1b0176809e36">

Zhirnoff:
Tested.

PR #3626, opened 2023-11-28
Macro: #3530 - Labels in attachment points selection popup are small for large molecules

chgayane:

## Check list
- [ ] unit-tests written
- [ ] e2e-tests written
- [ ] documentation updated


packages/ketcher-core/src/application/render/renderStruct.ts
```diff
@@ -24,6 +25,25 @@ export class RenderStruct {
     return struct;
   }
 
+  static removeSmallAttachemntPointLabelsInModal(render: Render, options) {
```
ivan-arkhipov-epam, line 28: should ```options``` have a type?

packages/ketcher-core/src/application/render/renderStruct.ts
```diff
@@ -24,6 +25,25 @@ export class RenderStruct {
...
+    if (!options.labelInMonomerConnectionsModal) {
+      return;
+    }
+
+    const attachmentPointRegExp = /^R[1-8]$/;
```
ivan-arkhipov-epam, line 33: I wonder if it makes sense to store this as a member of the class itself to make it reusable for other methods... Just thinking. That may be redundant.

packages/ketcher-core/src/application/render/renderStruct.ts
```diff
@@ -24,6 +25,25 @@ export class RenderStruct {
...
+
+    render.ctab.atoms.forEach((atom: ReAtom) => {
+      if (!atom.label) {
+        return;
+      }
+      const isSmall = atom.label.path.node.getBoundingClientRect().width < 8;
```
ivan-arkhipov-epam, line 39: I have a feeling that eslint is going to complain at this magic number. Maybe define a ```const``` with a meaningful name? Or a class property?

ivan-arkhipov-epam (approved):

packages/ketcher-core/src/application/render/renderStruct.ts
rrodionov91, line 39: I would suggest to move it under "if" statement to get boundingClientRect only for attachmentPointAtoms because it is very unefficient method from the performance point of view.

rrodionov91 (changes requested):

packages/ketcher-core/src/application/render/renderStruct.ts
```diff
@@ -24,6 +27,31 @@ export class RenderStruct {
     return struct;
   }
 
+  static removeSmallAttachemntPointLabelsInModal(
```
rrodionov91, line 30: typo in word "Attachemnt"

packages/ketcher-core/src/application/render/renderStruct.ts
```diff
@@ -24,6 +27,31 @@ export class RenderStruct {
...
+      const isAttachmentPointAtom = attachmentPointRegExp.test(atom.label.text);
+
+      if (isAttachmentPointAtom) {
+        const isSmall =
+          atom.label.path.node.getBoundingClientRect().width <
+          MIN_ATTACHMENT_POINT_SIZE;
+        if (isSmall) {
+          atom.label.path.node.remove();
+        }
+      }
```
rrodionov91, line 51: I believe it will look better if we'll write in that manner:
```
if (!isAttachmentPointAtom) {
      return;  
}

const isSmall =
          atom.label.path.node.getBoundingClientRect().width <
          MIN_ATTACHMENT_POINT_SIZE;
          
if (isSmall) {
  atom.label.path.node.remove();
}
```

rrodionov91 (changes requested):

Issue #3635, opened 2023-11-30
Macro: Misalignment of monomers when imposing on top of each other


chgayane:
Steps to reproduce:

1. Select two monomers that need to be imposed on top of each other.
2. Position the first monomer above the second one.
3. Use the bond tool to aim for the attachment points of the underlying monomer.
4. Observe the resulting misalignment and scattering of the attachment points.

![MicrosoftTeams-image](https://github.com/epam/ketcher/assets/72468284/b993145e-9c67-483c-a072-46cf6c762035)

Actual behavior:
The attachment points of the underlying monomer become scattered, resulting in distorted monomers and misalignment.

Expected behavior:
The attachment points should remain intact, and the monomers should align correctly without scattering or distortion.

Zhirnoff:
Autotest added to the table.

PR #3662, opened 2023-12-06
Macro: #3635 - Misalignment of monomers when imposing on top of each other

chgayane:
Before: The attachment points of the underlying monomer become scattered, resulting in distorted monomers and misalignment.
image (1)
After: The attachment points should remain intact, and the monomers should align correctly without scattering or distortion.
image

Updated the logic used to find the point on the monomer border. Replaced `document.elementFromPoint` with `document.elementsFromPoint` to accurately handle overlapping monomers. Added additional type checks to cater to 'null' conditions. This change resolves issues related to incorrect border point and angle calculation when monomers overlap.

## Check list
- [ ] unit-tests written
- [ ] e2e-tests written
- [ ] documentation updated


StarlaStarla:
Seems there may still be defects with the attachment points of the underlying monomer
image

chgayane:
> I checkouted this branch and tried, seems the attachment points of the underlying monomer are still not perfectly aligned image

@StarlaStarla 
Yes, thank you for noticing that. There have been changes in the constants within the getSearchFunction, specifically where newLength is computed by dividing by a value approximately 1.5 times smaller. Since the new constants are effectively 1.5 times larger, I've considered adjusting the original divisor, multiplying it by 1.5. Could you please check this out?

StarlaStarla:
> > I checkouted this branch and tried, seems the attachment points of the underlying monomer are still not perfectly aligned image
> 
> @StarlaStarla Yes, thank you for noticing that. There have been changes in the constants within the getSearchFunction, specifically where newLength is computed by dividing by a value approximately 1.5 times smaller. Since the new constants are effectively 1.5 times larger, I've considered adjusting the original divisor, multiplying it by 1.5. Could you please check this out?

@chgayane Hi, I am not sure I am 100% correct. In my opinion, the issue lies within this block of code, especially with the document.elementFromPoint which will returns the topmost element at the specified coordinates. When there are no top monomers and the `newPointCoord` is in current monomer,  the topmost element will be the current monomer. However, if there is a top monomer and the `newPointCoord` is in the intersection of the two monomers, the top monomer will be considered as the topmost element, leading to incorrect angle calculation.
image

packages/ketcher-core/src/domain/helpers/attachmentPointCalculations.ts
```diff
@@ -130,16 +130,27 @@ export function getSearchFunction(
...
+        newPointCoord.x,
+        newPointCoord.y,
+      );
+
+      for (const element of elementsAtPoint) {
+        if (element === monomer.renderer.bodyElement.node()) {
+          newPoint = element;
+          break;
+        }
+      }
```
StarlaStarla, line 147: I would recomend using Array.some here instead, which should be more concise.

eg: `const isMonomerAtPoint = elementsAtPoint.some(element =>element  === newPoint ); 
      if(isMonomerAtPoint ){
      newAngle = initialAngle;
      } else {
        newAngle = initialAngle - 180;
      } `

packages/ketcher-core/src/domain/helpers/attachmentPointCalculations.ts
```diff
@@ -130,16 +130,22 @@ export function getSearchFunction(
...
-    ) as Element;
-
-    let newAngle;
-    if (newPoint === monomer.renderer?.bodyElement?.node()) {
-      newAngle = initialAngle;
-    } else {
-      newAngle = initialAngle - 180;
+    let newAngle: number = initialAngle;
+
+    if (monomer.renderer?.bodyElement) {
```
StarlaStarla, line 135: I propose to remove this line since we already used Optional chaining (?.) for it

StarlaStarla (approved):

Nitvex (approved):

Issue #2922, opened 2023-07-14
Modal for global error handling should be moved to "ketcher-react"


Inga-Vishnivetskaia:
**Background**
Global error handling was added in [PR 730](https://github.com/epam/ketcher/pull/730). With this implementation errorHandler triggers Modal on the client / demo application which is not a reliable approach (as for example, revealed in [bug 2790](https://github.com/epam/ketcher/issues/2790))

**Solution**
For fixing [bug 2790](https://github.com/epam/ketcher/issues/2790) the Info / Error modal was moved to "ketcher-react" ([PR 2868](https://github.com/epam/ketcher/pull/2868)). Triggering of modal for all other errors handled in [PR 730](https://github.com/epam/ketcher/pull/730) should be also moved to "ketcher-react". In this case the client does not need to import the InfoModal component and take care about this error handling.

**Additional context**
The text displayed in the modal for other errors has to be verified / discussed.

PR #3723, opened 2023-12-18
#2922 - Modal for global error handling should be moved to "ketcher-react"

chgayane:

## Check list
- [ ] unit-tests written
- [ ] e2e-tests written
- [ ] documentation updated

### DIFF
--- a/example/src/App.tsx
+++ b/example/src/App.tsx
@@ -50,8 +50,7 @@ if (enablePolymerEditor) {
 
 const App = () => {
   const hiddenButtonsConfig = getHiddenButtonsConfig();
-  const [hasError, setHasError] = useState(false);
-  const [errorMessage, setErrorMessage] = useState('');
+  const [, setErrorMessage] = useState('');
   const [showPolymerEditor, setShowPolymerEditor] = useState(false);
 
   const togglePolymerEditor = (toggleValue: boolean) => {
@@ -74,7 +73,6 @@ const App = () => {
     <>
       <Editor
         errorHandler={(message: string) => {
-          setHasError(true);
           setErrorMessage(message.toString());
         }}
         buttons={hiddenButtonsConfig}
@@ -92,19 +90,6 @@ const App = () => {
         }}
         togglerComponent={togglerComponent}
       />
-      {hasError && (
-        <InfoModal
-          message={errorMessage}
-          close={() => {
-            setHasError(false);
-
-            // Focus on editor after modal is closed
-            const cliparea: HTMLElement | null =
-              document.querySelector('.cliparea');
-            cliparea?.focus();
-          }}
-        />
-      )}
     </>
   );
 };
--- a/example/src/App.tsx
+++ b/example/src/App.tsx
@@ -1,7 +1,7 @@
 import 'ketcher-react/dist/index.css';
 
 import { useState } from 'react';
-import { ButtonsConfig, Editor, InfoModal } from 'ketcher-react';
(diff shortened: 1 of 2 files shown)


example/src/App.tsx
```diff
@@ -73,10 +71,8 @@ const App = () => {
   ) : (
     <>
       <Editor
-        errorHandler={(message: string) => {
-          setHasError(true);
-          setErrorMessage(message.toString());
-        }}
+        // eslint-disable-next-line @typescript-eslint/no-empty-function
+        errorHandler={() => {}}
```
StarlaStarla, line 75: I checked soem other places are still using this `editor.errorHandler`, do we need to remove it completely? and all the corresponding places of usage?

StarlaStarla: Why don't we move this old errorHandler since it seems doing nothing here?

packages/ketcher-react/src/script/ui/views/modal/components/document/Open/Open.tsx
```diff
@@ -113,7 +116,19 @@ const Open: FC<Props> = (props) => {
...
+    (
+      dispatch(
+        load(res.structStr, { badHeaderRecover: true, fragment: res.fragment }),
+      ) as unknown as Promise<any>
+    ).catch((error) => {
+      if (ketcher && ketcher.eventBus) {
+        ketcher.eventBus.emit(GLOBAL_ERROR_HANDLER, error.message);
+      }
+    });
+  };
```
StarlaStarla, line 131: Maybe we could use` try catch`，which I checked the `load` func already implemented  in the parent component.

packages/ketcher-react/src/script/ui/state/server/index.js
```diff
@@ -132,6 +132,7 @@ export function analyse() {
       .catch((e) => {
         KetcherLogger.error('index.js::analyse', e);
         editor.errorHandler(e);
+        throw e;
```
StarlaStarla, line 135: Why dont we implement the same way as `ketcher.eventBus.emit(GLOBAL_ERROR_HANDLER, error.message);` here and remove the old `editor.errorHandler(e)`?

packages/ketcher-react/src/script/ui/state/shared.ts
```diff
@@ -184,6 +184,7 @@ export function load(struct: Struct, options?) {
       KetcherLogger.error('shared.ts::load', e);
       dispatch(setAnalyzingFile(false));
       e && errorHandler && errorHandler(e.message);
+      throw e;
```
StarlaStarla, line 187: same here

example/src/App.tsx
chgayane: I attempted to remove the editor.errorHandler, but it led to numerous crashes throughout the application because of its extensive usage in various scenarios.  I believe it's not ideal to remove it at this stage. I suggest keeping it for now.

packages/ketcher-react/src/script/ui/views/modal/components/document/Open/Open.tsx
StarlaStarla: Apologies for not making myself clear. What I meant was that there's already a try-catch for `load` func in `shared.ts`, so it may not be needed here.

chgayane: OK, got it, thanks @StarlaStarla))

Issue #3666, opened 2023-12-07
Macro: After opening a file in macro mode, structure is not in center of the screen and need scroll to find it


Zhirnoff:
**Steps to Reproduce**
1. Switch to Macro mode
2. Open file [peptide_Bzl.zip](https://github.com/epam/ketcher/files/13602282/peptide_Bzl.zip)

**Actual behavior**
After opening a file in macro mode, structure is not in center of the screen but at the top. You have to scroll to find the open structure.

**Expected behavior**
After opening a file in macro mode, structure is in center of the screen.

**Screenshots**

https://github.com/epam/ketcher/assets/62464241/05895f8f-6984-424a-82ac-ea70ba68dc1e

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser Chrome
 - Version 112.0.5615.138 (Official Build) (64-bit)

**Ketcher version** 
[Version 2.17.0-rc.1](https://lifescience.opensource.epam.com/ketcher/index.html)

Zhirnoff:
Autotests added.

Zhirnoff:
Tested. Fixed.

PR #3782, opened 2023-12-23
Macro: #3666 - After opening a file in macro mode, structure is not in center of the screen and need scroll to find it

StarlaStarla:

## Check list
- [ ] unit-tests written
- [x] e2e-tests written
- [ ] documentation updated


packages/ketcher-core/src/application/editor/shared/coordinates.ts
```diff
@@ -6,7 +6,7 @@ import ZoomTool from '../tools/Zoom';
  * `canvas` -- The real coordinates used to draw entities
  * `view` -- The zoomed canvas coordinates
  */
-class Coordinates {
+export class Coordinates {
```
Nitvex, line 9: Why have you decided to change it to named export?

packages/ketcher-core/src/domain/serializers/ket/ketSerializer.ts
```diff
@@ -234,7 +234,34 @@ export class KetSerializer implements Serializer<Struct> {
...
       },
     );
-    return { error, parsedFileContent };
+    return {
+      error,
+      parsedFileContent,
+    };
+  }
+
+  convertCoordinatesAndGetMinMax(parsedFileContent: IKetMacromoleculesContent) {
```
Nitvex, line 243: My proposal is to move it outside serializer. Serializer doesn't know anything about the canvas. Imagine you're sending these coordinates to indigo (backend), what is the size of canvas on backend?
So I believe we need to parse the structure from Indigo and then separately center it on our canvas.

Nitvex (changes requested):

packages/ketcher-core/src/application/editor/shared/coordinates.ts
StarlaStarla: Because I encountered error when I tried to import the it and I am not sure why. So i changed it.
image

Nitvex: you can try to run `npm run build` from the root of the project

StarlaStarla: Same error even run npm run build from the root of the project

chgayane (approved):

Nitvex (approved):

Issue #3756, opened 2023-12-21
Macro: Loading from file and than - Undo changes causes exceptions and breakes selection/moving fuctionality


AlexeyGirin:
**Steps to Reproduce**
1. Switch to Macromolecules view
2. Load this file: [PeptideEnumeration_One_Two_Three connections3.zip](https://github.com/epam/ketcher/files/13740246/PeptideEnumeration_One_Two_Three.connections3.zip) (unzip first)
3. Load the same file once again
4. Press Undo button ![image](https://github.com/epam/ketcher/assets/26869421/d2ddbcec-7642-4c50-934d-f3b55516f04a)
5. Select all elements on the canvas
![image](https://github.com/epam/ketcher/assets/26869421/9e8e3500-6ee5-4db8-bbfc-6064e6e88067)
6. Try to move selection 

**Actual behavior**
System throws exceptions to console: [polymer.test.lifescience.opensource.epam.com-1703157072453.log](https://github.com/epam/ketcher/files/13740632/polymer.test.lifescience.opensource.epam.com-1703157072453.log)
Moving functionality is broken
![image](https://github.com/epam/ketcher/assets/26869421/eeda28ef-501d-4fe0-9d20-823cccaf229e)

**Expected behavior**
No exception in log, app works as intended

**Desktop (please complete the following information):**
 - OS: win10
 - Browser Chrome 120.0.6099.110 (Официальная сборка), (64 бит)

**Ketcher version** [e.g. v2.4.2]. 
![image](https://github.com/epam/ketcher/assets/26869421/a04bc8c1-4141-47c9-82df-73498321c4b0)

Zhirnoff:
Test added to automation table.

AlexeyGirin:
Fixed.

PR #3817, opened 2023-12-28
 #3756 - Macro: Loading from file and than - Undo changes causes exceptions and breakes selection/moving fuctionality

StarlaStarla:

The root cause is `allEntities` in `DrawingEntitiesManager` of editor doesn't get updated when undo and redo so that the `SelectRectangle` tool can not decide correctly whether one entity is in the rectangle.

## Check list
- [ ] unit-tests written
- [x] e2e-tests written
- [ ] documentation updated


StarlaStarla:
The root cause is `allEntities` in `DrawingEntitiesManager` of editor doesn't get updated when undo and redo so that the `SelectRectangle` tool can not decide correctly whether one entity is in the rectangle.

rrodionov91 (changes requested):
I would propose another solution. Please check this PR https://github.com/epam/ketcher/pull/3822

Issue #3713, opened 2023-12-18
Macro: Hot keys for toolbar buttons are not implemented in Macromoleculas view


AlexeyGirin:
**Background**
At Molecular view we are able to use key bindings for every toolbar button - ctrl+z, ctrl+y and etc.
![image](https://github.com/epam/ketcher/assets/26869421/7c3af476-5f47-4c9a-961d-d02b2ed9300e)
Every button has hot key assigned and tooltip

**Solution**
We have to have the same on Macro view

Clear Canvas: Ctrl+Del, Ctrl+Backspace
Undo: Ctrl+Z
Redo: Ctrl+Shit+Z, Ctrl+Y
Erase: Del, Backspace
Select Rectangle: Shift+Tab, Esc
Zoom In: Ctrl+=
Zoom Out: Ctrl+_
Reset Zoom: Ctrl+0
"Select All" operation (not a button): Ctrl+A

Zhirnoff:
Tests added to automation table.

Zhirnoff:
Tested. Some hotkeys only work in micro mode or macro mode. We need consistency. An additional bug has been created.
https://github.com/epam/ketcher/issues/3917

PR #3836, opened 2024-01-04
Macro: #3713 - Hot keys for toolbar buttons are not implemented in Macromoleculas view

rrodionov91:

## Check list
- [ ] unit-tests written
- [ ] e2e-tests written
- [ ] documentation updated


packages/ketcher-core/src/application/editor/Editor.ts
```diff
@@ -74,6 +78,22 @@ export class CoreEditor {
     return editor;
   }
 
+  private handleHotKeyEvents(event) {
+    const keySettings = hotkeysConfiguration;
+    const hotKeys = initHotKeys(keySettings);
+    const shortcutKey = keyNorm.lookup(hotKeys, event);
+
+    if (keySettings[shortcutKey] && keySettings[shortcutKey].handler) {
```
Nitvex, line 86: We can simplify
```
if (keySettings[shortcutKey]?.handler)
```

packages/ketcher-core/src/application/editor/Editor.ts
```diff
@@ -74,6 +78,22 @@ export class CoreEditor {
...
+      keySettings[shortcutKey].handler(this);
+      event.preventDefault();
+    }
+  }
+
+  setupHotKeysEvents() {
+    this.hotKeyEventHandler = (event) => this.handleHotKeyEvents(event);
+    document.addEventListener('keydown', this.hotKeyEventHandler);
```
Nitvex, line 94: Can we just simply?
```
setupHotKeysEvents() {
  document.addEventListener('keydown', this. handleHotKeyEvents);
}
```

Nitvex (approved):

packages/ketcher-core/src/application/editor/Editor.ts
rrodionov91: Nope, because we need to bind 'this' to hotKeyEventHandler method. And this.hotKeyEventHandler.bind(this) also didn't work because it returns another function and it did not clear event listener in this case.

Issue #3728, opened 2023-12-19
Macro: Group selection using Shift+LClick is not implemented


AlexeyGirin:
**Background**
It is unable to select/unselect macromolecules using Shift+LClick. The only way to select group of elements is area selection (by mouse).

**Solution**
Implement group selection using Shift+LClick in the same manner as it is done in Molecules view

Zhirnoff:
Test added to the automation table.

Zhirnoff:
Test automated.

PR #3879, opened 2024-01-12
#3728 - Add group selection for macromolecules

Balzamova:

## Check list
- [x] unit-tests written
- [ ] e2e-tests written
- [ ] documentation updated


packages/ketcher-core/src/domain/entities/DrawingEntitiesManager.ts
```diff
@@ -175,6 +175,29 @@ export class DrawingEntitiesManager {
...
+      command.addOperation(operation);
+    } else {
+      drawingEntity.turnOnSelection();
+      const operation = new DrawingEntitySelectOperation(drawingEntity);
+      command.addOperation(operation);
+    }
+
+    return command;
+  }
+
```
ilya-asiyuk-epam, line 200: I suggest below changes:

`  private addOperationToCommand(
    command: Command,
    drawingEntity: DrawingEntity,
  ) {
    const operation = new DrawingEntitySelectOperation(drawingEntity);
    command.addOperation(operation);
  }

  public selectCurrentDrawingEntities(drawingEntity: DrawingEntity) {
    const command = new Command();

    this.allEntities.forEach(([, drawingEntity]) => {
      if (drawingEntity.selected) {
        this.addOperationToCommand(command, drawingEntity);
      }
    });

    if (drawingEntity.selected) {
      drawingEntity.turnOffSelection();
    } else {
      drawingEntity.turnOnSelection();
    }
    this.addOperationToCommand(command, drawingEntity);

    return command;
  }`

ilya-asiyuk-epam (approved):

packages/ketcher-core/src/domain/entities/DrawingEntitiesManager.ts
rrodionov91: I believe that addOperationToCommand is not the best name for this method, because it actually adds concrete DrawingEntitySelectOperation. 
However I would propose to not create this method at all. We can do like this in selectCurrentDrawingEntities method:
```
      command.addOperation(
         new DrawingEntitySelectOperation(drawingEntity)
      );
```

packages/ketcher-core/src/domain/entities/DrawingEntitiesManager.ts
```diff
@@ -175,6 +175,33 @@ export class DrawingEntitiesManager {
...
+    if (drawingEntity.selected) {
+      drawingEntity.turnOffSelection();
+    } else {
+      drawingEntity.turnOnSelection();
+    }
+    this.addOperationToCommand(command, drawingEntity);
+
+    return command;
+  }
```
rrodionov91, line 203: It looks like a bit overhead to create operations for already selected entities. Can we rework and rename this method to add only one drawing entity to selection?

Issue #2912, opened 2023-07-11
Fix memory leak for Ketcher when it is opened inside modal


Nitvex:
**Steps to Reproduce**
1. Open Ketcher inside modal
2. Close the modal
3. Open Ketcher again
4. Repeat steps 1-3 multiple times

**Actual behavior**
User receives memory leak warning:
![MicrosoftTeams-image (41)](https://github.com/epam/ketcher/assets/41272128/0412c5c6-c1d8-4139-83fb-f7daa59ce476)

**Expected behavior**
Memory leak is eliminated. No warning message is shown.

hellhorse123:
I still get error:
`MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 layout listeners added. Use emitter.setMaxListeners() to increase limit`
when open/close modal with ketcher at many times

rrodionov91:
Hi @hellhorse123 
Thank you for notifying us. We will check

PR #3887, opened 2024-01-14
#3711 - @react-refresh removes Ketcher from DOM

### LINKED ISSUE #3711: `@react-refresh` removes Ketcher from DOM
**Steps to Reproduce**
1. Do a new, clean React install (issue observed in latest Vite and latest CRA)
2. Install Ketcher v2.14+
3. Try to render `Editor` from `ketcher-react`.

**Actual behavior**
On very first render, Ketcher appears as normal, but on the first pass of React re-rendering itself Ketcher disappears. It's containing `div` is suddenly empty. 

**Expected behavior**
It doesn't disappear. 

**Screenshots**

https://github.com/epam/ketcher/assets/24811404/87a7499b-22d7-4def-99ed-b332f3607526

Here's me going through debugger and seeing where(ish) the problem is happening. 
https://www.loom.com/share/0d0fc86a067b42ec83e237a907245be2

 - OS: MacOS Sonoma 14.2
 - Observed in latest Chromium (Brave, Chrome), Firefox, and Safari

**Ketcher version** [e.g. v2.4.2]. =
```
    "ketcher-core": "^2.15.1",
    "ketcher-react": "^2.15.1",
    "ketcher-standalone": "^2.15.1",
    "react": "^18.2.0",
    "react-dom": "^18.2.0"
```

**Additional context**
Bug is NOT observed on a production build. 
Bug seems to have been introduced in `2.14`. `2.13` runs fine. 

Here's my barebones test component, being used in a new Vite setup (same behavior can also be seen in new CRA setup).
```typescript
import 'ketcher-react/dist/index.css'

import { Editor } from 'ketcher-react'
import { StandaloneStructServiceProvider } from 'ketcher-standalone'

const structServiceProvider = new StandaloneStructServiceProvider()

export const KetcherEditor = () => {
  return (
    <div>
      <Editor
        staticResourcesUrl="http://localhost:5173"
        onInit={() => {
          debugger
          console.log('hello from init')
        }}
        errorHandler={() => {
          console.log('error!')
        }}
      />
    </div>
  )
}

```

andy197:
The root cause was the `setTimeout`. Due to the double invoke in Strict Mode, it is possible for the cleanup from the first invocation to run after the second `useEffect` has already started. This means that before `setTimeout` get's executed from the first `useEffect`,  there is a new `appRoot` from the second `useEffect`, and it get's unmounted when `setTimeout` function actually executes.

~~The solution is to check if `appRoot` has been mounted or not. This will prevent unmounting twice.~~

`appRoot.unmount();` should be synchronous so that it executes immediately after the unmount. I think it's worth it even if we get the 
`// setTimeout is used to disable the warn msg from react "Attempted to synchronously unmount a root while React was already rendering"`
. Otherwise folks who use ketcher with strictmode would not be able to use ketcher at all.

closes #3711 

## Check list
- [ ] unit-tests written
- [ ] e2e-tests written
- [ ] documentation updated

### DIFF
--- a/packages/ketcher-react/src/Editor.tsx
+++ b/packages/ketcher-react/src/Editor.tsx
@@ -22,7 +22,7 @@ import 'whatwg-fetch';
 import './index.less';
 
 import init, { Config } from './script';
-import { useEffect, useRef } from 'react';
+import { useEffect, useRef, useState } from 'react';
 
 import { Ketcher } from 'ketcher-core';
 import classes from './Editor.module.less';
@@ -32,7 +32,7 @@ import {
   KETCHER_INIT_EVENT_NAME,
   KETCHER_ROOT_NODE_CLASS_NAME,
 } from './constants';
-import { createRoot } from 'react-dom/client';
+import { createRoot, Root } from 'react-dom/client';
 
 const mediaSizes = {
   smallWidth: 1040,
@@ -50,7 +50,7 @@ function Editor(props: EditorProps) {
     ref: rootElRef,
   });
   const ketcherInitEvent = new Event(KETCHER_INIT_EVENT_NAME);
-
+  const [appRoot, setAppRoot] = useState<Root | null>(null);
   useEffect(() => {
     const appRoot = createRoot(rootElRef.current as HTMLDivElement);
     init({
@@ -63,12 +63,20 @@ function Editor(props: EditorProps) {
         window.dispatchEvent(ketcherInitEvent);
       }
     });
-    return () => {
-      appRoot.unmount();
-    };
+    // set root to a state so we can use it in the other useEffect
+    setAppRoot(appRoot);
     // TODO: provide the list of dependencies after implementing unsubscribe function
   }, []);
-
+  // separate useEffect for cleanup This prevents funkyness in strictmode.
+  useEffect(() => {
+    return () => {
+      // setTimeout is used to disable the warn msg from react "Attempted to synchronously unmount a root while React was already rendering"
+      setTimeout(() => {
+        appRoot?.unmount();
+        setAppRoot(appRoot);
+      });
+    };
+  }, [appRoot]); // if appRoot changes, cleanup
   return (
     <div
       ref={rootElRef}
--- a/packages/ketcher-react/src/Editor.tsx
+++ b/packages/ketcher-react/src/Editor.tsx
@@ -22,7 +22,7 @@ import 'whatwg-fetch';
 import './index.less';
 
 import init, { Config } from './script';
-import { useEffect, useRef, useState } from 'react';
+import { useEffect, useRef } from 'react';
 
 import { Ketcher } from 'ketcher-core';
 import classes from './Editor.module.less';
@@ -32,7 +32,7 @@ import {
   KETCHER_INIT_EVENT_NAME,
   KETCHER_ROOT_NODE_CLASS_NAME,
 } from './constants';
-import { createRoot, Root } from 'react-dom/client';
+import { createRoot } from 'react-dom/client';
 
 const mediaSizes = {
   smallWidth: 1040,
@@ -50,7 +50,7 @@ function Editor(props: EditorProps) {
     ref: rootElRef,
   });
   const ketcherInitEvent = new Event(KETCHER_INIT_EVENT_NAME);
-  const [appRoot, setAppRoot] = useState<Root | null>(null);
+
   useEffect(() => {
     const appRoot = createRoot(rootElRef.current as HTMLDivElement);
     init({
@@ -63,20 +63,12 @@ function Editor(props: EditorProps) {
         window.dispatchEvent(ketcherInitEvent);
       }
     });
-    // set root to a state so we can use it in the other useEffect
-    setAppRoot(appRoot);
+    return () => {
+      appRoot.unmount();
+    };
     // TODO: provide the list of dependencies after implementing unsubscribe function
   }, []);
-  // separate useEffect for cleanup This prevents funkyness in strictmode.
-  useEffect(() => {
-    return () => {
-      // setTimeout is used to disable the warn msg from react "Attempted to synchronously unmount a root while React was already rendering"
-      setTimeout(() => {
-        appRoot?.unmount();
-        setAppRoot(appRoot);
-      });
-    };
-  }, [appRoot]); // if appRoot changes, cleanup
+
   return (
     <div
       ref={rootElRef}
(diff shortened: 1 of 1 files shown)


packages/ketcher-react/src/Editor.tsx
```diff
@@ -64,10 +64,14 @@ function Editor(props: EditorProps) {
       }
     });
     return () => {
-      // setTimeout is used to disable the warn msg from react "Attempted to synchronously unmount a root while React was already rendering"
-      setTimeout(() => {
-        appRoot.unmount();
-      });
+      isUnmounted.current = true;
+      // In StrictMode, React double invokes and the component could be unmounted again due to the timeout.
```
StarlaStarla, line 68: A question here is that seems the isUnmounted.current would always be true and the unmount could not be triggered, which also got verified(memory leak still occurs) after I tested it with my old memory-leak-fix branch:
[2912-example-fix-memory-leak-for-ketcher-when-it-is-opened-inside-modal-1](https://github.com/epam/ketcher/tree/2912-example-fix-memory-leak-for-ketcher-when-it-is-opened-inside-modal-1)

andy197: Great catch! And you're right, it should be below. Unfortunately, this doesn't fix the problem so I investigated further.

This actually exposed the "real" reason why strict mode doesn't work:
Render once, unmount. `setTimeout` is put into the next event loop. 
Second mount. We create the root. We don't really have control when `setTimeout` executes, But from testing  `setTimeout` function is called after the useEffect and `appRoot.unmount();` is executed. This removes the `Editor`.

andy197: `appRoot.unmount();` should be synchronous so that it executes immediately after the unmount. I think it's worth it even if we get the 
`// setTimeout is used to disable the warn msg from react "Attempted to synchronously unmount a root while React was already rendering"`
. Otherwise folks who use ketcher with strictmode would not be able to use ketcher at all.

packages/ketcher-react/src/Editor.tsx
```diff
@@ -64,10 +64,7 @@ function Editor(props: EditorProps) {
       }
     });
     return () => {
-      // setTimeout is used to disable the warn msg from react "Attempted to synchronously unmount a root while React was already rendering"
-      setTimeout(() => {
-        appRoot.unmount();
-      });
+      appRoot.unmount();
     };
```
StarlaStarla, line 68: After some deeper investigation , I found this info from the official site:
`createRoot() controls the contents of the container node you pass in.
Any existing DOM elements inside are replaced when render is called.
Later calls use React’s DOM diffing algorithm for efficient updates.`

My understanding is that DOM elements inside the appRoot would only be updated in the second render in strict mode,  while listeners will be attached twice which causes the memory leak.

While for now, I am not sure the best fix for this issue.

andy197: Hi Stella! Thank you for the reply.

Can you clarify what you mean by 
>while listeners will be attached twice which causes the memory leak.
I'm not very familiar with the code base, but wouldn't the listeners be removed when we unmount?

If the listeners aren't being destroyed and are attached twice, then this would only occur when we are in strict mode. i.e., dev mode. I think this is okay?

>While for now, I am not sure the best fix for this issue.

Would there be an issue with just removing the setTimeout? I'm seeing the error we're getting documented here https://github.com/facebook/react/issues/25675
"Attempted to synchronously unmount a root while React was already rendering".

StarlaStarla: Hi! 
By this `while listeners will be attached twice which causes the memory leak.`, I meant the reason for the original bug I fixed here: https://github.com/epam/ketcher/issues/2912.
Removing the setTimeout seems working here, just I am sure if it is the best practice. 
If we really follow this fix, we may also need to add some extra checks for this function, otherwise the first appRoot will be null after unmounted and there is another error for it.
image
image

andy197: Gotcha thanks! I will look into that. Can I ask how you replicated
>otherwise the first appRoot will be null after unmounted and there is another error for it.

Does this occur in strictmode only or in regular behavior? I'll see what I can do!

StarlaStarla: Yes, the error occurs in strictmode only. If it's not too much trouble, consider checking out this branch, running `npm i` in the root path, and then launching the project to try.

andy197: Okay I tried seeing what we can do, and I don't think theres a good solution taking account of strictMode.

We can't directly check for truthy in for `_internalRoot`. In general, we really shouldn't mess with the internal fibers. Future react updates would break this.

If we try to defer unmount, we get 
image

I thought of this, but it feels really hacky. I haven't tried it yet though.
https://dev.to/ag-grid/react-18-avoiding-use-effect-getting-called-twice-4i9e

What are your thoughts on this? In return for working in strict mode, we'll get this error along with the error you tried to silence.

andy197: I wonder if it is possible to not do this in a useEffect?

StarlaStarla: Hi, as for this solution: https://dev.to/ag-grid/react-18-avoiding-use-effect-getting-called-twice-4i9e, I agree with the comment by `Alex Chiu`, is not a good practice.
'I wonder if it is possible to not do this in a useEffect?' --- I am not sure how to achieve this. I've tried several methods to resolve this issue, but none have been successful so far. Do you have a clear thought yet?

andy197: Hi Starla,
I do not unfortunately :(. I think we just have to live with it? It's not worst solution imo, but at the minimum, strict mode is useable for people.

StarlaStarla: Hi Andy,
Yes, I agree with living with it for now untill we could find a better solution to hide the warnnings.

andy197: Hi Starla,
Just to clarify, this PR is good to go?

StarlaStarla (approved):

packages/ketcher-react/src/Editor.tsx
StarlaStarla: Yes, I think so if we agree to live with the warnnings now.

andy197:
@StarlaStarla Thank you Starla for the review! What would the next step be for merging this in? 🙂

andy197:
Hi @StarlaStarla!

Sorry for the ping, but just in case you missed this! Is there else that needs to be done? 
We would like to get this in so we can use the latest ketcher!

emripka:
Hi all! When can we expect to get this PR merged in?

rrodionov91:
Hi @emripka  @andy197,
We fixed errors in strict mode in this PR https://github.com/epam/ketcher/pull/5013

Issue #3900, opened 2024-01-15
Macro: The last letter in RNA Builder name field cannot be deleted


Zhirnoff:
**Steps to Reproduce**
1. Switch to Macro mode
2. Select RNA tab
3. Expand RNA Builder
4. Add any text to name field
5. Try delete text

**Actual behavior**
The last letter in RNA Builder name field cannot be deleted.

**Expected behavior**
The all letters in RNA Builder name field can be deleted.

**Screenshots**
![2024-01-15_22h48_43](https://github.com/epam/ketcher/assets/62464241/a4a70a74-6338-4a18-8d5b-2494ddf34710)

**Desktop (please complete the following information):**
 - OS: Windows 11
 - Browser Chrome
 - Version 112.0.5615.138 (Official Build) (64-bit)

**Ketcher version** 
[Version 2.18.0-rc.1]

Zhirnoff:
Test added to the automation table.

Zhirnoff:
Tested. Bug is fixed.

PR #3908, opened 2024-01-16
Macro: #3900 - The last letter in RNA Builder name field cannot be deleted

chgayane:
The trim() method in the `onChangeName` function prevented the deletion of the last letter. Adjusted logic to allow for the removal of all characters, while still trimming input for the new state.

https://github.com/epam/ketcher/assets/72468284/10317cc4-5156-4319-a00a-f6290879ae1c

## Check list
- [ ] unit-tests written
- [ ] e2e-tests written
- [ ] documentation updated

### DIFF
--- a/packages/ketcher-macromolecules/src/components/monomerLibrary/RnaBuilder/RnaEditor/RnaEditorExpanded/RnaEditorExpanded.tsx
+++ b/packages/ketcher-macromolecules/src/components/monomerLibrary/RnaBuilder/RnaEditor/RnaEditorExpanded/RnaEditorExpanded.tsx
@@ -146,10 +146,16 @@ export const RnaEditorExpanded = ({
   const onChangeName = (event: ChangeEvent<HTMLInputElement>) => {
     if (isEditMode) {
       const newPresetName = event.target.value;
-      if (newPresetName.trim() !== '') {
+      if (newPresetName !== '') {
         setNewPrest({
           ...newPreset,
           name: newPresetName.trim(),
+          editedName: true,
+        });
+      } else {
+        setNewPrest({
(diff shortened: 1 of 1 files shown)


Zhirnoff (approved):

packages/ketcher-macromolecules/src/components/monomerLibrary/RnaBuilder/RnaEditor/RnaEditorExpanded/RnaEditorExpanded.tsx
```diff
@@ -146,12 +146,18 @@ export const RnaEditorExpanded = ({
...
           name: newPresetName.trim(),
           editedName: true,
         });
+      } else {
+        setNewPrest({
+          ...newPreset,
+          name: '',
+          editedName: true,
+        });
       }
```
StarlaStarla, line 161: How about we just remove this judgement `if (newPresetName !== '')` and setNewPrest directly?

chgayane: @StarlaStarla Yes, we can simplify it that way. However, it will always trim names, removing leading/trailing spaces even on empty input. If we're okay with this, I'll update the code.

StarlaStarla: I would think it is ok. But maybe we could consult Roman on this.

rrodionov91 (approved):

Issue #3899, opened 2024-01-15
Macro: Attachment points do not disappear when hover is removed from some monomers.


Zhirnoff:
**Steps to Reproduce**
1. Switch to Macro mode
2. Add monomers to canvas
3. Select Bond tool
4. Hover monomers going through one to another
Note: Also reproduced when hovering monomers connected by bonds

**Actual behavior**
Attachment points do not disappear when hover is removed from some monomers.

**Expected behavior**
Attachment points disappear when hover is removed from monomers.

**Screenshots**

https://github.com/epam/ketcher/assets/62464241/e0fe9448-f69b-4b54-a044-fe855c4a53e1

**Desktop (please complete the following information):**
 - OS: Windows 11
 - Browser Chrome
 - Version 112.0.5615.138 (Official Build) (64-bit)

**Ketcher version** 
[Version 2.18.0-rc.1]

Zhirnoff:
Bug is still reproducing
![2024-02-09_18h02_03](https://github.com/epam/ketcher/assets/62464241/897a6f1d-a33a-4537-a1eb-224f36d87759)

rrodionov91:
Finally fixed in https://github.com/epam/ketcher/pull/4482

NadezhdaPeskun:
Verified as fixed for CI Version 
2.21.0-rc.1 Build at 2024-04-26
[Indigo Toolkit](http://lifescience.opensource.epam.com/indigo/)
Version 1.20.0-dev.1.25-gafde2b35-wasm32-wasm-clang-12.0.0

OS: Windows 11
Browser Chrome
Version 112.0.5615.138 (Official Build) (64-bit)

Zhirnoff:
Tested on release 2.21.0-rc.2
Fixed.

https://github.com/epam/ketcher/assets/62464241/c19993b8-ea75-4683-9679-9d1432e17cb3

PR #4190, opened 2024-03-04
#3899 - Attachment points do not disappear when hover is removed

StarlaStarla:

**Root cause:** 
When mouse over the current attachment points, they will be destoryed in a setTimeout function and new set of attachment points will be created. There would be no issue if current attachment points are destoryed after their mouseleave event be triggered because the mouseleave event will trigger actions to clear the new created attachment points. (**The important point here is that its always the old attachment point's mouseleave event being triggered to clear the new created attachment points** )
While if current attachment points are destoryed before its mouseleave event be triggered(because the execution time of the settimeout function couldn't be guaranteed), then their mouseleave event will not be triggered and new created set of attachment points will be left in the view.

**Solution:** 
Instead of creating new attachment points everytime when mouseover event listener triggered, just update attachment points' styles when there are already old attachment points existing. **That is to prevent the redrawing of attachment points.**

**Updates for this PR:**
While at first, I didn't realize two situations that will redraw the attachment points and trigger this problem to happen: hovering and leaving the attachment points, and mouse leaving a monomer to its points. This PR https://github.com/epam/ketcher/pull/4035 only solved the problem for first situation. 
So for the second scenario, a check was added to determine if redrawing the attachment points is necessary. If the mouse is moving from a monomer to its points, we don't need to redraw. But if it's moving to the outer canvas, we do.--------But obviously, there are some issues with this solution, because the the mouse could also leave the monomer and move to other elements like texts or polymerbonds etc.
The final fix is to check whether the mouse leaves the monomer and move to the `attachmentPoint-hoverable-element`, which is the `hoverableAreaElement` we created for attachment points.  If yes, then we dont need to redraw attachment points; otherwise, we do.

## Check list
- [ ] unit-tests written
- [ ] e2e-tests written
- [ ] documentation updated


Zhirnoff (approved):

ilya-asiyuk-epam (approved):

ilya-asiyuk-epam:
Check with @rrodionov91 if it can be merged.

rrodionov91:
Fixed in https://github.com/epam/ketcher/pull/4482

Issue #4369, opened 2024-03-28
Macro: Del button works as Backspace button that is wrong


AlexeyGirin:
**Steps to Reproduce**
1. Switch to the Macro mode - Sequence mode
2. Add chains to the canvas (ATUC) and put cursor in the middle
![image](https://github.com/epam/ketcher/assets/26869421/9ee00ee7-42ee-4951-9928-db5352e11c2b)
3. Press Del key

**Actual behavior**
System removes T (i.e. removes symbol BEFORE cursor - this is how Backspace key works)
![image](https://github.com/epam/ketcher/assets/26869421/59cd7927-1d29-4ffd-ad0c-4b04be066b09)

**Expected behavior**
System removes U (i.e. removes symbol AFTER cursor - this is how Del key intended to work)
![image](https://github.com/epam/ketcher/assets/26869421/3cb8f455-709b-48e7-8189-e06cd7b0dcb6)

**Desktop (please complete the following information):**
 - OS: Windows 10 Pro (version 22H2) x64
 - Browser: Сhrome Version: (122.0.6261.112) x64
 - Ketcher version 2.20.0-rc.2 Build at 2024-03-27; 08:25:13
 - Indigo Toolkit Version 1.19.0-rc.2.0-g6c0e3fecf-wasm32-wasm-clang-12.0.0
 
Issue found while testing https://github.com/epam/ketcher/issues/3650

AlexeyGirin:
Not fixed.
Delete button works as Backspace still

AlexeyGirin:
Fixed.
![image](https://github.com/epam/ketcher/assets/26869421/4d58340e-d874-4769-86ee-f6a19aa6952d)

Ketcher Version 2.20.0-rc.3 Build at 2024-04-26; 16:57:15
Indigo Toolkit Version 1.19.0-rc.3.0-gf2d7689b8-wasm32-wasm-clang-12.0.0

OS: Windows 10
Browser Chrome Version 112.0.5615.138 (Official Build) (64-bit)

PR #4408, opened 2024-04-04
#4369 - Del button works as Backspace button that is wrong

burcinkarimov:
Backspace and Delete keys both had the same deletion behavior (deleting left node). Since Delete key should remove the right side, according changes were applied to separate their behavior. 

## Check list
- [ ] unit-tests written
- [ ] e2e-tests written
- [ ] documentation updated


packages/ketcher-core/src/application/editor/modes/SequenceMode.ts
```diff
@@ -480,54 +480,65 @@ export class SequenceMode extends BaseMode {
...
-          this.finishNodesDeletion(
-            modelChanges,
-            previousCaretPosition,
-            nodesToDelete[0][0].nodeIndexOverall,
-          );
+      this.finishNodesDeletion(
+        modelChanges,
+        nodesToDelete[0][nodesToDelete[0].length - 1].nodeIndexOverall + 1,
+        nodesToDelete[0][0].nodeIndexOverall,
+      );
```
rrodionov91, line 525: Looks like this change causes bug found by autotest. I believe we need to fix it here instead of update screenshot.

packages/ketcher-core/src/application/editor/modes/SequenceMode.ts
```diff
@@ -480,54 +480,65 @@ export class SequenceMode extends BaseMode {
...
-                  node: previousNode,
-                  nodeIndexOverall:
-                    SequenceRenderer.previousCaretPosition as number,
-                },
-              ],
-            ];
-          } else {
-            return;
-          }
+    const deleteNode = (direction: 'left' | 'right') => {
```
rrodionov91, line 483: I would propose to use enum instead of string literal type to modify it easier in future

rrodionov91 (changes requested):

rrodionov91 (approved):

NataliaLoginova (approved):

Issue #4329, opened 2024-03-23
Macro: Cyclic chains disappear from canvas when switching to sequence mode


Zhirnoff:
**Steps to Reproduce**
1. Switch to Macro mode
2. Add RNA or DNA to canvas
3. Make chain cyclic by connecting sugars through R2-R1 attachment points,
4. Switch to sequence mode

**Actual behavior**
Cyclic chains disappear from canvas when switching to sequence mode

**Expected behavior**
Cyclic chains not disappear from canvas when switching to sequence mode.
Additional link should be visible.
![image](https://github.com/epam/ketcher/assets/161723514/ad953135-9d79-4b1e-9590-0a861e76084b)

**Screenshots**

https://github.com/epam/ketcher/assets/62464241/3758b673-769b-4289-b6e0-dc605120f675

**Desktop (please complete the following information):**
 - OS: Windows 11
 - Browser Chrome
 - Version 112.0.5615.138 (Official Build) (64-bit)

**Ketcher version** 
[Version 2.20.0-rc.1]

AlexeyGirin:
Fixed.
![image](https://github.com/epam/ketcher/assets/26869421/715b5f0e-53fe-44f1-9b6f-5215801da76f)
Ketcher Version 2.20.0-rc.3 Build at 2024-04-26; 16:57:15
Indigo Toolkit Version 1.19.0-rc.3.0-gf2d7689b8-wasm32-wasm-clang-12.0.0

OS: Windows 10
Browser Chrome Version 112.0.5615.138 (Official Build) (64-bit)

rrodionov91:
+ backmerge to 2.21 https://github.com/epam/ketcher/pull/4606

PR #4426, opened 2024-04-05
#4329 - Macro: Cyclic chains disappear from canvas when switching to sequence mode

baranovdv:
- add function that handles cycled chains
- add cycled link drawing

https://github.com/epam/ketcher/assets/117748331/252e02a1-8c71-4a7c-8552-babaff2e1f08

## Check list
- [ ] unit-tests written
- [x] e2e-tests written
- [ ] documentation updated


ilya-asiyuk-epam:
My main question: 
should we allow and, as a result, display cyclic chain for such case (without P at the end of the chain)?
chain without p
cc: @Zhirnoff 

Does current fix support such case?
cc: @baranovdv , @NataliaLoginova

baranovdv:
> Does current fix support such case? cc: @baranovdv , @NataliaLoginova

@ilya-asiyuk-epam no, this this case is not supported. I can implement this after desired behavior would be known

baranovdv:
> > Does current fix support such case? cc: @baranovdv , @NataliaLoginova
> 
> @ilya-asiyuk-epam no, this this case is not supported. I can implement this after desired behavior would be known

> > Does current fix support such case? cc: @baranovdv , @NataliaLoginova
> 
> @ilya-asiyuk-epam no, this this case is not supported. I can implement this after desired behavior would be known

Or maybe it's ok to implement mechanic similar to HELM Editor (Phosphate is added automatically):

https://github.com/epam/ketcher/assets/117748331/b19ef259-a0ed-4cc7-b770-b8ee63da62e9

rrodionov91 (approved):

Issue #4346, opened 2024-03-27
Macro: Monomers disappear when switching to sequence view if they are attached to bases via the R2 attachment point


Zhirnoff:
**Steps to Reproduce**
1. Switch to macro
2. Open file [file.zip](https://github.com/epam/ketcher/files/14770235/file.3.zip)
3. Switch to Sequence mode
![image](https://github.com/epam/ketcher/assets/26869421/6f791635-28dd-4a81-ab6d-0e007b7f2d44)

**Actual behavior**
Monomers disappear when switching to sequence view if they are attached to bases via the R2 attachment point
![image](https://github.com/epam/ketcher/assets/26869421/bf7096e0-2aca-4c85-8410-b8e94d869fbd)

**Expected behavior**
Monomers are not disappear when switching to sequence view if they are attached to bases via the R2 attachment point
![image](https://github.com/epam/ketcher/assets/26869421/888e9bb4-cdcd-4d0b-9291-9e5f89bcdab2)

**Screenshots**

https://github.com/epam/ketcher/assets/62464241/acb35238-2435-4c83-a305-c1c7d2364a63

**Desktop (please complete the following information):**
 - OS: Windows 11
 - Browser Chrome
 - Version 112.0.5615.138 (Official Build) (64-bit)

**Ketcher version** 
[Version 2.20.0-rc.1]

Zhirnoff:
Test added to the automation table.

Zhirnoff:
Test automated.

AlexeyGirin:
Fixed.
![image](https://github.com/user-attachments/assets/342dbc4b-91e3-43e9-91d4-1ca2bc3ab2ac)
- Ketcher Version 2.23.0-rc.3 Build at 2024-07-25; 05:30:25
- Indigo Toolkit Version 1.23.0-dev.1.14-g284c33a1-wasm32-wasm-clang-12.0.0
- Windows 10 Pro
- Chrome Version 127.0.6533.73 (Official Build) (64-bit)

PR #4468, opened 2024-04-16
#4346 - Monomers disappear when switching to sequence view if they are attached to bases via the R2 attachment point

sonozyurt:

## Check list
- [ ] unit-tests written
- [ ] e2e-tests written
- [ ] documentation updated


packages/ketcher-core/src/domain/entities/PolymerBond.ts
```diff
@@ -78,6 +81,14 @@ export class PolymerBond extends DrawingEntity {
   }
 
   public get isSideChainConnection() {
+    if (
+      (this.secondMonomer instanceof Chem ||
+        this.secondMonomer instanceof Peptide) &&
+      this.firstMonomer instanceof RNABase
```
rrodionov91, line 87: The order of monomers can be opposite (Rna base can be second monomer).
Also we need to handle Rna base connections with all types of monomers. So if Rna base connection is not with sugar through r1(base) -> r3(sugar) then it should be interpreted as side connection

rrodionov91 (changes requested):

Issue #4509, opened 2024-04-22
In the Text-editing mode, when inserting a fragment after pressing Ctrl+V twice, the cursor is moved to another row


ViktoriaTkacheva:
**Steps to Reproduce**
1.	Open two Ketcher browser tabs – the Macro mode
2.	Draw three connected nucleotides
3.	Switch to the Sequence mode - the Text-editing mode
      <img width="125" alt="image" src="https://github.com/epam/ketcher/assets/139227567/67832b57-3673-4f69-a6f4-ec74b14cc7c9">

4.	Copy this sequence and paste it several times into the Text-editing mode of another Ketcher browser tab, where there is already some sequence (for example, AAAp or AAA)

**Actual behavior**
Only two fragments are inserted in each row, and after that, the cursor moves to the next row and the next insertion occurs on the new row
<img width="156" alt="image" src="https://github.com/epam/ketcher/assets/139227567/a3758206-a04d-411d-98ce-a08b8c24106e">

**Expected behavior**
Insertion on one row occurs until 30 letters are entered. If the inserted fragment contains a nucleotide that will be 31 on the row, it must be moved to the next row
<img width="493" alt="image" src="https://github.com/epam/ketcher/assets/139227567/c7a1cdb9-a045-4774-9d98-d8e62f051cd1">

**Screenshots**

https://github.com/epam/ketcher/assets/139227567/94f4e94e-8a01-45ab-902d-436e03635cc9

**Desktop (please complete the following information):**
 - OS: Windows 10 Pro (version 22H2) x64
 - Browser: Сhrome
 - Version: (124.0.6367.60) x64

**Ketcher version** 2.21.0-rc.1, Build at 2024-04-15; 20:06:01

Issue found while testing https://github.com/epam/ketcher/issues/3916

Zhirnoff:
The issue has been moved from 2.21.0-rc.2 to 2.22.0-rc.1 after discussion with management and in accordance with the testing flow process.

ViktoriaTkacheva:
Verified.

Indigo Toolkit Version 1.21.0-dev.4.0-gb15b3f662-wasm32-wasm-clang-12.0.0
Ketcher Version 2.21.0-rc.2, Build at 2024-05-30; 08:32:11
Сhrome Version: (125.0.6422.113) x64
Windows 10 Pro (version 22H2) x64


https://github.com/epam/ketcher/assets/139227567/411f2603-3e3b-40bb-9aec-23219b5f2fe6

[...]

PR #4664, opened 2024-05-20
#4509 - In the Text-editing mode, when inserting a fragment after pressing Ctrl+V twice, the cursor is moved to another row

Guch1g0v:
## Overview
This pull request addresses the issue where, in Text-editing mode, the cursor moves to another row after inserting a fragment by pressing Ctrl+V twice.

## Solution
The problem has been resolved by implementing a check for the presence of phosphate in the current sequence and the inserted sequence. 

## Testing
Manual testing has been performed to validate the fix in various scenarios within the Text-editing mode.

## Impact
This fix improves the user experience by preventing the cursor from inadvertently moving to another row after inserting a fragment. It enhances the overall usability and reliability of the Text-editing mode.

## References
[In the Text-editing mode, when inserting a fragment after pressing Ctrl+V twice, the cursor is moved to another row #4509](https://github.com/epam/ketcher/issues/4509)

## Check list
- [ ] unit-tests written
- [ ]  e2e-tests written
- [ ]  documentation updated
- [x]  PR name follows the pattern `#1234 – issue name`
- [x]  branch name doesn't contain '#'
- [x]  PR is linked with the issue
- [x]  base branch (master or release/xx) is correct
- [x]  task status changed to "Code review"
- [x]  reviewers are notified about the pull request

### DIFF
--- a/packages/ketcher-core/src/application/editor/modes/SequenceMode.ts
+++ b/packages/ketcher-core/src/application/editor/modes/SequenceMode.ts
@@ -838,23 +838,14 @@ export class SequenceMode extends BaseMode {
 
     const currentSequence = SequenceRenderer.currentChain;
 
-    const currentSequenceHasPhosphate = currentSequence?.lastNonEmptyNode
-      ? this.hasPhosphateAtEndOfSequence(
-          currentSequence.lastNonEmptyNode.monomer,
-        )
-      : false;
-
-    const newSequenceHasPhosphate = chainsCollection.lastNode
-      ? this.hasPhosphateAtEndOfSequence(chainsCollection.lastNode.monomer)
-      : false;
+    const currentSequenceHasPhosphate =
+      currentSequence.lastNonEmptyNode?.monomer?.monomerItem?.props?.Name ===
+      'Phosphate';
 
     let nextCaretPosition =
       SequenceRenderer.caretPosition + chainsCollection.length;
 
-    if (
-      (currentSequenceHasPhosphate && newSequenceHasPhosphate) ||
-      currentSequenceHasPhosphate
-    ) {
+    if (currentSequenceHasPhosphate) {
       nextCaretPosition -= 1;
     }
 
@@ -870,14 +861,6 @@ export class SequenceMode extends BaseMode {
     );
 
     return modelChanges;
-  }
-
-  private hasPhosphateAtEndOfSequence(monomer: BaseMonomer): boolean {
-    if (!monomer) return false;
-    if (!monomer.monomerItem || !monomer.monomerItem.props) {
-      return false;
-    }
-    return monomer.monomerItem.props.Name === 'Phosphate';
   }
 
   private insertNewSequenceItem(editor: CoreEditor, enteredSymbol: string) {
--- a/packages/ketcher-core/src/application/editor/modes/SequenceMode.ts
+++ b/packages/ketcher-core/src/application/editor/modes/SequenceMode.ts
@@ -838,7 +838,7 @@ export class SequenceMode extends BaseMode {
 
     const currentSequence = SequenceRenderer.currentChain;
 
-    const currentSequenceHasPhosphate = currentSequence.lastNonEmptyNode
+    const currentSequenceHasPhosphate = currentSequence?.lastNonEmptyNode
       ? this.hasPhosphateAtEndOfSequence(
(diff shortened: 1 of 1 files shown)


packages/ketcher-core/src/application/editor/modes/SequenceMode.ts
```diff
@@ -835,20 +835,51 @@ export class SequenceMode extends BaseMode {
...
         SequenceRenderer.caretPosition,
-        SequenceRenderer.caretPosition + chainsCollection.length,
+        nextCaretPosition,
       ),
     );
 
     return modelChanges;
   }
 
+  private hasPhosphateAtEndOfSequence(monomer: BaseMonomer): boolean {
```
svvald, line 875: Could you please elaborate why exactly this fix works? For me it seems that you're targeting some edge case here rather than addressing the general issue, but I may be wrong

packages/ketcher-core/src/application/editor/modes/SequenceMode.ts
```diff
@@ -835,20 +835,51 @@ export class SequenceMode extends BaseMode {
...
+    if (!monomer) return false;
+    if (!monomer.monomerItem || !monomer.monomerItem.props) {
+      return false;
+    }
+    return monomer.monomerItem.props.Name === 'Phosphate';
```
svvald, line 880: It can be simplified by using conditional chaining: `return monomer?.monomerItem?.props?.Name === 'Phosphate'`, you won't need those conditions above

packages/ketcher-core/src/application/editor/modes/SequenceMode.ts
```diff
@@ -835,20 +835,51 @@ export class SequenceMode extends BaseMode {
...
+
+    const currentSequenceHasPhosphate = currentSequence?.lastNonEmptyNode
+      ? this.hasPhosphateAtEndOfSequence(
+          currentSequence.lastNonEmptyNode.monomer,
+        )
+      : false;
+
+    const newSequenceHasPhosphate = chainsCollection.lastNode
+      ? this.hasPhosphateAtEndOfSequence(chainsCollection.lastNode.monomer)
+      : false;
```
svvald, line 849: It seems that you may omit these ternaries here by passing monomer directly via conditional chaining, even if it's `undefined`/`null`, `hasPhosphateAtEndOfSequence` will return you false, if you apply conditional chaining there as well

packages/ketcher-core/src/application/editor/modes/SequenceMode.ts
```diff
@@ -835,20 +835,51 @@ export class SequenceMode extends BaseMode {
...
+
+    let nextCaretPosition =
+      SequenceRenderer.caretPosition + chainsCollection.length;
+
+    if (
+      (currentSequenceHasPhosphate && newSequenceHasPhosphate) ||
+      currentSequenceHasPhosphate
+    ) {
```
svvald, line 857: This condition seems to be either overcomplicated or incorrect for me, it's always `true` if `currentSequenceHasPhosphate` is and vice verse

packages/ketcher-core/src/application/editor/modes/SequenceMode.ts
Guch1g0v: The bug was due to incorrect calculation of the caret.


We solve, in fact, a particular problem of the position of the caret, if at the end of the sequence there is a Phosphate residue (p), then when inserting another sequence, this residual is transferred to the end of the resulting sequence. 


And we need to take into account several variants of events, whether there is a “p” on an existing one or on two simultaneous sequences. 

If the condition is true, we decrement the caret position by 1. Since "p" does not "add up". 

e.g. `AAAp + CCCp = AAACCCp (not AAACCCpp)`

But if “p” is only on the new sequence, then we do not change the position of the caret.

Therefore, we use the `hasPhosphateAtEndOfSequence` method to check at the end of the sequence. =)

packages/ketcher-core/src/application/editor/modes/SequenceMode.ts
Guch1g0v: I agree, if you simplify the code here, you don't have to use `hasPhosphateAtEndOfSequence`, thanks!

svvald (approved):

NataliaLoginova (approved):

Issue #4575, opened 2024-05-06
New Sequences are created while we add cyclic to middle of sequence.


NadezhdaPeskun:
**Steps to Reproduce**

1. Switch to Macro 
2. go to Sequence mode 
3. Open File 
[ketcher (12).zip](https://github.com/epam/ketcher/files/15221693/ketcher.12.zip)
![image](https://github.com/epam/ketcher/assets/161723514/13873efa-6317-4558-9f31-f3de57be4aa2)
4. Copy cyclic TTTTT >> 
5. Context menu >> Edit
6. Paste to midle of AAAA

![image](https://github.com/epam/ketcher/assets/161723514/2e81ebab-2d8c-438e-a271-52cae3019735)

**Actual behavior**
New Sequences are created
<img width="179" alt="image" src="https://github.com/epam/ketcher/assets/161723514/f38752ed-521b-48f3-b032-f3b878100262">

**Expected behavior**

**Versions:**
[Version 2.21.0-rc.1](https://lifescience.opensource.epam.com/ketcher/index.html)
    Build at 2024-04-29; 09:29:15

Indigo Toolkit
    Version 1.20.0-rc.1.0-g8e8ffc3c3-x86_64-linux-gnu-11.2.1

NadezhdaPeskun:
![image](https://github.com/epam/ketcher/assets/161723514/e04109df-e7a5-480a-9d0b-4d1bd7ccb63c)
Would be nice for future to have abbility

NadezhdaPeskun:
Should be verified with next Issues: (4708, 4575, 4506,4706) + 4328
Error Message should appear:
Error Message:
It is impossible to merge fragments. Attachment point to establish bonds are not available.

AlexeyGirin:
Verified.
![image](https://github.com/epam/ketcher/assets/26869421/ec0c2702-a8b7-422f-bc5b-f01640771873)


 - Indigo Toolkit Version 1.21.0-rc.1.0-g9194599b2-wasm32-wasm-clang-19.0.0
 - Ketcher Version 2.22.0-rc.2 Build at 2024-06-03; 18:43:00
 - Chrome Version 125.0.6422.142 (Official Build) (64-bit)
 - Windows 10

Zhirnoff:
Automated.

PR #4703, opened 2024-05-28
#4575 - New Sequences are created while we add cyclic to middle of sequence. 

Guch1g0v:
## Check list
- [ ] unit-tests written
- [ ] e2e-tests written
- [ ] documentation updated


packages/ketcher-core/src/application/editor/modes/SequenceMode.ts
```diff
@@ -883,6 +941,14 @@ export class SequenceMode extends BaseMode {
     return modelChanges;
   }
 
+  private showMergeWarningModal(editor: CoreEditor) {
```
svvald, line 944: I think it's rather an error than a warning message (which I covered by a pop-up in the bottom of the editor but doesn't block it)

packages/ketcher-core/src/application/editor/modes/SequenceMode.ts
```diff
@@ -821,6 +821,48 @@ export class SequenceMode extends BaseMode {
...
+    const isPasteInStart = !previousNodeInSameChain;
+    if (isPasteInEnd && !previousNodeInSameChain) return true;
+    if (isPasteInEnd) {
+      return (
+        this.isR1Free(firstNodeOfNewFragment) &&
+        this.isR2Free(previousNodeInSameChain)
+      );
+    }
+    if (isPasteInStart) {
+      return this.isR2Free(lastNodeOfNewFragment) && this.isR1Free(currentNode);
```
svvald, line 861: If you need to check both R1 and R2 when pasting into start or end of sequence, you may probably use `areR1R2Free` here rather than separate functions

svvald (approved):
Some minor comments, no critical, can be addressed later to avoid blocking into upcoming release. Overall looks good to me

Issue #4530, opened 2024-04-24
Add ability to define attachment points for molecules


rrodionov91:
**Background**
Currently there is no way to define attachment points for molecules, so it is not possible to connect them in macromolecules mode with monomer.

**Solution**
New items should be added to atom context menu: Add attachment point, Remove attachment point. It should appear only for atoms which are NOT part of s-group. 

When add attachment point, system should create new superatom s-group for whole structure implicitly without label. System should draw label(R1/.../R8) depends on next free attachment point number and single bond from atom to that label. If all 8 attachment points exist in structure system should NOT create attachment point.

System should NOT allow to show brackets and collapse superatoms without labels.

When remove attachment point system should delete label and corresponding bond. When last attachment point deleted system should delete s-group superatom.

Such attachment points should be converted into monomer attachment points during the switching to macromolecules mode and vice versa.

System should save such attachment points to sgroup.attachmentPoints field during the serialization to ket format and should set leavingGroup new atom H.

**Extra info:**
1. What happens if I apply R-Group on AP label?
        No, since AP is not an atom - its a label connected to atom by bond (we can do that with atom AP label connected to)
2. Can I wrap atom/molecule with AP to R-Group fragment tool?
        No, since AP is not an atom - its a label connected to atom by bond (we can do that with atom AP label connected to)
3. Can I attach "old" AP to new AP?
        No, since AP is not an atom - its a label connected to atom by bond (we can do that with atom AP label connected to)
4. Can I charge (A+/A-) AP?
        No, since AP is not an atom - its a label connected to atom by bond (we can do that with atom AP label connected to)
5. Can I bond AP with bond? (i.e. attach bond to AP)
        No, since AP is not an atom - its a label connected to atom by bond (we can do that with atom AP label connected to)
6. Erase logic (since system enumerate AP automatically). What if we have R1, R2, R3 and I delete R2 and add NEW AP? What if we have R1-R8 in place and I delete R2 and try to add new AP?
         Gaps should start be filled if we rich the R8 (later this will be changed to let user choose AP name himself)
7. Selection logic? Partial selection? Should selection be saved in KET?
         Selection should work the same as for atoms but without saving selection state in KET file
         _Extra work over selection login is needed here. Will be done separatedy_

8. Items in context menu? Context menu if selected together with bond and atom?
      Only Delete item should be remain in the context menu for now
9. Open from IMAGE?
     Load from image is not supported
10. Should we support save to: 
     We support all formats that support s-groups and attachment points
[...]

PR #4727, opened 2024-06-01
#4530 - Add ability to define attachment points for molecules fixes

rrodionov91:
- fixed atomId in molecule-monomer connections in serialization
- turned off structure rescaling during opening from file if there is monomer-molecule connection

## Check list
- [ ] unit-tests written
- [ ] e2e-tests written
- [ ] documentation updated


packages/ketcher-react/src/script/ui/state/shared.ts
```diff
@@ -149,6 +150,14 @@ export function load(struct: Struct, options?) {
       const hasUnsupportedGroups = parsedStruct.sgroups.some(
         (sGroup) => !supportedSGroupTypes[sGroup.type],
       );
+      const hasMoleculeToMonomerConnections = parsedStruct.bonds.find(
```
svvald, line 153: It's minor, but it seems that `.some` might fit better semantically here rather than `.find`Mi as you return some flag of whether any bond meets your condition, not the bond itself

svvald (approved):

Issue #4833, opened 2024-06-12
Support rendering multiple ketcher editors on a page without an iframe


ihuang8:
**Background**
Add the ability to render multiple ketcher editors on a page without an iframe. The iframe makes it more difficult to access the ketcher instance (e.g. https://github.com/epam/ketcher/issues/3693).

**Solution**

Instead of binding the ketcher instance to the window of an iframe, each instance could be stored in the React state instead.

AlexeyGirin:
Impossible to verify since it require non-trivial development of Demo app
Will be done in the context of https://github.com/epam/ketcher/issues/4145

PR #4834, opened 2024-06-12
#4833 - Support rendering multiple ketcher editors on a page without an iframe

ihuang8:

Add a ketcherId so that multiple ketcher instances can be added to the same page without using an iframe. Instead of binding each ketcher instance to the iframe, they can now be stored in the React state and easily accessed to obtain the data. 

## Check list
- [ ] unit-tests written
- [ ] e2e-tests written
- [ ] documentation updated

### DIFF
--- a/packages/ketcher-react/src/constants.ts
+++ b/packages/ketcher-react/src/constants.ts
@@ -14,8 +14,8 @@
  * limitations under the License.
  ***************************************************************************/
 
-export const ketcherInitEventName = (ketcherId: string) =>
-  `ketcher-init-${ketcherId}`;
+export const ketcherInitEventName = (ketcherId?: string) =>
+  ketcherId ? `ketcher-init-${ketcherId}` : 'ketcher-init';
 
 export const MODES = {
   FG: 'fg',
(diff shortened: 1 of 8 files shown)


packages/ketcher-react/src/constants.ts
```diff
@@ -14,7 +14,8 @@
  * limitations under the License.
  ***************************************************************************/
 
-export const KETCHER_INIT_EVENT_NAME = 'ketcher-init';
+export const ketcherInitEventName = (ketcherId: string) =>
+  `ketcher-init-${ketcherId}`;
```
rrodionov91, line 18: Could you please make it backward-compatible by using old 'ketcher-init' id by default?

rrodionov91 (changes requested):

ihuang8:
@rrodionov91 Is there anything else you need from me here?

rrodionov91 (approved):

rrodionov91:
Hi @ihuang8 
Thank you for fixing this issue.
Really appreciate it.

Issue #4837, opened 2024-06-13
Library filter does not reset when switching between Micro and Macro mode


Zhirnoff:
**Steps to Reproduce**
1. Enter a search term in the library's search field
2. Switch to micro mode
3. Switch back to macro mode

**Actual behavior**
The search field is empty, but the library remains filtered according to search term entered before switching modes.

**Expected behavior**
The search filter should reset when switching modes

**Screenshots**

![2024-06-13_16h46_22](https://github.com/epam/ketcher/assets/62464241/f490a88c-9be9-4538-9906-cbd52288dd0b)

**Desktop:**
 - OS: Windows 11
 - Browser Chrome
 - Version 124.0.6367.79 (Official Build) (64-bit)

**Ketcher version** 
[Version 2.22.0-rc.4]
**Indigo version** 
[Version 1.21.0-rc.1]

Zhirnoff:
Tested. Bug fixed.
**Desktop:**
 - OS: Windows 11
 - Browser Chrome
 - Version 124.0.6367.79 (Official Build) (64-bit)

**Ketcher version** 
[Version 2.23.0-rc.1]
**Indigo version** 
[Version 1.22.0-rc.1]

PR #4859, opened 2024-06-18
#4837 - Library filter does not reset when switching between Micro and Macro mode

Guch1g0v:
Adding logic to track the first rendering of a component

## Check list
- [ ] unit-tests written
- [ ] e2e-tests written
- [ ] documentation updated

### DIFF
--- a/packages/ketcher-macromolecules/src/components/monomerLibrary/MonomerLibrary.tsx
+++ b/packages/ketcher-macromolecules/src/components/monomerLibrary/MonomerLibrary.tsx
@@ -13,7 +13,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  ***************************************************************************/
-import React, { ChangeEvent, useEffect, useRef, useState } from 'react';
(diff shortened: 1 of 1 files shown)


packages/ketcher-macromolecules/src/components/monomerLibrary/MonomerLibrary.tsx
```diff
@@ -51,6 +51,14 @@ const MonomerLibrary = React.memo((props: MonomerLibraryProps) => {
   const isDisabledTabsPanels =
     isSequenceMode && !isSequenceEditInRNABuilderMode;
 
+  const [firstRender, setFirstRender] = useState(true);
+  useEffect(() => {
+    if (firstRender) {
+      setFirstRender(false);
+      dispatch(setSearchFilter(''));
+    }
+  }, [dispatch, firstRender]);
```
svvald, line 60: I think you don't need `firstRender` flag, if you just leave useEffect dependencies array empty, it will fire only once on mount, can you check will it work?

svvald (approved):

PR #4865, opened 2024-06-20
#4763 - It is possible to create forbidden RNA preset

### LINKED ISSUE #4763: It is possible to create forbidden RNA preset
**Steps to Reproduce**
1. Open RNA builder
2. Click New Preset button
3. Select Phosphate (P in my case)
4. Select Base (A in my case)
5. Select 3SS6 sugar
image
6. Press Add to Presets button

**Actual behavior**
Preset is created
image

**Expected behavior**
Corresponding error should be shown in a pop-up error box and 'Save'/'Add to presets' button should be disabled
Since 3SS6 sugar don't have R3 attachment point RNA preset is forbidden due to requirements:

> Sugar without R3 connection point cannot make this nucleotide
> If user validation rule is violated corresponding error should be shown in a pop-up error box and 'Save'/'Add to presets' button
> should be disabled

**Versions:**
 - Indigo Toolkit Version 1.21.0-rc.1.0-g9194599b2-wasm32-wasm-clang-19.0.0
 - Ketcher Version 2.22.0-rc.2 Build at 2024-06-03; 18:43:00
 - Chrome Version 125.0.6422.142 (Official Build) (64-bit)
 - Windows 10

Issue found while testing https://github.com/epam/ketcher/issues/3816

vitaepam:
Added validations to Accordion action

Closes #4763 

## Check list
- [x] unit-tests written
- [x] e2e-tests written
- [ ] documentation updated


packages/ketcher-macromolecules/src/components/monomerLibrary/RnaBuilder/RnaAccordion/RnaAccordion.tsx
```diff
@@ -91,11 +94,48 @@ export const RnaAccordion = ({ libraryName, duplicatePreset, editPreset }) => {
...
+    setNewPreset(activePreset);
+  }, [activePreset]);
+
+  const dispatch = useDispatch();
   const handleAccordionSummaryClick = (rnaBuilderItem: RnaBuilderItem) => {
     if (expandedAccordion === rnaBuilderItem) {
       setExpandedAccordion(null);
-    } else setExpandedAccordion(rnaBuilderItem);
+    } else {
+      dispatch(setActiveRnaBuilderItem(rnaBuilderItem));
```
svvald, line 108: Is this logic needed here? I mean it doesn't look so straightforward to call it on accordion entry click for me

svvald, line 108: Also I'm not quite sure about how to improve reusability here if it's really needed to use this logic, doing so React way would probably cause to restructure components composition to handle common logic
I'd propose to introduce some util function to encapsulate this logic there and call it here and in another component

packages/ketcher-macromolecules/src/components/monomerLibrary/RnaBuilder/RnaAccordion/RnaAccordion.tsx
```diff
@@ -91,11 +94,48 @@ export const RnaAccordion = ({ libraryName, duplicatePreset, editPreset }) => {
 
   const [expandedAccordion, setExpandedAccordion] =
     useState<RnaBuilderItem | null>(activeRnaBuilderItem);
+  const [newPreset, setNewPreset] = useState(activePreset);
 
+  useEffect(() => {
```
svvald, line 99: Is this useEffect needed here? There is the same in 
`RnaEditorExpanded.tsx` component, isn't it enough to cover this case? If not, I'd suggest to move this effect up components tree, put it to some common parent component, probably, `RnaBuilder.tsx`

packages/ketcher-macromolecules/src/components/monomerLibrary/RnaBuilder/RnaAccordion/RnaAccordion.tsx
vitaepam: Checked, we may remove this, it won't affect validations

vitaepam: Moved this to helper functions section

packages/ketcher-macromolecules/src/components/monomerLibrary/RnaBuilder/RnaAccordion/RnaAccordion.tsx
vitaepam: Yeap, you right it's not needed here, removed

svvald (approved):

Issue #4802, opened 2024-06-06
Copy/Past operation works wrong


AlexeyGirin:
**Steps to Reproduce**
1. Load from file: [Cut operation for attachment points makes canvas inaccessible and delete entire molecule2.zip](https://github.com/user-attachments/files/15635923/Cut.operation.for.attachment.points.makes.canvas.inaccessible.and.delete.entire.molecule2.zip)
OR manually create following configuration on the canvas: 
![image](https://github.com/epam/ketcher/assets/26869421/636af8db-d050-44bd-b820-095988f97a5c)
2. Select few atoms and at least one attachment point
![image](https://github.com/epam/ketcher/assets/26869421/11c1fdb5-c5fa-42e8-a4a7-6e9f7d03c907)
3. Press CTRL+C key
4. Press CTRL+V key

**Actual behavior**
System past selected elements with attachment points replaced with hydrogens
![image](https://github.com/epam/ketcher/assets/26869421/9d8652cf-0ecc-4d08-91d9-ebccfe084efc)
Canvas become inaccessible and inoperational (try to select the rest molecule)

**Expected behavior**
System pasts selected elements correct
![image](https://github.com/epam/ketcher/assets/26869421/531413c5-64ad-476a-9c3d-ca5afde90928)

**Versions:**
 - Indigo Toolkit Version 1.21.0-rc.1.0-g9194599b2-wasm32-wasm-clang-19.0.0
 - Ketcher Version 2.22.0-rc.2 Build at 2024-06-03; 18:43:00
 - Chrome Version 125.0.6422.142 (Official Build) (64-bit)
 - Windows 10

Issue found while testing https://github.com/epam/ketcher/issues/4530

iRadionova:
Can not reproduce: CTRL+V doesn't work

https://github.com/user-attachments/assets/7427d68c-32be-46db-ad06-172d4dffc25c

<img width="537" height="340" alt="Image" src="https://github.com/user-attachments/assets/6662d430-dd83-429e-bfb2-2ba7301c7715" />

PR #4910, opened 2024-06-27
#4802 - Copy/Past operation works wrong

Guch1g0v:

## Check list
- [ ] unit-tests written
- [ ] e2e-tests written
- [ ] documentation updated

### DIFF
--- a/packages/ketcher-core/src/domain/entities/struct.ts
+++ b/packages/ketcher-core/src/domain/entities/struct.ts
@@ -288,14 +288,15 @@ export class Struct {
 
     const sgroupIdMap = {};
     this.sgroups.forEach((sg, sgroupId) => {
-      if (sg.atoms.some((aid) => !atomSet!.has(aid))) return;
+      const notInGroup = sg.atoms.some((aid) => !atomSet!.has(aid));
+      const atomsInSet = sg.atoms.filter((aid) => atomSet!.has(aid));
+      if (!atomsInSet.length && notInGroup) return;
       const oldSgroup = sg;
 
       sg =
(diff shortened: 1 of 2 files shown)


NataliaLoginova (approved):

packages/ketcher-core/src/domain/entities/struct.ts
```diff
@@ -288,14 +288,15 @@ export class Struct {
 
     const sgroupIdMap = {};
     this.sgroups.forEach((sg, sgroupId) => {
-      if (sg.atoms.some((aid) => !atomSet!.has(aid))) return;
+      const notInGroup = sg.atoms.some((aid) => !atomSet!.has(aid));
+      const atomsInSet = sg.atoms.filter((aid) => atomSet!.has(aid));
+      if (!atomsInSet.length && notInGroup) return;
```
svvald, line 293: Maybe I'm missing something here but why do you need `notInGroup`? Conditions in `.some` and `.filter` are mutually exclusive so if there are no elements satisfying the `atomSet.has(aid)` condition, there is definitely at least one element which satisfies the inverse `!atomSet.has(aid)` condition

Guch1g0v: Sometimes (when we copy a attachment point label) we are not in the group, but we need still filter some atoms,
but if we are not in the group and there are no atoms, we exit. Also, we must support the previous realization of merge info. 

Screenshot from 2024-07-01 21-51-11
Screenshot from 2024-07-01 21-49-11

Perhaps there is another way to solve the problem.

Issue #4897, opened 2024-06-26
Insert image to canvas via "Add image" tool, Select, Move and Scale the image


yauhen-karmyzau:
As a **User**, I want to be able **to insert, select, scale and move images** on the canvas so that I can enhance my chemical structure diagrams with relevant visuals and adjust them to fit my needs.

**Acceptance Criteria:**

1. A user should be able to add an image using a dedicated 'Add Image' button located at the side panel with the static tools. The application should support at least the PNG image format and, if feasible, other commonly used image formats like JPEG, GIF, and SVG.

![Image](https://github.com/epam/ketcher/assets/173041434/2397d260-a5ec-4db1-9eda-362cf49b3f8b)

2. An image should be placed on the canvas on the click event under the mouse pointer.
3. A user should be able to select an image by clicking on it with the selection and "Add Image" tools. The selected image should be visibly highlighted or outlined and the Scale functionality must be visible.
4. A user should be able to scale (enlarge or reduce) a selected image using a dedicated 'Scale' option. The application should provide real-time feedback as the user scales an image.
5. A user should be able to move a selected image around the canvas.
6. After the image is inserted, moved or scaled, the user should be able to undo/redo his actions. The new size and position of the image must be saveable to supported formats (KET).
7. If a user tries to upload an unsupported image format, they should be shown an error message. Moving and rotating images are out of scope for this user story.

TBD: Design of the image selection and UI elements for scaling.
 
https://github.com/epam/ketcher/issues/4897#issuecomment-2215065068
> Ketcher is not support file insert from the system clipboard. This functionality may be implemented in scope of separate issue.

yauhen-karmyzau:
Drag-n-Drop is out of scope! New feature need to be created.

vanoprenko:
It doesn't make a lot of sense to deal with SVG as with raster. Let's either exclude it from scope (add with a separate ticket) or support as a vector.

daniil-sloboda:
In case some images are not working due to size\format they are out of scope

yauhen-karmyzau:
Team Agreement:
We can implement the same behaviour as for other "static tools:

1. A User selects Insert image tool, the tool in the sidebar becomes active.
2. A User clicks on the canvas and the system "select a file" window appears.
3. When an image is uploaded , it appears at the place of the click.

yauhen-karmyzau:
Missed AC:

- A User should be able to switch to Macro mode and back.
- Images shouldn’t be visible in Micro.

[...]

PR #5002, opened 2024-07-04
#4897 - Insert image to canvas via "Add image" tool, Select, Move and Scale the image

daniil-sloboda:

## Check list
- [ ] unit-tests written
- [ ] e2e-tests written
- [x] documentation updated


packages/ketcher-react/src/script/editor/tool/rasterImage.ts
```diff
@@ -0,0 +1,85 @@
...
+    const element = document.getElementById(RasterImageTool.INPUT_ID);
+    if (element) {
+      if (element instanceof HTMLInputElement) {
+        return element;
+      }
+      element.remove();
+    }
+    return this.createElement();
+  }
+}
```
rrodionov91, line 85: I would propose simplify a bit
```
  private getElement(): HTMLInputElement {
    const element = document.getElementById(RasterImageTool.INPUT_ID);

    if (element instanceof HTMLInputElement) {
      return element;
    }

    element?.remove();

    return this.createElement();
  }
```

rrodionov91 (changes requested):

packages/ketcher-core/src/domain/serializers/serializers.types.ts
```diff
@@ -13,8 +13,18 @@
...
 
 export interface Serializer<T> {
   deserialize: (content: string) => T;
   serialize: (struct: T) => string;
 }
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export interface KetFileNode<T = any> {
+  type: string;
+  fragment?: Struct;
```
rrodionov91, line 26: I don't see usage of fragment field. And usually fragment field is used for identifying that several atoms/anything else are the part of the same structure. Can we remove this field? Or maybe rename?

packages/ketcher-react/src/script/editor/shared/closest.ts
```diff
@@ -285,9 +310,9 @@ function findClosestEnhancedFlag(restruct, pos, skip, _minDist, options) {
   return ret;
 }
 
-function findClosestDataSGroupData(restruct, pos) {
-  let minDist = null;
-  let ret = null;
+function findClosestDataSGroupData(restruct: ReStruct, pos: Vec2) {
+  let minDist = Number.POSITIVE_INFINITY;
```
rrodionov91, line 314: Why not to keep null here? I'm not sure that it will not affect some functionality

packages/ketcher-react/src/script/editor/shared/closest.ts
```diff
@@ -386,7 +420,8 @@ function findClosestRgroupAttachmentPoints(
   let closestItemId;
 
   restruct.visibleRGroupAttachmentPoints.forEach((reItem, id) => {
-    const itemOutlinePoints = reItem.getOutlinePoints();
+    // Remove readonly from getOutlinePoints return type
+    const itemOutlinePoints = [...reItem.getOutlinePoints()];
```
rrodionov91, line 424: Why do we need this?

packages/ketcher-react/src/script/editor/shared/closest.ts
```diff
@@ -406,10 +441,10 @@ function findClosestRgroupAttachmentPoints(
       };
 }
 
-function findClosestRxnArrow(restruct, pos) {
-  let minDist = null;
-  let refPoint = null;
-  let ret = null;
+function findClosestRxnArrow(restruct: ReStruct, pos: Vec2) {
+  let minDist = Number.POSITIVE_INFINITY;
```
rrodionov91, line 445: Also here, can it potentially affect some existing functionality?

packages/ketcher-react/src/script/editor/tool/helper/locate.ts
```diff
@@ -279,7 +317,8 @@ function getElementsInPolygon(restruct, rr) {
   });
 
   const enhancedFlagList: Array<number> = [];
-  restruct.enhancedFlags.forEach((item, id) => {
+  // ReEnhancedFlag doesn't have pp item according to types but somehow it works
```
rrodionov91, line 320: :D

rrodionov91 (dismissed):

packages/ketcher-core/src/domain/serializers/serializers.types.ts
daniil-sloboda: We cannot, it is being used for bonds\atoms

packages/ketcher-react/src/script/editor/shared/closest.ts
daniil-sloboda: Because with types we need to additionally check whether this field is null. We do not return minDist across this file so this is safe to update

packages/ketcher-react/src/script/editor/shared/closest.ts
daniil-sloboda: We return readonly array but the function expects just an array, we need to copy it.
I can update the comment if it's not clear

packages/ketcher-react/src/script/editor/shared/closest.ts
daniil-sloboda: Same as above

Zhirnoff (approved):

Zhirnoff:
All screenshots are valid. Changes mainly due to the addition of a new button for adding Images

svvald (approved):

PR #5188, opened 2024-07-29
Preview: Update Image object in KET file to support CDX/CDXML format

daniil-sloboda:

## Check list
- [ ] unit-tests written
- [ ] e2e-tests written
- [ ] documentation updated


package-lock.json
```diff
@@ -34880,7 +34880,7 @@
       "license": "Apache-2.0",
       "dependencies": {
         "@babel/runtime": "^7.17.9",
-        "indigo-ketcher": "1.22.0-rc.3",
+        "indigo-ketcher": "^1.23.0-dev.4",
```
rrodionov91, line 34883: Let's keep version without "^" to have consistent version of indigo for all developers.

packages/ketcher-core/src/domain/entities/image.ts
```diff
@@ -15,16 +15,22 @@
...
-  bitmap: string;
-  halfSize: Point;
+import { IMAGE_SERIALIZE_KEY } from 'domain/constants';
+
+export interface KetFileImageNode extends KetFileNode<string> {
+  format: string;
+  boundingBox: {
+    x: number;
+    y: number;
+    z?: number;
```
rrodionov91, line 27: Do we need z coordinate?

packages/ketcher-core/src/domain/serializers/ket/schema.json
```diff
@@ -798,38 +781,43 @@
...
+              "exclusiveMinimum": 0
+            },
+            "x": {
+              "type": "number"
+            },
+            "y": {
+              "type": "number"
+            },
+            "z": {
+              "type": "number"
```
rrodionov91, line 814: Do we need z coordinate?

rrodionov91 (changes requested):

packages/ketcher-core/src/domain/entities/image.ts
daniil-sloboda: We have it by default across the app, whenever we are using Vec2

packages/ketcher-core/src/domain/serializers/ket/schema.json
daniil-sloboda: I've checked schema.json file - a lot of entities are using it.
I suppose we can omit it but all the code across the deserialization logic  have it.
We do not require it since it's not in the array of required fields, but if it's present it should be number - not string or object

rrodionov91 (approved):

Issue #4658, opened 2024-05-20
Some bonds remain selected on export to SVG in Snake mode


AlexeyGirin:
**Steps to Reproduce**
1. Switch to Macro
2. Load from file: [Some bonds remain selected.zip](https://github.com/epam/ketcher/files/15375542/Some.bonds.remain.selected.zip)
3. Switch to Flex mode
4. Select all elements on the canvas (press CTRL+A)
5. Export canvas to SVG

**Actual behavior**
Some bonds remain selected:
![image](https://github.com/epam/ketcher/assets/26869421/66f1eb1f-9ebe-44f1-bf98-c22851f70512)

**Expected behavior**
All selection is removed
![image](https://github.com/epam/ketcher/assets/26869421/697dd01b-3f94-4c79-bb24-c246d6b5b46e)

As per requirements:

> Temporary objects(monomer preview near cursor, css cursor, **selectors** and e.t.c) should not be included into image.

**Versions** [e.g. v2.4.2]. 
 - Indigo Toolkit Version 1.21.0-dev.3.0-g6f55c664-wasm32-wasm-clang-12.0.0
 - Ketcher Version 2.21.0-rc.2 Build at 2024-05-20; 05:30:45
 - Chrome Version 124.0.6367.201 (Official Build) (64-bit)
 - Windows 10

Issue found while testing https://github.com/epam/ketcher/issues/4539

KristinaKasyanovskaya:
> Not fixed.
> Bonds remain slightly selected

Fix https://github.com/epam/ketcher/pull/5219

Zhirnoff:
Tested. Bug fixed.

![2024-09-05_11h26_12](https://github.com/user-attachments/assets/e10f72b0-2d99-449a-9715-185b81b0c459)
![2024-09-05_11h26_51](https://github.com/user-attachments/assets/6e089347-98dc-464a-90f5-c2db40d5e855)

**Desktop:**
 - OS: Windows 11
 - Browser Chrome
 - Version 128.0.6613.120 (Official Build) (64-bit)

**Ketcher version** 
[Version 2.25.0-rc.1]
**Indigo version** 
[Version 1.24.0-rc.1]

PR #5219, opened 2024-08-06
 #4658-Some bonds remain selected on export to SVG in Snake mode

KristinaKasyanovskaya:
Issue => https://github.com/epam/ketcher/issues/4658

- Past behavior

- [ ] Not fixed.
Bonds remain slightly selected

- Actual behavior 

- [x] Fixed 
Fully Selected Bonds
Снимок экрана 2024-08-06 в 12 03 44

Снимок экрана 2024-08-06 в 12 03 56

## Check list
- [ ] unit-tests written
- [ ] e2e-tests written
- [ ] documentation updated


NataliaLoginova (approved):

packages/ketcher-core/src/utilities/getSvgFromDrawnStructures.tsx
```diff
@@ -5,6 +5,30 @@ const ADDITIONAL_TOP_MARGIN = 54;
 const ADDITIONAL_LEFT_MARGIN = 50;
 const DEFAULT_MARGIN = 10;
 
+const removeHoverEffects = (wrapper: SVGSVGElement): void => {
```
svvald, line 8: This logic is too complicated and too specific, I think you can remove it completely and just add `dynamic-element` class to snake bonds selection, which will be automatically covered by the existing logic

packages/ketcher-core/src/application/render/renderers/PolymerBondRenderer.ts
```diff
@@ -728,7 +728,8 @@ export class PolymerBondRenderer extends BaseRenderer {
           .attr('stroke', '#57FF8F')
           .attr('stroke-width', 2)
           .attr('fill-opacity', 0)
-          .attr('d', this.path);
+          .attr('d', this.path)
+          .attr('class', 'snake-bond-selection');
```
svvald, line 732: You can just replace this line with `.attr('class', 'dynamic-element')` and I think it will resolve the issue and make the logic easier

svvald (changes requested):

packages/ketcher-core/src/application/render/renderers/PolymerBondRenderer/SnakeModePolymerBondRenderer.ts
```diff
@@ -130,7 +130,7 @@ export class SnakeModePolymerBondRenderer extends BaseRenderer {
       .append('path')
       .attr('stroke', this.polymerBond.finished ? '#333333' : '#0097A8')
       .attr('stroke-width', 1)
-      .attr('class', 'selection-area')
+      .attr('class', 'selection-area snake-bond')
```
svvald, line 133: Perhaps this change is not needed then too?

KristinaKasyanovskaya: Fix

svvald (approved):

Issue #4853, opened 2024-06-17
Search monomer in library by IDT


olganaz:
**Background**
Since the IDT alias is displayed only by hover, it is almost impossible to find the desired monomer in the library.

**Requirements**

1. User should be able to search by IDT aliases in a library using search bar. 

2. User should be able to search by a part of IDT aliases.
> For example: `iPr` should return a monomer with alias `i6diPr`, and so should `6diP`

3. Single symbol `/` in a query should result in displaying all the monomers that have IDT alias.  

4. The symbol after `/` is the first symbol of the IDT alias.
> In example: /5deoxyU/, `5` is the first symbol of the alias.
In example: /i6diPr/, `i` is the first symbol of the alias.
Searching `/5deo` should return a monomer with alias `5deoxyU`, but searching `/deo/` should not.

5. The symbol before the second `/` is the last symbol of the IDT alias.
> In example: /5deoxyU/, `U` is the last symbol of the alias.
In example: /i6diPr/, `r` is the last symbol of the alias.
Searching `diPr/` should return a monomer with alias `i6diPr`, but searching diP/` should not.

6. If there is any symbol after the second slash, there shouldn't be any results.
> Searching `/5deoxyU/` should return a monomer with alias `5deoxyU`, but searching `/5deoxyU//` should return nothing.
> Searching `/i6diPr/` should return a monomer with alias `i6diPr`, but searching `/i6diPr/5deo` should return nothing.

7. The search is not case sensitive.
> Both `/5deo` and `/5DEO` should return the monomer with alias `5deoxyU`.

AlexeyGirin:
Verified.
**Versions**
- Ketcher Version 2.25.0-rc.1 Build at 2024-09-03; 15:31:03
- Indigo Toolkit Version 1.24.0-rc.1.0-gb8f099851-wasm32-wasm-clang-19.0.0
- Chrome Version 128.0.6613.120 (Official Build) (64-bit)
- Win10

PR #5373, opened 2024-08-27
#4853-Search monomer in library by IDT

KristinaKasyanovskaya:
Issue => https://github.com/epam/ketcher/issues/4853

Updates:

- [x] User be able to search by IDT aliases in a library using search bar.

- [x] Single symbol / in a query result in displaying all the monomers that have IDT alias.

- [x] The symbol after / is the first symbol of the IDT alias.

- [x] The symbol before the second / is the last symbol of the IDT alias.

- [x] Searching diPr/ should return a monomer with alias i6diPr, but searching diP/` should not.

- [x] If there is any symbol after the second slash, there shouldn't be any results.

- [x] The search is not case sensitive.

## Check list
- [ ] unit-tests written
- [ ] e2e-tests written
- [ ] documentation updated


rrodionov91 (approved):
Comments will be fixed in future PR

Issue #5322, opened 2024-08-20
Autotests: Convert IDT loading saving tests to clear convas execution


split-ends:
**Source task(s):**
 - https://github.com/epam/ketcher/issues/4495

PR #5381, opened 2024-08-28
#5322 - Refactor IDT import and save tests for optimized execution

split-ends:
## Overview
This pull request refines and optimizes the test suite for IDT import and saving by clearing the canvas after each test without reloading the page in accordance with [this task](https://github.com/epam/ketcher/issues/5322).

## Key changes

- Optimized tests to decrease execution time.
- Reworked `selectClearCanvasTool` to automatically close all interfering popups before clicking the «Clear Canvas» button, making it faster to write tests with this approach in the future.
- Added `resetZoomLevelToDefault` and `chooseTab` functions to encapsulate its logic.
- Implemented `markResetToDefaultState` and `processResetToDefaultState` - these allow for broader and more flexible resetting of the test environment post-execution.

The modifications have resulted in a reduction in the maximum execution time on a single worker — down to approximately 70 seconds (on my PC) from the previous ~170 seconds using the older page reload method. For tests where it is still necessary to reload the page, clear comments have been added to explain the rationale.

### Reset Logic Enhancements

The logic for resetting states to default has been moved to `afterEach` so that if a test fails, clearing is performed in any case:
- All active popups (combo boxes, various forms/windows) are closed using the Escape key until the «Clear Canvas» button can be clicked.
- Based on predefined annotations for specific tests, appropriate actions are performed to return the system to its default state, minimizing interference with subsequent tests.

In the future, this will allow tests to be quickly translated by clearing the canvas without reloading the page by extending `resetStateTypes`. 

## Check list
- [ ] unit-tests written
- [ ] e2e-tests written
- [ ] documentation updated

### DIFF
--- a/ketcher-autotests/tests/Macromolecule-editor/Import-Saving-Files/import-saving-idt.spec.ts
+++ b/ketcher-autotests/tests/Macromolecule-editor/Import-Saving-Files/import-saving-idt.spec.ts
@@ -97,7 +97,7 @@ test.beforeAll(async ({ browser }) => {
   await turnOnMacromoleculesEditor(page);
 });
 
-test.afterEach(async (_, testInfo) => {
+test.afterEach(async ({ context: _ }, testInfo) => {
   await selectClearCanvasTool(page);
   await resetZoomLevelToDefault(page);
   await processResetToDefaultState(testInfo, page);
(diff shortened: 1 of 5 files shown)


AlexeyGirin (approved):
Approved

Issue #5709, opened 2024-10-03
Unable to load variant sugar from HELM - system throws an error: Convert error! {}


AlexeyGirin:
**Steps to Reproduce**
1. Toggle to Macro - Flex
2. Load using paste from clipboard way following HELM: `RNA1{([25d3r],[25mo3r])(A)P}$$$$V2.0`
![image](https://github.com/user-attachments/assets/6d81a8de-7cc8-4039-a3e0-3c58e6e848f9)

**Actual behavior**
System throws an error: Convert error! {}
![image](https://github.com/user-attachments/assets/4dce5830-9aec-48d1-a2a8-18eeadc8b495)
and single sugar appears on the canvas (with APs occupied but with no phosphate and base)
![image](https://github.com/user-attachments/assets/c44d18bf-c314-4a01-8c96-cc7c939f2f48)

**Expected behavior**
No exception. Preset appears on the canvas
![image](https://github.com/user-attachments/assets/57ffb7c8-b947-4260-858a-fd99171b18db)

**Versions**
- Ketcher Version 2.25.0-rc.4 Build at 2024-09-23; 17:24:38
- Indigo Toolkit Version 1.24.0-rc.3.0-g692080c62-wasm32-wasm-clang-19.0.0
- Chrome Version 129.0.6668.59 (Official Build) (64-bit)
- Win10

Reference issues: https://github.com/epam/Indigo/issues/2188, https://github.com/epam/ketcher/issues/4554

PR #5716, opened 2024-10-10
#5709 - Unable to load variant sugar from HELM - system throws an error: Convert error! {}

rrodionov91:

- fixed sugar/phosphate/chem variant monomers render

## Check list
- [ ] unit-tests written
- [ ] e2e-tests written
- [ ] documentation updated


packages/ketcher-core/src/application/render/renderers/constants.ts
```diff
@@ -15,6 +15,7 @@ export const MONOMER_SYMBOLS_IDS = {
     hover: '#sugar-selection',
     selected: '#sugar-selection',
     body: '#sugar',
+    variant: '#sugar-variant',
```
svvald, line 18: Don't we want to use `ambiguous` here to avoid ambiguity (hehe) between `variant` and `ambiguous` in our code?

svvald (approved):

Issue #5988, opened 2024-11-18
Exception when modifying a functional group after adding a ketcher editor subscription


askyel:
**Steps to Reproduce**
1. Go to https://lifescience.opensource.epam.com/KetcherDemoSA/index.html or use ketcher v.2.25.0
2. Execute `ketcher.editor.subscribe("change", () => console.log("hello"));` in the console (or add any other change subscription)
3. In the canvas, add a functional group such as "CF3"
4. Click on another atom such as "Br" and click on the functional group
5. Exceptions are thrown and the canvas becomes blank

**Actual behavior**
After a ketcher editor subscription is added for a "change" event, trying to replace a functional group, expand a functional group, or contract a functional group causes exceptions to be thrown. In addition, the ketcher canvas becomes blank.
This is the stracktrace shown for the exception:
```
Uncaught TypeError: Cannot read properties of undefined (reading 'type')
    at eval (index.modern.js:17491:16)
    at Array.forEach (<anonymous>)
    at customOnChangeHandler (index.modern.js:17489:31)
    at subscribeFuncWrapper (index.modern.js:24527:22)
    at Subscription.dispatch (index.js:51:19)
    at Editor.update (index.modern.js:24463:29)
    at AtomTool.mouseup (index.modern.js:16678:14)
    at useToolIfNeeded (index.modern.js:24661:130)
    at Handler.eval [as f] (index.modern.js:24721:24)
    at DOMSubscription.dispatch (index.js:93:21)
    at HTMLDocument.eval (index.modern.js:24708:21)
    at HTMLDocument.sentryWrapped (helpers.js:106:17)
```

**Expected behavior**
The correct functional group modification on the canvas structure - in this case the "Br" atom replacing the "CF3" functional group. 

**Screenshots**
https://github.com/user-attachments/assets/ebccc96a-eb12-4d08-88c0-c0402fb0552c
<img width="1469" alt="image" src="https://github.com/user-attachments/assets/6950b813-1745-41fb-9225-e086138a3021">

**Desktop (please complete the following information):**
 - OS: macOS
 - Browser: chrome
 - Version: 130.0.6723.117 (Official Build) (arm64)

**Ketcher version**
v2.25.0

rrodionov91:
Hi @askyel 
Thank you for highlighting this issue.
We will check it out.

Zhirnoff:
Tested. Bug fixed.


[...]

PR #5993, opened 2024-11-19
5988 - perform s-group attribute action when an s-group is deleted

askyel:
In issue https://github.com/epam/ketcher/issues/5988, the structure in the canvas disappeared when the editor was configured with a custom change subscription and a functional group was replaced with a different atom.
The error thrown is that the `_inverted` field on the s-group attribute operations are undefined, but assumed to be defined here:
https://github.com/epam/ketcher/blob/f87331811d1b3eba7f93a804dee2b75706f7d8aa/packages/ketcher-react/src/script/editor/utils/customOnChangeHandler.ts#L46
An operation's `_inverted` field is only set after the `perform` function is executed here:
https://github.com/epam/ketcher/blob/f87331811d1b3eba7f93a804dee2b75706f7d8aa/packages/ketcher-core/src/application/editor/operations/base.ts#L41-L48
The change in this PR moves the `perform` function execution from before the s-group attribute operations were added to after, such that those operations also have their `_inverted` field set. 
I tested the same behavior as in the issue locally and this change solves the issue:
https://github.com/user-attachments/assets/8b9764a3-1074-40de-a821-9e44cb16a9d9

Note that the `_inverted` field still can be undefined in the `customOnChangeHandler` for other operations and this PR does not address that issue.

## Check list
- [ ] unit-tests written
- [ ] e2e-tests written
- [ ] documentation updated

### DIFF
--- a/packages/ketcher-core/src/application/editor/actions/sgroup.ts
+++ b/packages/ketcher-core/src/application/editor/actions/sgroup.ts
@@ -312,11 +312,11 @@ export function fromSgroupDeletion(restruct, id, needPerform = true) {
 
   action.addOp(new SGroupDelete(id));
 
+  action.mergeWith(sGroupAttributeAction(id, attrs));
+
   if (needPerform) {
     action = action.perform(restruct);
   }
-
-  action.mergeWith(sGroupAttributeAction(id, attrs));
 
   return action;
 }


rrodionov91:
Dear @askyel 
May I ask you to rebase from latest master please? It should fix failed tests.

askyel:
> Dear @askyel May I ask you to rebase from latest master please? It should fix failed tests.

Yes, it should be updated now

rrodionov91 (approved):

Issue #5618, opened 2024-10-01
Ketcher doesn't trigger change event in macromolecule mode


MariaDolotova:
How can I subscribe to Ketcher change event when using macromolecule mode.
I have used the demo stand, tried to subscribe to 'change', but it didn't work. 
Please see the gif below.

![ketcher_macromolecule_mode_change_event](https://github.com/user-attachments/assets/03ddc795-fd0a-40bb-97ae-fd4ac2c117fb)

Zhirnoff:
Tested. Bug fixed.

https://github.com/user-attachments/assets/c6285c1e-6f4f-4b43-995c-50fed95ad5fb



**Desktop:**
 - OS: Windows 11
 - Browser Chrome
 - Version 128.0.6613.120 (Official Build) (64-bit)

**Ketcher version** 
[Version 2.28.0-rc.1]
**Indigo version** 
[Version 1.27.0-rc.1]

PR #6012, opened 2024-11-26
#5618 ketcher doesnt trigger change event in macromolecule mode

Drimodaren:

## Check list
- [ ] unit-tests written
- [ ] e2e-tests written
- [ ] documentation updated


packages/ketcher-core/src/application/editor/EditorHistory.ts
```diff
@@ -25,6 +26,7 @@ export class EditorHistory {
   historyStack: Command[] | [] = [];
   historyPointer = 0;
   editor: CoreEditor | undefined;
+  ketcherInstance;
```
svvald, line 29: As we discussed, you don't need this class property for now

packages/ketcher-react/src/script/editor/Editor.ts
```diff
@@ -573,9 +573,9 @@ class Editor implements KetcherEditor {
     this.historyStack[this.historyPtr] = action;
 
     if (this._tool instanceof toolsMap.paste) {
-      this.event.change.dispatch();
+      ketcherProvider.getKetcher().changeEvent.dispatch();
     } else {
-      this.event.change.dispatch(action);
+      ketcherProvider.getKetcher().changeEvent.dispatch(action);
```
svvald, line 578: For a bit of readability I'd prefer to have some variable assigned to ketcher instance and it being reused as long as function uses ketcher more than once

packages/ketcher-react/src/script/editor/utils/customOnChangeHandler.ts
```diff
@@ -40,7 +40,9 @@ type Data = {
 
 export function customOnChangeHandler(action, handler) {
   const data: Data[] = [];
-
+  if (!action) {
```
svvald, line 43: Let's prepare this function a bit for some future adjustments. Let it be wrapper for both micro and macro 'changes' event handler and we will have two separate functions for micro and macro event handlers. The former will have all the existing functionality – preparing the data based on passed action and invoking the handler with this data. The latter will just have invoking the handler with empty arguments list – just what you've added here

Then, based on the currently active editor (you can use `window.isPolymerEditorTurnedOn`) just invoke either the first or the second function.

packages/ketcher-react/src/script/editor/utils/customOnChangeHandler.ts
```diff
@@ -40,7 +40,9 @@ type Data = {
 
 export function customOnChangeHandler(action, handler) {
```
svvald, line 41: For a fail safe check I'd add type for `action` parameter – let it be `Action | undefined` for now, as we could see, it's possible to trigger event without any data and it will silently fail on the following `forEach` call

svvald (changes requested):
Changes are needed, disregard the previous approval

NataliaLoginova (approved):

packages/ketcher-react/src/script/editor/Editor.ts
```diff
@@ -557,6 +557,7 @@ class Editor implements KetcherEditor {
   }
 
   undo() {
+    const ketcherProviderEditor = ketcherProvider.getKetcher().changeEvent;
```
svvald, line 560: It is rather event instance than editor instance, I'd change variable name

svvald (approved):

packages/ketcher-react/src/script/editor/Editor.ts
Drimodaren: Maby ketcherChangeEvent?

Zhirnoff (approved):

Zhirnoff (approved):

Issue #5969, opened 2024-11-13
Load from file having only micro structures on macro canvas causes  unnecessary zoom up to 200% and shift molecule to top left angle


AlexeyGirin:
**Steps to Reproduce**
1. Go to Macro - Flex
2. Load from file: [Benzene ring.zip](https://github.com/user-attachments/files/17733753/Benzene.ring.zip)

**Actual behavior**
Canvas zoomed to 200% and molecule shifted to top left angle
![image](https://github.com/user-attachments/assets/b1cd3025-52ee-4c08-af0f-c84a33ed0874)

**Expected behavior**
No zoom, molecule at the center
![image](https://github.com/user-attachments/assets/d6a2b50f-953c-4f99-a822-5008d716ad04)

**Versions**
- Ketcher Version 2.27.0-rc.2 Build at 2024-11-12; 19:26:10
- Indigo Toolkit Version 1.26.0-rc.2.0-gfec30ecd8-wasm32-wasm-clang-19.0.0
- Chrome Version 131.0.6778.70 (Official Build) (64-bit)
- Win10

Issue found while testing: https://github.com/epam/ketcher/issues/5600

rrodionov91:
Moved to backlog. Will be fixed in separate task: https://github.com/epam/ketcher/issues/5996

AlexeyGirin:
Verified.
![image](https://github.com/user-attachments/assets/408a650c-7db8-423a-8cea-9c555fb05bed)
- Ketcher Version 2.27.0-rc.3 Build at 2024-12-25; 10:08:04
- Indigo Version 1.26.0-rc.5.0-g42ee8510b-wasm32-wasm-clang-19.0.0
- Chrome Version 131.0.6778.205 (Official Build) (64-bit)
- Win10

PR #6057, opened 2024-12-04
#5969 - Load from file having only micro structures on macro canvas causes unnecessary zoom up to 200% and shift molecule to top left angle

rrodionov91:

## Check list
- [ ] unit-tests written
- [ ] e2e-tests written
- [ ] documentation updated

### DIFF
--- a/packages/ketcher-core/src/domain/entities/DrawingEntitiesManager.ts
+++ b/packages/ketcher-core/src/domain/entities/DrawingEntitiesManager.ts
@@ -166,6 +166,16 @@ export class DrawingEntitiesManager {
     return this.allEntities.length !== 0;
   }
 
+  public get hasMonomers() {
+    const monomers = [...this.monomers.values()].filter(
+      (monomer) =>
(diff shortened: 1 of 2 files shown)


packages/ketcher-core/src/domain/entities/DrawingEntitiesManager.ts
```diff
@@ -166,6 +166,16 @@ export class DrawingEntitiesManager {
     return this.allEntities.length !== 0;
   }
 
+  public get hasMonomers() {
+    const monomers = [...this.monomers.values()].filter(
```
svvald, line 170: `filter` method doesn't mutate the original array and creates a new copy, perhaps you can just use `this.monomers.values().filter()` here?

svvald (approved):

packages/ketcher-core/src/domain/entities/DrawingEntitiesManager.ts
rrodionov91: this.monomers.values() returns Iterator, so destructuring operator needs to create array from Iterator.

Issue #5978, opened 2024-11-15
Entire element bounding box should be clickable, not only black dots


AlexeyGirin:
**Steps to Reproduce**
1. Go to Macro - Flex mode
2. Load from file: [Entire element bounding box should be clickable, not only black dots.zip](https://github.com/user-attachments/files/17775562/Entire.element.bounding.box.should.be.clickable.not.only.black.dots.zip)
![image](https://github.com/user-attachments/assets/73c41a6c-8f42-4f92-bb8e-ed3a87d4000a)
3. Click on the center of double bond (e.g. between two black lines)

**Actual behavior**
Nothing happens

**Expected behavior**
Double bond got selected
![image](https://github.com/user-attachments/assets/af2bd493-8e11-4fd1-a122-7a6dd0cd96f3)

**Versions**
- Ketcher Version 2.27.0-rc.2 Build at 2024-11-12; 19:26:10
- Indigo Toolkit Version 1.26.0-rc.2.0-gfec30ecd8-wasm32-wasm-clang-19.0.0
- Chrome Version 131.0.6778.70 (Official Build) (64-bit)
- Win10

Issue found while testing: https://github.com/epam/ketcher/issues/5359

Zhirnoff:
Tested. Bug fixed.

![2025-01-03_12h04_32](https://github.com/user-attachments/assets/856f69d3-424b-4476-a228-91e07a2ae61b)

**Desktop:**
 - OS: Windows 11
 - Browser Chrome
 - Version 131.0.6778.205 (Official Build) (64-bit)

**Ketcher version** 
[Version 3.0.0-rc.1]
**Indigo version** 
[Version 1.28.0-rc.1]

PR #6059, opened 2024-12-04
#5978-entire-element-bounding-box-should-be-clickable-not-only-black-dots

Drimodaren:

image

## Check list
- [x] unit-tests written
- [ ] e2e-tests written
- [ ] documentation updated


packages/ketcher-core/src/application/render/renderers/BondRenderer.ts
```diff
@@ -628,6 +639,18 @@ export class BondRenderer extends BaseRenderer {
...
+          .attr('d', this.pathShape)
+          .attr('stroke', 'black')
+          .attr('stroke-width', `${BOND_WIDTH * 10}px`)
+          .attr('opacity', '0')
+          .on('mouseenter', () => {
+            this.appendHover();
+          })
+          .on('mouseleave', () => {
+            this.removeHover();
+          });
```
Guch1g0v, line 653: We have duplicated logic; it would be great to extract it into a private method, something like this:
```js
  private createPath(
    attrs: Record<string, string>,
    isHoverable: boolean,
  ): void {
    if (!this.rootElement) {
      return;
    }
    const path = this.rootElement.append('path').attr('d', this.pathShape);

    Object.entries(attrs).forEach(([key, value]) => {
      path.attr(key, value);
    });

    if (isHoverable) {
      path
        .on('mouseover', () => {
          this.appendHover();
        })
        .on('mouseout', () => {
          this.removeHover();
        });
    }
  }
```

And use it like this:
```js
        this.createPath(
          {
            stroke: 'transparent',
            'stroke-width': `${BOND_WIDTH * 10}px`,
            opacity: '0',
          },
          true,
        );

        this.createPath(
          {
            stroke: 'black',
            'stroke-width': `${BOND_WIDTH}px`,
          },
          true,
        );
```

Guch1g0v (changes requested):

packages/ketcher-core/src/application/render/renderers/BondRenderer.ts
```diff
@@ -374,23 +374,39 @@ export class BondRenderer extends BaseRenderer {
     this.move();
   }
 
-  private bondHoverablePath(attrs: Record<string, string>): void {
+  private createBondHoverablePath(
```
Guch1g0v, line 377:
```typescript
type MouseActionType = 'mouseover' | 'mouseenter';

const mouseEvents: Record<MouseActionType, { enter: string; leave: string }> = {
  mouseover: { enter: 'mouseover', leave: 'mouseout' },
  mouseenter: { enter: 'mouseenter', leave: 'mouseleave' },
};
  private createPath(
    attrs: Record<string, string>,
    mouseAction?: MouseActionType,
  ): void {
    if (!this.rootElement) {
      return;
    }

    const path = this.rootElement.append('path').attr('d', this.pathShape);

    Object.entries(attrs).forEach(([key, value]) => {
      path.attr(key, value);
    });

    if (mouseAction) {
      const events = mouseEvents[mouseAction];
      path
        .on(events.enter, () => {
          this.appendHover();
        })
        .on(events.leave, () => {
          this.removeHover();
        });
    }
  }
```

Guch1g0v (approved):

PR #4749, opened 2024-06-04
#853 - Content Security Policy - "unsafe-eval"

ihuang8:
## How the feature works? / How did you fix the issue?
(Screenshots, videos, or GIFs, if applicable)

Fixes the issue in #853.  AJV does not work in a browser page with a CSP that does not allow for "unsafe-eval". Switching to use the jsonschema library as the JSON schema validator fixes this. 

## Check list
- [ ] unit-tests written
- [ ] e2e-tests written
- [ ] documentation updated
- [ ] PR name follows the pattern `#1234 – issue name`
- [ ] branch name doesn't contain '#'
- [ ] PR is linked with the issue
- [ ] base branch (master or release/xx) is correct
- [ ] task status changed to "Code review"
- [ ] reviewers are notified about the pull request

Issue #3627, opened 2023-11-28
Ketcher requires unsafe-eval in order to run, which contradicts content security policy best practises


ahmepham:
**Steps to Reproduce**
1.  Disable unsafe-eval
2. Use ketcher-react

**Actual behavior**
Editor will not worker because browser will not allow scripts to run

**Expected behavior**
Editor will be fully functional

**Ketcher version** [e.g. v2.4.2]. 
2.13.0

**Additional context**
This appears to be because AJV schema compilation requires unsafe-eval.  The console shows that the schema compilation fails: 
![image](https://github.com/epam/ketcher/assets/112775251/a0a44dbd-a163-4e2a-b6f3-9cee7dc31a8b)

These schemas can be compiled ahead of time relatively easily : https://ajv.js.org/standalone.html 

Doing so should resolve the problem and allow ketcher to function without unsafe eval.

NicolasCARPi:
Hello,

We also run into this issue and adding "unsafe-eval" in the CSP is not possible obviously, as it would defeat the whole thing. Please let us know if we can sponsor this fix somehow or if there is a workaround or something.

Cheers,
~Nico

NicolasCARPi:
@Nitvex Would you please be able to look into this or point us in a direction on how you imagine this could be fixed?

rrodionov91:
Hi @NicolasCARPi 
We a discussion in this pull request https://github.com/epam/ketcher/pull/4749
I provided a link to potential solution there.
It would be great if you can help with implementing it.

NicolasCARPi:
@rrodionov91 Thank you very much for pointing to this PR. I've assigned @moustaphacamara on this issue.

NicolasCARPi:
Hello,

Just a heads-up, we now have a working fork that works with restrictive CSP. The changeset is pretty minimal, as can be seen here: https://github.com/epam/ketcher/compare/release/2.26...deltablot:ketcher-ajv:release/2.26-ajv?expand=1

We will make a proper PR based on master soon!

NicolasCARPi:
For information, the PR fixing this issue is opened here: https://github.com/epam/ketcher/pull/6269

Let's hope it can be reviewed and merged (relatively) quickly :)

Issue #853, opened 2021-10-19
Content Security Policy - "unsafe-eval"


onoufriosm:
**Steps to Reproduce**
Add a CSP (in your html entry file) that doesn't allow for "unsafe-eval"

**Expected behavior**
Not throw a CSP error

**Actual behavior**
Throws a CSP error 
```
Refused to evaluate a string as JavaScript because 'unsafe-eval' is not an allowed source of script
```

**Additional context**
I've tracked this down to `ketcher-standalone/src/generate/libindigo.js` line 1230. Specifically `new Function` will cause this. 

Don't have enough context to make a change to this but essentially `new Function` should go away otherwise CSP won't let it run.

AllaZhbanova:
related https://github.com/epam/Indigo/issues/567

AndreiMazol:
Moved to 2.5
cc: @AllaZhbanova

martinuncountable:
@AndreiMazol  @AllaZhbanova

My understanding is that if we introduce an option for ketcher-core/react package to compile the AVJ validation schemas at build time that will remove the `unsafe-eval` dependency.

Is this something being actively worked on?

ahmepham:
@martinuncountable Do you have any update on this.  It is currently blocking my ability to use ketcher in my app, which is a bummer since ketcher is a very nice library

martinuncountable:
> @martinuncountable Do you have any update on this. It is currently blocking my ability to use ketcher in my app, which is a bummer since ketcher is a very nice library

Yes we have forked Ketcher to use `json-schema` instead. We are working on creating a PR here to merge upstream.

NicolasCARPi:
For information, the PR fixing this issue is opened here: https://github.com/epam/ketcher/pull/6269

Let's hope it can be reviewed and merged (relatively) quickly :)

PR #6269, opened 2025-01-14
#3627 - Ketcher requires unsafe-eval in order to run, which contradicts content security policy best practises

MoustaphaCamara:

Fix the issue mentioned in:
- #3627
- #853

Supersedes:
- #4749
Referring to [rrodionov91's comment](https://github.com/epam/ketcher/pull/4749#issuecomment-2149694336) , it was preferred to provide a solution using ajv, to avoid switching to jsonschema.

This PR uses ajv-cli to pre-compile the validation schema and allow ketcher to run under restrictive CSP header.
Ajv is removed from ketcher-core dependencies as not used anymore, replaced by ajv-cli only.
Ajv persists in ketcher-react as it is used and doesn't need pre-compilation.

This change has been successfully used in production without any issues. Please consider merging this change so all projects with restrictive Content Security Policies can also benefit from this excellent piece of software. =)

## Check list
- [ ] unit-tests written
- [ ] e2e-tests written
- [ ] documentation updated

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -37,6 +37,9 @@ yarn-error.log*
 nohup.out
 ketcher-autotests/results.json
 
+# ajv validation file generated with npm run ajv
+/packages/ketcher-core/src/domain/serializers/ket/compiledSchema.js
+
 .pnp.*
 .yarn/*
 tests/test-data/KET/Nucleotide-Templates/01 - (R1) - Left only.ket
--- a/package-lock.json
+++ b/package-lock.json
@@ -8305,6 +8305,33 @@
         "url": "https://github.com/sponsors/epoberezkin"
       }
     },
+    "node_modules/ajv-cli": {
+      "version": "5.0.0",
+      "resolved": "https://registry.npmjs.org/ajv-cli/-/ajv-cli-5.0.0.tgz",
+      "integrity": "sha512-LY4m6dUv44HTyhV+u2z5uX4EhPYTM38Iv1jdgDJJJCyOOuqB8KtZEGjPZ2T+sh5ZIJrXUfgErYx/j3gLd3+PlQ==",
+      "dev": true,
+      "license": "MIT",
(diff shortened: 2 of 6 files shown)


rrodionov91:
Hi @MoustaphaCamara 
Thank you very much for this PR.

CI Build job failed
image

Coud you please check what is bad with prettier there?

NicolasCARPi:
> Coud you please check what is bad with prettier there?

TL;DR: prettier version is too old

In version 3+, prettier ignores files added to `.gitignore` ([source 1](https://github.com/prettier/prettier/issues/16421#issuecomment-2181997700), [source 2](https://prettier.io/docs/en/ignore.html)).

The current `package.json` pins it to `2.x`. @MoustaphaCamara will be back on Monday and work on this. Am I right to assume that you prefer a separate PR with the prettier upgrade (which might have side effects)?

rrodionov91:
> > Coud you please check what is bad with prettier there?
> 
> TL;DR: prettier version is too old
> 
> In version 3+, prettier ignores files added to `.gitignore` ([source 1](https://github.com/prettier/prettier/issues/16421#issuecomment-2181997700), [source 2](https://prettier.io/docs/en/ignore.html)).
> 
> The current `package.json` pins it to `2.x`. @MoustaphaCamara will be back on Monday and work on this. Am I right to assume that you prefer a separate PR with the prettier upgrade (which might have side effects)?

I think yes, it would be prefferable to have a separate PR for that.
Maybe for current PR we can use .prettierignore or .eslintignore files? Just to speed up merge

MoustaphaCamara:
>  Maybe for current PR we can use .prettierignore or .eslintignore files? Just to speed up merge

Seems good to me, i've updated the ignore files for the current PR @rrodionov91 .
Thanks for your feedback

rrodionov91 (approved):

Issue #6254, opened 2025-01-10
Update the representation of sense and antisense chains in sequence mode (2/2)


ljubica-milovic:
## Background 

Continuation of the ticket from the previous release: [#5942](https://github.com/epam/ketcher/issues/5942)

## Summary of differences between editing both chains and editing one chain

![image](https://github.com/user-attachments/assets/611da1c6-683b-4510-a14b-c67b22b0a3bd)

### Removing/adding H-bonds

1.1. Right clicking on a symbol that has no hydrogen bonds with a symbol bellow/above it, should give the option to "Establish Hydrogen Bonds". If a symbol has H-bonds with a symbol opposite it, or there is no monomer opposite it, the option should be disabled.
> A hydrogen bond should be established between the monomer that was clicked on and the monomer placed bellow/above it.
> "Establish Hydrogen Bonds" is disabled if any of the symbols (the one that was clicked on, or the one opposite it) is the @ symbol representing multiple monomers.
> If the symbol represents a nucleotide/nucleoside, a hydrogen bond should be established from the base (no need for a "sense base" check).

1.2. Right licking on a symbol that has any H-bonds should give the option to "Delete Hydrogen Bonds". If there are no H-bonds, the option should be disabled.
> All hydrogen bonds from monomers that symbol represents should be deleted.

1.3. If multiple monomers/symbols are selected and at least one of them does not have hydrogen bonds established with the monomer/symbol above/bellow it the option "Establish Hydrogen Bonds" should be available from the r-click menu. If all symbols have H-bonds, the option should be disabled.
> "Establish Hydrogen Bonds" is disabled if any of the symbols that do not have H-bonds already with the symbols opposite them are the @ symbol representing multiple monomers.
> For all monomers/symbols that do not have H-bonds with their pairs, a H-bond should be established (if the symbols represent nucleotides/nucleotides, an H-bond comes from the base).
> Edge case example:
> ![image](https://github.com/user-attachments/assets/3de0d158-9779-4862-9f89-c7c6375413a0)
> ![image](https://github.com/user-attachments/assets/876426b9-70ab-4009-8852-cbb8a44400d1)
> If all symbols are selected the option "Establish Hydrogen Bonds" should be enabled, because no H-bond will be established with the CHEMs (the @ symbol already has an H-bond with the symbol bellow), but just between nucleotides/nucleosides:
> ![image](https://github.com/user-attachments/assets/b522e0b6-dcfe-4b90-911e-105a268cb1c6)
[...]

PR #6575, opened 2025-02-26
 #6254 - Update the representation of sense and antisense chains in sequence mode (2/2)

rrodionov91:

## Check list
- [ ] unit-tests written
- [ ] e2e-tests written
- [ ] documentation updated


Zhirnoff (approved):

packages/ketcher-macromolecules/src/components/SequenceSyncEditModeButton/SequenceSyncEditModeButton.tsx
```diff
@@ -0,0 +1,164 @@
...
+
+    ':hover svg': {
+      fill: isActive ? 'white' : theme.ketcher.color.button.group.active,
+    },
+  }),
+);
+
+export const SequenceSyncEditModeButton = () => {
+  const editor = useAppSelector(selectEditor);
+  const [isSequenceSyncEditMode, setIsSequenceSyncEditMode] = useState(true);
```
svvald, line 53: I suppose we don't need to have this state here, it's not a responsibility of the button to track the sync mode state, it should just rely on the editor's internal state on render stage

packages/ketcher-macromolecules/src/components/SequenceSyncEditModeButton/SequenceSyncEditModeButton.tsx
```diff
@@ -0,0 +1,164 @@
...
+    const isSequenceSyncEditModeNewState = !isSequenceSyncEditMode;
+
+    setIsSequenceSyncEditMode(isSequenceSyncEditModeNewState);
+    editor.events.toggleIsSequenceSyncEditMode.dispatch(
+      isSequenceSyncEditModeNewState,
+    );
+    blurActiveElement();
+  };
+
+  useEffect(() => {
```
svvald, line 68: Following the above, this effect shouldn't exist at all

packages/ketcher-core/src/application/editor/Editor.ts
```diff
@@ -428,6 +459,20 @@ export class CoreEditor {
...
+    if (!(this.mode instanceof SequenceMode)) {
+      return;
+    }
+
+    if (isSequenceSyncEditMode) {
+      this.mode.turnOnSyncEditMode();
+    } else {
+      this.mode.turnOffSyncEditMode();
+    }
+  }
```
svvald, line 474: I assume `toggle` function shouldn't accept any value but just negate the current state of the control (compared to `set` function)
I think it would fit your logic as well, rather then explicitly setting the exact value when clicking the sync button you can just switch to the opposite value within sequence mode

svvald (approved):

Issue #5497, opened 2024-09-11
Refactor: Split logic of `SnakeModePolymerBondRenderer` by several classes


DmitriiP-EPAM:
Create:
- SnakeModeBackboneBondRenderer (black “snake” line)
- SnakeModeSideChainBondRenderer (blue “snake” line)
- SnakeModeRNABaseAndSugarBondRenderer (black straight line)

Next step after https://github.com/epam/ketcher/issues/5128

AlexeyGirin:
No testing required

PR #6756, opened 2025-03-25
#5497 – Refactor (SnakeModePolymerBondRenderer): Create `SnakeModeSideChainBondRenderer`

DmitriiP-EPAM:

## Check list
- [ ] unit-tests written
- [ ] e2e-tests written
- [ ] documentation updated

### DIFF
--- a/packages/ketcher-core/src/application/render/renderers/PolymerBondRenderer/SVGPathDAttributeUtil.ts
+++ b/packages/ketcher-core/src/application/render/renderers/PolymerBondRenderer/SVGPathDAttributeUtil.ts
@@ -7,6 +7,10 @@ export class SVGPathDAttributeUtil {
     return `H ${x}`;
   }
 
+  public static generateMoveTo(x: number, y: number): string {
+    return `M ${x},${y}`;
+  }
+
   public static generateVerticalAbsoluteLine(y: number): string {
     return `V ${y}`;
   }
--- a/packages/ketcher-core/src/application/render/renderers/PolymerBondRenderer/SVGPathDAttributeUtil.ts
+++ b/packages/ketcher-core/src/application/render/renderers/PolymerBondRenderer/SVGPathDAttributeUtil.ts
@@ -1,0 +1,13 @@
+export class SVGPathDAttributeUtil {
+  public static generateAbsoluteLine(x: number, y: number): string {
+    return `L ${x},${y}`;
+  }
+
+  public static generateHorizontalAbsoluteLine(x: number): string {
+    return `H ${x}`;
+  }
+
+  public static generateVerticalAbsoluteLine(y: number): string {
+    return `V ${y}`;
+  }
+}
--- a/packages/ketcher-core/src/application/render/renderers/PolymerBondRenderer/SnakeModeHydrogenBondRenderer.ts
+++ b/packages/ketcher-core/src/application/render/renderers/PolymerBondRenderer/SnakeModeHydrogenBondRenderer.ts
@@ -18,7 +18,6 @@ import { BaseRenderer } from '../BaseRenderer';
 import {
   CORNER_LENGTH,
   DOUBLE_CORNER_LENGTH,
-  generateBend,
   generateCornerFromBottomToRight,
   generateCornerFromLeftToBottom,
   generateCornerFromLeftToTop,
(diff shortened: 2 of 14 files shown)


packages/ketcher-core/src/application/render/renderers/PolymerBondRenderer/SnakeModeHydrogenBondRenderer.ts
```diff
@@ -0,0 +1,962 @@
...
+const LINE_FROM_MONOMER_LENGTH = 15;
+const VERTICAL_LINE_LENGTH = 21;
+const RNA_CHAIN_VERTICAL_LINE_LENGTH = 74;
+const RNA_ANTISENSE_CHAIN_VERTICAL_LINE_LENGTH = 20;
+const RNA_SENSE_CHAIN_VERTICAL_LINE_LENGTH = 210;
+
+const BOND_END_LENGTH = 15;
+const SIDE_CONNECTION_BODY_ELEMENT_CLASS = 'polymer-bond-body';
+
+export class SnakeModeHydrogenBondRenderer extends BaseRenderer {
```
rrodionov91, line 46: Render part of hydrogen bond is almost the same as for side connection bond. There is no need to separate them. >95% of code is the same as for side connection

rrodionov91 (changes requested):

packages/ketcher-core/src/application/render/renderers/PolymerBondRenderer/SnakeModePolymerBondRenderer.ts
```diff
@@ -226,141 +210,102 @@ export class SnakeModePolymerBondRenderer extends BaseRenderer {
...
-
-    if (areCellsOnSameRow) {
-      dAttributeForPath += `L ${startPosition.x},${
-        startPosition.y -
-        BOND_END_LENGTH -
-        horizontalPartIntersectionsOffset * 3
-      } `;
-      dAttributeForPath += generateBend(0, -1, cos, -1);
+    if (cellsAreOnSameRow) {
+      {
```
rrodionov91, line 257: Do we need these brackets?

packages/ketcher-core/src/application/render/renderers/PolymerBondRenderer/SnakeModePolymerBondRenderer.ts
```diff
@@ -226,141 +210,102 @@ export class SnakeModePolymerBondRenderer extends BaseRenderer {
...
+      }
+      pathDAttributeValue +=
+        SideChainConnectionBondRendererUtil.generateBend(0, -1, cos, -1) + ' ';
     } else {
-      dAttributeForPath += `L ${startPosition.x},${
-        startPosition.y +
-        BOND_END_LENGTH +
-        horizontalPartIntersectionsOffset * 3
-      } `;
+      {
```
rrodionov91, line 271: Do we need these brackets?

rrodionov91 (changes requested):

packages/ketcher-core/src/application/render/renderers/PolymerBondRenderer/SVGPathDAttributeUtil.ts
```diff
@@ -0,0 +1,17 @@
...
+  }
+
+  public static generateMoveTo(x: number, y: number): string {
+    return `M ${x},${y}`;
+  }
+
+  public static generateVerticalAbsoluteLine(y: number): string {
+    return `V ${y}`;
+  }
+}
```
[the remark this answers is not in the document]
DmitriiP-EPAM: Yes, this set of methods may be revised in the future to use a chain of calls. In my opinion, now only disparate static methods are needed, or even without the utility.

packages/ketcher-core/src/application/render/renderers/PolymerBondRenderer/SnakeModePolymerBondRenderer.ts
DmitriiP-EPAM: We can do without the curly braces. The braces are needed to create a visibility block that encapsulates a computed constant that is used once. This is useful because the method is very long and uses many constants - a quick glance at the method does not allow us to determine whether the constants are used further. This block indicates that this part of the code can be factored out into a function. The code of this long method should be transformed in the future.

packages/ketcher-core/src/application/render/renderers/PolymerBondRenderer/SnakeModePolymerBondRenderer.ts
DmitriiP-EPAM: We can do without the curly braces. The braces are needed to create a visibility block that encapsulates a computed constant that is used once. This is useful because the method is very long and uses many constants - a quick glance at the method does not allow us to determine whether the constants are used further. This block indicates that this part of the code can be factored out into a function. The code of this long method should be transformed in the future.

DmitriiP-EPAM:
https://github.com/epam/ketcher/pull/6819

Issue #7207, opened 2025-06-04
Rectangular input field should be wide enough to fit any (at least 4) digit number


AlexeyGirin:
**Steps to Reproduce**
1. Go to Molecules mode
2. Set monomers in line to `1000`

**Actual behavior**
Rectangular input field doesn't fit 4 digit number
![Image](https://github.com/user-attachments/assets/f8c29100-e0ab-409c-8e1c-a3c66aeafa04)

**Expected behavior**
Rectangular input field fits 4 digit number
![Image](https://github.com/user-attachments/assets/7f936ec5-b6aa-47a0-990c-e1afe0d1cbab)

**Environment details:**
- Ketcher Version 3.5.0-rc.1 Build at 2025-06-03; 14:37:03
- Indigo Version 1.33.0-rc.1.0-g1f546bed9-wasm32-wasm-clang-19.0.0
- Chrome Version 137.0.7151.69 (Official Build) (64-bit)
- Win10

Found while testing: https://github.com/epam/ketcher/issues/6738

AlexeyGirin:
Verified.
![Image](https://github.com/user-attachments/assets/1815996e-c613-40f2-a92c-90b51e43a29b)

- Ketcher Version 3.6.0-rc.1 Build at 2025-07-01; 19:04:05
- Indigo Version 1.34.0-rc.1.0-g7d0cdcdef-wasm32-wasm-clang-19.0.0
- Chrome Version 138.0.7204.97 (Official Build) (64-bit)
- Win10

PR #7296, opened 2025-06-20
 #7207 - Rectangular input field should be wide enough to fit any (at least 4) digit number

KristinaKasyanovskaya:
- [x]  Rectangular input field fits 4 digit number 

## Check list
- [ ] unit-tests written
- [ ] e2e-tests written
- [ ] documentation updated


packages/ketcher-macromolecules/src/components/Ruler/RulerArea.module.less
```diff
@@ -9,7 +9,7 @@
   position: absolute;
   top: 0;
   height: 14px;
-  width: 26px;
+  width: 35px;
```
svvald, line 12: There's a logic within `RulerArea.tsx` which adjusts input position to align with top toolbar, please change it as well:
image

svvald (changes requested):

packages/ketcher-macromolecules/src/components/Ruler/RulerArea.module.less
KristinaKasyanovskaya: Thanks for the review! I've fixed the code.

svvald (approved):

Issue #7700, opened 2025-08-22
System allow to delete monomers leaving connected bonds on the canvas


AlexeyGirin:
**Steps to Reproduce**
1. Go to Macromolecules mode - Flex canvas (empty)
2. Load from HELM: `RNA1{d([Hyp])p.r(A)p.d([Hyp])p.r(A)p.d([Hyp])}$$$$V2.0`
<img width="616" height="124" alt="Image" src="https://github.com/user-attachments/assets/8c4809aa-c070-40e2-b39e-04e5bbae4b38" />

3. Press `Erase` button
4. Press `Ctrl+a` key to select structure on the canvas
<img width="686" height="144" alt="Image" src="https://github.com/user-attachments/assets/b299d17e-c957-4b38-b613-57d3d57f58ff" />

5. Click on any monomer (top right `dR` sugar in my case)

**Actual behavior**
System removed monomer but bonds remain in place
<img width="668" height="162" alt="Image" src="https://github.com/user-attachments/assets/502d8ce4-7f88-4f50-b4f2-e0161039c226" />

**Expected behavior**
Monomer got removed with bonds
<img width="692" height="160" alt="Image" src="https://github.com/user-attachments/assets/d83e35d6-a1b1-4ebe-a560-3a1bf454e1a0" />

As per requirement:
> - If a user clicks on "Eraser" button and then click on a specific monomer, this monomer + all the bonds linked to it disappear from the canvas 

**Environment details:**
- Ketcher Version 3.7.0-rc.1 Build at 2025-08-22; 05:12:45
- Indigo 1.35.0-dev.1.27-g2b4a2091-wasm32-wasm-clang-19.0.0
- Chrome Version 139.0.7258.139 (Official Build) (64-bit)
- Win10

Found while testing: https://github.com/epam/ketcher/issues/7554
Related issue: https://github.com/epam/ketcher/issues/2370

Zhirnoff:
Tested. Bug fixed.

https://github.com/user-attachments/assets/6bb6d8a2-4563-423b-9a44-2ec9259c54c1

**Desktop:**
 - OS: Windows 11
 - Browser Chrome
 Version 140.0.7339.128 (Official Build) (64-bit)

**Ketcher version** 
[Version 3.9.0-rc.1]
**Indigo version** 
[Version 1.37.0-rc.1]

PR #7771, opened 2025-09-03
#7700 - System allow to delete monomers leaving connected bonds on the canvas

KukushkinDmi1riy:
Screenshot 2025-09-03 at 10 26 05

## Check list
- [ ] unit-tests written
- [ ] e2e-tests written
- [ ] documentation updated

### DIFF
--- a/packages/ketcher-core/src/application/editor/tools/Erase.ts
+++ b/packages/ketcher-core/src/application/editor/tools/Erase.ts
@@ -14,8 +14,6 @@
  * limitations under the License.
  ***************************************************************************/
 import { CoreEditor, EditorHistory } from 'application/editor/internal';
-import { Command } from 'domain/entities/Command';
-import { BaseMonomer } from 'domain/entities/BaseMonomer';
 import { BaseRenderer } from 'application/render/renderers/BaseRenderer';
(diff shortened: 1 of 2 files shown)


packages/ketcher-core/src/application/editor/tools/Erase.ts
```diff
@@ -46,10 +48,28 @@ class EraserTool implements BaseTool {
...
+          this.history.update(unselectCommand);
+          this.editor.renderersContainer.update(unselectCommand);
+        }
+      }
+
       const modelChanges =
-        this.editor.drawingEntitiesManager.deleteDrawingEntity(
-          selectedItemRenderer.drawingEntity,
-        );
+        this.editor.drawingEntitiesManager.deleteDrawingEntity(clickedEntity);
```
rrodionov91: Can we keep it as backup solution and think about the way of fixing that without unselecting it explicitly?

rrodionov91 (changes requested):

packages/ketcher-core/src/application/editor/tools/Erase.ts
KukushkinDmi1riy: You right! Fixed it with `deleteMonomer` method

Issue #7770, opened 2025-09-02
Pressing Right Scroll Arrow button of top menu cause exception to console


AlexeyGirin:
**Steps to Reproduce**
1. Open Ketcher in Popup mode
2. Toggle to Macro - Sequence mode
3. Press Right Scroll Arrow button for top menu
<img width="838" height="725" alt="Image" src="https://github.com/user-attachments/assets/506a3043-b723-4b60-88e1-3356f86654a9" />

**Actual behavior**
System throws exception to log: `Layout.tsx:157 Uncaught TypeError: Cannot read properties of undefined (reading 'stopPropagation')`
<img width="530" height="169" alt="Image" src="https://github.com/user-attachments/assets/dfbcf62f-cb8e-46b2-9430-b9f50cccb652" />

**Expected behavior**
No exception

**Versions**
- Ketcher Version 3.8.0-dev.1 Build at 2025-09-01; 15:10:48
- Indigo Version 1.37.0-dev.1.0-ged340212-wasm32-wasm-clang-19.0.0
- Chrome Version 139.0.7258.155 (Official Build) (64-bit)
- Win10

Related issue: https://github.com/epam/ketcher/issues/6270

PR #7847, opened 2025-09-09
#7770 - Pressing Right Scroll Arrow button of top menu cause exception to console

KukushkinDmi1riy:
issue >> https://github.com/epam/ketcher/issues/7770

- Split top bar scroll logic in Layout into no-arg functions and pass to ArrowScroll
- ArrowScroll: stop event propagation in onClick handlers; type callbacks as () => void
- Fixes runtime error: \"Cannot read properties of undefined (reading 'stopPropagation')\" when holding the right arrow (useInterval calls without event)

## Check list
- [ ] unit-tests written
- [ ] e2e-tests written
- [ ] documentation updated

### DIFF
--- a/packages/ketcher-react/src/hooks/useRaf.ts
+++ b/packages/ketcher-react/src/hooks/useRaf.ts
@@ -14,9 +14,28 @@
  * limitations under the License.
  ***************************************************************************/
 
-export * from './useSettingsContext';
-export * from './useResizeObserver';
-export * from './useFormContext';
-export * from './useAppContext';
-export * from './useRaf';
-export * from './useSubscribtionOnEvents';
+import { useEffect, useRef } from 'react';
+
+export const useRaf = (active: boolean, onFrame: (dtMs: number) => void) => {
+  const callbackRef = useRef(onFrame);
+  useEffect(() => {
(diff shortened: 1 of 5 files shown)


packages/ketcher-react/src/hooks/useRaf.ts
```diff
@@ -0,0 +1,41 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ ***************************************************************************/
+
+import { useEffect, useRef } from 'react';
+
+export const useRaf = (active: boolean, onFrame: (dtMs: number) => void) => {
```
rrodionov91: Let's not use shortens here. Just useRequestAnimationFrame, requestAnimationFrameId and deltaTimeInMilliseconds would be better because nobody will be lost after seeing it first time without additional context

rrodionov91 (changes requested):

rrodionov91 (approved):

Issue #7932, opened 2025-09-25
Add warning messages to the monomer creation wizard in case of unusual leaving group atoms


ljubica-milovic:
**Background**

There are some conventions in cheminformatics - for example R1 of a sugar is a hydrogen and R2 of a phosphate is a hydroxyl. Some chemists may not be aware of that convention.

This functionality is needed to warn/inform the users of a potential mistake they are performing.

**Requirements**

1. If the user is making an `Amino acid` monomer in the monomer creation wizard, and that monomer has:
- R1 attachment point with a leaving group ≠ H or
- R2 attachment point with a leaving group ≠ OH
after clicking on `Submit` in the monomer creation wizard, they should get a warning message in a modal window: "Amino acid monomers typically have a hydrogen as the leaving group for R1, and a hydroxyl as a leaving group for R2. Do you wish to proceed with the current attachment points?" with options "Cancel" (default) and "Yes" and a title "Non-typical attachment points".

2. If the user is making an `Sugar` monomer in the monomer creation wizard, and that monomer has:
- R1 attachment point with a leaving group ≠ H or
- R2 attachment point with a leaving group ≠ H or
- R3 attachment point with a leaving group ≠ OH
after clicking on `Submit` in the monomer creation wizard, they should get a warning message in a modal window: "Sugar monomers typically have a hydrogen as the leaving group for R1 and R2, and a hydroxyl as a leaving group for R3. Do you wish to proceed with the current attachment points?" with options "Cancel" (default) and "Yes" and a title "Non-typical attachment points".

3. If the user is making an `Base` monomer in the monomer creation wizard, and that monomer has:
- R1 attachment point with a leaving group ≠ OH
after clicking on `Submit` in the monomer creation wizard, they should get a warning message in a modal window: "Base monomers typically have a hydroxyl as the leaving group for R1. Do you wish to proceed with the current attachment points?" with options "Cancel" (default) and "Yes" and a title "Non-typical attachment points".

4. If the user is making an `Phosphate` monomer in the monomer creation wizard, and that monomer has:
- R1 attachment point with a leaving group ≠ OH
- R2 attachment point with a leaving group ≠ OH
after clicking on `Submit` in the monomer creation wizard, they should get a warning message in a modal window: "Phosphate monomers typically have a hydroxyl as the leaving group for R1 and R2. Do you wish to proceed with the current attachment points?" with options "Cancel" (default) and "Yes" and a title "Non-typical attachment points".

[...]

PR #8038, opened 2025-10-07
#7932 - Add warning messages to the monomer creation wizard in case of unusual leaving group atoms

KristinaKasyanovskaya:
## Check list
- [ ] unit-tests written
- [ ] e2e-tests written
- [ ] documentation updated


packages/ketcher-react/src/script/ui/views/components/MonomerCreationWizard/MonomerCreationWizard.tsx
```diff
@@ -329,6 +330,9 @@ const MonomerCreationWizard = () => {
   const { values, notifications, errors } = wizardState;
   const { type, symbol, name, naturalAnalogue } = values;
   const [modificationTypes, setModificationTypes] = useState<string[]>([]);
+  const [showLeavingGroupDialog, setShowLeavingGroupDialog] = useState(false);
+  const [leavingGroupDialogMessage, setLeavingGroupDialogMessage] =
```
svvald, line 350: I suppose having two states here is unnecessary, perhaps you could just rely on non-empty message as a flag to display a dialog?

packages/ketcher-react/src/script/ui/views/components/MonomerCreationWizard/MonomerCreationWizard.tsx
```diff
@@ -472,7 +476,7 @@ const MonomerCreationWizard = () => {
 
   const { assignedAttachmentPoints } = monomerCreationState;
 
-  const handleSubmit = () => {
+  const handleSubmit = (skipLeavingGroupWarning = false) => {
```
svvald: I don't see you pass this parameter as `true` anywhere, shall we keep it in function definition then?

packages/ketcher-react/src/script/ui/views/components/MonomerCreationWizard/MonomerCreationWizard.tsx
```diff
@@ -758,13 +779,46 @@ const MonomerCreationWizard = () => {
           </button>
           <button
             className={styles.buttonSubmit}
-            onClick={handleSubmit}
+            onClick={() => handleSubmit()}
```
svvald: I suppose this syntax is not needed here as you don't pass any additional params to function call

packages/ketcher-react/src/script/ui/views/components/MonomerCreationWizard/MonomerValidationRules.ts
```diff
@@ -0,0 +1,75 @@
...
+      },
+    ],
+    warningMessage:
+      'Phosphate monomers typically have a hydroxyl as the leaving group for R1 and R2. Do you wish to proceed with the current attachment points?" with options "Cancel" (default) and "Yes" and a title "Non-typical attachment points.',
+  },
+  {
+    monomerType: KetMonomerClass.RNA,
+    requirements: [
+      { attachmentPoint: 'R1', expectedLeavingGroup: 'H' },
+      { attachmentPoint: 'R2', expectedLeavingGroup: 'O' },
```
svvald: It's a nitpick but IIRC we have some enums/maps for both attachment point names and atoms names/leaving atoms options.
Perhaps it would be nice to use them here just as a good practice to avoid literals when possible

packages/ketcher-core/src/domain/entities/struct.ts
```diff
@@ -255,7 +255,8 @@ export class Struct {
     const bidMap = bidMapEntity ?? new Map();
 
     bondSet = bondSet.filter((bid) => {
-      const bond = this.bonds.get(bid)!;
+      const bond = this.bonds.get(bid);
+      if (!bond) return false;
```
svvald: Which case does it cover here? I suppose we should always have the queried bonds in current struct, if they're not there, perhaps something went wrong and at least we shouldn't just silently return

Issue #5570, opened 2024-09-24
PNG and SVG exported images should be identical to how the canvas looks like in case of Chiral label


AlexeyGirin:
**Steps to Reproduce**
1. Toggle to Molecules mode
2. Go to Settings - Stereochemistry tab
3. Switch off Show the Stereo flags (click Apply)
![image](https://github.com/user-attachments/assets/56d01963-ea5b-4159-8369-4cfc64c7f3cb)
4. Load from file: [PNG and SVG exported images should be identical to how the canvas looks like in case of Chiral label.zip](https://github.com/user-attachments/files/17116718/PNG.and.SVG.exported.images.should.be.identical.to.how.the.canvas.looks.like.in.case.of.Chiral.label.zip)
![image](https://github.com/user-attachments/assets/a4dd8251-4e82-4ba2-a3e4-8e839da5090d)
(no chiral flag on the canvas)
6. Save canvas to PNG/SVG

**Actual behavior**
Chiral flag present in export result
![image](https://github.com/user-attachments/assets/7bd50023-11fb-485b-85a0-4983eb07516d)

**Expected behavior**
No chiral flag
![image](https://github.com/user-attachments/assets/3ecc1da3-f2dc-439a-9f5f-9f9b4bbcba22)

**Versions**
- Ketcher Version 2.25.0-rc.4 Build at 2024-09-23; 17:24:38
- Indigo Toolkit Version 1.24.0-rc.3.0-g692080c62-wasm32-wasm-clang-19.0.0
- Chrome Version 129.0.6668.59 (Official Build) (64-bit)
- Win10

rrodionov91:
Related conversation on github https://github.com/epam/ketcher/discussions/4655

Zhirnoff:
Preliminary tested.

<img width="1920" height="945" alt="Image" src="https://github.com/user-attachments/assets/7486ad25-05d8-450c-884c-d6a9a8526062" />
<img width="1920" height="945" alt="Image" src="https://github.com/user-attachments/assets/aa875644-ac28-4e60-9bae-d4f1258225cf" />

**Desktop:**
 - OS: Windows 11
 - Browser Chrome
 Version 141.0.7390.108 (Official Build) (64-bit)

**Ketcher version** 
[Version master]
**Indigo version** 
[Version master]

PR #8266, opened 2025-10-16
 #5570 - PNG and SVG exported images should be identical to how the canvas looks like in case of Chiral label 

KristinaKasyanovskaya:

Issue => https://github.com/epam/ketcher/discussions/4655 
Issue => #5570 

Снимок экрана 2025-10-16 в 09 50 00

## Check list
- [ ] unit-tests written
- [ ] e2e-tests written
- [ ] documentation updated

### DIFF
--- a/packages/ketcher-react/src/script/ui/state/options/index.js
+++ b/packages/ketcher-react/src/script/ui/state/options/index.js
@@ -81,11 +81,14 @@ export const initOptionsState = {
 };
 
 function getSerilizedServerOptions(options) {
-  const renderStereoStyle = !options.showStereoFlags
-    ? 'none'
-    : options.ignoreChiralFlag
-    ? 'ext'
-    : 'old';
+  let renderStereoStyle;
+  if (!options.showStereoFlags) {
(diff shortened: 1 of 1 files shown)


packages/ketcher-react/src/script/ui/state/options/index.js
```diff
@@ -81,6 +81,12 @@ export const initOptionsState = {
 };
 
 function getSerilizedServerOptions(options) {
+  const renderStereoStyle = !options.showStereoFlags
+    ? 'none'
+    : options.ignoreChiralFlag
+    ? 'ext'
+    : 'old';
```
svvald: We're currently working on refactoring these nested ternary operators. Could you please rewrite it to the if-else statements sequence?

svvald (changes requested):

packages/ketcher-react/src/script/ui/state/options/index.js
KristinaKasyanovskaya: Fixed

Issue #8200, opened 2025-10-13
Add parameter to Editor to disable saving new monomers to local storage


svvald:
**Background**
After creation of new monomer it stores in browser local storage. To support the use case where ketcher consumer application manages the library by itself and receives new monomers through the event(https://github.com/epam/ketcher/issues/8281), we need to make parameter to disable this functionality. 

**Solution**
By default the functionality of saving new monomers from creation wizard in local storage should be enabled.
Add new parameter to Editor component which when can be used to disable it.

**Additional details**
Previously, `updateMonomerLibrary` function of Ketcher API was saving the updates into browser local storage. In scope of https://github.com/epam/ketcher/issues/8067 this functionality was removed.
However, this API function was also used by monomer creation wizard.

PR #8364, opened 2025-10-27
#8200 - Add parameter to Editor to disable saving new monomers to local storage  

KristinaKasyanovskaya:

## Check list
- [ ] unit-tests written
- [ ] e2e-tests written
- [ ] documentation updated

### DIFF
--- a/packages/ketcher-core/src/application/ketcher.ts
+++ b/packages/ketcher-core/src/application/ketcher.ts
@@ -41,6 +41,7 @@ import {
   SettingsManager,
   getSvgFromDrawnStructures,
   KetcherLogger,
+  ensureString,
 } from 'utilities';
 import {
   deleteAllEntitiesOnCanvas,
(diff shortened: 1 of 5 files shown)


packages/ketcher-core/src/application/ketcher.ts
```diff
@@ -714,6 +719,13 @@ export class Ketcher {
     );
 
     editor.updateMonomersLibrary(dataInKetFormat);
+    if (SettingsManager.persistMonomerLibraryUpdates) {
+      const updateString =
+        typeof dataInKetFormat === 'string'
+          ? dataInKetFormat
+          : JSON.stringify(dataInKetFormat);
```
svvald: This check is used in a few places in code right now, specifically it's used in `ensureMonomersLibraryDataInKetFormat` function
Maybe we could use or create some utility function for this?

packages/ketcher-core/src/application/ketcher.ts
```diff
@@ -714,6 +719,13 @@ export class Ketcher {
     );
 
     editor.updateMonomersLibrary(dataInKetFormat);
+    if (SettingsManager.persistMonomerLibraryUpdates) {
```
svvald: This check is not enough unfortunately, currently it will lead to each `updateMonomersLibrary` call to store data in local storage by default.

However, this function might be called in two cases – as an API call or by monomer creation wizard

We want to avoid storing this data always when it is an API call and have it optionally stored when using monomer creation wizard

Perhaps you can just pass additional parameter to params object when this function is invoked from monomer creation wizard and use this parameter in this check

Issue #8281, opened 2025-10-16
Adjust subscription to library update event of Ketcher API


svvald:
**Background**
Currently when subscribed to library update event (`libraryUpdate`) using Ketcher API, when a new monomer is added, the whole library is pushed to the subscribers as per original requirements (#6552). However, it makes it quite hard to distinguish the monomers that have been added/updated.

**Solution**
Change `libraryUpdate` event subscription so it only pushes the actual update to the library in sdf format.
Event subscription should work using existing [subscribe method](https://github.com/epam/ketcher?tab=readme-ov-file#events).
Add documentation similar to [change event](https://github.com/epam/ketcher?tab=readme-ov-file#change-event).

**Example**
```
ketcher.editor.subscribe('libraryUpdate', (eventData) => {
    console.log('Created monomer in sdf:', eventData);
});
```

AlexeyGirin:
Verified.
- Ketcher Version 3.10.0-rc.1 Build at 2025-11-03; 14:05:05
- Indigo Version 1.38.0-rc.1.0-g155cc4da8-wasm32-wasm-clang-19.0.0
- Chrome Version 142.0.7444.60 (Official Build) (64-bit)
- Win10

PR #8382, opened 2025-10-30
#8281 - Adjust subscription to library update event of Ketcher API 

KristinaKasyanovskaya:
Indigo ticket https://github.com/epam/Indigo/issues/3272 

## Check list
- [ ] unit-tests written
- [ ] e2e-tests written
- [ ] documentation updated

### DIFF
--- a/packages/ketcher-core/src/application/formatters/formatProperties.ts
+++ b/packages/ketcher-core/src/application/formatters/formatProperties.ts
@@ -183,17 +183,6 @@ const formatProperties: FormatPropertiesMap = {
     true,
     { 'molfile-saving-mode': '3000' },
   ),
-  'monomer-library-sdf': new SupportedFormatProperties(
-    'Monomer Library SDF',
-    ChemicalMimeType.MonomerLibrary,
-    ['.sdf'],
-    true,
-    {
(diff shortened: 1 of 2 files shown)


packages/ketcher-core/src/application/formatters/structFormatter.types.ts
```diff
@@ -58,6 +58,7 @@ export enum SupportedFormat {
   unknown = 'unknown',
   rdf = 'rdf',
   rdfV3000 = 'rdfV3000',
+  monomerLibrarySdf = 'monomer-library-sdf',
```
svvald: I don't see this value is used anywhere, perhaps you can remove it

packages/ketcher-core/src/application/formatters/formatProperties.ts
```diff
@@ -183,6 +183,17 @@ const formatProperties: FormatPropertiesMap = {
     true,
     { 'molfile-saving-mode': '3000' },
   ),
+  'monomer-library-sdf': new SupportedFormatProperties(
```
svvald: I'm not sure this value is needed here either, from what I can see it is mostly used for UI configuring

You're making a request internally and specify these parameters in the call code, so I assume it can be safely removed but please check it

Issue #4303, opened 2024-03-21
Macro: When editing sequences over 2000 monomers, slight slowdowns occur


Zhirnoff:
**Steps to Reproduce**
1. Switch to macro
2. Add sequence more than 2000 monomers
3. Switch to Sequence layout
4. Click right mouse on sequence and select 'edit sequence'
5. Add any monomers by typing on keyboard

**Actual behavior**
When editing sequences over 2000 monomers, slight slowdowns occur. If you quickly enter letters, this becomes very noticeable because you stopped typing and letters still appear in the grid

**Expected behavior**
When editing sequences over 2000 monomers, slight slowdowns not occur. Even if the user quickly enters letters, they are entered instantly and when he stops entering, new letters do not appear.

**Screenshots**

https://github.com/epam/ketcher/assets/62464241/5fe8ff1d-4e18-49b5-9665-6ee16b3114fc

**Desktop (please complete the following information):**
 - OS: Windows 11
 - Browser Chrome
 - Version 112.0.5615.138 (Official Build) (64-bit)

**Ketcher version** 
[Version 2.20.0-rc.1]

iRadionova:
Reproduced:

https://github.com/user-attachments/assets/d79b5994-8f3a-426f-8383-38b490ebc248

<img width="384" height="281" alt="Image" src="https://github.com/user-attachments/assets/d857675e-57c7-4409-956d-f1b60cbc39c0" />

PR #8938, opened 2025-12-18
Improve sequence mode performance on large (500+) chains

AnatoliiSazanov:
- cache allocations in BaseMonomer
- remove array allocation in Command merge
- cache arrays in Node classes
- memoize properties in Chain
- delegate event handling to SequenceRenderer

Initial 
500 symbols paste - 800ms indigo, 450ms render
1000 symbols paste - 800ms indigo,  900ms render
print symbol in 500 seq - avg 150ms

With memoizations/cache
500 symbols paste - 800ms indigo, 400ms render
1000 symbols paste - 800ms indigo,  700ms render
print symbol in 500 seq - avg 110ms

Summary:
25-30% improvements of import and type operations, better responsiveness of mouse events on large sequences.

Note: this still doesnt solve #4303 fully, element pool is going to be in next PR to ease refactoring and testing

### DIFF
--- a/packages/ketcher-core/src/application/render/renderers/sequence/SequenceEventDelegationManager.ts
+++ b/packages/ketcher-core/src/application/render/renderers/sequence/SequenceEventDelegationManager.ts
@@ -65,7 +65,7 @@ export class SequenceEventDelegationManager {
     if (!target || !(target instanceof SVGElement)) return null;
 
     const sequenceItemElement = (target as SVGElement).closest(
-      '[data-testid="sequence-item"]',
+      '.sequence-item',
     ) as SVGGElement;
 
     if (!sequenceItemElement) return null;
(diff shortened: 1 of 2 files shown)


packages/ketcher-core/src/application/render/renderers/sequence/SequenceEventDelegationManager.ts
```diff
@@ -0,0 +1,182 @@
...
+    this.canvas?.on(eventType, handler);
+  }
+
+  private findSequenceItemRenderer(
+    target: EventTarget | null,
+  ): { renderer: BaseSequenceItemRenderer; elementType: ElementType } | null {
+    if (!target || !(target instanceof SVGElement)) return null;
+
+    const sequenceItemElement = (target as SVGElement).closest(
+      '[data-testid="sequence-item"]',
```
rrodionov91: I think that we should not use data-testid in application code. This attribute should be used only in tests

AnatoliiSazanov: done

Issue #9193, opened 2026-02-16
Refactor:(TypeScript) Objects and classes converted or coerced to strings should define a "toString()" method #9183


guliaisaeva:
Problem:
Objects and classes converted or coerced to strings should define a toString() method

Why is this an issue?

When an object without a custom toString() implementation is converted to a string, JavaScript uses Object.prototype.toString() and returns "[object Object]". This is usually not meaningful and may indicate unintended behavior. Sonar flags these cases because string coercion should produce useful output.

Problem locations:
packages/ketcher-core/src/application/render/renderers/sequence/SequenceRenderer.ts

How to fix it?
You can simply define a toString() method for the object or class.
Noncompliant code example (class):
``
class Foo {};
const foo = new Foo();

foo + ''; // Noncompliant - evaluates to "[object Object]"
`Foo: ${foo}`; // Noncompliant - evaluates to "Foo: [object Object]"
foo.toString(); // Noncompliant - evaluates to "[object Object]"

const foo = {};
foo + ''; // Noncompliant - evaluates to "[object Object]"
`Foo: ${foo}`; // Noncompliant - evaluates to "Foo: [object Object]"
foo.toString(); // Noncompliant - evaluates to "[object Object]"
``
Compliant solution
``
class Foo {
  toString() {
    return 'Foo';
  }
}
const foo = new Foo();

foo + '';
`Foo: ${foo}`;
foo.toString();

const foo = {
  toString: () => {
    return 'Foo';
  }
}
foo + '';
`Foo: ${foo}`;
foo.toString();
``

PR #9196, opened 2026-02-16
#9193 - Refactor:(TypeScript) Objects and classes converted or coerced t…

guliaisaeva:
Problem:
Objects and classes converted or coerced to strings should define a toString() method

Why is this an issue?

When an object without a custom toString() implementation is converted to a string, JavaScript uses Object.prototype.toString() and returns "[object Object]". This is usually not meaningful and may indicate unintended behavior. Sonar flags these cases because string coercion should produce useful output.

How to fix it?
You can simply define a toString() method for the object or class.
Noncompliant code example (class):
```
class Foo {};
const foo = new Foo();

foo + ''; // Noncompliant - evaluates to "[object Object]"
Foo: ${foo}; // Noncompliant - evaluates to "Foo: [object Object]"
foo.toString(); // Noncompliant - evaluates to "[object Object]"

const foo = {};
foo + ''; // Noncompliant - evaluates to "[object Object]"
Foo: ${foo}; // Noncompliant - evaluates to "Foo: [object Object]"
foo.toString(); // Noncompliant - evaluates to "[object Object]"

```
Compliant solution

```
class Foo {
toString() {
return 'Foo';
}
}
const foo = new Foo();

foo + '';
Foo: ${foo};
foo.toString();

const foo = {
toString: () => {
return 'Foo';
}
}
foo + '';
Foo: ${foo};
foo.toString();
```
Problem locations:
packages/ketcher-core/src/application/render/renderers/sequence/SequenceRenderer.ts

### DIFF
--- a/packages/ketcher-core/src/application/render/renderers/sequence/SequenceRenderer.ts
+++ b/packages/ketcher-core/src/application/render/renderers/sequence/SequenceRenderer.ts
@@ -9,7 +9,6 @@ import {
   MonomerToAtomBond,
   Nucleotide,
   Phosphate,
-  Pool,
   Sugar,
   Vec2,
 } from 'domain/entities';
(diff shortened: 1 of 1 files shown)


packages/ketcher-core/src/application/render/renderers/sequence/SequenceRenderer.ts
```diff
@@ -1265,6 +1265,17 @@ export class SequenceRenderer {
   }
 }
 
+const getCustomToString = (value: object): string | undefined => {
+  const toStringCandidate = (value as { toString?: () => string }).toString;
```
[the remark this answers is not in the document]
guliaisaeva: fixed ✅

packages/ketcher-core/src/application/render/renderers/sequence/SequenceRenderer.ts
```diff
@@ -1265,6 +1265,17 @@ export class SequenceRenderer {
...
+  if (
+    typeof toStringCandidate === 'function' &&
+    toStringCandidate !== Object.prototype.toString
+  ) {
+    return toStringCandidate.call(value);
+  }
+  return undefined;
+};
+
 export function sequenceReplacer(key: string, value: unknown): unknown {
```
svvald: I'm not sure what's the purpose of this function to be honest, I do not see it being used anywhere in the code, perhaps it's better to just remove this function completely so we won't need to create a custom to string conversion function

guliaisaeva: done ✅

Issue #9202, opened 2026-02-17
Refactor: Constructors should not return values


azamatalt:
**Problem:** 
Constructors should not return values

**Why is this an issue?**
JavaScript allows returning a value from a class constructor. This obscure feature is rarely used and is more likely a bug than the developer’s intention.

Do not return anything from the constructor.

Noncompliant code example
```
class TextMessage {
    constructor(msg) {
        this.text = msg;
        return msg; // Noncompliant
    }
}
```
Compliant solution
```
class TextMsg1 {
    constructor(msg) {
        this.text = msg;
    }
}


class TextMsg2 {
    constructor(msg) {
        if (!msg) {
            return; // ok to return nothing for flow control
        }
        this.text = msg;
    }
}
```

**Problem locations:**
packages/ketcher-core/src/application/editor/EditorHistory.ts:33
packages/ketcher-core/src/application/editor/EditorHistory.ts:39

PR #9213, opened 2026-02-17
#9202 - Refactor: Constructors should not return values

azamatalt:
### Motivation
- Prevent constructors from returning values by removing the previous `return`-from-constructor pattern in `EditorHistory` and provide a clear singleton accessor instead.
- Ensure TypeScript typing and control-flow are safe when obtaining the shared history instance.

### Description
- Made `EditorHistory` constructor `private`, added `private static _instance: EditorHistory | null = null`, and implemented `static getInstance(editor: CoreEditor): EditorHistory` to lazily create/return the singleton instance.
- Replaced all `new EditorHistory(...)` call sites with `EditorHistory.getInstance(...)` across `ketcher-core` and `ketcher-macromolecules`, including code paths in `Editor`, modes, tools, utilities, and the Open modal.
- Adjusted `ClearTool` to use a local `const history = EditorHistory.getInstance(editor)` after the early-return guard to avoid undefined-flow typing issues.
- Updated unit tests to use `EditorHistory.getInstance(editor)` and kept `destroy()` to reset the singleton (`_instance = null`).

### Testing
- Ran `npm run ajv -w packages/ketcher-core` to generate compiled schema and it succeeded.
- Ran `npm run test:types -w packages/ketcher-core` (which invokes `tsc --noEmit`) and it succeeded after the schema generation.
- Ran `npm run test:unit -w packages/ketcher-core -- EditorHistory.test.ts SelectBase.test.ts` and both test suites passed.

------
[Codex Task](https://chatgpt.com/codex/tasks/task_e_69942ec7d7048332882996b3d689ba48)

### DIFF
--- a/packages/ketcher-core/src/application/editor/EditorHistory.ts
+++ b/packages/ketcher-core/src/application/editor/EditorHistory.ts
@@ -27,8 +27,7 @@ export class EditorHistory {
   historyPointer = 0;
   editor!: CoreEditor;
 
-  // eslint-disable-next-line no-use-before-define
-  private static _instance: EditorHistory | null = null;
+  private static _instance: object | null = null;
 
   private constructor(editor: CoreEditor) {
     this.editor = editor;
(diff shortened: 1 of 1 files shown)


packages/ketcher-core/src/application/editor/EditorHistory.ts
```diff
@@ -27,16 +27,20 @@ export class EditorHistory {
   historyPointer = 0;
   editor!: CoreEditor;
 
-  private static _instance;
-  constructor(editor: CoreEditor) {
-    if (EditorHistory._instance) {
-      return EditorHistory._instance;
-    }
+  // eslint-disable-next-line no-use-before-define
```
svvald: I know that we have a lot of eslint suppression in our code already, but is there a way to avoiding adding one more here as we're trying to reduce the amount of code smells?

azamatalt: fixed

Issue #9207, opened 2026-02-17
Refactor:(TypeScript) Unused React typed props should be removed


guliaisaeva:
Problem:
Unused React typed props should be removed

Why is this an issue?
Leaving unused props in a React component can make the code harder to understand and maintain. Other developers may wonder why certain props are passed to a component if they are not used. Unused props can also increase the size of the component’s memory footprint and impact performance. This is especially true if the unused props are large objects or arrays. Furthermore, if a prop is unused, it may indicate that the developer did not complete the implementation as he intended initially or made a mistake while writing the component.

To avoid these issues, you should remove any unused props from React components. This helps keep the codebase clean, improves performance, and enhances code readability.

How can I fix it in PropTypes?
Noncompliant code example

```
 import PropTypes from 'prop-types';
import React from 'react';

class Hello extends React.Component {
  render() {
    return (
      <h1>Hello</h1>
    );
  }
}

Hello.propTypes = {
  name: PropTypes.string
}; 
```

Compliant solution

```
 import PropTypes from 'prop-types';
import React from 'react';

class Hello extends React.Component {
  render() {
    return (
      <h1>Hello {this.props.name}</h1>
    );
  }
}
Hello.propTypes = {
  name: PropTypes.string
}; 
```

How can I fix it in TypeScript?
Noncompliant code example

```
 import React from 'react';

type Props = {
  name: string;
}

class Hello extends React.Component<Props> {
  render() {
    return <div>Hello</div>;
  }
}  
```

Compliant solution

```  import React from 'react';
type Props = {
  name: string;
};

class Hello extends React.Component<Props> {
  render() {
    return <div>Hello {this.props.name}</div>;
  }
}
```
Problem locations:
packages/ketcher-react/src/script/ui/views/modal/components/ExtendedTable/ExtendedTable.tsx
packages/ketcher-react/src/script/ui/views/modal/components/ExtendedTable/ExtendedTable.tsx
packages/ketcher-react/src/script/ui/views/modal/components/PeriodTable/PeriodTable.tsx

PR #9214, opened 2026-02-17
#9207- Refactor: Unused React typed props should be removed

guliaisaeva:
Problem:
Unused React typed props should be removed

Why is this an issue?
Leaving unused props in a React component can make the code harder to understand and maintain. Other developers may wonder why certain props are passed to a component if they are not used. Unused props can also increase the size of the component’s memory footprint and impact performance. This is especially true if the unused props are large objects or arrays. Furthermore, if a prop is unused, it may indicate that the developer did not complete the implementation as he intended initially or made a mistake while writing the component.

To avoid these issues, you should remove any unused props from React components. This helps keep the codebase clean, improves performance, and enhances code readability.

How can I fix it in PropTypes?
Noncompliant code example

```
 import PropTypes from 'prop-types';
import React from 'react';

class Hello extends React.Component {
  render() {
    return (
      <h1>Hello</h1>
    );
  }
}

Hello.propTypes = {
  name: PropTypes.string
}; 
```

Compliant solution

```
 import PropTypes from 'prop-types';
import React from 'react';

class Hello extends React.Component {
  render() {
    return (
      <h1>Hello {this.props.name}</h1>
    );
  }
}
Hello.propTypes = {
  name: PropTypes.string
}; 
```

How can I fix it in TypeScript?
Noncompliant code example

```
 import React from 'react';

type Props = {
  name: string;
}

class Hello extends React.Component<Props> {
  render() {
    return <div>Hello</div>;
  }
}  
```

Compliant solution

```  import React from 'react';
type Props = {
  name: string;
};

class Hello extends React.Component<Props> {
  render() {
    return <div>Hello {this.props.name}</div>;
  }
}
```
Problem locations:
packages/ketcher-react/src/script/ui/views/modal/components/ExtendedTable/ExtendedTable.tsx
packages/ketcher-react/src/script/ui/views/modal/components/ExtendedTable/ExtendedTable.tsx
packages/ketcher-react/src/script/ui/views/modal/components/PeriodTable/PeriodTable.tsx

### DIFF
--- a/packages/ketcher-react/src/script/ui/views/modal/components/ExtendedTable/ExtendedTable.tsx
+++ b/packages/ketcher-react/src/script/ui/views/modal/components/ExtendedTable/ExtendedTable.tsx
@@ -40,10 +40,8 @@ interface TableProps {
 }
 
 const Table = (props: TableProps) => {
-  const { pseudo, label, disabledQueryElements, onOk, onCancel } = props;
-
   const [value, setValue] = useState<string | null>(
-    pseudo ? label ?? null : null,
+    props.pseudo ? props.label ?? null : null,
   );
 
   const selected = (label: string): boolean => value === label;
@@ -59,7 +57,7 @@ const Table = (props: TableProps) => {
     setValue(label);
 
     if (activateImmediately) {
-      onOk(result());
+      props.onOk(result());
     }
   };
 
@@ -68,7 +66,7 @@ const Table = (props: TableProps) => {
       title="Extended Table"
       withDivider
       className={classes.extendedTable}
-      params={{ onOk, onCancel }}
+      params={props}
       result={result}
       buttons={['Cancel', 'OK']}
       buttonsNameMap={{ OK: 'Add' }}
@@ -77,7 +75,7 @@ const Table = (props: TableProps) => {
       <GenericGroups
         selected={selected}
         onAtomSelect={onAtomSelect}
-        disabledQueryElements={disabledQueryElements ?? null}
+        disabledQueryElements={props.disabledQueryElements ?? null}
       ></GenericGroups>
     </Dialog>
   );


packages/ketcher-react/src/components/Dialog/Dialog.tsx
```diff
@@ -41,7 +41,7 @@ export interface DialogParams extends DialogParamsCallProps {
 
 interface DialogProps {
   title?: string;
-  params: DialogParams;
+  params?: DialogParams;
```
svvald, line 44: Is this change actually needed here?

packages/ketcher-react/src/components/Dialog/Dialog.tsx
```diff
@@ -83,13 +83,14 @@ export const Dialog: FC<PropsWithChildren & Props> = (props) => {
...
+    const current = dialogRef.current;
     if (focusable) {
-      (dialogRef.current as HTMLElement).focus();
+      (current as HTMLElement).focus();
     }
 
     return () => {
       (
-        dialogRef.current
+        current
```
svvald: Keeping `dialogRef.current` would be better as it bears more sense rather than just `current`

packages/ketcher-react/src/script/ui/views/modal/components/ExtendedTable/ExtendedTable.tsx
```diff
@@ -62,12 +61,19 @@ const Table = (props: TableProps) => {
     }
   };
 
+  // Cast props to satisfy the Dialog component's expectations
+  const dialogParams = props as unknown as {
+    onOk: (result: unknown) => void;
+    onCancel: () => void;
```
svvald: It's weird, why remove `onCancel` property from props type only to forcefully cast the type through `unknown` and add it back?

guliaisaeva: Removed ✅

packages/ketcher-react/src/components/Dialog/Dialog.tsx
guliaisaeva: I've renamed it  now:  ``  const dialogElement = dialogRef.current; `` When i use only `` dialogRef.current `` It gives warning.

Issue #9218, opened 2026-02-17
Refactor: Unused methods of React components should be removed


azamatalt:
**Problem:** 
Unused methods of React components should be removed

**Why is this an issue?**

When using React class components, all non-React lifecycle methods should be called within the scope of the component. If a method is only called from outside the class, consider using props to interact with the component and re-render if needed, as React encourages data-driven components.

```
class Profile extends React.Component {
  render(props) {
    return <h1>{ props.name }</h1>;
  }

  getDefaultName() { // Noncompliant: this method is never used and is a dead code
    return 'John Smith';
  }
}
```
To fix the issue, remove the dead code or call the method from within the component scope.

```
class Profile extends React.Component {
  render(props) {
    return <h1>{ props.name || getDefaultName() }</h1>;
  }

  getDefaultName() {
    return 'John Smith';
  }
}
```

**Problem locations:**
packages/ketcher-react/src/script/ui/views/modal/components/process/Analyse/Analyse.tsx:71
packages/ketcher-react/src/script/ui/views/modal/components/process/Analyse/Analyse.tsx:78

PR #9221, opened 2026-02-17
#9218 Refactor: Unused methods of React components should be removed

azamatalt:
### Motivation
- Remove dead/unused class component plumbing and lifecycle methods from the Analyse dialog to satisfy the rule that non-lifecycle class methods should not remain unused. 
- Simplify the component to a function component so initialization logic and input rendering live inside component scope.

### Description
- Convert `AnalyseDialog` from a React class component into a function component and replace `componentDidMount` with a `useEffect` that calls `onAnalyse` on mount. 
- Extract input rendering into a local `renderInputComponent` helper function and use it directly from the function component. 
- Remove the unused `ErrorsContext` class wiring and other class-only constructs so there are no unused class methods left in the file. 
- Affected file: `packages/ketcher-react/src/script/ui/views/modal/components/process/Analyse/Analyse.tsx`.

### Testing
- Attempted `npm run build:react`, which failed in this environment due to missing git tag metadata and unresolved workspace module setup unrelated to the code changes. 
- Attempted `npx eslint packages/ketcher-react/src/script/ui/views/modal/components/process/Analyse/Analyse.tsx`, which failed because the ESLint shared config `standard` is not available in this environment.

------
[Codex Task](https://chatgpt.com/codex/tasks/task_e_69944ea859c0833285ef83a489f697de)

### DIFF
--- a/packages/ketcher-react/src/script/ui/views/modal/components/process/Analyse/Analyse.tsx
+++ b/packages/ketcher-react/src/script/ui/views/modal/components/process/Analyse/Analyse.tsx
@@ -69,6 +69,32 @@ function roundOff(value: string | number, round: number): string {
 
 const selectOptions = getSelectOptionsFromSchema({ enum: range(0, 8) });
 
+const analyseItems: AnalyseItem[] = [
+  {
+    name: 'Chemical Formula',
+    key: 'gross',
+    withSelector: false,
+  },
+  {
+    name: 'Molecular Weight',
+    key: 'molecular-weight',
+    round: 'roundWeight',
+    withSelector: true,
+  },
+  {
+    name: 'Exact Mass',
+    key: 'monoisotopic-mass',
+    round: 'roundMass',
+    withSelector: true,
+  },
+  {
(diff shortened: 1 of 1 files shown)


packages/ketcher-react/src/script/ui/views/modal/components/process/Analyse/Analyse.tsx
```diff
@@ -70,138 +69,129 @@ function roundOff(value: string | number, round: number): string {
...
+  loading,
+  onAnalyse,
+  onChangeRound,
+  ...props
+}: Props) {
+  useEffect(() => {
+    onAnalyse();
+  }, [onAnalyse]);
+
+  const analyseItems: AnalyseItem[] = [
```
svvald: These items do not depend on any component props/state and could be put outside of the component

azamatalt: fixed

svvald (approved):

PR #9226, opened 2026-02-17
#9160 - Refactor: (TypeScript) Unused assignments should be removed

guliaisaeva:
Problem: (TypeScript) Unused assignments should be removed.

What is the potential impact?
Dead code
An unused variable or local function usually occurs because some logic is no longer required after a code change. In that case, such code becomes unused and never executed.

Also, if you are writing code for the front-end, every unused variable or function remaining in your codebase is just extra bytes you have to send over the wire to your users. Unused code bloats your codebase unnecessarily and impacts the performance of your application.

Wrong logic
It could happen that due to a bad copy-paste or autocompletion, the wrong variable is used, while the right one is only declared. In that case, the unused variable should be used instead of deleted from the codebase.

Memory leaks
Finally, unused functions can also cause memory leaks. For example, an unused function can create a closure over a variable that would otherwise be released to the garbage collector.
```
let theThing = null;
const replaceThing = function () {
  const originalThing = theThing;
  const unused = function () {
    if (originalThing) {
      console.log("hi");
    }
  };
  theThing = {
    longStr: new Array(1000000).join("*"),
    someMethod: function () {
      console.log(someMessage);
    },
  };
};
setInterval(replaceThing, 1000);
```
How fix it?
Usually, the fix for this issue is straightforward, you just need to remove the unused variable declaration, or its name from the declaration statement if it is declared along with other variables.

Noncompliant code example
```
function numberOfMinutes(hours) {
  var seconds = 0;   // seconds is never used
  return hours * 60;
}
```
Compliant solution

```
function numberOfMinutes(hours) {
  return hours * 60;
}
```
Noncompliant code example
When an array destructuring is used and some element of the array is never referenced, one might simply remove it from the destructuring.

```
const [_, params] = url.split(path);
```
Compliant solution
```
const [, params] = url.split(path);
```
Problem locations:
packages/ketcher-react/src/script/ui/views/components/StructEditor/StructEditor.jsx
packages/ketcher-react/src/script/ui/component/form/form/form.jsx

### DIFF
--- a/packages/ketcher-react/src/script/ui/component/form/form/form.jsx
+++ b/packages/ketcher-react/src/script/ui/component/form/form/form.jsx
@@ -27,7 +27,7 @@ import { connect } from 'react-redux';
 import { getSelectOptionsFromSchema } from '../../../utils';
 import { updateFormState } from '../../../state/modal/form';
 import { useFormContext } from '../../../../../hooks';
-import { cloneDeep } from 'lodash';
+import { cloneDeep, omit } from 'lodash';
 import { Icon, IconButton } from 'components';
 import { Tooltip } from '@mui/material';
 
@@ -418,11 +418,12 @@ function propSchema(schema, { customValid, serialize = {}, deserialize = {} }) {
     Object.entries(customValid).forEach(([formatName, formatValidator]) => {
       ajv.addFormat(formatName, formatValidator);
 
-      const rest = { ...schemaCopy.properties[formatName] };
-      delete rest.pattern;
-      delete rest.maxLength;
-      delete rest.enum;
-      delete rest.enumNames;
+      const rest = omit(schemaCopy.properties[formatName], [
+        'pattern',
+        'maxLength',
+        'enum',
+        'enumNames',
+      ]);
 
       schemaCopy.properties[formatName] = {
         ...rest,
--- a/packages/ketcher-react/src/script/ui/views/components/StructEditor/StructEditor.jsx
+++ b/packages/ketcher-react/src/script/ui/views/components/StructEditor/StructEditor.jsx
@@ -21,6 +21,7 @@ import { LoadingCircles } from '../Spinner/LoadingCircles';
 import classes from './StructEditor.module.less';
 import clsx from 'clsx';
 import { upperFirst } from 'lodash/fp';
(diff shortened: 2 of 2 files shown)


packages/ketcher-react/src/script/ui/component/form/form/form.jsx
```diff
@@ -46,7 +46,6 @@ class Form extends Component {
     }
     this.updateState = this.updateState.bind(this);
 
-    // Cache the context value to avoid creating new objects on every render
```
svvald, line 49: Is this the only change in this pull request? If so, I'd close it

guliaisaeva: Hi! I had a problem after changing 2 files i deleted unused assignments,but it gives me error,after pushing into github by testing ,so i reverted it.

guliaisaeva: @svvald could you please re=review it?!🙌

packages/ketcher-react/src/script/ui/views/components/StructEditor/StructEditor.jsx
```diff
@@ -288,44 +288,46 @@ class StructEditor extends Component {
...
+      'onCipChange',
+      'onConfirm',
+      'onShowInfo',
+      'onApiSettings',
+      'showAttachmentPoints',
+      'onUpdateFloatingTools',
+      'onShowMacromoleculesErrorMessage',
+      'serverSettings',
+    ].forEach((propName) => {
+      delete props[propName];
```
svvald: This approach looks slightly verbose, I'd suggest to use `omit` function from `lodash` package which is already installed in our app, something like the following:
```
const {
  Tag = 'div',
  className,
  indigoVerification,
  ...props
} = omit(this.props, [ ...list of all keys to omit]);
```

And the same approach could be used for `form.jsx` file changes as well

guliaisaeva: @svvald HI! I've used  suggested approach with loadash's ` omit` .Please could you take  a look one more time ?!

svvald (approved):

PR #9273, opened 2026-02-19
Fixed context menu

GAsdeX:
**The Problem**: Ketcher was attaching a contextmenu event listener (likely to a global object) that was calling event.preventDefault() unconditionally. This resulted in two issues:

**Scope Creep**: Right-clicking anywhere in the host app (outside Ketcher) was blocked.

**Persistence Bug**: Because the listener wasn't being properly scoped or removed, the context menu remained blocked even after Ketcher was unmounted from the DOM.

**The Fix**:
The `setupContextMenuEvents` has been updated to include a guard clause. The handler now verifies that the event target is contained within this.ketcherRootElement.

**Impact**:

If the user right-clicks outside of Ketcher, the event is ignored by this handler, allowing the global context menu to work.

If Ketcher is unmounted, this.ketcherRootElement will be null/absent, preventing the "zombie" listener from blocking clicks on the rest of the application.

### DIFF
--- a/packages/ketcher-core/src/application/editor/Editor.ts
+++ b/packages/ketcher-core/src/application/editor/Editor.ts
@@ -129,9 +129,6 @@ const debouncedTurnOffScrollAnimation = debounce(
   SCROLL_SMOOTHNESS_IM_MS,
 );
 
-interface KetcherTarget extends EventTarget {
-  parent: HTMLElement | SVGElement | null; // Define what parent actually is
-}
 interface ICoreEditorConstructorParams {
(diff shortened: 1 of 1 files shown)


GAsdeX:
> Hi @GAsdeX Can you please tell us what problem does your PR resolve?

Hi @AlexeyGirin nice to meet you. Description has been provided. Please let me know if you need anything else.

AlexeyGirin:
Resolved here: https://github.com/epam/ketcher/pull/9672

PR #9277, opened 2026-02-19
#9276- Refactor: Literals should not be used for promise rejection

### LINKED ISSUE #9276: Refactor: (TypeScript) Literals should not be used for promise rejection
**Problem:** 
Expected the Promise rejection reason to be an Error.

**Why is this an issue?**
The use of literals (primitive values such as strings, numbers, booleans, etc.) for promise rejection is generally discouraged. While it is syntactically valid to provide literals as a rejected promise value, it is considered best practice to use instances of the Error class or its subclasses instead.

Using an instance of the Error class allows you to provide more meaningful information about the error. The Error class and its subclasses provide properties such as message and stack that can be used to convey useful details about the error, such as a description of the problem, the context in which it occurred, or a stack trace for debugging.

```
new Promise(function(resolve, reject) {
  reject(); // Noncompliant: use Error object to provide rejection reason
});

new Promise(function(resolve, reject) {
  reject('Something went wrong'); // Noncompliant: use Error object instead of literal
});
```

To fix your code provide an instanse of the Error class to the promise reject function.

```
new Promise(function(resolve, reject) {
  reject(new Error('Network timeout'));
});

new Promise(function(resolve, reject) {
  reject(new Error('Something went wrong'));
});
```

**Problem locations:**
packages/ketcher-core/src/infrastructure/services/struct/remoteStructService.ts
packages/ketcher-standalone/src/infrastructure/services/struct/standaloneStructService.ts

oguzhanduran:

## Check list
- [ ] unit-tests written
- [ ] e2e-tests written
- [ ] documentation updated


packages/ketcher-core/src/infrastructure/services/struct/remoteStructService.ts
```diff
@@ -60,10 +60,10 @@ function pollDeferred(process, complete, timeGap, startTimeGap) {
             else setTimeout(iterate, timeGap);
           } catch (error) {
             KetcherLogger.error('remoteStructService.ts::pollDeferred', error);
-            reject(error);
+            reject(error instanceof Error ? error : new Error(String(error)));
```
Gyv4ik: I'd recommend extending error handling by doing normalization.
Here is a utility that should cover all cases.

```
 const normalizeError = (error: unknown): Error => {
    if (error instanceof Error) return error;
    if (typeof error === 'string') return new Error(error);

    try {
      return new Error(JSON.stringify(error));
    } catch {
      return new Error(String(error));
    }
  };

``` 

I suppose it can be placed inside ketcher-core/utilities

packages/ketcher-core/src/application/editor/modes/BaseMode.ts
```diff
@@ -237,8 +237,12 @@ export abstract class BaseMode {
 
       return this.pasteKetFormatFragment(ketStruct.struct);
     } catch (error) {
-      const stringError =
-        typeof error === 'string' ? error : JSON.stringify(error);
+      let stringError;
```
svvald: I see that there is a `normalizeError` function defined and used in one of the other files – can we use it here and in every other changed file where some custom error normalization is duplicated instead?

svvald (approved):

Issue #9282, opened 2026-02-20
Refactor: Empty collections should not be accessed or iterated


azamatalt:
**Problem:** 
Empty collections should not be accessed or iterated

**Why is this an issue?**
When a collection is empty with absolute certainty, it makes no sense to access or iterate it. Doing so can lead to unexpected behavior or errors in the code. The most common cause is that population was accidentally omitted or removed.
```
const strings = [];
if (strings.includes("foo")) {}  // Noncompliant: strings is always emptyfor (const str of strings) {}  // Noncompliantstrings.forEach(str => doSomething(str)); // Noncompliant

```
Make sure your code provides some way to populate the collection if their elements are to be accessed.
```
const strings = [];
strings.push("foo");if (strings.includes("foo")) {}for (const str of strings) {}
strings.forEach(str => doSomething(str));
```

**Problem locations:**
ketcher-autotests/tests/specs/Macromolecule-editor/Sequence-Mode/new-sequence-representation2.spec.ts
packages/ketcher-core/src/application/editor/actions/sgroup.ts

PR #9284, opened 2026-02-20
#9282 Refactor: Empty collections should not be accessed or iterated

azamatalt:
### Motivation
- Prevent accessing or iterating collections that are known to be empty which can lead to unexpected behavior or runtime errors in sequence tests and SGroup handling.

### Description
- Add an early return in `filterBugsInTests` (file: `ketcher-autotests/tests/specs/Macromolecule-editor/Sequence-Mode/new-sequence-representation2.spec.ts`) that immediately returns an empty array when `FailedTestNewSequenceRepresentation` is empty to avoid pointless filtering/iteration.
- Initialize `attachmentPoints` from `attrs.attachmentPoints ?? []` in `fromSeveralSgroupAddition` (file: `packages/ketcher-core/src/application/editor/actions/sgroup.ts`) instead of using a hardcoded always-empty array so downstream logic does not assume a non-empty collection.
- Changes touch the two files above and limit the scope to guarding against empty collections.

### Testing
- Ran `npx eslint` on the modified files and the run failed due to missing local configuration `eslint-config-standard` in the execution environment, so linting could not be validated here.
- Pre-commit hooks could not execute because `lint-staged`/husky tooling was not available in the environment, so those checks were not performed.
- No other automated tests were executed in this environment after the changes due to missing local dev dependencies and config.

------
[Codex Task](https://chatgpt.com/codex/tasks/task_e_69981e8e94d08332845ff96c50a10085)

### DIFF
--- a/packages/ketcher-core/src/application/editor/actions/sgroup.ts
+++ b/packages/ketcher-core/src/application/editor/actions/sgroup.ts
@@ -88,7 +88,7 @@ export function fromSeveralSgroupAddition(
   atoms,
   attrs,
 ) {
-  const attachmentPoints = attrs.attachmentPoints ?? [];
+  const attachmentPoints = [];
 
   const descriptors = attrs.fieldValue;
   if (typeof descriptors === 'string' || type !== 'DAT') {


ketcher-autotests/tests/specs/Macromolecule-editor/Sequence-Mode/new-sequence-representation2.spec.ts
```diff
@@ -81,6 +81,10 @@ function filterBugsInTests(
   sequenceId: number,
   monomerId?: number,
 ): string[] {
+  if (FailedTestNewSequenceRepresentation.length === 0) {
```
svvald, line 84: It is an autotest change, not anything functional, but okay, if it help to silent sonar here :)

packages/ketcher-core/src/application/editor/actions/sgroup.ts
```diff
@@ -88,7 +88,7 @@ export function fromSeveralSgroupAddition(
   atoms,
   attrs,
 ) {
-  const attachmentPoints = [];
+  const attachmentPoints = attrs.attachmentPoints ?? [];
```
svvald: It's hard to say because of the lacking typing, but I suppose there is no `attachmentPoints` property in `attr`

PR #9308, opened 2026-02-24
#9306: Refactor: fixed binary expression should not always return same value issue

### LINKED ISSUE #9306: Refactor: (TypeScript) Binary expressions should not always return the same value
**Problem:** 
Unexpected constant binary expression. Compares constantly with the right-hand side of the `!=`

**Why is this an issue?**
Comparisons that always evaluate to true or to false, logical expressions that either always or never short-circuit and comparisons to a newly constructed object should not be used.

An expression that always produces the same result, regardless of the inputs, is unnecessary and likely indicates a programmer’s error. This can come from

confusing operator precedence
expecting strict equality between different types
expecting objects to be compared by value
expecting empty objects to be false or null
mistyping >= for ⇒
This can also happen when you put an assignment in a logical sub-expression. While not strictly a bug, this practice is confusing and should be avoided.

**How can I fix it?**
Review the code around the issue to find out why the expression always produces the same result. Pay attention to the operator precedence, comparing objects of different types, and comparing objects by reference (not by value!).

**Noncompliant code example**

```
!foo == null;
a + b ?? c;
x === [];
(foo=0) && bar;
```

**Compliant solution**

```
foo != null;
a + (b ?? c);
x.length === 0;
```

**Problem locations:**
packages/ketcher-core/src/application/editor/actions/erase.ts

oguzhanduran:

## Check list
- [ ] unit-tests written
- [ ] e2e-tests written
- [ ] documentation updated

### DIFF
--- a/packages/ketcher-core/src/application/editor/actions/erase.ts
+++ b/packages/ketcher-core/src/application/editor/actions/erase.ts
@@ -118,7 +118,7 @@ export function fromOneBondDeletion(restruct, id) {
 }
 
 export function fromFragmentDeletion(restruct, rawSelection) {
-  assert(!!rawSelection != null);
+  assert(rawSelection != null);
 
   let action = new Action();
   const atomsToRemove: Array<number> = [];


packages/ketcher-core/src/application/editor/actions/erase.ts
```diff
@@ -118,7 +118,7 @@ export function fromOneBondDeletion(restruct, id) {
 }
 
 export function fromFragmentDeletion(restruct, rawSelection) {
-  assert(!!rawSelection != null);
+  assert(rawSelection != null);
```
svvald, line 121: Maybe we could explore if it's safe to use strict comparison `!==` here for better clarity?

oguzhanduran: The rawSelection parameter has no type annotation, so it could be null or undefined. Using != null guards against both cases, so I think better to leave as it is

svvald (approved):

Issue #9350, opened 2026-03-04
Refactor: (Typescript) Methods should not contain selector parameters


umutzd-alt:
A selector parameter is a boolean parameter that’s used to determine which of two paths to take through a method. Specifying such a parameter may seem innocuous, particularly if it’s well named.

Unfortunately, developers calling the method won’t see the parameter name, only its value. They’ll be forced either to guess at the meaning or to take extra time to look the method up.

This rule finds methods with a boolean that’s used to determine which path to take through the method.

Noncompliant code example
```ts
function tempt(name: string, ofAge: boolean) {
  if (ofAge) {
    offerLiquor(name);
  } else {
    offerCandy(name);
  }
}

// ...
function corrupt() {
  tempt("Joe", false); // does this mean not to temp Joe?
}
```

Compliant solution
Instead, separate methods should be written.

```ts
function temptAdult(name: string) {
  offerLiquor(name);
}

function temptChild(name: string) {
  offerCandy(name);
}

// ...
function corrupt() {
  age < legalAge ? temptChild("Joe") : temptAdult("Joe");
}
```

Files to fix:

`packages/ketcher-core/src/application/render/restruct/resimpleObject.ts`
`packages/ketcher-core/src/domain/entities/struct.ts`
`packages/ketcher-react/src/script/ui/views/modal/components/PeriodTable/PeriodTable.tsx`

PR #9351, opened 2026-03-04
#9350 - Refactor: (Typescript) Methods should not contain selector parameters

umutzd-alt:
## Check list
- [ ] unit-tests written
- [ ] e2e-tests written
- [ ] documentation updated

### DIFF
--- a/packages/ketcher-core/src/domain/entities/struct.ts
+++ b/packages/ketcher-core/src/domain/entities/struct.ts
@@ -1285,7 +1285,11 @@ export class Struct {
   }
 
   getGroupFromAtomIdBySgroups(atomId: number | undefined): SGroup | undefined {
-    const sgroupId = this.getGroupIdFromAtomIdBySgroups(atomId as number);
+    if (!isNumber(atomId)) {
+      return undefined;
+    }
+
+    const sgroupId = this.getGroupIdFromAtomIdBySgroups(atomId);
     return this.sgroups?.get(sgroupId as number);
   }
 


packages/ketcher-core/src/domain/entities/atom.ts
```diff
@@ -951,7 +953,9 @@ export class Atom extends BaseMicromoleculeEntity {
     const sgroup =
       structOrSgroup instanceof SGroup
         ? structOrSgroup
-        : structOrSgroup.getGroupFromAtomId(atomId, searchBySgroups);
+        : searchBySgroups
+        ? structOrSgroup.getGroupFromAtomIdBySgroups(atomId)
+        : structOrSgroup.getGroupFromAtomId(atomId);
```
svvald: Let's avoid nested ternary operators here please

packages/ketcher-core/src/domain/entities/struct.ts
```diff
@@ -1255,32 +1258,29 @@ export class Struct {
...
-    atomId: number | undefined,
-    searchBySgroups = false,
-  ): SGroup | undefined {
-    const sgroupId = this.getGroupIdFromAtomId(
-      atomId as number,
-      searchBySgroups,
-    );
+  getGroupFromAtomId(atomId: number | undefined): SGroup | undefined {
+    const sgroupId = this.getGroupIdFromAtomId(atomId as number);
+    return this.sgroups?.get(sgroupId as number);
```
svvald: I'd also prefer to avoid using `as` as we know exactly what the possible outcomes are and add a clear short-circuit paths to leave the function:
```
if (atomId === undefined) {
return;
}
...
if (sgroupId === null) {
return;
}

return ...;
```

packages/ketcher-core/src/domain/entities/struct.ts
```diff
@@ -1255,32 +1258,29 @@ export class Struct {
...
+  }
+
+  getGroupFromAtomIdBySgroups(atomId: number | undefined): SGroup | undefined {
+    const sgroupId = this.getGroupIdFromAtomIdBySgroups(atomId as number);
```
svvald: Same here

PR #9440, opened 2026-03-13
#9362 - Refactor: Variables are missing in props validation (let's rewrite JS to TS) 27

mariam-khutuashvili:
 - Converted Tabs.jsx to TypeScript (Tabs.tsx)
  - Added TabPanel interface defining the shape of tab objects (caption, component, props, tabIndex)
  - Added TabsProps interface defining all component props (tabs, tabIndex, changeTab, className, contentClassName,
  captions)
  - Added TabsState interface for component state (tabIndex)
  - Typed all method parameters: constructor, changeTab, handleKeyDown, componentDidUpdate
  - Renamed local Component variable to TabComponent to avoid shadowing the React Component import

## Check list
- [ ] unit-tests written
- [ ] e2e-tests written
- [ ] documentation updated

### DIFF
--- a/packages/ketcher-react/src/script/ui/component/view/Tabs/Tabs.tsx
+++ b/packages/ketcher-react/src/script/ui/component/view/Tabs/Tabs.tsx
@@ -92,7 +92,7 @@ class Tabs extends Component<TabsProps, TabsState> {
             ))}
           </li>
         </ul>
-        {tabPanel && TabComponent && (
+        {TabComponent && (
           <div className={contentClassName}>
             <TabComponent {...componentProps} />
           </div>


packages/ketcher-react/src/script/ui/component/view/Tabs/Tabs.tsx
```diff
@@ -71,9 +92,9 @@ class Tabs extends Component {
             ))}
           </li>
         </ul>
-        {tabPanel && (
+        {tabPanel && TabComponent && (
```
svvald: Provided `TabComponent` is being accessed from within `tabPanel` it means that it's only sufficient to check for `TabComponent` here as `tabPanel` will be defined in this case

PR #9452, opened 2026-03-16
#9364 - Refactor: Variables are missing in props validation (let's rewrite JS to TS) 29

mariam-khutuashvili:
 - Converted template-attach.jsx to TypeScript (template-attach.tsx)
  - Added AttachProps (composed from AttachOwnProps, AttachStateProps, AttachDispatchProps) to type all component props
  - Added TemplateItem, NormalizedTemplate, AttachPoints, and FormState helper interfaces
  - Typed all class fields (mode, tmpl, oldKetcherEditor), method parameters (checkIsValidName, onResult), and lifecycle
   methods
  - Typed helper functions initTmpl, structNormalization, and getScale with proper Struct types
  - Created state/store.types.ts with separate interfaces per reducer slice (TemplatesState, ModalState, OptionsState)
  and a combined StoreState type, exported for reuse across the codebase
  - Used Boolean() instead of !! for clarity
  - Replaced any in connect() with typed StoreState and Dispatch from redux
  - Removed MODES constant in favor of inline 'save' | 'edit' literal values
  - Extended FieldProps in form.d.ts with onChange and placeholder props
  - Replaced unary + coercion with Number() for clarity on template prop conversion

## Check list
- [ ] unit-tests written
- [ ] e2e-tests written
- [ ] documentation updated

### DIFF
--- a/packages/ketcher-react/src/script/ui/dialog/template/template-attach.tsx
+++ b/packages/ketcher-react/src/script/ui/dialog/template/template-attach.tsx
@@ -25,6 +25,7 @@ import {
 
 import { connect } from 'react-redux';
 import { Dispatch } from 'redux';
+import { StoreState } from '../../state/store.types';
 import { storage } from '../../storage-ext';
 import Form, { Field } from '../../component/form/form/form';
 import { attachSchema } from '../../data/schema/struct-schema';
@@ -83,23 +84,6 @@ interface AttachDispatchProps {
 }
 
 type AttachProps = AttachOwnProps & AttachStateProps & AttachDispatchProps;
-
-interface StoreState {
-  templates: {
-    attach: {
-      name: string;
-      atomid: number;
-      bondid: number;
-    };
-    lib: TemplateItem[];
-  };
-  modal: {
-    form: FormState;
-  };
-  options: {
-    settings: Record<string, unknown>;
-  };
-}
 
 // @TODO When theming is implemented, use theme wherever possible
 const TemplateEditDialog = styled(Dialog)`
--- a/packages/ketcher-react/src/script/ui/state/store.types.ts
+++ b/packages/ketcher-react/src/script/ui/state/store.types.ts
@@ -1,0 +1,56 @@
+/****************************************************************************
+ * Copyright 2021 EPAM Systems
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ ***************************************************************************/
+
+// TODO: This is a partial type definition of the Redux store.
+// The actual store structure is more complex and should be expanded
+// as more components are migrated to TypeScript.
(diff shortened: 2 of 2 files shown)


packages/ketcher-react/src/script/ui/dialog/template/template-attach.tsx
```diff
@@ -266,7 +321,7 @@ class Attach extends Component {
       <TemplateEditDialog
         title={dialogTitle}
         result={this.onResult}
-        valid={() => this.props.formState.valid && name}
+        valid={() => this.props.formState.valid && !!name}
```
svvald: Let's use `Boolean` instead of `!!` for better clarity

packages/ketcher-react/src/script/ui/dialog/template/template-attach.tsx
```diff
@@ -306,12 +361,16 @@ class Attach extends Component {
...
             <NameInput
               name="name"
-              value={name}
-              onChange={this.props.onNameEdit}
-              placeholder="template"
+              {...({
+                value: name,
+                onChange: this.props.onNameEdit,
+                placeholder: 'template',
+              } as any)}
```
svvald: I don't see any point of doing so here, what's wrong with the previous approach of providing each prop separately?
Especially provided we have to silent eslint on explicit any which I'd like to avoid in new code

packages/ketcher-react/src/script/ui/dialog/template/template-attach.tsx
```diff
@@ -347,33 +406,35 @@ class Attach extends Component {
...
     ...store.templates.attach,
     templateLib: store.templates.lib,
     formState: store.modal.form,
     globalSettings: store.options.settings,
   }),
-  (dispatch) => ({
-    onInit: (name, ap) => dispatch(initAttach(name, ap)),
-    onAttachEdit: (ap) => dispatch(setAttachPoints(ap)),
-    onNameEdit: (name) => dispatch(setTmplName(name)),
+  (dispatch: any) => ({
```
svvald: Can we avoid using explicit any here and above? Perhaps it would require to specify types for Redux store and dispatch somewhere but let's also do this in this PR rather than ending up with some tech debt

packages/ketcher-react/src/script/ui/dialog/template/template-attach.tsx
```diff
@@ -196,19 +246,24 @@ const editorStyles = {
...
-    EDIT: 'edit',
+    SAVE: 'save' as const,
+    EDIT: 'edit' as const,
   };
 
-  constructor({ onInit, ...props }) {
-    super();
-    this.mode = isEmpty(props.tmpl.props) ? this.MODES.SAVE : this.MODES.EDIT;
-    this.tmpl = initTmpl(props.tmpl);
+  mode: 'save' | 'edit';
```
svvald, line 252: Given we specify the exact possible literal values for `mode` here, we may get rid of `MODES` constant above and use inline values when assigning/checking in this component as this will work just fine with Typescript now in place

packages/ketcher-react/src/script/ui/dialog/template/template-attach.tsx
```diff
@@ -83,6 +84,23 @@ interface AttachDispatchProps {
 
 type AttachProps = AttachOwnProps & AttachStateProps & AttachDispatchProps;
 
+interface StoreState {
```
svvald: I suppose it's better to put this type somewhere close to Redux setup (packages/ketcher-react/src/script/ui/state/index.js) and export it for it to be used by other code pieces as well

Also it would be nice to add some TODO mark there as apparently store structure is much more complex as this interface and it will be filled with more data in future

Ideally even create a separate type for each reducer and then combine them within Store type

svvald (approved):

PR #9472, opened 2026-03-17
#9447 Refactor: Variables are missing in props validation (let's rewrite JS to TS) 32

Nizomkhujaev:

Converted `OpenButton` from `.jsx` to `.tsx` and added proper TypeScript typing across the component and related utilities.

#### What’s included
- Defined `OpenButtonOwnProps` for component-specific props (`server`, `type`, `className`, `onLoad`, `onError`)
- Extended standard button props using `ButtonHTMLAttributes<HTMLButtonElement>` (via `Omit`) so things like `title`, `disabled`, `data-testid`, etc. can be passed through naturally
- Typed component state (`OpenButtonState`) and the internal `btn` ref
- Introduced `openButton.types.ts` with shared types:
  - `FileContent`
  - `OpenerFunction`
- Reused those types in `Recognize.tsx` for consistency
- Updated `onImage` in `Recognize.tsx` to accept `File | FileContent | null` (to match what `onLoad` can return)
- Fixed a small type issue by replacing `ev.target.value = null` with an empty string
- Prevented `onLoad` and `onError` from being accidentally passed down to the native `<button>`

## Check list
- [ ] unit-tests written
- [ ] e2e-tests written
- [ ] documentation updated

### DIFF
--- a/packages/ketcher-react/src/script/ui/component/view/openbutton.tsx
+++ b/packages/ketcher-react/src/script/ui/component/view/openbutton.tsx
@@ -14,7 +14,12 @@
  * limitations under the License.
  ***************************************************************************/
 
-import React, { Component, PropsWithChildren } from 'react';
+import {
+  ButtonHTMLAttributes,
+  ChangeEvent,
+  Component,
+  PropsWithChildren,
+} from 'react';
 import clsx from 'clsx';
 
 import { fileOpener } from '../../utils';
@@ -30,10 +35,7 @@ type OpenButtonOwnProps = {
 };
 
 type OpenButtonProps = PropsWithChildren<
-  Omit<
-    React.ButtonHTMLAttributes<HTMLButtonElement>,
-    keyof OpenButtonOwnProps
-  > &
+  Omit<ButtonHTMLAttributes<HTMLButtonElement>, keyof OpenButtonOwnProps> &
     OpenButtonOwnProps
 >;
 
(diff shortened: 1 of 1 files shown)


packages/ketcher-react/src/script/ui/component/view/openbutton.tsx
```diff
@@ -14,32 +14,55 @@
...
     if (props.server) {
-      fileOpener(props.server).then((opener) => {
+      fileOpener(props.server).then((opener: OpenerFunction) => {
         this.setState({ opener });
       });
     }
   }
 
-  open(ev) {
+  open(ev: React.ChangeEvent<HTMLInputElement>) {
```
svvald: `ChangeEvent` could be imported from `react` package directly without React prefix

packages/ketcher-react/src/script/ui/component/view/openbutton.tsx
```diff
@@ -14,32 +14,55 @@
...
     const files = ev.target.files;
     const noop = () => null;
     const { onLoad = noop, onError = noop } = this.props;
 
-    if (this.state.opener && files.length)
+    if (this.state.opener && files?.length)
       this.state.opener(files[0]).then(onLoad, onError);
-    else if (files.length) onLoad(files[0]);
-    ev.target.value = null;
+    else if (files?.length) onLoad(files[0]);
```
svvald: I know it was already like that but could you wrap this `if ... else` statement into `{ ... }` block for better readability?

packages/ketcher-react/src/script/ui/component/view/openbutton.tsx
```diff
@@ -49,17 +72,20 @@ class OpenButton extends Component {
...
 
     return (
       <button
-        onClick={() => this.btn.click()}
+        onClick={() => this.btn?.click()}
         className={clsx(classes.openButton, className)}
-        {...props}
+        {...buttonProps}
       >
+        open
```
svvald: Does this text have to be here?

packages/ketcher-react/src/script/ui/component/view/openbutton.tsx
Nizomkhujaev: fixed

packages/ketcher-react/src/script/ui/component/view/openbutton.tsx
Nizomkhujaev: fixed

packages/ketcher-react/src/script/ui/component/view/openbutton.tsx
Nizomkhujaev: fixed

PR #9473, opened 2026-03-17
#9435-Refactor: Explicit Any type clean up (part 2.2)

### LINKED ISSUE #9435: Refactor: Explicit Any type clean up (part 2.2)
**Problem:** 
Unspecified any.

**Why is this an issue?**

**How can it be fixed?**

**Problem locations:**
packages/ketcher-react/src/script/ui/views/modal/components/process/Analyse/Analyse.tsx:179
packages/ketcher-react/src/script/ui/views/modal/components/process/Analyse/Analyse.tsx:198
packages/ketcher-react/src/script/ui/views/modal/components/process/Analyse/Analyse.tsx:209
packages/ketcher-react/src/script/ui/views/modal/components/process/Analyse/Analyse.tsx:220

St-Permiakov:
- Removed all usages of "any" type from "Analyse.tsx".
- Added explicit typing and interfaces for Analyse state and mapping.
- Modified Select.tsx to accept both `string` and `number` as value to align with actual implementation in Analyse.

## Check list
- [ ] unit-tests written
- [ ] e2e-tests written
- [ ] documentation updated

### DIFF
--- a/packages/ketcher-react/src/script/ui/views/modal/components/process/Analyse/Analyse.tsx
+++ b/packages/ketcher-react/src/script/ui/views/modal/components/process/Analyse/Analyse.tsx
@@ -26,6 +26,8 @@ import { connect } from 'react-redux';
 import { range } from 'lodash/fp';
 import Select from '../../../../../component/form/Select';
 import { getSelectOptionsFromSchema } from '../../../../../utils';
+import { Action } from 'redux';
+import { ThunkDispatch } from 'redux-thunk';
 
 interface AnalyseValues {
   gross?: string;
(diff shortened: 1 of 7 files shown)


packages/ketcher-react/src/script/ui/views/modal/components/process/Analyse/Analyse.tsx
```diff
@@ -194,8 +195,21 @@ function AnalyseDialog({
...
+  roundWeight: number;
+  roundMass: number;
+  roundElAnalysis: number;
+}
+
+interface RootState {
+  options: {
+    analyse: AnalyseState;
+  };
+}
```
svvald: There are already defined types for redux store states at packages/ketcher-react/src/script/ui/state/store.types.ts, please use them and extend if needed

svvald (changes requested):

packages/ketcher-react/src/script/ui/views/modal/components/process/Analyse/Analyse.tsx
St-Permiakov: Thanks for pointing that out. It wasn't there when the changes were made.
I updated the PR accordingly.

svvald (approved):

PR #9477, opened 2026-03-17
#9449-Refactor: Variables are missing in props validation (let's rewrite JS to TS) 34

guliaisaeva:
This PR migrates modal form state from JavaScript to TypeScript to improve prop/form payload validation and type safety in modal workflows.

What was changed
Converted modal form state module from JS to TS:
form.ts
Added explicit typings for:
form state shape
action payloads
reducer input/output
Kept all new type declarations in the same file (per request), without introducing a separate shared types file.

### DIFF
--- a/packages/ketcher-react/src/script/ui/views/modal/components/document/Save/Save.test.utils.tsx
+++ b/packages/ketcher-react/src/script/ui/views/modal/components/document/Save/Save.test.utils.tsx
@@ -1,7 +1,6 @@
 import { createStore } from 'redux';
 import { Provider } from 'react-redux';
 import { render } from '@testing-library/react';
-import { formReducer } from 'src/script/ui/state/modal/form';
 import { ErrorsContext } from 'src/contexts';
 import { defaultBondThickness } from 'ketcher-core';
 
@@ -36,7 +35,7 @@ const initialState = {
 
 export function renderWithMockContext(component: React.ReactNode) {
   const store = {
-    ...createStore(formReducer, initialState),
+    ...createStore((state = initialState) => state),
     dispatch: jest.fn(),
   };
   return render(


packages/ketcher-react/src/script/ui/state/modal/form.ts
```diff
@@ -123,40 +144,60 @@ export const formsState = {
...
+      errors: {} as ModalFormErrors,
     },
   };
 }
 
-export function formReducer(state, action) {
-  const newType = action.data?.result?.type;
-  if (newType === 'DAT') return sdataReducer(state, action);
+export function formReducer(
+  state: unknown = { errors: {} },
```
svvald: This type doesn't make much sense, `state` should be of `ModalFormState` probably

packages/ketcher-react/src/script/ui/state/modal/form.ts
```diff
@@ -123,40 +144,60 @@ export const formsState = {
...
-export function setDefaultSettings() {
+export function setDefaultSettings(): UpdateFormAction {
   return {
     type: 'UPDATE_FORM',
     data: {
-      result: getDefaultOptions(sdataCustomSchema),
+      result: getDefaultOptions(),
       valid: true,
-      errors: {},
+      errors: {} as ModalFormErrors,
```
svvald: I'd avoid having type assertion here

packages/ketcher-react/src/script/ui/state/modal/form.ts
```diff
@@ -123,40 +144,60 @@ export const formsState = {
...
+    )
   )
-    return nucleotideComponentReducer(state, action);
+    return nucleotideComponentReducer(
+      formState as Parameters<typeof nucleotideComponentReducer>[0],
+      action as Parameters<typeof nucleotideComponentReducer>[1],
+    );
 
-  return { ...state, ...action.data };
+  return { ...(state as object), ...actionData };
```
svvald: This type assertion doesn't make much sense here, if `state` is defined properly, it should satisfy constraints just with `...state`

packages/ketcher-react/src/script/ui/state/modal/form.ts
```diff
@@ -123,40 +144,60 @@ export const formsState = {
...
   };
 }
 
-export function formReducer(state, action) {
-  const newType = action.data?.result?.type;
-  if (newType === 'DAT') return sdataReducer(state, action);
+export function formReducer(
+  state: unknown = { errors: {} },
+  action: ModalReducerAction,
+): unknown {
```
svvald: Reducer always returns a new state, return type of unknown is not quite correct here, it should be `ModalFormState`

packages/ketcher-react/src/script/ui/state/modal/form.ts
guliaisaeva: done✅

packages/ketcher-react/src/script/ui/state/modal/form.ts
guliaisaeva: done✅

packages/ketcher-react/src/script/ui/state/modal/form.ts
guliaisaeva: done ✅

packages/ketcher-react/src/script/ui/state/modal/form.ts
guliaisaeva: done✅

PR #9482, opened 2026-03-18
#9446 - Refactor: Variables are missing in props validation (let's rewrite JS to TS) 31

mariam-khutuashvili:
 - Converted StructEditor.jsx to TypeScript (StructEditor.tsx)
  - Added StructEditorProps interface defining all component props including ketcherId, struct, tool, toolOpts, options,
   event handlers (onZoomIn, onZoomOut, onInit, etc.), and an index signature for dynamic on[Event] props
  - Added StructEditorState interface for component state (enableCursor, clientX, clientY, tooltip)
  - Typed class fields (editor, editorRef, logRef), all lifecycle methods, and helper functions (setupEditor,
  removeEditorHandlers)
  - Typed handleWheel event handler with WheelEvent
  - Used optional chaining for ref access instead of non-null assertions
  - Added null-safe fallback for getBoundingClientRect() when ref may be null
  - Fixed Tooltip position to use nullish coalescing (?? 0) for optional cursor coordinates
  - Guarded logRef.current with early return in message handler

## Check list
- [ ] unit-tests written
- [ ] e2e-tests written
- [ ] documentation updated

### DIFF
--- a/packages/ketcher-react/src/script/ui/views/components/StructEditor/StructEditor.tsx
+++ b/packages/ketcher-react/src/script/ui/views/components/StructEditor/StructEditor.tsx
@@ -61,8 +61,6 @@ interface StructEditorProps {
 
 interface StructEditorState {
   enableCursor: boolean;
-  clientX: number;
-  clientY: number;
   tooltip: string;
 }
 
@@ -123,8 +121,6 @@ class StructEditor extends Component<StructEditorProps, StructEditorState> {
     super(props);
     this.state = {
       enableCursor: false,
-      clientX: 0,
-      clientY: 0,
       tooltip: '',
     };
     this.editorRef = createRef();
@@ -261,8 +257,6 @@ class StructEditor extends Component<StructEditorProps, StructEditorState> {
     });
 
     this.editor.event.cursor.add((csr) => {
-      let clientX, clientY;
-
       switch (csr.status) {
         case 'enable': {
           this.editorRef.current?.classList.add(classes.enableCursor);
@@ -274,8 +268,8 @@ class StructEditor extends Component<StructEditorProps, StructEditorState> {
               bottom: 0,
             };
 
-          clientX = csr.cursorPosition.clientX;
-          clientY = csr.cursorPosition.clientY;
+          const clientX = csr.cursorPosition.clientX;
+          const clientY = csr.cursorPosition.clientY;
 
           const handShouldBeShown =
             clientX >= left &&
@@ -294,8 +288,6 @@ class StructEditor extends Component<StructEditorProps, StructEditorState> {
           this.editorRef.current?.classList.add(classes.enableCursor);
           this.setState({
             enableCursor: true,
-            clientX,
-            clientY,
           });
           break;
         }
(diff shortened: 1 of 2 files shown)


packages/ketcher-react/src/script/ui/views/components/StructEditor/StructEditor.tsx
```diff
@@ -349,26 +405,36 @@ class StructEditor extends Component {
...
-          render={this.props.render}
-          groupStruct={this.props.groupStruct}
-          sGroup={this.props.sGroup}
+          {...({
+            clientX,
+            clientY,
+            render: this.props.render,
+            groupStruct: this.props.groupStruct,
+            sGroup: this.props.sGroup,
+          } as Record<string, unknown>)}
```
svvald: Why this destructuring is used here, what's wrong with line-by-line props assignment as it was done before?

packages/ketcher-react/src/script/ui/views/components/StructEditor/StructEditor.tsx
```diff
@@ -349,26 +405,36 @@ class StructEditor extends Component {
...
+        />
+        <InfoTooltip
+          {...({ render: this.props.render } as Record<string, unknown>)}
```
svvald: Same here

packages/ketcher-react/src/script/ui/views/components/StructEditor/StructEditor.tsx
```diff
@@ -88,22 +136,24 @@ class StructEditor extends Component {
...
-        this.props.onZoomOut(event);
+        this.props.onZoomOut?.(event);
       }
     } else {
       this.scrollCanvas(event);
       this.editor.rotateController.updateFloatingToolsPosition();
       this.editor.hoverIcon.updatePosition();
-      this.editor.tool()?.mousemove(this.editor.lastEvent);
+      (
+        this.editor.tool() as { mousemove?: (event: unknown) => void }
```
svvald: This type assertion doesn't sit well with me here, `tool()` returns `Tool | null` type with `Tool` having `mousemove` defined so there shouldn't be need in this assertion

packages/ketcher-react/src/script/ui/views/components/StructEditor/StructEditor.tsx
```diff
@@ -25,14 +31,52 @@ import { omit } from 'lodash';
...
 import { getSmoothScrollDelta } from './helpers';
 import InfoTooltip from './InfoTooltip';
 import MonomerCreationWizard from '../MonomerCreationWizard/MonomerCreationWizard';
-import { Tooltip } from '../Tooltip';
+import { Tooltip as TooltipComponent } from '../Tooltip';
+
+const Tooltip = TooltipComponent as React.FC<{
+  message: string;
+  position: { x: number; y: number };
+}>;
```
svvald: This is unnecessary, Tooltip is a self-sufficient component which already uses typescript and has prop types defined in its own file

svvald (changes requested):

PR #9490, opened 2026-03-18
#9030 – Fix: CSP script-src 'self' compatibility in example dev server

### LINKED ISSUE #9030: unsafe-eval ketcher-react
**Steps to Reproduce**
1. Go to examples foder 
2. Edit vite config to have CSP headeras `script-src 'self'`
3. Try to run locally

**Actual behavior**
App crashes with an error in console:
````
Executing inline script violates the following Content Security Policy directive 'script-src 'nonce-somerandomnonce' 'self''. Either the 'unsafe-inline' keyword, a hash ('sha256-LFCG3QCmCedCg1JYelEj6C+b/vwdFtnkcuVc3RakDfw='), or a nonce ('nonce-...') is required to enable inline execution. The action has been blocked.
(index):43

Uncaught Error: @vitejs/plugin-react can't detect preamble. Something is wrong.
Button.tsx:22
at Button.tsx:22:10
options-schema.ts:476

Error compiling schema, function code: const schema2 = scope.schema[2];const schema1 = scope.schema[1];return function validate1(data, {instancePath="", parentData, parentDataProperty, rootData=data}={}) {let vErrors = null;let errors = 0;if(!(((typeof data == "number") && (!(data % 1) && !isNaN(data))) && (isFinite(data)))){const err0 = {
instancePath, schemaPath:"#/definitions/nonNegativeInteger/type", keyword:"type", params: {type: "integer"}, message:"must be integer"};if(vErrors === null){vErrors = [err0]; }else
{vErrors.push(err0);}errors++;}if((typeof data == "number") && (isFinite(data))){if(data < 0 || isNaN(data)) {const err1 = {instancePath, schemaPath:"#/definitions/nonNegativeInteger/minimum", keyword:"minimum", params: {comparison: ">=", limit: 0},message:"must be >= 0"};if(vErrors === null){vErrors = [err1]; }else
{vErrors.push(err1);}errors++;}}validate1.errors = vErrors; return errors === 0;}
````

Image

**Expected behavior**
App doesn't crash

3.8.0 

**Additional context**
This happens due to ketcher-react relying on ajv. My project relies on ketcher-react to render the editor, and im trying to set up secure headers, including CSP header. I can't bypass this error due to ketcher-react. I saw in the description of https://github.com/epam/ketcher/pull/6269 that it was on purpose due to ketcher-react relying on it, but that creates the same issue as originally.

bekaChaduneli:

Fixed a crash that occurs when running the example app locally with a strict `Content-Security-Policy: script-src 'self'` header. The app failed with three distinct errors caused by different inline script execution requirements that CSP blocked.

**Root causes and fixes:**

1. **`var global = global || window` polyfill blocked** — `createHtmlPlugin` injected this as a raw inline `<script>` with no nonce attribute. CSP blocked it.
   - **Fix:** Added `attrs: { nonce: DEV_NONCE }` to the injected tag so it renders as `<script nonce="ketcher-dev-nonce">` which the CSP nonce allowlist permits.

2. **`@vitejs/plugin-react` preamble blocked → `can't detect preamble` crash** — In dev mode, `@vitejs/plugin-react` injects a React Fast Refresh bootstrap as an inline `<script type="module">`. CSP blocked it, making every React component throw on load. Passing `nonce` via `react({ nonce: DEV_NONCE })` is silently ignored on the version used (`^4.0.0`).
   - **Fix:** Added `CspNoncePlugin` — a custom Vite plugin with `enforce: 'post'` that runs after all other plugins have finished and stamps `nonce="ketcher-dev-nonce"` onto every `<script>` tag in the HTML that doesn't already carry a nonce. This is version-agnostic and covers all plugin-injected scripts.

3. **AJV `new Function()` blocked → `Error compiling schema` crash** — AJV v8 (used by `ketcher-react` in `options-schema.ts` and `form.jsx`) compiles JSON Schema validators at runtime using `new Function()`, which is equivalent to `eval` and blocked by CSP without `'unsafe-eval'`.
   - **Fix:** Added `'unsafe-eval'` to the CSP directive. Removing it would require pre-compiling all AJV schemas at build time using `ajv-standalone`/`ajv-cli` which is a separate, larger refactor.

**Changes made to `example/vite.config.js`:**

- Added `server.headers` with `Content-Security-Policy: script-src 'self' 'unsafe-eval' 'nonce-ketcher-dev-nonce'`
- Added `CspNoncePlugin` — post-phase `transformIndexHtml` that nonces all injected scripts
- Added `attrs: { nonce: DEV_NONCE }` to the `createHtmlPlugin` global polyfill script tag
- Added `react({ nonce: DEV_NONCE })` as belt-and-suspenders for React plugin versions that support it

**Note:** `DEV_NONCE` is a static string intentionally — it is only used in local development. In production, a cryptographically random nonce must be generated per HTTP request by the server.

## Check list
- [ ] unit-tests written
- [ ] e2e-tests written
- [ ] documentation updated


AlexeyGirin:
Already fixed by @svvald

PR #9676, opened 2026-04-06
#9625 - Refactor: Variables are missing in props validation (let's rewrite JS to TS) 43

### LINKED ISSUE #9625: Refactor: Variables are missing in props validation (let's rewrite JS to TS) 43
**Problem:** Variables are missing in props validation
React components should validate prop types

**Why is this an issue?**
In JavaScript, props are typically passed as plain objects, which can lead to errors and confusion when working with components that have specific prop requirements. However, it lacks of type safety and clarity when passing props to components in a codebase.

By defining types for component props, developers can enforce type safety and provide clear documentation for the expected props of a component. This helps catch potential errors at compile-time. It also improves code maintainability by making it easier to understand how components should be used and what props they accept.

**How can I fix it?**
Let re-write the entire file on TypeScript and define types for all props.

**Problem locations:**
packages/ketcher-core/src/application/render/restruct/reloop.js

**IMPORTANT:** Please do not delete old jsx file and create new tsx one, but add changes to existed file and rename it. That let me easily compare changes you made

mariam-khutuashvili:
Summary

  Converts packages/ketcher-core/src/application/render/restruct/reloop.js to TypeScript with full prop/parameter type annotations.
  Pure type refactor — no logic changes.

  Changes

  Single file: packages/ketcher-core/src/application/render/restruct/reloop.js → reloop.ts (renamed via git mv to preserve history).

  - ReLoop class fields typed: loop: Loop, centre: Vec2, radius: number
  - All method signatures typed: constructor(loop: Loop), static isSelectable(): boolean, show(restruct: ReStruct, _rlid: number, options: RenderOptions): void, isValid(struct: Struct, rlid: number): boolean
  - ReStruct and RenderOptions imported as import type (used only at type positions, matches the convention used by all 17 sibling files in restruct/)
  - Added a small getFromMap<T>(map, key): T helper that throws on missing keys — replaces 11 implicit non-null Map lookups with typed access and clear error messages
  - Fixed dormant JS bug: this.radius = new Vec2() → this.radius = 0 (radius is always treated as a number; the initial Vec2 was
  overwritten before being read)

  Out of scope

  - Tests — no other files modified per task scope
  - restruct.ts non-null assertion (new ReLoop(...!)) — separate file, not in scope of this PR
  - The pre-existing minSin === 0 → NaN latent edge case — out of scope for a pure JS→TS conversion (Copilot flagged this as low-confidence in a previous round; the same logic exists in master's reloop.js)

  Verification

  - npm run test — passes (prettier + eslint + types + jest for 4 packages)
  - npm run test:types — passes across all workspaces
  - Full Playwright autotest suite (Docker) — passes

## Check list
- [ ] unit-tests written
- [ ] e2e-tests written
- [ ] documentation updated

### DIFF
--- a/packages/ketcher-core/src/application/render/restruct/reloop.ts
+++ b/packages/ketcher-core/src/application/render/restruct/reloop.ts
@@ -27,14 +27,6 @@ import type { Struct } from 'domain/entities/struct';
 import type ReStruct from './restruct';
 import type { RenderOptions } from '../render.types';
 
-function getFromMap<T>(map: Map<number, T>, key: number): T {
-  const value = map.get(key);
-  if (value === undefined) {
-    throw new Error(`Map entry not found for key ${key}`);
-  }
-  return value;
-}
-
 class ReLoop extends ReObject {
   loop: Loop;
   centre: Vec2;
@@ -59,22 +51,24 @@ class ReLoop extends ReObject {
     const { hbs } = loop;
 
     this.centre = new Vec2();
-    hbs.forEach((hbid) => {
-      const hb = getFromMap(molecule.halfBonds, hbid);
-      const bond = getFromMap(restruct.bonds, hb.bid);
-      const apos = Scale.modelToCanvas(
-        getFromMap(restruct.atoms, hb.begin).a.pp,
-        options,
-      );
+    for (const hbid of hbs) {
+      const hb = molecule.halfBonds.get(hbid);
+      if (hb === undefined) return;
+      const bond = restruct.bonds.get(hb.bid);
+      if (bond === undefined) return;
+      const beginAtom = restruct.atoms.get(hb.begin);
+      if (beginAtom === undefined) return;
+      const apos = Scale.modelToCanvas(beginAtom.a.pp, options);
       if (bond.b.type !== Bond.PATTERN.TYPE.AROMATIC) loop.aromatic = false;
       // eslint-disable-next-line no-underscore-dangle
       this.centre.add_(apos);
-    });
+    }
 
     loop.convex = true;
     for (let k = 0; k < hbs.length; ++k) {
-      const hba = getFromMap(molecule.halfBonds, hbs[k]);
-      const hbb = getFromMap(molecule.halfBonds, hbs[(k + 1) % hbs.length]);
+      const hba = molecule.halfBonds.get(hbs[k]);
+      const hbb = molecule.halfBonds.get(hbs[(k + 1) % hbs.length]);
+      if (hba === undefined || hbb === undefined) return;
       const angle = Math.atan2(
         Vec2.cross(hba.dir, hbb.dir),
         Vec2.dot(hba.dir, hbb.dir),
@@ -84,30 +78,29 @@ class ReLoop extends ReObject {
 
     this.centre = this.centre.scaled(1.0 / hbs.length);
     this.radius = -1;
-    hbs.forEach((hbid) => {
-      const hb = getFromMap(molecule.halfBonds, hbid);
-      const apos = Scale.modelToCanvas(
-        getFromMap(restruct.atoms, hb.begin).a.pp,
-        options,
-      );
-      const bpos = Scale.modelToCanvas(
-        getFromMap(restruct.atoms, hb.end).a.pp,
-        options,
-      );
+    for (const hbid of hbs) {
+      const hb = molecule.halfBonds.get(hbid);
+      if (hb === undefined) return;
+      const beginAtom = restruct.atoms.get(hb.begin);
+      const endAtom = restruct.atoms.get(hb.end);
+      if (beginAtom === undefined || endAtom === undefined) return;
+      const apos = Scale.modelToCanvas(beginAtom.a.pp, options);
+      const bpos = Scale.modelToCanvas(endAtom.a.pp, options);
       const n = Vec2.diff(bpos, apos).rotateSC(1, 0).normalized();
       const dist = Vec2.dot(Vec2.diff(apos, this.centre), n);
       this.radius = this.radius < 0 ? dist : Math.min(this.radius, dist);
-    });
+    }
     this.radius *= 0.7;
     if (!loop.aromatic) return;
 
     // Skip rendering the aromatic circle when the loop sits entirely inside one contracted sgroup
     const atomIds = new Set<number>();
-    hbs.forEach((hbid) => {
-      const hb = getFromMap(molecule.halfBonds, hbid);
+    for (const hbid of hbs) {
+      const hb = molecule.halfBonds.get(hbid);
+      if (hb === undefined) return;
       atomIds.add(hb.begin);
       atomIds.add(hb.end);
-    });
+    }
 
     const firstAtomId = atomIds.values().next().value;
     if (firstAtomId === undefined) return;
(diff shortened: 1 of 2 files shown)


AlexeyGirin:
Look in the editor

packages/ketcher-core/src/application/render/restruct/reloop.ts
```diff
@@ -14,64 +14,87 @@
...
+import type ReStruct from './restruct';
+import type { RenderOptions } from '../render.types';
+
+function getFromMap<T>(map: Map<number, T>, key: number): T {
+  const value = map.get(key);
+  if (value === undefined) {
+    throw new Error(`Map entry not found for key ${key}`);
+  }
+  return value;
+}
```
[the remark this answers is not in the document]
mariam-khutuashvili: dropped the helper and switched all the Map.get() calls in this file to ! to match the convention used in rebond.ts, reatom.ts, and restruct.ts. 
 Tried the suggestion in 0b82209 but had to revert in 84b9911. CI failed with 12 @typescript-eslint/no-non-null-assertion errors in this file. Turns out the rule is "error" package-wide at the ketcher-core root and .eslintrc.json has a per-file override list (~50 entries: reatom.ts, rebond.ts, restruct.ts etc.) that downgrades it to "warn" for those legacy files — reloop.ts is not in that list, so ! here breaks CI even though the sibling files compile. The visible "convention" in restruct/ is actually grandfathered legacy, not current policy. Keeping getFromMap for now as the lint-compliant equivalent.

packages/ketcher-core/src/application/render/restruct/reloop.ts
```diff
@@ -14,64 +14,87 @@
...
 class ReLoop extends ReObject {
-  constructor(loop) {
+  loop: Loop;
+  centre: Vec2;
+  radius: number;
+
+  constructor(loop: Loop | undefined) {
     super('loop');
+    assert(loop !== undefined, 'ReLoop requires a Loop instance');
     this.loop = loop;
```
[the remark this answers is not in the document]
mariam-khutuashvili: Agreed — the constructor now takes Loop directly, the runtime assert is gone, and the single caller in restruct.ts:627 uses this.molecule.loops.get(loopId)! (which is allowed there since restruct.ts is in the per-file "warn" override list).

packages/ketcher-core/src/application/render/restruct/reloop.ts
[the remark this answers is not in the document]
AlexeyGirin: Drop function and do not use Clause suggestion but use check on undefined in every place

packages/ketcher-core/src/application/render/restruct/restruct.ts
```diff
@@ -624,7 +624,7 @@ class ReStruct {
       this.markBond(bid, 1);
     });
     ret.newLoops.forEach((loopId) => {
-      this.reloops.set(loopId, new ReLoop(this.molecule.loops.get(loopId)));
+      this.reloops.set(loopId, new ReLoop(this.molecule.loops.get(loopId)!));
```
AlexeyGirin: Use ===undefined check and call return if so

AlexeyGirin (dismissed):

packages/ketcher-core/src/application/render/restruct/restruct.ts
mariam-khutuashvili: Done in e7c33440:

  ret.newLoops.forEach((loopId) => {
    const loop = this.molecule.loops.get(loopId);
    if (loop === undefined) return;
    this.reloops.set(loopId, new ReLoop(loop));
  });

packages/ketcher-core/src/application/render/restruct/reloop.ts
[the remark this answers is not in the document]
mariam-khutuashvili: Removed getFromMap and replaced every Map.get() call site with an explicit if (… === undefined) return; check. Converted the three forEach loops to for…of so the early return aborts the whole show method (matching the original throw-on-missing semantics) instead of just skipping one iteration.

packages/ketcher-core/src/application/render/restruct/reloop.ts
```diff
@@ -16,95 +16,100 @@
...
 
-  static isSelectable() {
+  static isSelectable(): boolean {
     return false;
   }
 
-  show(restruct, rlid, options) {
-    // eslint-disable-line max-statements
-    const render = restruct.render;
+  // eslint-disable-next-line max-statements
```
svvald: I know it was already present but while we're here maybe we could get rid of this eslint bypass?

packages/ketcher-core/src/application/render/restruct/reloop.ts
```diff
@@ -16,95 +16,100 @@
...
-        options,
-      );
+      if (bond === undefined) return;
+      const beginAtom = restruct.atoms.get(hb.begin);
+      if (beginAtom === undefined) return;
+      const apos = Scale.modelToCanvas(beginAtom.a.pp, options);
       if (bond.b.type !== Bond.PATTERN.TYPE.AROMATIC) loop.aromatic = false;
-      this.centre.add_(apos); // eslint-disable-line no-underscore-dangle
-    });
+      // eslint-disable-next-line no-underscore-dangle
```
svvald: And this one as well. I wonder why `add_` was used here, why not `add`?

packages/ketcher-core/src/application/render/restruct/reloop.ts
```diff
@@ -16,95 +16,100 @@
...
-    const molecule = restruct.molecule;
-    const loop = this.loop;
+    const { loop } = this;
+    const { hbs } = loop;
+
     this.centre = new Vec2();
-    loop.hbs.forEach((hbid) => {
+    for (const hbid of hbs) {
       const hb = molecule.halfBonds.get(hbid);
+      if (hb === undefined) return;
```
svvald: A nitpick on this and all other `if` oneliners here, let's stick to having all of them in newline:
```
if (...) {
return; // or any other action e.g. loop.aromatic = false
}
```

It will drastically improve the code clarity and readability

packages/ketcher-core/src/application/render/restruct/reloop.ts
```diff
@@ -16,95 +16,100 @@
...
+    const { render, molecule } = restruct;
     const paper = render.paper;
-    const molecule = restruct.molecule;
-    const loop = this.loop;
+    const { loop } = this;
+    const { hbs } = loop;
+
     this.centre = new Vec2();
-    loop.hbs.forEach((hbid) => {
+    for (const hbid of hbs) {
```
svvald: Also while we're here I'd suggest to refactor local variables names and make them more verbose for better clarity (hbs -> halfBonds or halfBondsIds, hbid -> halfBondId etc.)

packages/ketcher-core/src/application/render/restruct/reloop.ts
svvald: Also I think this (and most or all the check below) could be simplified as `(!hb)` as `halfBonds` pool returns either `HalfBond` entity or `undefined` and the former could not be falsy but meaningful (in contrast with entity id which can be 0 which is falsy but meaningful)

svvald (changes requested):

packages/ketcher-core/src/application/render/restruct/reloop.ts
mariam-khutuashvili: Turns out max-statements isn't configured anywhere in our eslint chain (not in `standard`, `@typescript-eslint/recommended`, or the package rules), so the directive was a no-op — eslint passes clean with it removed, no extraction needed.

packages/ketcher-core/src/application/render/restruct/reloop.ts
mariam-khutuashvili: Switched to `this.centre = this.centre.add(beginPosition)` — Vec2.add returns a new vector, so it's equivalent, reads clearer, and the no-underscore-dangle bypass is gone.

packages/ketcher-core/src/application/render/restruct/reloop.ts
mariam-khutuashvili: Done — every `if` body is now on its own line with braces.

packages/ketcher-core/src/application/render/restruct/reloop.ts
mariam-khutuashvili: Done — renamed throughout: hbs→halfBondIds, hbid→halfBondId, hb→halfBond, hba/hbb→currentHalfBond/nextHalfBond, k→index, apos/bpos→beginPosition/endPosition, n→normal, dist→distance, pi/qi→atomPosition/innerPosition, dir→direction, pathStr→pathString.

packages/ketcher-core/src/application/render/restruct/reloop.ts
mariam-khutuashvili: Done — all the half-bond / bond / atom entity checks now use `!entity`. I kept `firstAtomId === undefined` as an explicit check, since that one's an atom id and 0 is falsy-but-meaningful — exactly the distinction you called out.

svvald (approved):

PR #9681, opened 2026-04-06
#5740 - Establishing connection between ambiguous phosphate and sugar works wrong

### LINKED ISSUE #5740: Establishing conennection between ambiguous phosphate and sugar works wrong
**Steps to Reproduce**
1. Toggle to Macro - Flex
2. Load using paste from clipboard way following HELM: `RNA1{R(A)([bnn],[bP])}|RNA2{R(A)([bnn],[bP])}$$$$V2.0`
image
image
3. Try to connect ambiguous phosphate with sugar using center to center way or ph(R2) to center way or sg(R1) to center way

**Actual behavior**
1. Attachment points are not shown
image
image
image
image

2. In case of center-to-center way - Select Connection point dialog appears
image

**Expected behavior**
1. Attachment points are shown
2. Select Connection point dialog doesn't appear - connection rules for phosphates and sugars applied (i.e. ph(R2)-sg(R1) connection establishes by default)

As per [requirements](https://github.com/epam/ketcher/issues/4554): 

> 6.2. If a variant monomer is a base, all rules for establishing bonds with bases are followed. The same logic for all other types.

**Versions**
- Ketcher Version 2.26.0-rc.1 Build at 2024-10-14; 07:11:59
- Indigo Toolkit Version 1.25.0-rc.1.0-g704673fcb-wasm32-wasm-clang-19.0.0
- Chrome Version 129.0.6668.101 (Official Build) (64-bit)
- Win10

Reference issues: https://github.com/epam/Indigo/issues/2188, https://github.com/epam/ketcher/issues/4554

beyzaevcen:

 Summary

- Fixes a `TypeError: this.getValidPoint is not a function` crash when
  hovering a sugar over an ambiguous phosphate (or vice versa) in Macro
  Flex mode.
- Fixes attachment points not being shown and the connection dialog
  incorrectly appearing when connecting an ambiguous phosphate/sugar
  with its counterpart.
- The default `ph(R2)–sg(R1)` bond rule now applies automatically for
  variant monomers, per requirement §6.2.

Root cause

Two separate bugs:

**1. TypeError in `AmbiguousMonomer`**

`AmbiguousMonomer.getValidSourcePoint` delegates to the real class
prototype via `.call(this, ...)`. For `Phosphate` and `Sugar`, those
prototype methods internally call a **private** helper `getValidPoint`.
Because `this` is an `AmbiguousMonomer` instance at call time,
`getValidPoint` is not in its prototype chain → `TypeError`.

**2. Failed `instanceof` checks in `Sugar` and `Phosphate`**

`Sugar.getValidPoint` checks `instanceof Phosphate` and
`Phosphate.getValidPoint` checks `instanceof Sugar`. Both return
`false` for `AmbiguousMonomer` instances, so ambiguous monomers were
never recognised as their equivalent type — causing the modal to open
instead of the bond being created automatically.

 Changes

| File | Change |
|---|---|
| `AmbiguousMonomer.ts` | Added `_createContextProxy`: a `Proxy` that intercepts property lookups and falls through to the target prototype for methods absent from `AmbiguousMonomer` (e.g. private `getValidPoint`) |
| `monomers.ts` | Added `isPhosphateOrAmbiguousPhosphate` and `isSugarOrAmbiguousSugar` type guard helpers, following the existing `isRnaBaseOrAmbiguousRnaBase` / `isPeptideOrAmbiguousPeptide` pattern |
| `Sugar.ts` | Replaced `instanceof Phosphate` with `isPhosphateOrAmbiguousPhosphate` |
| `Phosphate.ts` | Replaced `instanceof Sugar` with `isSugarOrAmbiguousSugar` |

## Check list
- [ ] unit-tests written
- [ ] e2e-tests written
- [ ] documentation updated

### DIFF
--- a/packages/ketcher-core/src/domain/entities/AmbiguousMonomer.ts
+++ b/packages/ketcher-core/src/domain/entities/AmbiguousMonomer.ts
@@ -135,15 +135,38 @@ export class AmbiguousMonomer extends BaseMonomer implements IVariantMonomer {
   }
 
   public getValidSourcePoint(_secondMonomer?: BaseMonomer) {
-    return MONOMER_CLASS_TO_CONSTRUCTOR[
-      this.monomerClass
-    ].prototype.getValidSourcePoint.call(this, _secondMonomer);
+    const proto = MONOMER_CLASS_TO_CONSTRUCTOR[this.monomerClass].prototype;
+    return proto.getValidSourcePoint.call(
+      this._createContextProxy(proto),
+      _secondMonomer,
+    );
   }
 
   public getValidTargetPoint(_firstMonomer: BaseMonomer) {
-    return MONOMER_CLASS_TO_CONSTRUCTOR[
-      this.monomerClass
-    ].prototype.getValidTargetPoint.call(this, _firstMonomer);
+    const proto = MONOMER_CLASS_TO_CONSTRUCTOR[this.monomerClass].prototype;
+    return proto.getValidTargetPoint.call(
+      this._createContextProxy(proto),
+      _firstMonomer,
+    );
+  }
+
+  private _createContextProxy(proto: Record<string, unknown>) {
+    return new Proxy(this, {
+      get(target, prop, receiver) {
+        if (
+          typeof prop === 'string' &&
+          !(prop in AmbiguousMonomer.prototype) &&
+          prop in proto
+        ) {
+          const value = proto[prop];
+          if (typeof value === 'function') {
+            return (...args: unknown[]) => value.call(receiver, ...args);
+          }
+          return value;
+        }
+        return Reflect.get(target, prop, receiver);
+      },
+    });
   }
 
   public get SubChainConstructor() {
--- a/packages/ketcher-core/src/domain/entities/AmbiguousMonomer.ts
+++ b/packages/ketcher-core/src/domain/entities/AmbiguousMonomer.ts
@@ -135,38 +135,15 @@ export class AmbiguousMonomer extends BaseMonomer implements IVariantMonomer {
   }
 
   public getValidSourcePoint(_secondMonomer?: BaseMonomer) {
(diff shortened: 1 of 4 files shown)


packages/ketcher-core/src/domain/entities/AmbiguousMonomer.ts
```diff
@@ -135,15 +135,38 @@ export class AmbiguousMonomer extends BaseMonomer implements IVariantMonomer {
...
-      this.monomerClass
-    ].prototype.getValidTargetPoint.call(this, _firstMonomer);
+    const proto = MONOMER_CLASS_TO_CONSTRUCTOR[this.monomerClass].prototype;
+    return proto.getValidTargetPoint.call(
+      this._createContextProxy(proto),
+      _firstMonomer,
+    );
+  }
+
+  private _createContextProxy(proto: Record<string, unknown>) {
```
svvald: I don't like this approach very much to be honest, it is not quite readable, adds an additional layer of complexity in runtime and overall violates SOLID principles.

As a solution for the following issue:
> AmbiguousMonomer.getValidSourcePoint delegates to the real class
prototype via .call(this, ...). For Phosphate and Sugar, those
prototype methods internally call a private helper getValidPoint.
Because this is an AmbiguousMonomer instance at call time,
getValidPoint is not in its prototype chain → TypeError.

I can propose making `getValidPoint` in `Sugar` and `Phosphate` classes `private static` and make it accept the monomer instance as `self` parameter:
```
private static getValidPoint(
    self: BaseMonomer,
    otherMonomer: BaseMonomer & IVariantMonomer,
    potentialPointOnOther: string | null,
  ) {
  // Replace each _this_ usage to _self_
}
```
And then call it like the following:
```
return Sugar.getValidPoint(
      this,
      secondMonomer,
      secondMonomer.potentialSecondAttachmentPointForBond,
    );
```

It should have the same effect while keeping the code nice and clean

svvald (dismissed):

rrodionov91 (approved):

PR #9784, opened 2026-04-17
6053:fix D peptide with asterik is marked as modified

### LINKED ISSUE #6053: D* peptide is marked as modified
**Steps to Reproduce**
1. Go to Macro - Flex mode
2. Add to canvas `D*` peptide
image

**Actual behavior**
`D*` peptide is marked as modified
image

**Expected behavior**
`D*` peptide is NOT marked as modified
image

As per [requirement](https://github.com/epam/ketcher/issues/5630):
> The list of non-modified amino acids can be found here: [#5629](https://github.com/epam/ketcher/issues/5629)
image

**Versions**
- Ketcher Version 2.28.0-rc.1 Build at 2024-12-03; 17:04:52
- Indigo Version 1.27.0-rc.1.0-g8d656a341-wasm32-wasm-clang-19.0.0
- Chrome Version 131.0.6778.109 (Official Build) (64-bit)
- Win10

Issue found while testing - https://github.com/epam/ketcher/issues/5630

oguzhanduran:

## Check list
- [ ] unit-tests written
- [ ] e2e-tests written
- [ ] documentation updated


packages/ketcher-core/src/application/render/renderers/PeptideRenderer.ts
```diff
@@ -91,9 +91,12 @@ export class PeptideRenderer extends BaseMonomerRenderer {
     };
 
     const monomerCode = this.monomer.monomerItem.props.MonomerNaturalAnalogCode;
+    const monomerLabel = this.monomer.monomerItem.label;
     let baseColor = peptideColorsMap[monomerCode] ?? super.textColor;
 
-    if (this.monomer.isModification) {
+    if (monomerLabel?.includes('*')) {
```
svvald: Is this check really needed? I think just the change of `isModification` getter is only is sufficient.
This check is even leading to a worse result - on the screens you updated the white label has worse contrast with orange backrgound

svvald (changes requested):

packages/ketcher-core/src/application/render/renderers/PeptideRenderer.ts
oguzhanduran: You are right, I discussed it with Alexey and we realized that expected behaviour was not correct on the ticket and fixed it so based on the new expected behaviour, I removed the previous implementation.

svvald (approved):

PR #9827, opened 2026-04-21
#4057 Drop-down menus are not shown if add Ketcher to dialog

### LINKED ISSUE #4057: Drop-down menus are not shown if add Ketcher to dialog
We create Ketcher and show it inside a dialog element, which has its z-index. And if I undestand correctly, because of this some drop-down menus (like file format for saving the structure is not shown. To be exact it is shoen but behind the dialog...)

I have investigated a bit and found that elements of drop-down menu are placed inside div with role="presentation" (please see the screenshot below). And basically, setting z-index will solve the problem. But I'm not sure this is the best way to solve the problem. Couls you please advise how to better handle this? Should I add some styles? And to which class it should be added?

image

guliaisaeva:
Description:
Restores correct dropdown stacking in dialogs by removing the custom z-index override from the dialog root. This lets MUI’s default z-index system handle dialog and dropdown layering, ensuring dropdown menus (Select, Popover, etc.) always appear above dialogs as intended. No functional regressions; all tests pass.


AlexeyGirin:

AlexeyGirin:

PR #9832, opened 2026-04-21
#9761 - Autotests: Circular structures improvement - formation of fused rings

### LINKED ISSUE #9761: Autotests: Circular structures improvement - formation of fused rings
**Source task(s):**
 - https://github.com/epam/Indigo/issues/3273

- [ ] Check that when there are more than one chain of monomers connected to each other and they are arranged as ring structures, the layout applied in a way that allows for the first structure to be fixed in its place and the rest of the monomers arranged in an n-agon in relation to it
- [ ] Check that when one circular structure is already formed and the user makes a selection that includes monomers connected to it outside of the circle and also monomers that are a part the circle in order to make a bicircular structure, the layout applied in a way that allows for the first structure to be fixed in its place and the rest of the monomers arranged in an n-agon in relation to it
- [ ] Check that Indigo first check if part of the selection belongs to an already formed regular n-agon
- [ ] Verify that the part of the selection that already belongs to an n-agon stay fixed and the second circular structure formed in relation to it
- [ ] Check that when forming the second circular structure, the bond length stay the same as it was between the monomers in the first layouted n-agon, 1,5Å

SergeiBakirov:
- added two tests
- added two test data files
- added expected screenshots

## Check list
- [ ] unit-tests written
- [ ] e2e-tests written
- [ ] documentation updated


AlexeyGirin (approved):
Approved

PR #9895, opened 2026-04-28
#8564 - Refactor: Deprecated APIs should not be used (part 3)

### LINKED ISSUE #8564: Refactor: Deprecated APIs should not be used (part 3)
**Problem:** 
['cancelBubble' is deprecated.](https://sonar.epam.com/sonarqube/coding_rules?open=typescript%3AS1874&rule_key=typescript%3AS1874)

**Why is this an issue?**
Code is sometimes annotated as deprecated by developers maintaining libraries or APIs to indicate that the method, class, or other programming element is no longer recommended for use. This is typically due to the introduction of a newer or more effective alternative. For example, when a better solution has been identified, or when the existing code presents potential errors or security risks.

Deprecation is a good practice because it helps to phase out obsolete code in a controlled manner, without breaking existing software that may still depend on it. It is a way to warn other developers not to use the deprecated element in new code, and to replace it in existing code when possible.

Deprecated classes, interfaces, and their members should not be used, inherited or extended because they will eventually be removed. The deprecation period allows you to make a smooth transition away from the aging, soon-to-be-retired technology.

Check the documentation or the deprecation message to understand why the code was deprecated and what the recommended alternative is.
```
/**
 * @deprecated Use newFunction instead.
 */
function oldFunction() {
  console.log("This is the old function.");
}

function newFunction() {
  console.log("This is the new function.");
}
oldFunction(); // Noncompliant: "oldFunction is deprecated"
```

**Problem location:**
packages/ketcher-core/src/application/editor/Editor.ts:548

sabaPailodze:
## Summary
Removes deprecated `cancelBubble` property by tracking `stopPropagation()` calls directly. Eliminates TypeScript deprecation warnings while maintaining exact behavioral compatibility.

### Problem
The `setupKeyboardEvents()` method used the deprecated `cancelBubble` property to check if event propagation was stopped by other handlers. Modern DOM API has no direct equivalent to read this state after the fact, causing linter warnings.

### Solution
- Wrapped `event.stopPropagation()` to track when it's called
- Set `isPropagationStopped` flag when any handler calls `stopPropagation()`
- Use the flag to determine whether to invoke `mode.onKeyDown()`
- Maintains identical original behavior with modern standard API

### Key Points
- ✅ Eliminates deprecation warning
- ✅ No behavior changes (semantic equivalent to original logic)
- ✅ Uses standard DOM API (`stopPropagation()`)
- ✅ Zero performance impact
- ✅ Cross-browser compatible

## Files Modified
- `packages/ketcher-core/src/application/editor/Editor.ts`

## Check list
- [ ] unit-tests written
- [ ] e2e-tests written
- [ ] documentation updated

### DIFF
--- a/packages/ketcher-core/src/application/editor/Editor.ts
+++ b/packages/ketcher-core/src/application/editor/Editor.ts
@@ -637,8 +637,16 @@ export class CoreEditor {
 
   private setupKeyboardEvents() {
     this.keydownEventHandler = (event: KeyboardEvent) => {
+      let isPropagationStopped = false;
+      const originalStopPropagation = event.stopPropagation.bind(event);
+      event.stopPropagation = () => {
+        isPropagationStopped = true;
+        originalStopPropagation();
+      };
+
       this.events.keyDown.dispatch(event);
-      if (!event.cancelBubble) {
+
+      if (!isPropagationStopped) {
         this.mode.onKeyDown(event).catch((error) => {
           KetcherLogger.error('Editor.ts::keydownEventHandler', error);
         });


AlexeyGirin:
@sabaPailodze please find community approved way to cancelBubble replacement

sabaPailodze:
https://developer.mozilla.org/en-US/docs/Web/API/Event/cancelBubble

rrodionov91 (approved):

PR #9897, opened 2026-04-28
#9743 – TODO | Bug: Action.isDummy() always returns false for most operation types

### LINKED ISSUE #9743: TODO | Bug: Action.isDummy() always returns false for most operation types
BaseOperation.isDummy() returns false by default and only 4 operation types override it. This makes Action.isDummy() effectively broken for the majority of operations. Requires auditing all operation types and adding meaningful isDummy() overrides.

editor/actions/action.ts:59

DmirtyUsov:

The claim is partially correct in facts but overstated in conclusion. Here's a precise analysis:
**What's true:**
`BaseOperation.isDummy()` returns false by default — confirmed.
Exactly 4 operations override it: `AtomAttr`, `BondAttr`, `AtomMove`, `RGroupAttr`.
The existing TODO in the code already acknowledges this:` // TODO [RB] the condition is always true for op.* operations.`
**Why "effectively broken" is too strong:**
For structural operations (add/delete atom, bond, fragment, sgroup, etc.) — `false` is semantically correct. They always produce a real state change; no override is needed or meaningful.
### Attribute-setting operations
Operation | Has isDummy? | Notes
-- | -- | --
AtomAttr | ✅ | atom[attribute] === value
BondAttr | ✅ | bond[attribute] === value
RGroupAttr | ✅ | rgroup[attribute] === value
SGroupAttr | ❌ | Missing
TextUpdate | ❌ | Missing

**Fix**
`SGroupAttr.ts` — delegates to the existing `checkAttr` method (`this.data[attr] === value`).
`TextUpdate.ts` — compares current `text.content` against the incoming value.
### Move operations
Operation | Has isDummy? | Delta field
-- | -- | --
AtomMove | ✅ | d.x === 0 && d.y === 0
BondMove | ❌ | d
RxnArrowMove | ❌ | d
RxnPlusMove | ❌ | d
TextMove | ❌ | d
LoopMove | ❌ | d
SGroupDataMove | ❌ | d
EnhancedFlagMove | ❌ | p
ImageMove | ❌ | offset (readonly)
MultitailArrowMove | ❌ | offset (readonly)

**Fix:**
Every move operation now follows the same pattern as `AtomMove.isDummy()` returns true when the delta is` (0, 0)`, preventing zero-move gestures from polluting the undo stack.

**Comments**:
The existing overrides fall into two fundamentally different patterns:
Attr ops (`AtomAttr`, `BondAttr`, etc.) — must compare against live `restruct` state → signature `isDummy(restruct: ReStruct): boolean`
Move ops — pure check on the delta vector, no `restruct` needed  
**making `isDummy()` universal isn't viable** — the two patterns are incompatible. What is viable is eliminating the 9 identical move-op overrides by extracting a `MoveOperation` base class.
The individual overrides are only 3 lines each and the pattern is obvious. The refactor would be justified only **if more move operations are expected to be added** — otherwise it's over-engineering

**'any'** has been partially narrowed in '...Move.ts'.

## Check list
- [ ] unit-tests written
- [ ] e2e-tests written
- [ ] documentation updated

Fix #9743


DmirtyUsov:
See PR #9904

PR #9960, opened 2026-05-05
#9920 – Refactor: Redundant casts and non-null assertions should be a…

### LINKED ISSUE #9920: Refactor: Redundant casts and non-null assertions should be avoided
**Problem:**
Redundant casts and non-null assertions should be avoided

**Why is this an issue?**

In TypeScript, casts and non-null assertions are two mechanisms used to inform the TypeScript compiler about the intended types of variables, expressions, or values in the code. They are used to help the compiler understand the types more accurately and to handle certain situations where type inference might not be sufficient:

A type assertion tells the compiler to treat the value as the specified type, even if the compiler’s type inference suggests a different type.
A non-null assertion is a way to tell the TypeScript compiler explicitly that you are certain that a variable will not be null or undefined, and you want to access its properties or methods without any null checks.
However, a redundant cast occurs when you perform a type assertion that is not needed because the compiler already knows the type based on the context or explicit type declarations. Similarly, a redundant non-null assertion occurs when you use the non-null assertion operator ! to assert that a variable is not null or undefined, but the compiler already knows that based on the context.

Both redundant casts and redundant non-null assertions should be avoided in TypeScript code, as they add unnecessary noise, clutter the code, and lead to confusion.

```js
function getName(x?: string | Person) {
  if (x) {
    console.log("Getting name for " + x!); // Noncompliant: 'x' is known to be defined here

    if (typeof x === "string") {
      return (x as string); // Noncompliant: 'x' is known to be a string here
    } else {
      return (x as Person).name; // Noncompliant: 'x' is defined and not a string, thus a 'Person' here
    }
  }
  return "NoName";
}
```
Remove all the redundant casts and non-null assertions based on the contextual typing information, as inferred by the TypeScript compiler.

```js
function getName(x?: string | Person) {
  if (x) {
    console.log("Getting name for " + x);

    if (typeof x === "string") {
      return x;
    } else {
      return x.name;
    }
  }
  return "NoName";
}

```

**Problem locations:**
packages/ketcher-core/src/application/editor/actions/bond.ts:194
packages/ketcher-core/src/application/editor/actions/chain.ts:81
packages/ketcher-core/src/application/editor/actions/chain.ts:82
packages/ketcher-core/src/application/editor/actions/template.ts:81
packages/ketcher-core/src/application/editor/actions/template.ts:101
packages/ketcher-core/src/application/render/restruct/rergroup.ts:206
packages/ketcher-core/src/application/render/restruct/visel.ts:74
packages/ketcher-macromolecules/src/components/monomerLibrary/RnaBuilder/RnaEditor/RnaEditorExpanded/RnaEditorExpanded.tsx:537
packages/ketcher-react/src/components/Dialog/Dialog.tsx:94
packages/ketcher-react/src/script/ui/component/form/form/form.tsx:150
packages/ketcher-react/src/script/ui/component/form/form/form.tsx:176

aslmbk:
Address Sonar findings by replacing unsafe casts ```!``` with narrowing and
clearer types.

- ketcher-core: template bond placement, rergroup drawing iteration,
  related editor/render tweaks
- ketcher-react: form state updates without Record casts; Dialog if touched
- ketcher-macromolecules: RNA preset save uses narrowed preset name for
  setUniqueNameError (fixes TS2345 in production build)

## Check list
- [ ] unit-tests written
- [ ] e2e-tests written
- [ ] documentation updated

### DIFF
--- a/packages/ketcher-core/src/application/editor/actions/template.ts
+++ b/packages/ketcher-core/src/application/editor/actions/template.ts
@@ -32,6 +32,7 @@ import { fromPaste } from './paste';
 import utils from '../shared/utils';
 import { fromSgroupAddition } from './sgroup';
 import type { ReStruct } from 'application/render';
+import { KetcherLogger } from 'utilities';
 
 const benzeneMoleculeName = 'Benzene';
 const cyclopentadieneMoleculeName = 'Cyclopentadiene';
@@ -82,7 +83,10 @@ function extraBondAction(
   } else {
     const pivotAtom = restruct.molecule.atoms.get(aid);
     if (!pivotAtom) {
-      throw new Error(`Atom ${aid} not found`);
+      KetcherLogger.error(
+        `template.ts::extraBondAction: atom ${aid} not found`,
+      );
+      return { action, aid1: aid };
     }
 
     const operation = new AtomAdd(
@@ -93,7 +97,10 @@ function extraBondAction(
     action.addOp(operation);
     const newAid = operation.data.aid;
     if (typeof newAid !== 'number') {
-      throw new Error('Atom id was not assigned after AtomAdd');
+      KetcherLogger.error(
+        'template.ts::extraBondAction: atom id was not assigned after AtomAdd',
+      );
+      return { action, aid1: aid };
     }
 
     action.addOp(new BondAdd(aid, newAid, { type: 1 }).perform(restruct));
--- a/packages/ketcher-core/src/application/editor/actions/template.ts
+++ b/packages/ketcher-core/src/application/editor/actions/template.ts
@@ -78,27 +78,30 @@ function extraBondAction(
     );
     action = actionRes[0];
     action.operations.reverse();
-    additionalAtom = actionRes[2] as number;
+    additionalAtom = actionRes[2];
   } else {
+    const pivotAtom = restruct.molecule.atoms.get(aid);
+    if (!pivotAtom) {
+      throw new Error(`Atom ${aid} not found`);
+    }
+
     const operation = new AtomAdd(
       { label: 'C', fragment: frid },
-      new Vec2(1, 0)
-        .rotate(angle)
-        .add((restruct.molecule.atoms.get(aid) as Atom).pp)
-        .get_xy0(),
-    ).perform(restruct) as AtomAdd;
+      new Vec2(1, 0).rotate(angle).add(pivotAtom.pp).get_xy0(),
+    ).perform(restruct);
 
     action.addOp(operation);
-    action.addOp(
-      new BondAdd(aid, operation.data.aid as number, { type: 1 }).perform(
-        restruct,
-      ),
-    );
+    const newAid = operation.data.aid;
+    if (typeof newAid !== 'number') {
+      throw new Error('Atom id was not assigned after AtomAdd');
+    }
 
-    additionalAtom = operation.data.aid as number;
+    action.addOp(new BondAdd(aid, newAid, { type: 1 }).perform(restruct));
+
+    additionalAtom = newAid;
   }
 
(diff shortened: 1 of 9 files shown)


packages/ketcher-core/src/application/editor/actions/template.ts
```diff
@@ -78,27 +78,30 @@ function extraBondAction(
     );
     action = actionRes[0];
     action.operations.reverse();
-    additionalAtom = actionRes[2] as number;
+    additionalAtom = actionRes[2];
   } else {
+    const pivotAtom = restruct.molecule.atoms.get(aid);
+    if (!pivotAtom) {
+      throw new Error(`Atom ${aid} not found`);
```
rrodionov91: Instead of throwing error please Use KetcherLogger to log error and then return from function.

packages/ketcher-core/src/application/editor/actions/template.ts
```diff
@@ -78,27 +78,30 @@ function extraBondAction(
...
 
     action.addOp(operation);
-    action.addOp(
-      new BondAdd(aid, operation.data.aid as number, { type: 1 }).perform(
-        restruct,
-      ),
-    );
+    const newAid = operation.data.aid;
+    if (typeof newAid !== 'number') {
+      throw new Error('Atom id was not assigned after AtomAdd');
```
rrodionov91: Instead of throwing error please Use KetcherLogger to log error and then return from function.

packages/ketcher-react/src/script/ui/component/form/form/form.tsx
```diff
@@ -644,20 +644,26 @@ function propSchema(
...
+    return {};
   }
 
-  const instanceObj = instance as Record<string, unknown>;
+  const res: Record<string, unknown> = {};
   Object.keys(schema.properties).forEach((p) => {
-    if (p in instanceObj)
-      res[p] = instanceObj[serializeMap[p]] ?? instanceObj[p];
+    if (p in instance) res[p] = instance[serializeMap[p]] ?? instance[p];
   });
```
rrodionov91: I think it's better to revert this change and try to do it in separate PR in more readable way

rrodionov91 (changes requested):

packages/ketcher-core/src/application/editor/actions/template.ts
aslmbk: done

packages/ketcher-core/src/application/editor/actions/template.ts
aslmbk: done

packages/ketcher-react/src/script/ui/component/form/form/form.tsx
aslmbk: done

AlexeyGirin:

packages/ketcher-core/src/application/editor/actions/template.ts
```diff
@@ -78,27 +79,36 @@ function extraBondAction(
...
+    ).perform(restruct);
 
     action.addOp(operation);
-    action.addOp(
-      new BondAdd(aid, operation.data.aid as number, { type: 1 }).perform(
-        restruct,
-      ),
-    );
+    const newAid = operation.data.aid;
+    if (typeof newAid !== 'number') {
```
AlexeyGirin: Please use isNumber instead of == 'number'
and
rename `newAid` to `newAtomId`

AlexeyGirin (dismissed):

packages/ketcher-core/src/application/editor/actions/template.ts
aslmbk: done

rrodionov91 (approved):

PR #9981, opened 2026-05-07
#9927 - Refactor: Bunch of sonar problems

### LINKED ISSUE #9927: Refactor: Bunch of sonar problems
### **Problem 1:**
Unused React typed props should be removed

**Why is this an issue?**
Methods that are never executed are dead code and should be removed. Cleaning out dead code decreases the size of the maintained codebase, making it easier to understand the program and preventing bugs from being introduced.

When using React class components, all non-React lifecycle methods should be called within the scope of the component. If a method is only called from outside the class, consider using props to interact with the component and re-render if needed, as React encourages data-driven components.

```
class Profile extends React.Component {
  render(props) {
    return <h1>{ props.name }</h1>;
  }

  getDefaultName() { // Noncompliant: this method is never used and is a dead code
    
return 'John Smith';
  }
}
```

To fix the issue, remove the dead code or call the method from within the component scope.

```
class Profile extends React.Component {
  render(props) {
    return <h1>{ props.name || getDefaultName() }</h1>;
  }

  getDefaultName() {
    
return 'John Smith';
  }
}
```

**Problem locations:**
packages/ketcher-react/src/script/ui/component/form/form/form.tsx:180

-------------------

### **Problem 2:**
Unused assignments should be removed

**Why is this an issue?**
Dead stores refer to assignments made to local variables that are subsequently never used or immediately overwritten. Such assignments are unnecessary and don’t contribute to the functionality or clarity of the code. They may even negatively impact performance. Removing them enhances code cleanliness and readability. Even if the unnecessary operations do not do any harm in terms of the program’s correctness, they are - at best - a waste of computing resources.

**Exceptions**

The rule ignores

- Initializations to `-1`, `0`, `1`, `undefined`, `[]`, `{}`, `true`, `false` and `""`.
- Variables that start with an underscore (e.g. `_unused`) are ignored.
- Assignment of `null` is ignored because it is sometimes used to help garbage collection
- Increment and decrement expressions are ignored because they are often used idiomatically instead of x+1
- This rule also ignores variables declared with object destructuring using rest syntax (used to exclude some properties from object)

```
let {a, b, ...rest} = obj;  // 'a' and 'b' are compliant
doSomething(rest);

let [x1, x2, x3] = arr;     // 'x1' is noncompliant, as omitting syntax can be used: "let [, x2, x3] = arr;"
doSomething(x2, x3);
```

**How can I fix it**
Remove the unnecesarry assignment, then test the code to make sure that the right-hand side of a given assignment had no side effects (e.g. a method that writes certain data to a file and returns the number of written bytes).

**Noncompliant code example**
```
function foo(y) {
  let x = 100; // Noncompliant: dead store
  x = 150;     // Noncompliant: dead store
  x = 200;
  
return x + y;
}
```

**Compliant solution**
```
function foo(y) {
  let x = 200; // Compliant: no unnecessary assignment
  
return x + y;
}
```

**Problem locations:**
packages/ketcher-core/src/domain/serializers/mol/v2000.ts:591

-------------------

### **Problem 3:**
Type constituents of unions and intersections should not be redundant

**Why is this an issue?**
In unions and intersections, redundant types should not be used.

When defining a union or intersection in TypeScript, it is possible to mistakenly include type constituents that encompass other constituents, that don’t have any effect, or that are more restrictive. For instance,

- The type `something` in `any | something` is redundant because `any` covers all possible types, whatever `something` is.
- The types `never` in unions like `never | something` or `unknown` in intersections like `unknown & something` are effectless.
- More restrictive types in intersections like the literal type `1` in `1 & number` reduce the set of possible values to specific ones.

Eliminating redundant types from a union or intersection type simplifies the code and enhances its readability. Moreover, it provides a clearer representation of the actual values that a variable can hold.

**How can I fix it**
The redundant and overridden types should be removed.
**Noncompliant code example**

```
type UnionWithAny = any | 'redundant'; // Noncompliant
type UnionWithNever = never | 'override'; // Noncompliant
type UnionWithLiteral = number | 1; // Noncompliant

type IntersectionWithAny = any & 'redundant'; // Noncompliant
type IntersectionWithUnknown = string & unknown; // Noncompliant
type IntersectionWithLiteral = string & 'override'; // Noncompliant

```

**Compliant solution**
```
type UnionWithAny = any;
type UnionWithNever = never;
type UnionWithLiteral = number;

type IntersectionWithAny = any;
type IntersectionWithUnknown = string;
type IntersectionWithLiteral = 'override';
```

**Problem locations:**
packages/ketcher-react/src/script/ui/views/modal/components/document/Save/Save.tsx;144

-------------------

### **Problem 4:**
Type aliases should be used

**Why is this an issue?**
In TypeScript, a type alias is a way to give a name to a specific type. It allows you to create a new name for an existing type, making your code more expressive and readable. This is especially useful when you are working with complex or lengthy types that you use frequently.

Type aliases should be preferred over complex types like unions or intersections for several reasons:

- Type aliases can make your code more readable and maintainable by giving meaningful names to complex types. When you use a type alias, it becomes clear what the type represents, making it easier for other developers (including your future self) to understand the code.
- Type aliases promote code reusability. When you define a type alias for a complex type, you can use that alias in multiple places throughout your codebase, reducing duplication and promoting consistency.
- Type aliases allow you to abstract away the underlying complexity of types. This promotes encapsulation by hiding implementation details behind a well-named alias, allowing you to change the underlying type in the future without affecting the code that uses the alias.
- If you need to modify a complex type, using a type alias means you only need to change the type definition in one place. This change will automatically apply to all usages of the alias.
- Type aliases communicate the intent of the type, making it easier for other developers to understand what the type represents. Complex unions and intersections, on the other hand, might require additional comments or documentation to explain their purpose.
- Using type aliases can help you avoid excessive nesting of complex types. Nested unions and intersections can quickly become hard to read and maintain, and type aliases can simplify the type definitions.

This rule enforces the rule of three for code refactoring and reports unions and intersections with three or more constituents appearing at least three times in the codebase.

```
function foo(x: string | null | number) { // Noncompliant: The union has three constituents and is duplicated three times in the code
  
/* ... */
}

let bar: string | null | number = /* ... */;

function baz(): string | null | number {
  
/* ... */
}
```

You should define a type alias for the union or intersection duplicated in the code and replace all their occurrences with the alias name.

```
type MyType = string | null | number;

function foo(x: MyType) {
  
/* ... */
}

let bar: MyType = /* ... */;

function baz(): MyType {
  
/* ... */
}
```

The rule disregards nullable types, irrespective of their frequency within the codebase. This includes types of the form T | null | undefined, where T can represent any type.

**Problem locations:**
packages/ketcher-react/src/script/ui/views/modal/components/process/Recognize/Recognize.tsx:46
packages/ketcher-react/src/script/ui/views/modal/components/process/Recognize/Recognize.tsx:102
packages/ketcher-react/src/script/ui/views/modal/components/process/Recognize/Recognize.tsx:274

-------------------

### **Problem 5:**
Two branches in a conditional structure should not have exactly the same implementation

**Why is this an issue?**
When the same code is duplicated in two or more separate branches of a conditional, it can make the code harder to understand, maintain, and can potentially introduce bugs if one instance of the code is changed but others are not.

Having two `cases` in a `switch` statement or two branches in an `if` chain with the same implementation is at best duplicate code, and at worst a coding error.

```
if (a >= 0 && a < 10) {
  
doFirstThing();
  doTheThing();
}
else if (a >= 10 && a < 20) {
  doTheOtherThing();
}
else if (a >= 20 && a < 50) {
  doFirstThing();
  doTheThing();  // Noncompliant; duplicates first condition
}
else {
  doTheRest();
}
```

```
switch (i) {
  case 1:
    doFirstThing();
    doSomething();
    break;
  case 2:
    doSomethingDifferent();
    break;
  case 3:  // Noncompliant; duplicates case 1's implementation
    doFirstThing();
    doSomething();
    break;
  
default:
    doTheRest();
}
```

If the same logic is truly needed for both instances, then:

- in an `if` chain they should be combined

```
if ((a >= 0 && a < 10) || (a >= 20 && a < 50)) { // Compliant
  
doFirstThing();
  doTheThing();
}
else if (a >= 10 && a < 20) {
  doTheOtherThing();
}
else {
  doTheRest();
}
```

- for a `switch`, one should fall through to the other

```
switch (i) {
  case 1:
  case 3: // Compliant
    
doFirstThing();
    doSomething();
    break;
  case 2:
    doSomethingDifferent();
    break;
  default:
    doTheRest();
}
```

When all blocks are identical, either this rule will trigger if there is no default clause or rule [S3923](https://sonar.epam.com/sonarqube/coding_rules#rule_key=javascript%3AS3923) will raise if there is a default clause.

**Exceptions**

Unless all blocks are identical, blocks in an if chain that contain a single line of code are ignored. The same applies to blocks in a switch statement that contains a single line of code with or without a following break.
```
if (a == 1) {
  doSomething();  // Compliant, usually this is done on purpose to increase the readability
} else if (a == 2) {
  doSomethingElse();
} else {
  doSomething();
}
```

**Problem locations:**
packages/ketcher-react/src/script/ui/views/components/StructEditor/StructEditor.tsx:305 - This case's code block is the same as the block for the case on line 288.

-------------------

### **Problem 6:**
React legacy lifecycle methods should not be used

**Why is this an issue?**
React works in two phases: render and commit.

- The render phase determines what changes need to be made to e.g. the DOM. During this phase, React calls render and then compares the result to the previous render.
- The commit phase is when React applies any changes. (In the case of React DOM, this is when React inserts, updates, and removes DOM nodes.) React also calls lifecycles like `componentDidMount` and `componentDidUpdate` during this phase.

Render phase lifecycles include among others, the following lifecycle methods:

- `componentWillMount` (or its alias `UNSAFE_componentWillMount`)
- `componentWillReceiveProps` (or its alias `UNSAFE_componentWillReceiveProps`)
- `componentWillUpdate` (or its alias `UNSAFE_componentWillUpdate`)

These are considered unsafe and also happen to be the lifecycles that cause the most confusion within the React community and tend to encourage unsafe coding practices.

```
class Foo extends React.Component {
  UNSAFE_componentWillMount() {}         // Noncompliant
  UNSAFE_componentWillReceiveProps() {}  // Noncompliant
  UNSAFE_componentWillUpdate() {}        // Noncompliant
}
```

**How can I fix it**
Instead of `componentWillUpdate`, use `getSnapshotBeforeUpdate` together with `componentDidUpdate`. The `getSnapshotBeforeUpdate` lifecycle is called right before mutations are made. The return value for this lifecycle will be passed as the third parameter to `componentDidUpdate`.

Instead of `componentWillReceiveProps`, Use `getDerivedStateFromProps` together with `componentDidUpdate`. The `getDerivedStateFromProps` lifecycle is invoked after a component is instantiated as well as before it is re-rendered. It can return an object to update state, or null to indicate that the new props do not require any state updates.

As for `componentWillMount`, React will call it immediately after the constructor. It only exists for historical reasons and should not be used. Instead, use one of the alternatives:

- To initialize state, declare `state` as a class field or set `this.state` inside the `constructor`.
- If you need to run a side effect or set up a subscription, move that logic to `componentDidMount` instead.

**Noncompliant code example**
```
class myComponent extends React.Component {
  constructor(props) {
      super(props);
  }

  componentWillMount() { // Noncompliant: "componentWillMount" is deprecated
      
if (localStorage.getItem("token")) {
          this.setState({logged_in: true});
      }
  }
  // ...
}
```

**Compliant solution**
```
class myComponent extends React.Component {
  constructor(props) {
      super(props);

      if (localStorage.getItem("token")) {
          this.setState({logged_in: true});
      }
  }
  // ...
}
```

**Problem locations:**
packages/ketcher-react/src/script/ui/views/components/StructEditor/StructEditor.tsx:200

-------------------

### **Problem 7:**
Promises should not be misused

**Why is this an issue?**
Promises need to be resolved or awaited to return the expected value, otherwise, they return the promise object.

**Unresolved promises:**

Forgetting to await a promise is a frequent mistake. There are places where the use of a promise object is confusing or unclear because the developer forgot to resolve it.

This rule forbids returning promises where another type is expected such as in:

- conditionals
- void returns
- spread operators

**What is the potential impact?**

Using a promise instead of its resolved value can have unexpected results leading to bugs.

- In conditionals, it will always return a truthy value.
- In places where the expected type is void, returning a promise is often a mistake.
- Using the spread operator on a promise will raise an exception.

The executor function of a promise can also be an async function. However, this usually denotes a mistake:

- If an async executor function throws an error, the error won’t cause the created promise to reject and will be lost. Therefore, this could make it difficult to debug and handle runtime errors.
- If a promise executor function is using await, this means that it’s not necessary to use the Promise constructor, or the scope of the Promise constructor can be reduced.

**Exceptions**

This rule can be ignored for promises that you know will always resolve like timers.

````
await new Promise(resolve => time.setTimeout(1000));
````

**How can I fix it**
If you mistakenly treated a promise as its resolved value, you can ensure it is properly resolved by using await or resolve on the promise. In some cases, you may need to use an "immediately invoked function expression" (IIFE):

```
(async function foo() {
  const result = await bar();
  // work with result
})();
```

**Noncompliant code example**

```
const promise = new Promise((resolve, reject) => {
  // ...
  resolve(false)
});
if (promise) {
  
// ...
}
```

**Compliant solution**

```
const promise = new Promise((resolve, reject) => {
  // ...
  resolve(false)
});
if (await promise) {
  
// ...
}
```

**Noncompliant code example**

```
const p = new Promise(async (resolve, reject) => {
  
doSomething('Hey, there!', function(error, result) {
    if (error) {
      reject(error);
      return;
    }
    await saveResult(result)
    
resolve(result);
  });
});

await p;
```

**Compliant solution**

```
const p = new Promise((resolve, reject) => {
  
doSomething('Hey, there!', function(error, result) {
    if (error) {
      reject(error);
      return;
    }
    resolve(result);
  });
});

const result = await p;
await saveResult(result);
```

**Noncompliant code example**

```
apiCalls.forEach(async (apiCall) => {
  
await apiCall.send();
});
```

**Compliant solution**

```
for (const apiCall of apiCalls) {
  
await apiCall.send();
}
```

**How does this work?**

In JavaScript, a promise is a mechanism to perform tasks asynchronously. To this end, the language provides the Promise object which represents the eventual completion or failure of an asynchronous operation and its resulting value. A promise can be created with the Promise constructor accepting an executor function as an argument, which has resolve and reject parameters that are invoked when the promise completes or fails.

The logic of the promise is executed when it is called, however, its result is obtained only when the promise is resolved or awaited.

**Problem locations:**
packages/ketcher-react/src/script/ui/views/modal/components/document/Save/Save.tsx:520

-------------------

### **Problem 8:**
Parameters should be passed in the correct order

**Why is this an issue?**
Consistent naming between arguments and parameters reduces the chances of making mistakes, such as passing the wrong value to a parameter or omitting an argument in a function call. When the names of arguments in a function call match the names of the function parameters, it contributes to clearer, more readable code.

However, when the names match but are passed in a different order than their declaration in the function signature, it may indicate a mistake in the parameter order, likely leading to unexpected results.

```
function divide(dividend, divisor) {
  return dividend / divisor;
}

function doTheThing() {
  const dividend = 15;
  const divisor = 5;

  const result = divide(divisor, dividend); // Noncompliant: not the expected result
  //...
}
```

Ensure that the arguments passed to the function are in the correct order, according to the function signature.

```
function divide(dividend, divisor) {
  return dividend / divisor;
}

function doTheThing() {
  const dividend = 15;
  const divisor = 5;

  const result = divide(dividend, divisor);
  //...
}
```

**Exceptions**

Swapped arguments that are compared beforehand in an enclosing if statement are ignored:

```
function divide(dividend, divisor) {
  return dividend / divisor;
}

function doTheThing() {
  const dividend = 5;
  const divisor = 15;
  if (divisor > dividend) {
    const result = divide(divisor, dividend);
    //...
  }
}
```

**Problem locations:**
packages/ketcher-react/src/script/editor/Editor.ts:1451

-------------------

### **Problem 9:**
Non-interactive DOM elements should not have interactive ARIA roles

**Why is this an issue?**
Non-interactive DOM elements are those that do not have built-in interactivity or do not require user interaction. Examples include `<div>`, `<p>`, ``, `<h1>` to `<h6>`, and `<ul>`, among others. These elements are typically used to structure content and layout but do not have any inherent interactive behavior.

Interactive ARIA roles, on the other hand, are used to make elements interactive and accessible. These roles include `button`, `link`, `checkbox`, `menuitem`, `tab`, and others. They are used to communicate the type of interaction that users can expect from an element.

Non-interactive DOM elements should not use interactive ARIA roles because it can confuse assistive technologies and their users. For example, if a `<div>` (a non-interactive element) is given an interactive role like "button", assistive technologies like screen readers will announce it as a button. However, since `<div>` doesn’t have the inherent behavior of a button, it can confuse users who expect it to behave like a button when interacted with.

This can lead to a poor user experience and can make the website less accessible for users relying on assistive technologies. Therefore, it’s important to ensure that non-interactive DOM elements are not given interactive ARIA roles.

**How can I fix it**
Ensure that non-interactive DOM elements do not use interactive ARIA roles.
**Noncompliant code example**

```
function myListElement() {
    return <li role="button">Foo</li>; // Noncompliant; "li" isn't interactive, but "button" is
}
```

**Compliant solution**

```
function myListElement() {
    return <li role="listitem">Foo</li>;
}
```

**Problem locations:**
packages/ketcher-react/src/script/ui/component/actionmenu.tsx:297

-------------------

### **Problem 10:**
Constructors should not contain asynchronous operations

**Why is this an issue?**
Constructors should not be asynchronous because they are meant to initialize an instance of a class synchronously. While you can technically run a promise in a constructor, the instance will be returned before the asynchronous operation completes, which can lead to potential issues if the rest of your code expects the object to be fully initialized. While returning the promise may seem the solution for this problem, it is not standard practice and can lead to unexpected behavior. Returning a promise (or, in general, any object from another class) from a constructor can be misleading and is considered a bad practice as this leads to unexpected results with inheritance and the `instanceof` operator.

```
class MyClass {
  constructor() {
    Promise.resolve().then(() => this.data = fetchData()); // Noncompliant, this.data will be undefined in the new instance
  }
}
```

Instead, you can create an asynchronous method that performs the necessary asynchronous operations after the object has been constructed

```
class MyClass {
  constructor() {
    this.data = null;
  }

  async initialize() {
    this.data = await fetchData();
  }
}

(async () => {
  const myObject = new MyClass();
  await myObject.initialize();
})();
```

Otherwise, you can use a static factory method that returns a promise resolving to an instance of the class after performing the asynchronous operations.

```
class MyClass {
  constructor(data) {
    this.data = data;
  }

  static async create() {
    const data = await fetchData();
    return new MyClass(data);
  }
}

(async () => {
  const myObject = await MyClass.create();
})();
```

Using an asynchronous factory function or an initialization method provides a more conventional, readable, and maintainable approach to handling asynchronous initialization in JavaScript.

**Problem locations:**
packages/ketcher-react/src/script/ui/component/view/openbutton.tsx:53

-------------------

### **Problem 11:**
Boolean checks should not be inverted
Use the opposite operator (!==) instead.

**Why is this an issue?**
It is needlessly complex to invert the result of a boolean comparison. The opposite comparison should be made instead.
**Noncompliant code example**
````
if (!(a === 2)) { ... }  // Noncompliant
````

**Compliant solution**

````
if (a !== 2) { ... }
````

**Problem locations:**
packages/ketcher-core/src/application/editor/Editor.ts:1052

-------------------

### **Problem 12:**
All branches in a conditional structure should not have exactly the same implementation

**Why is this an issue?**
Having all branches of a `switch` or `if` chain with the same implementation indicates a problem.

```
In the following code:

if (b == 0) {  // Noncompliant
  doOneMoreThing();
}
else {
  doOneMoreThing();
}

let a = (b == 0) ? getValue() : getValue();   // Noncompliant

switch (i) {  // Noncompliant
  case 1:
    doSomething();
    break;
  case 2:
    doSomething();
    break;
  case 3:
    doSomething();
    break;
  default:
    doSomething();
}
```

Either there is a copy-paste error that needs fixing or an unnecessary switch or if chain that should be removed.
**Exceptions**

This rule does not apply to `if` chains without `else`, nor to `switch` without a `default` clause.

```
if(b == 0) {    //no issue, this could have been done on purpose to make the code more readable
  doSomething();
} else if(b == 1) {
  doSomething();
}
```

**Problem locations:**
packages/ketcher-core/src/application/render/restruct/remultitailArrow.ts:298

-------------------

### **Problem 13:**
"undefined" should not be passed as the value of optional parameters

**Why is this an issue?**
In TypeScript, optional and default parameters are features that enhance the flexibility and usability of functions by allowing you to define parameters that can be omitted during function calls or have default values assigned to them when not provided explicitly:

- Optional parameters are denoted by adding a question mark (`?`) after the parameter name in the function declaration. These parameters can be omitted when calling the function, and TypeScript will assign them a value of undefined if they are not provided.
- Default parameters are used to assign a default value to a parameter in case the caller does not provide a value for that parameter. Default values are specified in the function declaration using the assignment operator (`=`) followed by the default value.

When a parameter is defined as optional, it automatically allows that parameter to be omitted during function calls. If you explicitly pass undefined as the value for an optional parameter, it contradicts the purpose of making the parameter optional, making the code less readable and maintainable.

Similarly, when a parameter has a default value, it means that if the caller omits the argument during function calls, the default value will be used automatically. There is no need to pass `undefined` explicitly for default parameters, except when they come before a required parameter.

This rule checks that the last argument of a function call is not redundant with regard to the function’s signature.

```
function foo(x: number, y: string = "default", z?: number) {
  // ...
}

foo(42, undefined); // Noncompliant: 'undefined' is redundant
foo(42, undefined, undefined); // Noncompliant: Both 'undefined' are redundant
foo(42, undefined, 5); // Compliant: 'undefined' is required to get the second parameter's default value
```

Instead of explicitly passing `undefined`, simply omit the optional argument when calling the function to let TypeScript handle the default value correctly.

```
function foo(x: number, y: string = "default", z?: number) {
  // ...
}

foo(42);
```

**Problem locations:**
packages/ketcher-react/src/script/ui/views/components/MonomerCreationWizard/RnaPresetAttachmentPointValidation.test.ts:112

-------------------

### **Problem 14:**
"RegExp.exec()" should be preferred over "String.match()"

**Why is this an issue?**
`String.match()` behaves the same way as `RegExp.exec()` when the regular expression does not include the global flag `g`. While they work the same, `RegExp.exec()` can be slightly faster than `String.match()`. Therefore, it should be preferred for better performance.

The rule reports an issue on a call to `String.match()` whenever it can be replaced with semantically equivalent `RegExp.exec()`.
````
'foo'.match(/bar/);
````

Rewrite the pattern matching from `string.match(regex)` to `regex.exec(string)`.
````
/bar/.exec('foo');
````

**Problem locations:**
packages/ketcher-core/src/domain/serializers/mol/v3000.ts:51

### PR DESCRIPTION
**Pull request by sabaPailodze**
## Summary

Fixes Sonar code quality findings across core and React packages by removing dead/redundant code, suppressing one justified legacy lifecycle with a documented rationale, improving async handling, and tightening accessibility/type usage.

### Code Quality Cleanups

**Problem:**
Several Sonar findings reported dead assignments, redundant conditions, redundant union types, duplicated switch branches, inverted boolean checks, and non-preferred regex usage.

**Solution:**
- Removed unused assignment in MOL V2000 parsing
- Replaced `String.match()` with `RegExp.exec()` in MOL V3000 parsing
- Simplified inverted boolean comparison in core editor logic
- Removed redundant ternary branch in multitail arrow rendering (both branches returned the same value)
- Simplified redundant `Record<string, unknown> | unknown` type to `unknown`
- Added a reusable type alias for repeated recognize image file union type
- Removed explicit `undefined` arguments for optional test parameters

### React Lifecycle and Async Handling

**Problem:**
Some React class components used deprecated or unsafe patterns, including async work in a constructor and externally consumed class methods. `StructEditor` also had a bug where `serverSettings`-only changes were not applied independently of `options` changes.

**Solution:**
- Kept `UNSAFE_componentWillReceiveProps` in `StructEditor` with `// NOSONAR` suppression — `componentDidUpdate` fires after React commits and causes extra re-renders on struct changes, which reintroduced white-screen states in CI screenshot tests; the suppression is intentional and documented
- Fixed `setupEditor` to apply `serverSettings` changes independently when `options` has not changed
- Moved async `fileOpener` initialization from `OpenButton` constructor into `componentDidMount`/`componentDidUpdate` lifecycle methods
- Added unmount guard and request ID counter to prevent stale async opener results from overwriting a newer initialization
- Extracted form field state creation from the `Form` class method inline, keeping the context API compatible

### API and Accessibility Improvements

**Problem:**
Optional positional parameters were easy to misuse, and menu list items used interactive ARIA semantics on non-interactive DOM elements.

**Solution:**
- Replaced optional positional monomer creation parameters with a named options object
- Updated the monomer creation wizard call site accordingly
- Changed action menu list items to presentational role and removed non-interactive tab focus
- Wrapped async save format change handler so the form `onChange` callback does not return a `Promise`

## Validation

- TypeScript checks passed for `ketcher-core` and `ketcher-react`
- ESLint checks passed for `ketcher-core` and `ketcher-react`
- Focused unit test passed for RNA preset attachment point validation
- Prettier check passed for touched files
- All previously failing CI screenshot tests verified locally and in Docker (linux)

## Files Modified

- `packages/ketcher-core/src/application/editor/Editor.ts`
- `packages/ketcher-core/src/application/render/restruct/remultitailArrow.ts`
- `packages/ketcher-core/src/domain/serializers/mol/v2000.ts`
- `packages/ketcher-core/src/domain/serializers/mol/v3000.ts`
- `packages/ketcher-react/src/script/editor/Editor.ts`
- `packages/ketcher-react/src/script/ui/component/actionmenu.tsx`
- `packages/ketcher-react/src/script/ui/component/form/form/form.tsx`
- `packages/ketcher-react/src/script/ui/component/view/openbutton.tsx`
- `packages/ketcher-react/src/script/ui/views/components/MonomerCreationWizard/MonomerCreationWizard.tsx`
- `packages/ketcher-react/src/script/ui/views/components/MonomerCreationWizard/RnaPresetAttachmentPointValidation.test.ts`
- `packages/ketcher-react/src/script/ui/views/components/StructEditor/StructEditor.tsx`
- `packages/ketcher-react/src/script/ui/views/modal/components/document/Save/Save.tsx`
- `packages/ketcher-react/src/script/ui/views/modal/components/process/Recognize/Recognize.tsx`
- `ketcher-autotests/tests/specs/Structure-Creating-&-Editing/Actions-With-Structures/R-group-tool/r-group-label-dialog.spec.ts-snapshots/R-Group-Label-Tool-Copy-Paste-actions-Structure-with-R-Group-label-1-chromium-linux.png`

## Check list
- [ ] unit-tests written
- [ ] e2e-tests written
- [ ] documentation updated

### CODE REVIEW

File: packages/ketcher-react/src/script/editor/Editor.ts [hunk 1]
```diff
@@ -1449,8 +1449,10 @@ class Editor implements KetcherEditor {
 
   finishNewMonomersCreation(
     monomersData,
-    rnaPresetName?: string,
-    phosphatePosition?: '3' | '5',
+    {
+      rnaPresetName,
+      phosphatePosition,
+    }: { rnaPresetName?: string; phosphatePosition?: '3' | '5' } = {},
```
svvald: Please create a type alias here

svvald (approved):

AlexeyGirin:

PR #10081, opened 2026-05-18
#9121 - The Attributes panel should be disabled for a non-continuous selection of the Nucleotide (preset) type

### LINKED ISSUE #9121: The Attributes panel should be disabled for a non-continuous selection of the Nucleotide (preset) type
**Source task:**
https://github.com/epam/ketcher/issues/8850

**Steps to Reproduce**
1. Go to Molecules mode
2. add to the Canvas `CCCCCC`
3. click on Create monomer button on the Left panel
4. choose Nucleotide (preset) type on the Attributes panel
5. make a non-continuous selection
Image

**Actual behavior**
Mark as component buttons are active
Image

**Expected behavior**
Mark as component buttons are disabled
Image

 - OS: MacOs Tahoe version 26.2
- Browser Chrome Version 143.0.7499.170
- Ketcher Version 3.13.0-rc.2
- Indigo Version 1.40.0-rc.1.0-g63d8f2068-wasm32-wasm-clang-19.0.0

mariam-khutuashvili:
Mark as component buttons are disabled

Summary

  When the user picks "Nucleotide (preset)" in the Monomer Creation Wizard and makes a non-continuous selection on
  canvas, the "Mark as base/sugar/phosphate" buttons, the Code input on the preset tab, and the form fields in
  MonomerCreationWizardFields are now disabled. A continuous selection keeps them enabled.

  Changes

  RnaPresetTabs.tsx — Compute isSelectionContinuous by passing the selected atoms plus the bonds derived from them
  (bonds whose both endpoints are in the selection) to Editor.isStructureContinuous. Derive isFieldsDisabled =
  hasSelectedAtoms && !isSelectionContinuous. Apply it to the "Mark as" button, the Code input, and pass it as a new
  prop to MonomerCreationWizardFields.

  MonomerCreationWizardFields.tsx — Add optional disabled?: boolean prop and combine it with the existing !type checks
  on the four input fields (disabled={!type || disabled}).

  Why derive bonds from atoms

  Editor.isStructureContinuous builds its adjacency list from the selection's bonds. Canvas selections often track only
  atoms (no bonds), which would make chemically connected atoms look isolated and falsely trigger the disabled state.
  Deriving bonds from the struct based on the atom selection gives the correct adjacency for the connectivity check.

  Scope

  Only affects the Nucleotide (preset) component. Other monomer types (Amino Acid, Sugar, Base, Phosphate, CHEM as
  single monomers) are unchanged. The existing submit-time "incontinuousStructure" notification still fires.

  Verification

  npm run test, npm run test:types, npm run build:packages, full Playwright autotests — all pass. Manual repro confirms
  buttons + fields disable on non-continuous selection and re-enable on continuous selection.

## Check list
- [ ] unit-tests written
- [ ] e2e-tests written
- [ ] documentation updated


packages/ketcher-react/src/script/ui/views/components/MonomerCreationWizard/MonomerCreationWizardFields.tsx
```diff
@@ -196,10 +198,10 @@ const MonomerCreationWizardFields = (
               onChange={(event: ChangeEvent<HTMLInputElement>) =>
                 onFieldChange('name', event.target.value)
               }
-              disabled={!type}
+              disabled={!type || disabled}
             />
           }
-          disabled={!type}
+          disabled={!type || disabled}
```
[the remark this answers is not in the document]
mariam-khutuashvili: disabled not wired to all fields — Propagated disabled to NaturalAnaloguePicker (added an optional disabled prop, OR'd with its internal !isNaturalAnalogueRequired gate) and to both HELM/BILN Autocomplete inputs. ModificationTypeDropdown / "Add modification type" are intentionally exempt: that section renders only for amino-acid monomers (displayModificationTypes = type === AminoAcid), which never coincides with the RNA-preset flow that passes disabled. Documented this in the new JSDoc.

packages/ketcher-react/src/script/ui/views/components/MonomerCreationWizard/RnaPresetTabs.tsx
```diff
@@ -76,6 +76,22 @@ export const RnaPresetTabs = (props: IRnaPresetTabsProps) => {
...
+        selectedAtomIdsSet.has(bond.begin) && selectedAtomIdsSet.has(bond.end),
+    )
+    .map(([id]) => id);
+  const isSelectionContinuous =
+    !hasSelectedAtoms ||
+    Editor.isStructureContinuous(struct, {
+      atoms: selectedAtomIds,
+      bonds: derivedBondIds,
+    });
+  const isFieldsDisabled = hasSelectedAtoms && !isSelectionContinuous;
```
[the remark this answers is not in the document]
mariam-khutuashvili: 1. Bond derivation runs every render — Wrapped the derivation + continuity check in useMemo keyed on [hasSelectedAtoms, selectedAtomIds, struct], with an early return false when nothing is selected.
2. Rationale not in code — Added a comment above the derivation explaining why bonds are derived from selected atoms instead of structureSelection.bonds (canvas selections often carry only atoms; an empty bond list would make a connected selection look non-continuous).

packages/ketcher-react/src/script/ui/views/components/MonomerCreationWizard/MonomerCreationWizardFields.tsx
```diff
@@ -46,6 +46,7 @@ interface IMonomerCreationWizardFieldsProps {
   ) => void;
   showNaturalAnalogue?: boolean;
   attachmentPointsExtra?: ReactNode;
+  disabled?: boolean;
```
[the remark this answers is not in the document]
mariam-khutuashvili: JSDoc on disabled prop — Added, documenting that callers layer it on the existing !type gate and noting the modification-types exemption.

packages/ketcher-react/src/script/ui/views/components/MonomerCreationWizard/MonomerCreationWizardFields.tsx
```diff
@@ -212,9 +220,10 @@ const MonomerCreationWizardFields = (
                   onFieldChange('naturalAnalogue', value);
                 }}
                 error={errors.naturalAnalogue}
+                disabled={disabled}
```
AlexeyGirin: Please do not disable ALL components but "Mark as ..." button only

AlexeyGirin (dismissed):

packages/ketcher-react/src/script/ui/views/components/MonomerCreationWizard/MonomerCreationWizardFields.tsx
mariam-khutuashvili: Reverted the disabled-propagation to all form fields and limited it to the "Mark as ..." button only. The Code, Name, Natural analogue, HELM and BILN inputs are now editable regardless of selection continuity; the only thing that gates on non-continuous selection is the "Mark as {component}" button. Renamed the helper from
  isFieldsDisabled to isSelectionNonContinuous to match the new scope, and removed the disabled prop from
  MonomerCreationWizardFields and NaturalAnaloguePicker since no one consumes it anymore.

packages/ketcher-react/src/script/ui/views/components/MonomerCreationWizard/RnaPresetTabs.tsx
```diff
@@ -91,6 +92,30 @@ export const RnaPresetTabs = (props: IRnaPresetTabsProps) => {
...
+      return false;
+    }
+    const selectedAtomIdsSet = new Set(selectedAtomIds);
+    const derivedBondIds = Array.from(struct.bonds.entries())
+      .filter(
+        ([, bond]) =>
+          selectedAtomIdsSet.has(bond.begin) &&
+          selectedAtomIdsSet.has(bond.end),
+      )
+      .map(([id]) => id);
```
svvald: I'm not sure why do we have to derive bonds from atoms if we could just use `structureSelection.bonds`?

packages/ketcher-react/src/script/ui/views/components/MonomerCreationWizard/RnaPresetTabs.tsx
```diff
@@ -91,6 +92,30 @@ export const RnaPresetTabs = (props: IRnaPresetTabsProps) => {
...
     monomerCreationState?.assignedAttachmentPoints ?? new Map();
   const struct = editor.struct();
+  const selectedAtomIds = structureSelection?.atoms ?? [];
+  // Derive bonds from the selected atoms instead of using
+  // `structureSelection.bonds`: canvas selections frequently carry only atoms
+  // (e.g. click-selection), and `Editor.isStructureContinuous` builds its
+  // adjacency list from bonds — an empty bond list would make a chemically
+  // connected selection look non-continuous. Memoized so the bond iteration
+  // does not re-run on every wizard keystroke / tab switch.
+  const isSelectionNonContinuous = useMemo(() => {
```
svvald: It's better to rename it to `isSelectionContinuous` as usually it's easier to perceive a "positive" boolean naming rather than "negative", it will also match with `Editor.isStructureContinuous` method naming

packages/ketcher-react/src/script/ui/views/components/MonomerCreationWizard/RnaPresetTabs.tsx
```diff
@@ -91,6 +92,30 @@ export const RnaPresetTabs = (props: IRnaPresetTabsProps) => {
...
+    }
+    const selectedAtomIdsSet = new Set(selectedAtomIds);
+    const derivedBondIds = Array.from(struct.bonds.entries())
+      .filter(
+        ([, bond]) =>
+          selectedAtomIdsSet.has(bond.begin) &&
+          selectedAtomIdsSet.has(bond.end),
+      )
+      .map(([id]) => id);
+    return !Editor.isStructureContinuous(struct, {
```
svvald: I'd suggest to extract `isStructureContinuous` method as a separate utility function so we won't have to import the whole `Editor` component as it could lead to the circular dependencies which we have recently refactored

svvald (changes requested):

packages/ketcher-react/src/script/ui/views/components/MonomerCreationWizard/RnaPresetTabs.tsx
mariam-khutuashvili: renamed isSelectionNonContinuous to the positive isSelectionContinuous and inverted it at the call site (disabled={!hasSelectedAtoms || !isSelectionContinuous}). This also keeps it consistent with the existing
  Editor.isStructureContinuous API.

packages/ketcher-react/src/script/ui/views/components/MonomerCreationWizard/RnaPresetTabs.tsx
mariam-khutuashvili: Extracted the connectivity check into a standalone util at packages/ketcher-react/src/script/editor/utils/structureContinuity.ts (isStructureContinuous + isAtomSelectionContinuous). RnaPresetTabs now imports that util instead of the Editor class, and Editor is a type-only import there, so there's no value import of Editor and no circular-dependency risk.
Editor.isStructureContinuous just delegates to the util, so the other two callers (useMarkAs, MonomerCreationWizard) are unchanged. Verified with npm run test:circ (dpdm) — no circular dependency found. Also added unit tests for the util.

packages/ketcher-react/src/script/ui/views/components/MonomerCreationWizard/RnaPresetTabs.tsx
mariam-khutuashvili: Switched to using structureSelection.bonds directly via isStructureContinuous(struct, structureSelection), so the bonds are no longer derived from the selected atoms. This also brings it in line with useMarkAs, which already passes the selection straight to Editor.isStructureContinuous. I removed the now-unused isAtomSelectionContinuous helper and updated the unit tests to cover isStructureContinuous directly.

packages/ketcher-react/src/script/editor/Editor.ts
```diff
@@ -911,54 +912,7 @@ class Editor implements KetcherEditor {
...
-        for (const neighbor of adjacencyList.get(nextAtomId) ?? []) {
-          if (!visited.has(neighbor)) {
-            queue.push(neighbor);
-          }
-        }
-      }
-    }
-
-    return visited.size === atomIds.length;
+    return isStructureContinuousUtil(struct, selection);
```
svvald: Nitpick but I assume you don't need to have this static method anymore just be a wrapper around an util function, just use util function as is where needed (you'll also avoid the name clash in import)

svvald (changes requested):

packages/ketcher-react/src/script/editor/Editor.ts
mariam-khutuashvili: Dropped the wrapper and its aliased import; both call sites now use the isStructureContinuous util directly, which removes the name clash.

svvald (approved):

PR #10091, opened 2026-05-19
#6784 -  Home and End keys support in Sequence mode

### LINKED ISSUE #6784: `Home` and `End` keys support in Sequence mode
**Background**
Pressing `Home` and `End` keys do nothing while in sequence edit mode

**Solution**
Add support of `Home` and `End` keys in edit mode
`Home` should move cursor to the beginning of current row (INPORTANT: Not to beginning of sequence, but row)
Image

`End` should move cursor to the end of current row (INPORTANT: Not to the end of sequence, but row)
Image

sabaPailodze:
## Summary

Adds `Home` and `End` key support in Sequence edit mode. `Home` moves the caret to the beginning of the current row, and `End` moves it to the end — neither jumps across row boundaries.

### Home and End Key Support

**Problem:** Pressing `Home` or `End` in sequence edit mode had no effect.

**Solution:** Added `move-caret-to-row-start` and `move-caret-to-row-end` keyboard actions bound to `Home` and `End`. Both are guarded against RNA Builder mode and non-edit mode, and reset `lastUserDefinedCaretPosition` so subsequent vertical arrow navigation stays consistent.

### "After Row End" Caret State

**Problem:** At the last position of a full row, `ArrowRight` jumped straight to the next row, skipping the visually expected "end of row" caret. `ArrowLeft` from the start of a row skipped the same position on the previous row. `End` had to land in this same state on full rows.

**Solution:** Introduced an `isCaretAfterRowEnd` flag on `SequenceRenderer` representing a virtual caret position after the last item of a full row. In this state the caret is rendered with a positive x-offset on the last node (`CARET_X_OFFSET_AFTER_NODE`) instead of the default left-side offset (`CARET_X_OFFSET_BEFORE_NODE`). `ArrowRight` / `ArrowLeft` now delegate to `moveCaretForwardOrToRowEnd` / `moveCaretBackOrFromRowEnd` wrappers that pause at this state on row boundaries. `setCaretPosition` accepts an optional `{ afterRowEnd }` flag so row-end transitions complete in a single redraw.

### Caret Render Helper

**Problem:** The sense + antisense caret redraw pattern was duplicated across multiple call sites, and the offset values `-17` and `3` were embedded inline without explanation.

**Solution:** Extracted `redrawCaretOnBothStrands(node, afterRowEnd)` on `SequenceRenderer` to centralize the dual-strand caret update, and replaced the inline offsets with `CARET_X_OFFSET_BEFORE_NODE` / `CARET_X_OFFSET_AFTER_NODE` named constants.

### Files Modified

**`ketcher-core`**
- `SequenceMode.ts` — Added `Home` / `End` keyboard actions; switched `ArrowRight` / `ArrowLeft` handlers and the right-side-click branch to the new `moveCaretForwardOrToRowEnd` / `moveCaretBackOrFromRowEnd` wrappers to pick up the row-boundary pause behavior
- `SequenceRenderer.ts` — Added `isCaretAfterRowEnd` flag, `isCurrentCaretAtLastInFullRow` getter, `redrawCaretOnBothStrands` helper, the four new caret-move methods, and an optional `{ afterRowEnd }` parameter on `setCaretPosition`
- `BaseSequenceItemRenderer.ts` — Added `CARET_X_OFFSET_BEFORE_NODE` / `CARET_X_OFFSET_AFTER_NODE` constants; parameterized `showCaret(xOffset)`; added `showCaretAfterNode()` wrapper; updated `redrawCaret()` to pick the right offset when at the last position of a full row

## Check list
- [ ] unit-tests written
- [ ] e2e-tests written
- [ ] documentation updated


packages/ketcher-core/src/application/render/renderers/sequence/SequenceRenderer.ts
```diff
@@ -756,6 +770,148 @@ export class SequenceRenderer {
...
+    afterRowEnd: boolean,
+  ) {
+    this.isCaretAfterRowEnd = afterRowEnd;
+    const renderer = node.senseNode?.renderer;
+
+    if (!(renderer instanceof BaseSequenceItemRenderer)) {
+      return;
+    }
+
+    const redraw = (r: BaseSequenceItemRenderer) => {
```
svvald: Pretty minor, but can we make it a separate function outside? Maybe also make it private static
Also there is no sense in making short name for parameter, please use full `renderer`

packages/ketcher-core/src/application/render/renderers/sequence/SequenceRenderer.ts
```diff
@@ -456,7 +456,21 @@ export class SequenceRenderer {
...
+    return this.isCaretAfterRowEndValue;
+  }
+
+  private static set isCaretAfterRowEnd(value: boolean) {
+    this.isCaretAfterRowEndValue = value;
+  }
+
+  public static setCaretPosition(
+    caretPosition: number,
+    options?: { afterRowEnd?: boolean },
```
svvald: Please create a type alias for this parameter

packages/ketcher-core/src/application/render/renderers/sequence/SequenceRenderer.ts
```diff
@@ -756,6 +770,148 @@ export class SequenceRenderer {
...
+    }
+  }
+
+  public static get isCurrentCaretAtLastInFullRow(): boolean {
+    const currentNode = this.currentEdittingNode;
+    const currentRow = this.currentChainRow;
+    const lastNodeInRow = currentRow[currentRow.length - 1];
+
+    return (
+      !!currentNode &&
```
svvald: Let's use `Boolean(currentNode)` instead

svvald (approved):

PR #10100, opened 2026-05-21
#6013 - System reposition monomers on the canvas if ACS setting were applied

### LINKED ISSUE #6013: System reposition monomers on the canvas if ACS setting were applied
**Steps to Reproduce**
1. Go to Macro - Flex mode
2. Load from file: [System reposition monomers on the canvas if ACS setting were applied.zip](https://github.com/user-attachments/files/17919974/System.reposition.monomers.on.the.canvas.if.ACS.setting.were.applied.zip)
image
3. Go to Molecules mode, open Settings and press Set ACS Settings button, Press Apply
4. Go back to Macro - Flex mode

**Actual behavior**
Monomer chain is outside the view frame
image

**Expected behavior**
Monomer chain is inside the view frame and it's position remain unchenged

**Versions**
- Ketcher Version 2.27.0-rc.2 Build at 2024-11-12; 19:26:10
- Indigo Toolkit Version 1.26.0-rc.2.0-gfec30ecd8-wasm32-wasm-clang-19.0.0
- Chrome Version 131.0.6778.70 (Official Build) (64-bit)
- Win10

Lianamirru:
###   Problem

  When switching between micro and macro editor modes, canvas entities were not repositioned correctly if microModeScale and macroModeScale differed.

###   Fix

  Added scale compensation on mode transition by comparing the two scale values and applying the appropriate factor to all  struct entities (atoms, sgroups, monomers, arrows, texts, etc.).

  The initial implementation introduced a dedicated scaleForModeTransition method on Struct, which was later consolidated into the existing scale method via an includeMonomerSgroups flag — also fixing a bug where MonomerMicromolecule sgroups were skipped during scaling.

## Check list
- [ ] unit-tests written
- [ ] e2e-tests written
- [ ] documentation updated


packages/ketcher-core/src/domain/entities/struct.ts
```diff
@@ -925,7 +925,7 @@ export class Struct {
     });
   }
 
-  scale(scale: number) {
+  scale(scale: number, { includeMonomerSgroups = false } = {}) {
```
svvald: Nitpick, but I'd prefer to specify a type alias, something like
```
type ScaleOptions = {
  includeMonomerSgroups?: boolean;
}

const DefaultScaleOptions = {
  includeMonomerSgroups: false
}
```

And specify the param as `options = DefaultScaleOptions` for better readability

packages/ketcher-core/src/domain/entities/struct.ts
```diff
@@ -941,7 +941,7 @@ export class Struct {
     });
 
     this.sgroups.forEach((item) => {
-      if (item instanceof MonomerMicromolecule) {
+      if (!includeMonomerSgroups && item instanceof MonomerMicromolecule) {
```
svvald: I also think that maybe we don't need this flag in this function at all? Maybe we could adjust `applyMonomersTransformations` instead by calculating the scaleFactor there as well and applying it to monomers scale transformation?

svvald (changes requested):

svvald (approved):

PR #10126, opened 2026-05-26
#7909 - Some structure elements disappear when saving as SVG or PNG i…

PaulinaKacz:
This PR fixes an export regression where structures saved as SVG or PNG did not match what was visible on the canvas.

The issue affected both expanded and contracted macromolecule states. In expanded view, brackets, labels, and other annotations could disappear in the preview and in the final exported file. In contracted view, only part of the structure could be exported, for example partial bonds without the full surrounding annotation set.

The root cause was not a single missing element type, but the export pipeline itself. Export bounds were being calculated too narrowly, some transient SVG layers and clipping-related attributes were leaking into the exported markup, and the preview path did not fully match the real file-export path. As a result, labels rendered through foreignObject, bracket annotations, and other painted elements could be clipped or omitted even though they were visible on the canvas.

What Was Changed

The main fix was implemented in getSvgFromDrawnStructures.tsx:73. The export utility was reworked to:
- sanitize the SVG before export by removing transient UI-only layers and elements such as selection overlays, hover layers, blinking elements, masks, and clip paths
- compute export bounds from the real rendered content instead of relying on a narrower subset of the canvas
- use both getBBox-based geometry and client-rect fallback logic mapped back into SVG coordinates, so text-like and foreignObject-based content is included in the final bounds
- apply safer margins and build the exported SVG from the measured structure area rather than from UI viewport assumptions

This is the key reason the fix works for both expanded and contracted states: it addresses the export geometry and content selection at the source instead of patching individual missing elements one by one.

The macromolecule save modal was updated in Save.tsx:104 so that SVG preview generation and actual file export both use the same export path from the canvas. That removes the old mismatch where preview and final export could behave differently. The preview container styling was also adjusted in Save.styles.ts so the generated SVG is centered, scaled safely, and no longer clipped by the modal layout.

The regular save dialog in ketcher-react was also aligned with the same approach in Save.tsx:561 and Save.tsx:864. Instead of depending only on server-generated image payloads for preview and save, the dialog now:

- builds a sanitized SVG snapshot directly from the current canvas
- uses that same snapshot for SVG saving
- rasterizes that same snapshot for PNG saving
- shows the preview from the same rendered export source

This was important because it guarantees that preview and saved output come from the same content model. That makes the exported result much closer to "what the user actually sees" and avoids format-specific drift between preview and final output.

Why This Solution Was Chosen

This solution was chosen because the problem was systemic, not isolated.

A narrower fix, such as manually restoring brackets, labels, or specific bond groups in a few cases, would have been fragile and would likely fail again for other macromolecule layouts, popup mode, or future annotation types. The missing content was a symptom of incorrect export selection and bounding behavior, not a problem with one specific label renderer.

Likewise, relying only on server-side image generation would not solve the mismatch between what is painted in the browser canvas and what gets exported from the Save dialog. The chosen approach makes the export derive from the actual client-side SVG canvas, which is the closest representation of the on-screen structure.

In short, the branch fixes the root cause by:

- exporting from the rendered canvas state
- measuring the real visual bounds of the structure
- stripping UI-only overlays from export
- unifying preview and final export generation

That is more robust than adding special-case handling for individual disappearing elements.

Tests and Validation

A new core-level test suite was added in getSvgFromDrawnStructures.test.ts:40 to cover:
- SVG cleanup of dynamic and non-export elements
- margin and viewBox generation
- export behavior when drawn structures are missing

A new regression autotest was added in save-structure-image-regression.spec.ts:40 with fixture data from save-structure-macro-labels.ket. This test verifies that exported SVG output preserves macro labels and bracket-related annotations instead of checking only image snapshots.

The branch also includes snapshot updates across the affected export scenarios, including the popup flow, to reflect the corrected rendering behavior.

The branch has been rebased onto the latest `master` to resolve merge conflicts.

## Check list
- [x] unit-tests written
- [x] e2e-tests written
- [ ] documentation updated


AlexeyGirin:

AlexeyGirin:

AlexeyGirin:
Should be fixed on indigo side

PR #10158, opened 2026-06-02
#7973 -  Hover mouse over AA, LGA, or the bond connecting them should cause highlight of corresponding AP in the Attributes panel

### LINKED ISSUE #7973: Hover mouse over AA, LGA, or the bond connecting them should cause highlight of corresponding AP in the Attributes panel
**Steps to Reproduce**
1. Open Molecules mode (clear canvas)
2. Load from Extended SMILES using paste from clipboard way: [*:1]C |$_R1;$|`
Image

3. Select whole structure
Image

4. Press `Create monomer` button
5. Hover mouse over AA, LGA and bond connecting them

**Actual behavior**
Every hover doesn't cause sustainable selection of corresponding AP in the Attributes panel
Image

**Expected behavior**
Every hover causes sustainable selection of corresponding AP in the Attributes panel in the same manner as it does in case of R-number hover
Image

As per [requirement](https://github.com/epam/ketcher/issues/7441):

> 3.4. Hovering over an already set AP on the canvas (over an R-number, AA, LGA, or the bond connecting them), highlights the corresponding AP in the Attributes panel.
> 

**Environment details:**
- Ketcher Version 3.8.0-rc.1 Build at 2025-09-03; 08:33:53
- Indigo Version 1.36.0-rc.1.0-gc8cb7f883-wasm32-wasm-clang-19.0.0
- Chrome Version 140.0.7339.81 (Official Build) (64-bit)
- Win10

Found while testing: https://github.com/epam/ketcher/issues/7441

sabaPailodze:
## Summary

When hovering over an already-set attachment point on the canvas — the attachment atom (AA), the leaving group atom (LGA), or the bond connecting them — the editor now highlights the corresponding attachment point row in the Attributes panel. Previously only hovering the R-number label triggered the highlight.

### Canvas hover → panel highlight for AA, LGA, and the connecting bond

**Problem:** Hovering the R-number label highlighted its panel row, but hovering the attachment atom, the leaving-group atom, or the bond between them did not. The highlight was wired up through a Raphael DOM `.hover()` on the hover plate, which stops firing when the cursor sits directly over the atom's text label, so the highlight broke at the center of the atom.

**Solution:** Moved the highlight dispatch off the DOM hover plate and into the editor's geometric `setHover` tracking in `ReAtom` and `ReBond`. This is driven by distance-based hit detection rather than DOM pointer events, so it stays stable over the whole element, including the label center. A `getAttachmentPointName` helper resolves which AP the hovered atom/bond belongs to, and a highlight/reset `CustomEvent` is dispatched only on real hover transitions.

### Order-independent highlight via hover reference counting

**Problem:** Moving the cursor directly between two elements of the same AP (e.g. from R1 to the H atom) could deliver a "reset" event from the element being left right after the "highlight" event from the element being entered. Since both targeted the same panel row, the late reset wrongly cleared the highlight.

**Solution:** The panel components now keep an active-hover counter instead of a boolean. Each highlight increments it and each reset decrements it; the row only clears when the count reaches zero — once the cursor has truly left every element of that AP. This makes the result independent of event ordering.

### Files Modified

**`ketcher-core`**
- `reatom.ts` — removed the DOM-`.hover()`-based highlight trigger; added a `getAttachmentPointName` helper and a `setHover` override that dispatches the panel highlight events on hover transitions
- `rebond.ts` — added a `setHover` override and `getAttachmentPointName` helper so the bond connecting AA↔LGA dispatches the same panel highlight events

**`ketcher-react`**
- `AttachmentPoint.tsx` — replaced the boolean highlight handling with active-hover reference counting so the row stays highlighted across transitions between the AP's elements
- `ReadonlyAttachmentPoint.tsx` — applied the same reference-counting logic for readonly (inter-component connection) attachment point rows

## Check list
- [ ] unit-tests written
- [ ] e2e-tests written
- [ ] documentation updated


AlexeyGirin:
Already fixed

PR #10159, opened 2026-06-02
#9821-Collapsing a monomer alters the length and direction of its connecting bond 

### LINKED ISSUE #9821: Collapsing a monomer alters the length and direction of its connecting bond
Discovered while testing issue #9105

**Steps to Reproduce:**

1. Open Molecules mode (clean canvas)
2. Add any chain to the canvas (e.g., using `Chain` button)
Image

3. Select a part of the chain
4. Click `Create a monomer` button
5. For CHEM: enter a code in the attributes panel and click `Submit` button
6. Hover over the newly created monomer 
7. Right-click and collapse the monomer
Image

**Actual Result:**
The bond connecting the monomer to the main structure changes: it becomes visibly longer and changes its direction

Image

**Expected Result:**
Collapsing the monomer should not alter the appearance, length, or direction of the connecting bond. The bond should remain visually and structurally consistent before and after collapsing the monomer. Collapsing should work in the same manner as s-group collapsing works (and probably use common functions for that)

Image

**Found in Environment:**
- Ketcher Version 3.15.0-rc.1 Build at 2026-04-02; 15:41:30
- Indigo Version 1.43.0-rc.1.0-gd74653443-wasm32-wasm-clang-19.0.0
- Chrome Version 146.0.7680.178 (Official Build) (64-bit)
- Win11

guliaisaeva:
In Molecules mode, after a user creates a CHEM monomer from part of a chain via the "Create a monomer" wizard and then right-clicks → Collapse monomer, the bond connecting the collapsed monomer to the rest of the chain visibly changes both length and direction. It should stay identical.

Root cause
setExpandMonomerSGroup (in packages/ketcher-core/src/application/editor/actions/sgroup.ts) was designed symmetrically — expand pushes outside atoms/sgroups away, collapse pulls them back — and never realigned sGroup.pp after collapsing. Two side effects:

Outside pull-in on collapse: the surrounding chain atoms were shifted toward the monomer center, shortening the connecting bond.
Stale pp: the rendered contracted-bond endpoint reads sGroup.pp, which stayed at the (now meaningless) bounding-box center — so the bond also pointed in the wrong direction.
For monomers created from existing chain atoms (the wizard flow), the outside atoms were never "pushed out" in the first place, so pulling them in on collapse compresses geometry that should not move.


AlexeyGirin:
@guliaisaeva investigate why monomer creation works different from S-Group(superatom) creation

See MonomerMicromolecule class (seems we set wrong position while create monomer)

PR #10184, opened 2026-06-05
8139 Skipping malformed monomers

### LINKED ISSUE #3161: SDF files for library/monomer information
## Background

Current API for monomers library update can apply the data only in KET format. It needs to extend the API to have an ability to provide the data in SDF v3000 format.

## Solution

API for monomers library update will receive new parameter “format” which specifies the format of provided data for new library items.

The following data structure of sdf file should be used, according to CTFile formats specification:

Image

Each compound (*c) describes one library item.

In **Molfile section** the monomer/preset MOL V3000 file should be provided. It contains library item structure. It can be one or more templates and common CTAB section with connections between them. This section also contains some metadata of monomer according to MOL specification, f.e. class, name, fullname, attachment points.

In **Data section** additional metadata can be passed:
 
  - Common fields:

    - type - optional. Default: `monomerTemplate`. Possible values: `monomerTemplate`, `monomerGroupTemplate`(used only for presets now. Can be used for other groups of monomers in future).

    - aliasHELM - string, optional

    - idtAliases - string, optional.
Format: `base` or `ep5`, `ep3`, i set of fields can be provided. Meaning: `ep5` - endpoint5, ep3 - endpoint3, `i` - internal

If base provided, then default rules for endpoints definition will be applied. F.e. `base=Sp9`, then `5Sp9`, `iSp9`, `3Sp9` alias will be used depending on the monomer position inside the chain.

If any combination of `ep5`, `ep3`, i is defined, then it will be used by ketcher/indigo directly for endpoints definition. F.e. `ep5=5Phos`, `ep3=3Phos`, then only 5 and 3 positions are defined and will be used for features related to IDT (conversion to/from IDT, displaying in preview window, e.t.c).
 
  - monomerTemplate specific fields:

    - modificationType - string, optional. Used for aminoacids modification through the context menu.
       
  - monomerGroupTemplate specific fields:

    - groupClass - mandatory. Possible values: RNA

    - groupName - string, mandatory. Used for preset name.

## Examples

[Examples.zip](https://github.com/user-attachments/files/22175404/Examples.zip)

Bidzinashvili:

**Issue:** When calling `ketcher.updateMonomersLibrary(sdf, { format: 'sdf' })` with an SDF containing multiple monomer items where one item has malformed data (e.g., bad COUNTS line), the entire library fails to load instead of gracefully skipping the broken item.

**Root Cause:** The upstream SDF→KET batch conversion via Indigo's `structService.convert()` was atomic — if one record was malformed, the whole call threw an error. This prevented the existing per-item graceful skipping logic downstream in `Editor.updateMonomersLibrary` from ever running.

**Solution:** Implemented a **lazy fallback** strategy:
1. **Fast path (common case):** Try converting the entire SDF batch in one backend round-trip (unchanged behavior, one network call)
2. **Fallback path (batch fails):** If the batch conversion throws, split the SDF into individual `$$$$`-delimited records, convert each one separately in a try/catch, and merge the successful KET results
3. **Per-item handling:** Invalid items are skipped with a console warning: `Monomer item could not be loaded because of an error: [details]`

**Changes:**
- Added `splitSdfRecords()` helper to split SDF by `$$$$` delimiter
- Added `mergeKetMonomerLibraries()` helper to merge multiple KET monomer-library JSON objects
- Wrapped `structService.convert()` in try/catch with per-item fallback in `ensureMonomersLibraryDataInKetFormat()`
- Added test case 21a to verify valid items load despite one broken item

Result: Valid monomers load successfully, broken ones are skipped with a warning, and the user gets clear feedback in the console.

## Check list
- [x] unit-tests written
- [x] e2e-tests written
- [ ] documentation updated
- [x] PR name follows the pattern `#8139 – Gracefully handle malformed items in monomer library batch load`

**Related Issues:** Fixes epam/Indigo#3755, resolves epam/Indigo#3161


AlexeyGirin:
This is indigo

PR #10213, opened 2026-06-09
#10156 - Refactor: give each error toast an independent 6-second auto-dismiss timer

### LINKED ISSUE #10156: Refactor: A single `Snackbar` with one `autoHideDuration={6000}` now wraps a list of toasts. The timer is shared across the whole list, which has two awkward consequences when warnings are pushed at different moments:
A single `Snackbar` with one `autoHideDuration={6000}` now wraps a list of toasts. The timer is shared across the whole list, which has two awkward consequences when warnings are pushed at different moments:

- A toast added shortly before the timer expires gets cut short.
- When the timer fires, `onClose` calls `handleCloseErrorTooltip()` with no argument, which clears **all** tooltips at once — so a freshly added toast can disappear immediately.

For independent per-toast lifetimes, render each toast in its own `Snackbar` (or manage per-message timeouts in the slice and dispatch `closeErrorTooltip(text)` per item).

_Originally posted by @claude[bot] in https://github.com/epam/ketcher/pull/10140#discussion_r3340290064_

guranda2626:
## Summary

- Removed the shared `autoHideDuration={6000}` and `onClose` from the single `Snackbar` that wrapped all toasts.
- Added a per-toast `setTimeout` (6 s) started the moment each toast is pushed; only that toast is dismissed when its timer fires.
- `handleCloseErrorTooltip` now requires an explicit `text` argument, eliminating the "clear all" code path that the shared timer triggered.

## Problem

A single `Snackbar` with one `autoHideDuration={6000}` wrapped the full toast list. This caused two bugs:

1. A toast added shortly before the timer expired was cut short.
2. When the timer fired, `onClose` called `handleCloseErrorTooltip()` with no argument, which cleared **all** toasts — so a freshly added toast could vanish immediately.

## Solution

Replaced the shared MUI timer with a `Map<string, timeout>` ref that tracks one independent `setTimeout` per toast message. A `useEffect` watching `errorTooltips`:
- starts a new 6 s timer for each newly added toast
- cancels the timer for any toast that was manually dismissed before it expired

The `Snackbar` is now a pure positioning container; per-toast lifetime is managed in the component.

[The issue](https://github.com/orgs/epam/projects/45/views/2?pane=issue&itemId=195073043&issue=epam%7Cketcher%7C10156)

## Check list
- [ ] unit-tests written
- [ ] e2e-tests written
- [ ] documentation updated


packages/ketcher-macromolecules/src/Editor.tsx
```diff
@@ -331,7 +331,30 @@ function Editor({
...
+      }
+    });
+
+    toastTimers.current.forEach((timer, text) => {
+      if (!errorTooltips.includes(text)) {
+        clearTimeout(timer);
+        toastTimers.current.delete(text);
+      }
+    });
+  }, [errorTooltips, dispatch]);
```
svvald: I'd add a separate "cleanup" effect that will clear all tooltips timeout and clear timers map on unmount just to be on a safer side

packages/ketcher-macromolecules/src/Editor.tsx
```diff
@@ -331,7 +331,30 @@ function Editor({
     };
   }, [dispatch]);
 
-  const handleCloseErrorTooltip = (text?: string) => {
+  const toastTimers = useRef<Map<string, ReturnType<typeof setTimeout>>>(
+    new Map(),
+  );
```
svvald: And while overall it is a fine working solution, I think that would be better to put it into redux/middlewares layers so the component always gets only the relevant tooltips and shows them

svvald (changes requested):

svvald (approved):

Issue #9422, opened 2026-03-12
Refactor: Explicit Any type clean up (part 5)


AlexeyGirin:
**Problem:** 
Unspecified any.

**Why is this an issue?**
The any type in TypeScript is a dangerous "escape hatch" from the type system. Using any disables many type checking rules and is generally best used only as a last resort or when prototyping code. This rule reports on explicit uses of the any keyword as a type annotation.

**How can it be fixed?**
Preferable alternatives to any include:

- If the type is known, describing it in an interface or type
- If the type is not known, using the safer unknown type

Please check if types you you want to declare already exist and if no - create new declatations at types.ts if they can be used elsewhere (if yes - use existed one)

**Problem locations:**
packages/ketcher-react/src/script/ui/views/modal/components/process/Recognize/Recognize.tsx:243
packages/ketcher-react/src/script/ui/views/modal/components/process/Recognize/Recognize.tsx:271
packages/ketcher-react/src/script/ui/views/modal/components/process/Check/Check.tsx:315
packages/ketcher-react/src/script/ui/component/form/Select/Select.tsx:87
packages/ketcher-react/src/script/editor/tool/helper/isMacroMolecule.ts:10
packages/ketcher-react/src/script/editor/tool/Tool.ts:38
packages/ketcher-react/src/script/editor/tool/templatePreview.ts:200
packages/ketcher-react/src/script/editor/shared/closest.ts:635
packages/ketcher-macromolecules/src/components/shared/Tabs/Tabs.types.ts:7
packages/ketcher-core/src/utilities/SettingsManager.ts:37
packages/ketcher-core/src/domain/serializers/ket/fromKet/multitailArrowToStruct.ts:4
packages/ketcher-core/src/domain/serializers/ket/fromKet/imageToStruct.ts:21
packages/ketcher-core/src/domain/serializers/ket/multitailArrowsValidator.ts:6
packages/ketcher-core/src/domain/serializers/serializers.types.ts:24

PR #10217, opened 2026-06-10
#10216 - Refactor: Explicit Any type clean up (part 5.1)

### LINKED ISSUE #10216: Refactor: Explicit Any type clean up (part 5.1)
**Problem:** 
Unspecified any.

**Why is this an issue?**

**How can it be fixed?**

**Problem locations:**
packages/ketcher-react/src/script/ui/views/modal/components/process/Recognize/Recognize.tsx:243
packages/ketcher-react/src/script/ui/views/modal/components/process/Recognize/Recognize.tsx:271
packages/ketcher-react/src/script/ui/views/modal/components/process/Check/Check.tsx:315
packages/ketcher-react/src/script/editor/tool/helper/isMacroMolecule.ts:10
packages/ketcher-react/src/script/editor/tool/templatePreview.ts:200
packages/ketcher-react/src/script/editor/shared/closest.ts:635
packages/ketcher-core/src/domain/serializers/ket/fromKet/multitailArrowToStruct.ts:4
packages/ketcher-core/src/domain/serializers/ket/fromKet/imageToStruct.ts:21
packages/ketcher-core/src/domain/serializers/ket/multitailArrowsValidator.ts:6
packages/ketcher-core/src/domain/serializers/serializers.types.ts:24

mariam-khutuashvili:
Splitted task: Refactor: Explicit Any type clean up (part 5)
https://github.com/epam/ketcher/issues/9422
 (verified against the full build that they need a refactor/dependency change, not just a type swap):
- `Tool.ts:38` `...args: any[]` — `unknown[]` breaks the tool registry, `never[]` breaks construction; needs the tool constructors refactored.
- `Select.tsx:87` `as any` — duplicate `@types/react` (v18 for MUI vs v19); needs the dependency de-duplicated.
- `Tabs.types.ts:7` `FC<any>` — heterogeneous component list; needs the Tabs API made generic.
- `SettingsManager.ts:37` `selectionTool?: any` — a clean type ripples into ketcher-react consumers outside this issue's scope.

Replaced explicit `any` type annotations with proper types (reusing an existing type where one fit, creating a new reusable one otherwise) or the safer
`unknown`, per the Sonar `typescript:S1444`/no-explicit-any guidance.

This PR covers the 10 locations that can be fixed with a type change alone:

ketcher-core (KET serializers)
- `serializers.types.ts` — `KetFileNode<T = unknown>` default (was `any`); added a reusable `KetFileRootContent` type. The `unknown` default required narrowing the `rgroup` case in `ketSerializer.ts` (it read `item.data.rgnumber`).
- `multitailArrowToStruct.ts` / `imageToStruct.ts` — typed `ketItem` via the entity `fromKetNode` parameter types (`KetFileNode<KetFileMultitailArrowNode>`,
  `KetFileImageNode`).
- `multitailArrowsValidator.ts` — typed `json` as `KetFileRootContent`.

ketcher-react (editor / UI)
- `isMacroMolecule.ts` — typed `dragCtx` via the existing `DragContext`.
- `templatePreview.ts` — typed the action promise as `Promise<[Action, …]>`.
- `closest.ts` — `Map<number, number>` for the merge reducer.
- `Recognize.tsx` / `Check.tsx` — `ThunkDispatch<…>` for `mapDispatchToProps` (matches the existing `Save.tsx`/`Analyse.tsx` pattern); typed `window.webkitURL`.

Verification: `test:types` = 0 errors (ketcher-core + ketcher-react), ESLint = 0 errors, no circular dependencies, ketcher-core unit suite passes (374 tests).
The change is type-only (plus one trivial `rgroup` destructure), so runtime behavior is unchanged.

## Check list
- [ ] unit-tests written
- [ ] e2e-tests written
- [ ] documentation updated

### DIFF
--- a/packages/ketcher-core/src/domain/serializers/ket/ketSerializer.ts
+++ b/packages/ketcher-core/src/domain/serializers/ket/ketSerializer.ts
@@ -206,11 +206,9 @@ export class KetSerializer implements Serializer<Struct> {
           break;
         }
         case 'rgroup': {
-          result.root.nodes.push({ $ref: `rg${item.data!.rgnumber}` });
-          result[`rg${item.data!.rgnumber}`] = rgroupToKet(
-            item.fragment!,
-            item.data,
-          );
+          const { rgnumber } = item.data as { rgnumber: number };
+          result.root.nodes.push({ $ref: `rg${rgnumber}` });
+          result[`rg${rgnumber}`] = rgroupToKet(item.fragment!, item.data);
           break;
         }
         case 'plus': {
--- a/packages/ketcher-core/src/domain/serializers/serializers.types.ts
+++ b/packages/ketcher-core/src/domain/serializers/serializers.types.ts
@@ -29,6 +29,10 @@ export interface KetFileNode<T = unknown> {
   selected?: boolean;
 }
 
+export interface KetFileRoot {
+  nodes: KetFileNode[];
+}
+
 export interface KetFileRootContent {
-  root: { nodes: KetFileNode[] };
+  root: KetFileRoot;
 }
--- a/packages/ketcher-core/src/domain/serializers/serializers.types.ts
+++ b/packages/ketcher-core/src/domain/serializers/serializers.types.ts
@@ -21,11 +21,14 @@ export interface Serializer<T> {
   serialize: (struct: T) => string;
 }
 
-// eslint-disable-next-line @typescript-eslint/no-explicit-any
-export interface KetFileNode<T = any> {
+export interface KetFileNode<T = unknown> {
   type: string;
   fragment?: Struct;
   center: Vec2;
   data?: T;
   selected?: boolean;
+}
+
+export interface KetFileRootContent {
+  root: { nodes: KetFileNode[] };
 }
--- a/packages/ketcher-react/src/script/editor/tool/helper/isMacroMolecule.ts
+++ b/packages/ketcher-react/src/script/editor/tool/helper/isMacroMolecule.ts
@@ -1,13 +1,15 @@
 import type { Editor } from '../../Editor';
+import type { DragContext } from '../select/select.types';
 
 const isMacroMolecule = (editor: Editor, id: number): boolean => {
   const struct = editor.struct();
   return struct.isFunctionalGroupFromMacromolecule(id);
 };
 
-// dragCtx is actually "any" in the code
-// eslint-disable-next-line @typescript-eslint/no-explicit-any
-const isMergingToMacroMolecule = (editor: Editor, dragCtx: any): boolean => {
+const isMergingToMacroMolecule = (
+  editor: Editor,
+  dragCtx: Pick<NonNullable<DragContext>, 'mergeItems'> | null | undefined,
+): boolean => {
   const funcGroups = dragCtx?.mergeItems?.atomToFunctionalGroup;
   if (!funcGroups?.size) {
     return false;
--- a/packages/ketcher-react/src/script/editor/tool/helper/isMacroMolecule.ts
+++ b/packages/ketcher-react/src/script/editor/tool/helper/isMacroMolecule.ts
@@ -8,7 +8,7 @@ const isMacroMolecule = (editor: Editor, id: number): boolean => {
 
 const isMergingToMacroMolecule = (
   editor: Editor,
-  dragCtx: Pick<NonNullable<DragContext>, 'mergeItems'> | null | undefined,
+  dragCtx: DragContext,
 ): boolean => {
   const funcGroups = dragCtx?.mergeItems?.atomToFunctionalGroup;
   if (!funcGroups?.size) {
--- a/packages/ketcher-react/src/script/editor/tool/templatePreview.ts
+++ b/packages/ketcher-react/src/script/editor/tool/templatePreview.ts
@@ -196,8 +196,7 @@ class TemplatePreview {
         shouldFlip,
         true,
         true,
-        // eslint-disable-next-line @typescript-eslint/no-explicit-any
-      ) as Promise<any>;
+      ) as Promise<[Action, { atoms: number[]; bonds: number[] }]>;
 
       promise.then(([action, pasteItems]) => {
         if (!this.isModeFunctionalGroup) {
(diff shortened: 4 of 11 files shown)


packages/ketcher-react/src/script/editor/tool/helper/isMacroMolecule.ts
```diff
@@ -1,13 +1,15 @@
...
   const struct = editor.struct();
   return struct.isFunctionalGroupFromMacromolecule(id);
 };
 
-// dragCtx is actually "any" in the code
-// eslint-disable-next-line @typescript-eslint/no-explicit-any
-const isMergingToMacroMolecule = (editor: Editor, dragCtx: any): boolean => {
+const isMergingToMacroMolecule = (
+  editor: Editor,
+  dragCtx: Pick<NonNullable<DragContext>, 'mergeItems'> | null | undefined,
```
svvald: That's a strange type definition here, why couldn't we use just `DragContext` here?

packages/ketcher-core/src/domain/serializers/serializers.types.ts
```diff
@@ -21,11 +21,14 @@ export interface Serializer<T> {
...
+export interface KetFileNode<T = unknown> {
   type: string;
   fragment?: Struct;
   center: Vec2;
   data?: T;
   selected?: boolean;
 }
+
+export interface KetFileRootContent {
+  root: { nodes: KetFileNode[] };
```
svvald: Please avoid inlining nested properties, format it properly

packages/ketcher-core/src/domain/serializers/ket/ketSerializer.ts
```diff
@@ -206,11 +206,9 @@ export class KetSerializer implements Serializer<Struct> {
...
         }
         case 'rgroup': {
-          result.root.nodes.push({ $ref: `rg${item.data!.rgnumber}` });
-          result[`rg${item.data!.rgnumber}`] = rgroupToKet(
-            item.fragment!,
-            item.data,
-          );
+          const { rgnumber } = item.data as { rgnumber: number };
+          result.root.nodes.push({ $ref: `rg${rgnumber}` });
+          result[`rg${rgnumber}`] = rgroupToKet(item.fragment!, item.data);
```
svvald, line 211: I don't mind these changes (except for `as` casting on line 209 probably but we had a non-null assertion there originally anyway) but I assume there were no explicit `any`-s here?

packages/ketcher-react/src/script/ui/views/modal/components/process/Recognize/Recognize.tsx
```diff
@@ -241,8 +243,9 @@ function RecognizeDialog(prop: Readonly<RecognizeDialogProps>) {
 
 function url(file: File | null): string | null {
   if (!file) return null;
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  const URL = window.URL || (window as any).webkitURL;
+  const URL =
+    window.URL ||
+    (window as Window & { webkitURL?: typeof globalThis.URL }).webkitURL;
```
svvald: I'm too fond of having this type and `as` casting to be fair, but I assume we won't get away from them in this particular case. Let's then declare a type alias for this window with `webkitURL` property rather than defining it inline just for a better readability

packages/ketcher-react/src/script/editor/tool/templatePreview.ts
```diff
@@ -196,8 +196,7 @@ class TemplatePreview {
         shouldFlip,
         true,
         true,
-        // eslint-disable-next-line @typescript-eslint/no-explicit-any
-      ) as Promise<any>;
+      ) as Promise<[Action, { atoms: number[]; bonds: number[] }]>;
```
svvald: Is there a way we could properly specify the return type for `fromTemplateOnBondAction` rather than type casting its result?

svvald (changes requested):

packages/ketcher-react/src/script/editor/tool/helper/isMacroMolecule.ts
mariam-khutuashvili: Switched it to DragContext directly. mergeItems is declared on the shared SharedDragContext base, so it's accessible across the whole union via optional chaining, and the only caller passes this.dragCtx (typed DragContext). (The param was any originally.)

packages/ketcher-core/src/domain/serializers/serializers.types.ts
mariam-khutuashvili: Extracted the nested object into a named KetFileRoot interface (root: KetFileRoot) instead of inlining it.

packages/ketcher-core/src/domain/serializers/ket/ketSerializer.ts
mariam-khutuashvili: The original was item.data!.rgnumber (non-null assertion). This hunk is a consequence of the core change: KetFileNode.data is now typed unknown (KetFileNode<T = unknown>) instead of effectively any, so item.data is unknown here and .rgnumber needs a cast. I can't drop it without re-introducing the any on data, so I kept it minimal — one as { rgnumber: number } destructure replacing the original !.

packages/ketcher-react/src/script/ui/views/modal/components/process/Recognize/Recognize.tsx
mariam-khutuashvili: Extracted a WindowWithWebkitURL type alias and use (window as WindowWithWebkitURL) instead of the inline type.

packages/ketcher-react/src/script/editor/tool/templatePreview.ts
mariam-khutuashvili: Added force-keyed overloads to fromTemplateOnBondAction (sync [Action, {atoms, bonds}] when force is false, Promise<...> when true), so the result no longer needs casting here. The 3 remaining casts in tool/template.ts are pre-existing as Promise<any>/as Array<any> outside this PR's scope — happy to drop those in a follow-up now that the overloads make them removable.

svvald (approved):

PR #10233, opened 2026-06-11
#5502 - Structures shift to center of canvas when enabling viewOnlyMo…

### LINKED ISSUE #5502: Structures shift to center of canvas when enabling viewOnlyMode through the API
**Description:** When using the API call window.ketcher.editor.setOptions(JSON.stringify({ viewOnlyMode: true })), structures already present on canvas are unexpectedly moved to the center, as if a layout function is being applied. This behavior is unintended and disrupts the original positioning of structures.

**Steps to Reproduce:**

1) Add any structures onto the canvas.
2) Use the following API call: window.ketcher.editor.setOptions(JSON.stringify({ viewOnlyMode: true })).
3) Observe the position of the structures on the canvas.

**Actual Result:** The structures shift to center of the canvas, behaving as though a layout is being applied.

**Expected Result:** The structures should remain in their original position without any movement or layout changes.

**Screenshots**

https://github.com/user-attachments/assets/73bc2eb1-0619-4e34-bbf0-7542645001fd

**Desktop:**
 - OS: Windows 11
 - Browser Chrome
 - Version 128.0.6613.120 (Official Build) (64-bit)

**Ketcher version** 
[Version 2.25.0-rc.1]
**Indigo version** 
[Version 1.24.0-rc.1]

**Bug found during testing:** https://github.com/epam/ketcher/issues/4965

PaulinaKacz:
Issue                                                                                                                                                    
                                  
  Fixes #5502 - Structures shift to center of canvas when enabling viewOnlyMode through the API

  Problem
  When enabling or disabling viewOnlyMode via the API (using setOptions({ viewOnlyMode: true/false })), existing structures on the canvas were being
  repositioned to the center. This was unexpected behavior - users expected structures to remain in their original positions when toggling the view mode.

  Root Cause
  The setOptions() method in Editor.ts was calling this.struct(struct.clone()) which by default centers the structure on the canvas. This centering
  behavior was appropriate for operations like opening files or initial setup, but not when simply toggling viewOnlyMode on an existing structure.

  Solution
  Modified Editor.ts:setOptions() to conditionally preserve atom coordinates when viewOnlyMode is being set/changed on an existing structure:

  const hasAtoms = struct.atoms.size > 0;
  const isSettingViewOnlyMode = value && 'viewOnlyMode' in value;
  const shouldPreservePosition = hasAtoms && isSettingViewOnlyMode;

  this.struct(struct.clone(), !shouldPreservePosition);

  The fix:
  - ✅ Preserves structure position when enabling/disabling viewOnlyMode
  - ✅ Maintains original centering behavior for other operations (opening files, changing other settings)
  - ✅ Only applies when there's an existing structure on canvas

  Test Changes

  Updated benzene-ring-saved-in-view-only-mode-expected.ket test fixture:
  - Removed stereoFlagPosition field from expected KET export in viewOnlyMode
  - This aligns with the actual export behavior where stereoFlagPosition is not included in view-only mode

  Testing

  - Manual testing: Toggle viewOnlyMode on structures - they stay in place ✅
  - Automated test: Check saving in KET format in View only mode now passes ✅
  - Regression check: Opening files and other operations still center structures correctly ✅

## Check list
- [ ] unit-tests written
- [ ] e2e-tests written
- [ ] documentation updated


packages/ketcher-react/src/script/editor/Editor.ts
```diff
@@ -450,7 +450,15 @@ class Editor implements KetcherEditor {
...
     this.updateToolAfterOptionsChange(wasViewOnlyEnabled);
     this.render.setMolecule(struct);
-    this.struct(struct.clone());
+
+    // Preserve atom coordinates only when viewOnlyMode is being set/changed on existing structure.
+    // This prevents structures from shifting when enabling/disabling viewOnlyMode.
+    // In all other cases (opening files, changing settings), center as before.
+    const hasAtoms = struct.atoms.size > 0;
+    const isSettingViewOnlyMode = value && 'viewOnlyMode' in value;
+    const shouldPreservePosition = hasAtoms && isSettingViewOnlyMode;
```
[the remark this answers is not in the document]
PaulinaKacz: The suggested change is too pedantic. It requires an actual value transition, which causes the canvas to unexpectedly auto-center on subsequent API calls with the same value. The original solution correctly respects the user's intent to set/maintain the mode without shifting the structure.

packages/ketcher-react/src/script/editor/Editor.ts
```diff
@@ -450,7 +450,15 @@ class Editor implements KetcherEditor {
...
+
+    this.struct(struct.clone(), !shouldPreservePosition);
```
AlexeyGirin: Make fix at one level above from section where options is called

AlexeyGirin (dismissed):

packages/ketcher-react/src/script/editor/Editor.ts
PaulinaKacz: I've refactored the code by moving the logic "one level above" in terms of abstraction.
What changed:
- Created a private method `shouldPreserveStructPosition(options: any, struct: Struct): boolean` (lines 475-482)
- The decision logic is now at a higher abstraction level with clear intent and documentation
- The `options()` method remains the correct location for the fix, as it's where the actual issue occurs

svvald (approved):

PR #10290, opened 2026-06-22
#5857, #5676 - Fix coordinates calculation for v2000 format 

### LINKED ISSUE #5676: Distorted structure representation when opening saved macromolecules in micro mode in MOL V2000 file
**Steps to Reproduce**
1. Open file in Macro mode (Flex mode) [monomers-cycled.zip](https://github.com/user-attachments/files/17281569/monomers-cycled.zip)
2. Switch to Micro mode
3. Save as MOL V2000
4. Open saved file

**Actual behavior**
Upon reopening the MOL V2000 file, the structure appears distorted, with bonds incorrectly placed, resulting in a non-cyclic structure (as shown in the screenshot). The original intended shape and arrangement of monomers are lost.
2024-10-07_19h16_21

**Expected behavior**
The saved structure should maintain its original form upon reopening, appearing as a cyclic structure similar to the screenshot.

2024-10-07_19h16_31

**Desktop:**
 - OS: Windows 11
 - Browser Chrome
 - Version 128.0.6613.120 (Official Build) (64-bit)

**Ketcher version** 
[Version 2.26.0-rc.1]
**Indigo version** 
[Version 1.25.0-rc.1]
Bug found during testing: https://github.com/epam/ketcher/issues/5400

### LINKED ISSUE #5857: Export of monomers to Mol v2000 works wrong (Open as New Project way)
**Steps to Reproduce**
1. Go to Micro
2. Load from file: [All type of monomers in horisontal chain and large micromolecule in the middle.zip](https://github.com/user-attachments/files/17508856/All.type.of.monomers.in.horisontal.chain.and.large.micromolecule.in.the.middle.zip)
image
3. Save canvas to Mol v2000
4. Load it back using Open as New Project button

**Actual behavior**
System puts on the canvas:
image

**Expected behavior**
System let you put on the canvas
image

**Versions**
- Ketcher Version 2.26.0-rc.5 Build at 2024-10-21; 15:27:55
- Indigo Toolkit Version 1.25.0-rc.3.0-g795119115-wasm32-wasm-clang-19.0.0
- Chrome Version Version 130.0.6723.59 (Official Build) (64-bit)
- Win10

Found while testing - https://github.com/epam/ketcher/issues/5400

kpustakhod:

1. `getContractedPosition` — geometric center fallback (`sgroup.ts`)

  Problem: When a contracted sgroup has no attachment points, the representative position fell back to the
  first externally-connected atom or just the first atom, which caused incorrect collapsed positions.

  Fix: Changed the fallback to return the geometric center of all atoms in the group instead.

  ---
  2. v2000 atom centering before save (`molfile.ts`)
  
  Problem: The v2000 format doesn't store `sgroup.pp` (monomer position). When saving to v2000, atom
  coordinates weren't adjusted, so the collapsed sgroup position in the output didn't match the monomer's
  position in the KET file.

  Fix: Added `centerMonomerMicromoleculeAtoms()` — called on the struct clone before writing v2000 — which
  translates all atoms of each MonomerMicromolecule so their geometric center equals sgroup.pp (= the
  monomer's KET position). A verification test was added to confirm the round-trip (KET position → v2000 →
  parse → geometric center) is correct.

  ---
  3. Contracted sgroup clickable area misalignment (`resgroup.ts`)

  Problem: The green highlight/click area around a contracted sgroup label was displaced from the actual
  label because `getTextHighlightDimensions` computed its position from the model-space center (x - width/2, y 
  - height/2), but SVG text uses baseline anchoring, so the actual rendered bounding box didn't match.

  Fix: Changed `getTextHighlightDimensions` to derive position directly from the visel's bounding box (p0.x - 
  padding, p0.y - padding), making the clickable area exactly overlay the rendered label.

  ---
  4. `getContractedPosition` — always use geometric center as position (`sgroup.ts`)
  
  Problem: Even after fix bullet point 2, the contracted label in v2000 still appeared at the attachment point atom's
  position (not the geometric center), because getContractedPosition returned the attachment point atom's pp
  as the position when attachment points existed.

  Fix: Changed `getContractedPosition` so that position is always the geometric center of all atoms, regardless
  of whether attachment points exist. The atomId (which atom slot renders the label) still prefers the
  attachment point atom. MonomerMicromolecule continues to override position to return sgroup.pp unchanged.

  ---
  5. Label rendering uses contracted position (`reatom.ts`)
  
  Problem: The sgroup label was rendered at `this.a.pp` (the representative atom's canvas position), not at the
  position returned by `getContractedPosition`. For regular sgroups this meant the label appeared at the
  attachment point atom, not the geometric center.

  Fix: Changed the label rendering to use `contractedPosition` from `getContractedPosition()` for all sgroup
  types, removing the separate `MonomerMicromolecule` special-case (which is already handled by its override
  returning `sgroup.pp`).

  ---
  6. Bond endpoints align with contracted position (`rebond.ts`)
  
  Problem: Inter-monomer bonds still connected to the attachment point atom's actual coordinates, not the
  geometric center where the label was now rendered. This caused bonds to appear disconnected from the
  labels.

  Fix: Changed `bondRecalc` to use `getContractedPosition(struct).position` for all contracted sgroups (not just
  `MonomerMicromolecule`), so bond endpoints always route to the same visual position as the label — geometric
  center for regular `sgroups`, `sgroup.pp` for `MonomerMicromolecule`.

## Check list
- [ ] unit-tests written
- [ ] e2e-tests written
- [ ] documentation updated


packages/ketcher-core/src/domain/entities/sgroup.ts
```diff
@@ -334,37 +334,36 @@ export class SGroup {
...
+    }
 
-      atomId = isNumber(externalConnectionAtom)
-        ? externalConnectionAtom
-        : this.atoms[0];
-      representAtom = struct.atoms.get(atomId);
+    // position: always the geometric center of all atoms in the group.
+    const positions = this.atoms
+      .map((id) => struct.atoms.get(id)?.pp)
+      .filter((pp): pp is Vec2 => pp != null);
```
svvald: Is non-strict comparison necessary here? I suppose strict equal check or even `Boolean` should work just fine?

packages/ketcher-core/src/domain/entities/sgroup.ts
```diff
@@ -334,37 +334,36 @@ export class SGroup {
...
+      const center = positions
+        .reduce((sum, pp) => sum.add(pp), new Vec2(0, 0))
+        .scaled(1 / positions.length);
+      return { atomId, position: center };
     }
-    assert(representAtom != null);
-    return { atomId, position: representAtom.pp };
+
+    const atom = struct.atoms.get(atomId);
+    assert(atom != null);
```
svvald: Same here

packages/ketcher-core/src/domain/serializers/mol/__tests__/molfile.monomerPosition.test.ts
```diff
@@ -0,0 +1,127 @@
...
+}
+
+function sgroupGeometricCenter(sgroup: SGroup, struct: Struct): Vec2 {
+  const positions = sgroup.atoms
+    .map((id) => struct.atoms.get(id)?.pp)
+    .filter((pp): pp is Vec2 => pp != null);
+  return positions
+    .reduce((sum, pp) => sum.add(pp), new Vec2(0, 0))
+    .scaled(1 / positions.length);
+}
```
svvald: This function implementation is the same as the one used in `getContractedPosition` in SGroup class. I'd probably extract it as some util function and reused

packages/ketcher-core/src/domain/serializers/mol/molfile.ts
```diff
@@ -541,6 +545,31 @@ export class Molfile {
...
+      if (!(sgroup instanceof MonomerMicromolecule) || !sgroup.pp) return;
+
+      const positions = sgroup.atoms
+        .map((id: number) => mol.atoms.get(id)?.pp)
+        .filter((pp): pp is Vec2 => pp != null);
+      if (!positions.length) return;
+
+      const center = positions
+        .reduce((sum, pp) => sum.add(pp), new Vec2(0, 0))
+        .scaled(1 / positions.length);
```
svvald: And same here

packages/ketcher-core/src/application/render/restruct/resgroup.ts
```diff
@@ -188,20 +188,17 @@ class ReSGroup extends ReObject {
     let height = 0;
     const sGroup = this.item;
     if (sGroup) {
-      const { atomId, position } = sGroup.getContractedPosition(
-        render.ctab.molecule,
-      );
-      if (sGroup?.isContracted() && position) {
+      const { atomId } = sGroup.getContractedPosition(render.ctab.molecule);
```
svvald: `atomId` is now only used inside `if` statement, let's move it there so we won't perform unnecessary calculations

packages/ketcher-core/src/domain/serializers/mol/molfile.ts
```diff
@@ -541,6 +545,31 @@ export class Molfile {
     this.writeCR('M  END');
   }
 
+  private centerMonomerMicromoleculeAtoms() {
+    if (!this.molecule) return;
```
svvald: Nitpick but for better readability let's add newline for all one-liner if statements like
```
if (...) {
return;
}
```

svvald (changes requested):

packages/ketcher-core/src/domain/serializers/mol/__tests__/molfile.monomerPosition.test.ts
```diff
@@ -0,0 +1,125 @@
...
+      buildMonomerStruct('TestMon', monomerPosition, atomOffsets),
+    );
+
+    const sgroup = findSgroupByName(parsed, 'TestMon');
+    expect(sgroup).toBeDefined();
+
+    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+    const positions = sgroup!.atoms
+      .map((id) => parsed.atoms.get(id)?.pp)
+      .filter((pp): pp is Vec2 => pp != null);
```
svvald: This non strict equality check `!=` still persists here and below a few times

Also I'm think whether we could extract this `map-filter` sequence into some utility function as well as it is used and duplicated several times already?

packages/ketcher-core/src/domain/entities/geometry.ts
```diff
@@ -0,0 +1,16 @@
...
+
+export function geometricCenter(positions: Vec2[]): Vec2 {
+  return positions
+    .reduce((sum, pp) => sum.add(pp), new Vec2(0, 0))
+    .scaled(1 / positions.length);
+}
+
+export function getAtomPositions(
+  atomIds: number[],
+  atoms: Map<number, { pp: Vec2 }>,
```
svvald: I suppose this parameter will always be `Struct`'s `atoms` property which is declared as `Pool<Atom>`, I suppose we can have the same typing here

packages/ketcher-core/src/domain/entities/sgroup.ts
```diff
@@ -334,37 +335,33 @@ export class SGroup {
...
+    }
 
-      atomId = isNumber(externalConnectionAtom)
-        ? externalConnectionAtom
-        : this.atoms[0];
-      representAtom = struct.atoms.get(atomId);
+    // position: always the geometric center of all atoms in the group.
+    const positions = this.atoms
+      .map((id) => struct.atoms.get(id)?.pp)
+      .filter((pp): pp is Vec2 => Boolean(pp));
```
svvald: I think we can use `getAtomPositions` here as well

svvald (approved):

PR #10298, opened 2026-06-23
#3890 - System doesn't show problem valence atom

### LINKED ISSUE #3890: System doesn't show problem valence atom
**Steps to Reproduce**
1. Load from file: [WrongAnyAtom.zip](https://github.com/epam/ketcher/files/13928020/WrongAnyAtom.zip) (unzip first)

**Actual behavior**
System doesn't show "C" label underlined with red on the canvas:
image

**Expected behavior**
System shows "C" label underlined with red on the canvas:
image

 - OS: Win10
 - Browser Chrome Version 120.0.6099.130 (Official Build) (64-bit)

Ketcher Version 2.18.0+dev.1 Build at 2024-01-11; 13:21:50
Indigo Toolkit Version 1.18.0-dev.4+1452.0-gaf1efce67-x86_64-linux-gnu-11.4.0

kpustakhod:

Screenshot 2026-06-23 at 10 21 40

Gaps found

Gap A — the bug still exists in macro mode. `CoreAtom.calculateValence()` has no aromatic branch at all. For an aromatic C(charge 0) with 4 connections: `connectionAmount = 4 → hydrogenAmount = 4 − 4 = 0 → hasBadValence = false`. That is precisely the pre-fix micro behavior (before your fix, aromatic C conn=4 fell through to `calcValence(4)` with the same result; `conn≥5` was already caught as `4−5 = −1`). Loading the issue's structure onto the macro canvas would show no warning.

Gap B — bad-valence mid-chain carbons are invisible in macro mode. `isLabelVisible` (`AtomRenderer.ts:313`) hides the label of a non-terminal carbon without `charge/radical/alias/explicit-valence/isotope`, and it does not consider `hasBadValence`. Since `appendBadValenceWarning` early-returns when the label is hidden, a carbon with five single bonds mid-chain renders as a bare vertex with no warning. Micro mode explicitly handles this by forcing label visibility (`reatom.ts:1428`).

Gap C — dead setting in macro mode. `showValenceWarnings` exists in the unified settings schema (`application/settings/schema.ts:17`, default `true`) and is shown as a checkbox in the macro Settings modal (`ketcher-macromolecules/.../fieldGroups.ts:42`), but no macro rendering code reads it — the underline is unconditionally on. The only consumers of any `showValenceWarnings` flag are in micro's `reatom.ts`.

Gap D (micro, pre-existing) — non-carbon aromatic atoms are never flagged. In the aromatic branch (`struct.ts:1170-1181`), aromatic O(0) at any connectivity, N(0/+1) and S(0) at conn 3, and any aromatic atom whose `implicitH` is currently falsy short-circuit to `implicitH = 0`; return without a `badConn` check. E.g. an aromatic O with 3 connections (genuinely bad valence for neutral O) is silently accepted in micro — while macro does flag it (2 − 3 = −1). Note this branch exists for a reason: the `correctConn++` heuristic (approximating the extra π bond) would otherwise produce false positives on every furan/pyrrole heteroatom. The `!atom.implicitH` catch-all is also stateful — it reads the previous `implicitH` value, making results order-dependent.

---------

Claude fixed: 
(1) port the aromatic handling into `CoreAtom.calculateValence` — or better, extract one shared valence module used by both Atom and CoreAtom so the tables can't drift; 
(2) add `hasBadValence` to macro's `isLabelVisible`; 
(3) wire the macro `showValenceWarnings` setting into `appendBadValenceWarning`. Gap D (aromatic heteroatoms in micro) needs a product decision, since flagging them without kekulization risks false positives.

----------

1. Shared valence module (no more drift)

New file packages/ketcher-core/src/domain/entities/atomValence.ts — pure functions with no dependencies:
- calculateValenceResult(groupno, context) — the per-group valence tables, moved verbatim from atom.ts (I also restored the PubChem reference comments that the macro copy had preserved).
- calculateValenceMinusHydrogen(groupno, context) — the charge-participates-in-bonding exceptions used for explicit valence.
- resolveAromaticAtomValence(context, options) — the aromatic branch extracted from Struct.calcImplicitHydrogen, returning one of 'calculated' (final hydrogen count + explicit badValence verdict), 'correctedConnectivity' (the correctConn++ heuristic), or 'standard'.

Micro mode (atom.ts, struct.ts) now delegates to it with byte-identical behavior — the existing 83 atom tests pass unchanged. Macro mode (CoreAtom.ts) shrank from ~485 to ~230 lines: calculateConnections() now also reports isAromatic, calculateValence() runs the shared aromatic resolver first and returns an explicit badValence flag, and hasBadValence uses that flag. This fixes #3890 on the macro canvas: aromatic C with 4 connections is now flagged there too.

2. Bad-valence mid-chain carbons now visible in macro

AtomRenderer.isLabelVisible considers a new shouldDisplayBadValenceWarning getter, so a carbon with e.g. five bonds mid-chain now renders its "C" label with the red underline instead of a bare vertex — matching micro's reatom.ts:1428 behavior.

3. showValenceWarnings setting actually works in macro

- appendBadValenceWarning and the label-forcing both go through shouldDisplayBadValenceWarning, which reads the setting from the settings service (defaulting to true if unavailable) — same pattern as the existing stereo-label settings.
- Since nothing re-rendered the macro canvas on settings changes, CoreEditor now subscribes to the settings service in its constructor and calls drawingEntitiesManager.rerenderMolecules() when showValenceWarnings changes (unsubscribed in unsubscribeEvents/destroy), so the toggle takes effect immediately.

Deliberate behavior changes in macro mode (aligned to micro as reference)

- Aromatic heteroatoms are no longer flagged (e.g. aromatic O with 3 connections was red-underlined before, micro never flagged it) — this also prevents false positives on furan/pyrrole-type patterns.
- Benzene-like aromatic carbons now compute 1 implicit hydrogen instead of 2 (visible in the data-atomHCount attribute and terminal-atom labels).
- Aromatic resolution and the indeterminate-connectivity guard now take precedence over explicit valence, mirroring micro's ordering.

Gap D from the investigation (micro never flags aromatic heteroatoms) is untouched, as we agreed it needs a product decision — but if it's ever decided, it's now a one-line change in resolveAromaticAtomValence that fixes both modes at once.

## Check list
- [ ] unit-tests written
- [ ] e2e-tests written
- [ ] documentation updated


AlexeyGirin:

packages/ketcher-core/src/application/render/renderers/RenderersManager.ts
```diff
@@ -423,9 +417,29 @@ export class RenderersManager {
...
+  public addBond(bond: Bond) {
+    if (bond.renderer) {
+      bond.renderer.remove();
+    }
+
+    const bondRenderer = new BondRenderer(bond);
+
+    this.bonds.set(bond.id, bondRenderer);
+
+    this.syncAndRedrawBondAtomLabels(bond, true);
```
AlexeyGirin, line 483: Please use post renderer methonds for defining bad valence (see how it is done for atomatic rings and base counters)

AlexeyGirin (changes requested):

AlexeyGirin:

AlexeyGirin:

Issue #5149, opened 2024-07-22
Only elements are moving with the cursor while adding User Template to the Canvas


ivanmaslow:
**Steps to Reproduce**
1. Open from .ket file - there are some reaction tools, text, shapes and elements: 
[test.zip](https://github.com/user-attachments/files/16332685/test.zip)
![image](https://github.com/user-attachments/assets/cb5eb383-4086-4644-b2dd-214cdbcecbc9)
2. Click on Save - Save to Templates
3. Add Molecule Name and Save
4. Click on SL to open Structure Library, scroll down and click on the saved template and move the cursor

**Actual behavior**
Only elements are moving, reaction tools, text, shapes are not moving while moving the cursor:
![image](https://github.com/user-attachments/assets/0134f27d-6f41-43b9-a361-75daa2bb5828)
Recording is attached:

https://github.com/user-attachments/assets/6bcc7aee-b1e5-4c33-b209-5b247eed98d3

**Expected behavior**
All entities are moving together

**Desktop (please complete the following information):**
- OS: Windows 11 Enterprise
- Browser: Chrome 126.0.6478.128 (Official Build) (64-bit)

**Ketcher version:**
- [Version 2.23.0-rc.3] Build at 2024-07-14; 20:34:10

**Indigo version:**
- Version 1.22.0-rc.3.0-gbb5c51fc2-wasm32-wasm-clang-19.0.0

**Additional context**
- Detected during preview tests for Image feature - https://github.com/epam/ketcher/issues/4897, this issue is reproducible for images as well, as for now, this feature is not in the master. The fix should be verified for images as well.

AlexeyGirin:
Verified.
<img width="861" height="555" alt="Image" src="https://github.com/user-attachments/assets/15c0afbe-1bb9-4f5e-ab47-04b2b49a05ee" />

- Ketcher Version 3.19.0-rc.1 Build at 2026-08-03; 17:10:30
- Indigo Version 1.46.0-rc.1.0-gf3408986b-x86_64-linux-gnu-11.2.1
- Chrome Version 151.0.7922.72 (Official Build) (64-bit)
- Win11

PR #10310, opened 2026-06-25
#6004 - Fix stereo flag (ABS) position ignoring text and shapes on ca…

### LINKED ISSUE #6004: ABS sign is added to Canvas with a wrong place if there are text or shape
**Steps to Reproduce**
1. Add any text or shape to Canvas:
image
2. Click on SL and select any sugar with ABS sign
3. Select the place on Canvas under the shape and paste - you can notice that only element is moving with the cursor and ABS sign isn't - it's reported there - https://github.com/epam/ketcher/issues/5149

**Actual Result**
- Chemical element is added to Canvas, but ABS sign isn't near chemical element:
image

**Expected Result**
- Shape or text doesn't affect position of ABS sign
image

**Environment details**
- Tested on RC SA env;
- Ketcher [Version 2.27.0-rc.2] Build at 2024-11-12; 19:26:10
- Indigo Version 1.26.0-rc.2.0-gfec30ecd8-wasm32-wasm-clang-19.0.0
- Desktop, Windows 11 Enterprise
- Chrome Version 131.0.6778.86 (Official Build) (64-bit)

**Notes**
- Issue is detected during exploratory testing of save text for reactions - https://github.com/epam/Indigo/issues/2404, but it's not related to this feature
    - It affected the case of loading cascade reaction with text and ABS signs:    
[!Synthia_path_8_tails.zip](https://github.com/user-attachments/files/17846437/Synthia_path_8_tails.zip)
image

aslmbk:
Closes #6004

`Fragment.getDefaultStereoFlagPosition` computed the default enhanced-stereo flag
(ABS/AND/OR/MIXED) position from `struct.getFragment(fragmentId).getCoordBoundingBox()`.
That path clones **all** texts and shapes of the canvas into the fragment and then
computes the bounding box globally, so any unrelated text/shape pulled the flag away
from the molecule. It was most visible when pasting a structure that carries an ABS
flag onto a canvas that already contains text or a shape: the molecule followed the
cursor while the `ABS` label stayed near the text/shape (also reported as #5149).

**Fix:** compute the bounding box from the fragment's own atoms only, reusing the
existing `struct.getFragmentIds()` together with the `atomSet` parameter of
`struct.getCoordBoundingBox()`. The flag position now depends solely on the fragment's
atoms, regardless of any other canvas content. As a bonus, this removes a full
`Struct.clone()` from the render / paste-drag hot path.

No behaviour change when no text/shape is present; reaction arrows/pluses were never
involved (the previous clone dropped them via `dropRxnSymbols=true`).

Verified: new unit test `fragment.test.ts` (4 cases, incl. text/shape regressions),
`ketcher-core` entity + render suites green, `tsc --noEmit` and ESLint clean.

## Screenshots
Screenshot 2026-06-25 at 10 41 43

## Check list
- [x] unit-tests written
- [ ] e2e-tests written
- [ ] documentation updated


packages/ketcher-core/src/domain/entities/fragment.ts
```diff
@@ -102,9 +102,9 @@ export class Fragment {
     struct: Struct,
     fragmentId: number,
   ): Vec2 | undefined {
-    const fragment = struct.getFragment(fragmentId);
-    if (!fragment) return undefined;
-    const bb = fragment.getCoordBoundingBox();
+    const fragmentAtomIds = struct.getFragmentIds(fragmentId);
+    if (fragmentAtomIds.size === 0) return undefined;
```
svvald, line 106: Nitpick – let's put `return` into a new line

svvald (approved):

PR #10316, opened 2026-06-25
#1449 - The Template Library sections are opening slowly

### LINKED ISSUE #1449: The Template Library sections are opening slowly
**Steps to Reproduce**
1. Launch Ketcher
2. Click on the "TP" icon on the down menu to open Templates dialog
3. Open some Template Library sections by clicking on the arrow to extend drop down (e.x: alfa-D-Sugars; beta-D-Sugars; D-Amigo Acids)

**Expected behavior**
The sections drop down in the Templates dialog is extended right away

**Actual behavior**
The sections drop down in the Templates dialog is extended slowly (1-2 second)

https://user-images.githubusercontent.com/103422815/163378359-e0fd4b5c-5dcc-4032-9a8b-8e6dbb2eccb2.mp4

OS: [Microsoft Windows]
Browsers: Google Chrome: Version 100.0.4896.75/FireFox: Version 99.0 (64-bit)
Ketcher v2.4.2-158-g2312c8e8

mb-mbe:

## Check list
- [ ] unit-tests written
- [ ] e2e-tests written
- [ ] documentation updated


packages/ketcher-react/src/script/ui/dialog/template/TemplateDialog.tsx
```diff
@@ -311,6 +338,9 @@ const TemplateDialog: FC<Props> = (props) => {
                     key={groupName}
                     onChange={handleAccordionChange(groupName)}
                     expanded={shouldGroupBeRended}
+                    slotProps={{
+                      transition: { unmountOnExit: true },
+                    }}
```
AlexeyGirin: Check if that really needed

AlexeyGirin:

packages/ketcher-react/src/script/ui/dialog/template/TemplateTable.tsx
```diff
@@ -77,6 +85,119 @@ function createKeyDownHandler(callback: () => void) {
...
+            }}
+            onKeyDown={createKeyDownHandler(() => onDelete?.(tmpl))}
+          >
+            <Icon name="delete" />
+          </button>
+        )}
+        {!isFunctionalGroupTemplate(tmpl) && !isSaltOrSolventTemplate(tmpl) && (
+          <button
+            type="button"
+            data-testid={'edit-template-button'}
```
svvald: You can pass this (and other values above) as a pure string `data-testid="edit-template-button"`

packages/ketcher-react/src/script/ui/dialog/template/TemplateTable.tsx
```diff
@@ -77,6 +85,119 @@ function createKeyDownHandler(callback: () => void) {
...
+                    downScale: true,
+                  }}
+                />
+              </div>
+            )}
+          </div>
+          <div
+            className={`${classes.structTitle} ${
+              isSelected ? classes.selectedTitle : ''
+            }`}
```
svvald: We have `clsx` package to nicely combine classes, please use it here and everywhere where template strings are used for classnames

packages/ketcher-react/src/script/ui/dialog/template/TemplateTable.tsx
```diff
@@ -77,6 +85,119 @@ function createKeyDownHandler(callback: () => void) {
...
+  index: number;
+  isSelected: boolean;
+  shouldRenderPreview: boolean;
+  renderOptions?: any;
+  onSelect: (tmpl: Template) => void;
+  onDelete?: (tmpl: Template) => void;
+  onAttach?: (tmpl: Template) => void;
+}
+
+const TemplateItem: FC<TemplateItemProps> = memo(
```
svvald: Let's keep one component per file please

packages/ketcher-react/src/script/ui/dialog/template/TemplateTable.tsx
```diff
@@ -77,6 +85,119 @@ function createKeyDownHandler(callback: () => void) {
...
+      prevProps.shouldRenderPreview === nextProps.shouldRenderPreview &&
+      prevProps.tmpl === nextProps.tmpl &&
+      prevProps.index === nextProps.index &&
+      prevProps.renderOptions === nextProps.renderOptions &&
+      prevProps.onSelect === nextProps.onSelect &&
+      prevProps.onDelete === nextProps.onDelete &&
+      prevProps.onAttach === nextProps.onAttach
+    );
+  },
+);
```
svvald: Aren't those checks excessive? You just list all existing props of this component and by default `memo` will perform the same checks

svvald (changes requested):

packages/ketcher-react/src/script/ui/dialog/template/TemplateTable.tsx
```diff
@@ -88,79 +209,107 @@ const TemplateTable: FC<TemplateTableProps> = (props) => {
     renderOptions,
   } = props;
 
+  // Render all tiles immediately to keep accordion height stable,
+  // then progressively mount heavy SVG previews with low-priority updates.
+  const INITIAL_PREVIEW_COUNT = 4;
+  const PREVIEW_BATCH_SIZE = 4;
```
svvald: These values are defined on each render, it's pretty harmless as they're primitives but I'd still move them out of component and just have them file-scoped

packages/ketcher-react/src/script/ui/dialog/template/TemplateTable.tsx
```diff
@@ -77,6 +85,119 @@ function createKeyDownHandler(callback: () => void) {
...
+          data-testid={tmpl.struct.name}
+          className={classes.selectButton}
+          title={greekify(getTemplateTitle(tmpl, index))}
+          onClick={() => onSelect(tmpl)}
+        >
+          <div className={classes.structSlot}>
+            {!shouldRenderPreview && (
+              <div className={classes.structPlaceholder} />
+            )}
+            {shouldRenderPreview && (
```
svvald: Could be replaced by ternary

packages/ketcher-react/src/script/ui/dialog/template/TemplateTable.tsx
```diff
@@ -88,79 +209,107 @@ const TemplateTable: FC<TemplateTableProps> = (props) => {
...
+    // Recalculate on resize
+    const resizeObserver = new ResizeObserver(calculateSize);
+    resizeObserver.observe(containerRef.current);
+
+    return () => resizeObserver.disconnect();
+  }, []);
+
+  useEffect(() => {
+    setPreviewRenderedCount(Math.min(INITIAL_PREVIEW_COUNT, templates.length));
+  }, [templates]);
```
svvald: This useEffect usage is an anti-pattern, more on it [https://react.dev/learn/you-might-not-need-an-effect](here). In short, if `templates` is changed, the component will re-render once with stale `previewRenderedCount` and then will do it once again caused by state change as effect fires after component render

The better approach would be
```
const [prevTemplates, setPrevTemplates] = useState(templates);
const [previewRenderedCount, setPreviewRenderedCount] = useState(() =>
  Math.min(INITIAL_PREVIEW_COUNT, templates.length),
);

if (templates !== prevTemplates) {
  setPrevTemplates(templates);
  setPreviewRenderedCount(Math.min(INITIAL_PREVIEW_COUNT, templates.length));
}
```

svvald: And to get even further, ideally `templates` should have a stable reference so when they have identical content (e.g. filtered by the same filter), that would return the same reference, now it's not the case I suppose but we can leave it at that and have a tech debt/investigation ticket on this particular point

packages/ketcher-react/src/script/ui/dialog/template/TemplateTable.tsx
```diff
@@ -88,79 +209,107 @@ const TemplateTable: FC<TemplateTableProps> = (props) => {
...
+  const [, startTransition] = useTransition();
+  const [containerSize, setContainerSize] = useState<{
+    width: number;
+    height: number;
+  } | null>(null);
+  const containerRef = useRef<HTMLDivElement>(null);
+
+  // Calculate container size once to avoid getBoundingClientRect() calls during rendering
+  // This prevents forced reflows when rendering multiple items in a batch
+  useEffect(() => {
```
svvald: Let's use `useLayoutEffect` here instead, it happens before any regular effects in the render tree and before the actual browser paint thus calculating and providing `containerSize` correctly. Otherwise, initial batch of `RenderStruct` will render first and will have to fallback to `getBoundingClientRect` anyway

packages/ketcher-react/src/script/ui/dialog/template/TemplateTable.tsx
```diff
@@ -88,79 +209,107 @@ const TemplateTable: FC<TemplateTableProps> = (props) => {
...
+    if (!containerRef.current) return;
```
svvald: Pretty minor, but I'd prefer to have this short-circuit return on a newline. Also please check the same for other places in your changes

svvald (changes requested):

packages/ketcher-react/src/script/ui/dialog/template/TemplateTable.tsx
```diff
@@ -88,79 +209,107 @@ const TemplateTable: FC<TemplateTableProps> = (props) => {
...
+    setPreviewRenderedCount(Math.min(INITIAL_PREVIEW_COUNT, templates.length));
+  }, [templates]);
+
+  useEffect(() => {
+    if (previewRenderedCount >= templates.length) {
+      return;
+    }
+
+    // Use idle/frame scheduling for progressive preview hydration.
+    const requestIdle = window.requestIdleCallback;
```
svvald, line 114: Also maybe I missed it on our discussion (or I didn't attend some), but what's the benefit of using `requestIdleCallback`? Why can't we just use rAF in all cases?

packages/ketcher-react/src/script/ui/dialog/template/TemplateTable.tsx
mb-mbe: @AlexeyGirin , please create a ticket with tech-debt/investigation.

packages/ketcher-react/src/script/ui/dialog/template/TemplateTable.tsx
mb-mbe: requestIdleCallback is preferable here, since this task is rather low priority, there is no need to clog animation frame with it. But if this API method is not available, than requestAnimationFrame is a fallback choice, way less suitable, to be honest, compared to idle-callback. The logic hidden behind requestIdleCallback is not interaction-critical.

svvald (approved):

PR #10338, opened 2026-06-29
#7043 - Fix multi-line textbox vertical position when taller than viewport

### LINKED ISSUE #7043: Textbox position is wrong if textbox size greater than view port (?)
Follow up on https://github.com/epam/Indigo/issues/1679

**Steps to Reproduce**
1. Go to Macro - Sequence mode (clean canvas)
2. Load from KET: [tm 1.zip](https://github.com/user-attachments/files/20061955/tm.1.zip)

**Actual behavior**
`Huge TEXT message` textbox position is wrong
Image
Image
Image

**Expected behavior**
`Huge TEXT message` textbox position is correct
Image

**Environment details:**
- Ketcher Version 3.4.0-rc.1 Build at 2025-05-01; 08:56:06
- Indigo Version 1.32.0-rc.2.0
- Chrome Version 136.0.7103.49 (Official Build) (64-bit)
- Win10

Feature: https://github.com/epam/ketcher/issues/496

aslmbk:
### Problem
A multi-line text annotation imported from KET/CDX (e.g. the "Huge TEXT Message" block from the issue) was rendered far above its intended position whenever the text box was taller than the viewport, pushing it out of view.

Follow-up to [[epam/Indigo#1679](https://github.com/epam/Indigo/issues/1679)](https://github.com/epam/Indigo/issues/1679).

### Root cause
Imported multi-line content is stored as a single paragraph with embedded newlines, so the renderer (`ReText.show`) draws it as **one** Raphael `<text>` element. Raphael **vertically centers** an element on its anchor `y`, but the text model anchors a box at its **top** (rows flow downward — that's how `position`/`pos`/export and the click-to-create tool treat it).

The result is a vertical offset of **half the element's height**: negligible (~10px) for single-line labels, but ~874px for the 81-line block in the issue — larger than the viewport, so the text lands completely off-position.

### Fix
Compensate for Raphael's vertical centering: when a rendered text element spans more than one line, shift it down so its **first line sits at the anchor** and the remaining lines flow downward — exactly matching single-line and multi-paragraph (typed) text.

- Localized to the renderer; the data model (`position`/`pos`/KET round-trip) is untouched.
- Single-line text and text typed in the editor (which produces separate paragraphs, not embedded newlines) are **pixel-identical** to before — no behavior change elsewhere.

### Screenshots
Screenshot 2026-06-29 at 11 43 41

## Check list
- [x] unit-tests written
- [ ] e2e-tests written
- [ ] documentation updated


packages/ketcher-core/src/application/render/restruct/retext.ts
```diff
@@ -238,6 +239,13 @@ class ReText extends ReObject {
             restruct.molecule.texts.keyOf(this.item),
           );
           path.translateAbs(shiftX, shiftY + (styles.shiftY || 0));
+          const lineCount = countTextLines(text);
+          if (lineCount > 1) {
+            path.translateAbs(
+              0,
+              getMultilineTopAnchorOffset(path.getBBox().height, lineCount),
+            );
+          }
```
svvald: It seems that you try to fix the issue on render-time when the core reason seems to be that text coordinates are not saved properly when serializing text nodes. Could you explore that please?

svvald (dismissed):

packages/ketcher-core/src/application/render/restruct/retext.ts
aslmbk: I looked into how the text coordinates are saved, as you asked. I saved the file and loaded it back, and the coordinates stayed the same, so the coordinates themselves are saved correctly. The text is placed wrong only when it is drawn: the renderer puts a multi-line text centered on the point, but we need the top of the text to sit on that point. The current change fixes this at draw time, and it works correctly now.

Still, your comment is fair — there is a cleaner place to fix it. Right now the whole multi-line text is stored as one block with the line breaks inside it. When we read the file, we could split it into separate lines (separate paragraphs). Then the usual drawing code would place each line correctly, with no adjustment needed at draw time. This also matches how newer files already store multi-line text (one line per paragraph).

I can rework the fix this way instead. Would you like me to do that?

rrodionov91: Okay, you are right. I see the root cause now in how raphael process text rendering. We pass y coordinate of top left corner of text, but raphael moves the first tspan inside it by ~half.
image

I think the fix in this case could be removing this "dy" attribute from **first** tspan inside text. In this case we don't need to calculate shifts for each row manually and it's much shorter. I think that the raphael's behaviour is buggy and it does not allow to configure it somehow.

rrodionov91 (changes requested):

packages/ketcher-core/src/application/render/restruct/retext.ts
aslmbk: done

rrodionov91 (approved):

PR #10355, opened 2026-07-01
#5275 - Size of elements is reduced after Save to KET format if the element is corrupted

### LINKED ISSUE #3406: Structures reduce size after paste if there are functional groups
**Steps to Reproduce**
1. Draw 2 functional groups linked by bond and any other structures.
image
2. Save to file.
3. Clear canvas
4. Open saved file

**Actual behavior**
Structures  reduce size and moved.
image

**Expected behavior**
Structures added to canvas with same size and position.

### LINKED ISSUE #5275: Size of elements is reduced after Save to KET format if the element is corrupted
**Steps to Reproduce**
1. Add Benzene ring to the Canvas and corrupt it:
image
2. Add text and different shapes:
image
3. Click on Save as...
4. Select KET Format
5. Copy JSON from preview
6. Paste to the Canvas and observe

**Expected behavior**
- The size of elements is the same 

**Actual behavior**
- The sizes of shapes and corrupted ring are reduced:
image

**Environment:**
- **Ketcher** [Version 2.24.0-rc.1] Build at 2024-08-06; 19:25:43
- **Indigo** Version 1.23.0-rc.1.0-g85f89b8f5-x86_64-linux-gnu-11.2.1

**Desktop:**
 - OS: Windows 11 Enterprise
 - Browser: Chrome
 - Version: Version 127.0.6533.100 (Official Build) (64-bit)

**Additional context**
- This case should be automated, tests are added to xls
- Detected during preview tests for Image feature - https://github.com/epam/Indigo/issues/2144. But this issue isn;t related to new changes this behavior was before. **The fix should be verified for images, shapes and text as well**.

la-nuta:
Closes #5275
Closes #3406

Both issues share one root cause: `Struct.rescale()` normalized coordinates so the **average** bond length became 1 (Ketcher's canvas unit). The average isn't robust — a few atypical bonds drag it away from 1 and the whole canvas gets scaled by that wrong factor. In #5275 one atom dragged out of a ring lengthens two bonds, so everything shrinks, shapes and text included. In #3406 the long bonds linking abbreviated functional groups do it in the other direction — in a 46-bond structure, 4 linkers of 3.4–4.2 pull the mean to 1.25, so the whole drawing is scaled to 80%.

Fixed by normalizing on the **median** bond length, which ignores a minority of outlier bonds.

Note this is **not KET-specific** — an earlier version of this branch gated rescaling per format; that was dropped in `a2ac3e6c` since the mean was wrong for every format.

## Before / after

Captured from the standalone build at 100% zoom — *before* from `origin/master`, *after* from this branch, same input file in each pair.

**#5275** — benzene ring with one atom dragged out, plus a rectangle, an ellipse and a text label. Master renders the whole drawing at **45.6%**: ring side 0.4557 vs 1.0000, rectangle diagonal 1.0191 vs 2.2361.

| before (master) | after (this PR) |
|---|---|
| 5275-before | 5275-after |

**#3406** — 42 bonds at ~1.0 plus 4 linkers of 3.4–4.2 joining six abbreviated groups. Master renders it at **80.2%**: median bond 0.8015 vs 1.0000, drawing extent 19.5235 vs 24.3581.

| before (master) | after (this PR) |
|---|---|
| 3406-before | 3406-after |

## Summary

**`struct.ts`** — new `getBondLengths()`, `getMedianBondLength()` and `static median()`; `rescale()` now uses the median, and refuses scale factors outside `[0.01, 100]` (degenerate geometry, better left alone) — replacing the old `if (avg < 1e-3) avg = 1` case.

**`serializers/mol/utils.js`** — `calculateAverageBondLength()` → `calculateMedianBondLength()`, pooling lengths across all fragments of a reaction so reactants, agents and products keep their relative sizes.

**`ui/state/server/index.js`** — dropped the dead `rescale: method === 'layout'` option (`parseStruct()` destructures it into `_rescale` and ignores it).

**`__tests__/struct.test.ts`** — new, 14 tests: `median()` edge cases, `rescale()` normalization/idempotence/no-bonds, the `[0.01, 100]` boundaries, and a #5275 regression (ring with one dragged atom plus a rectangle — mean > 1.5, median = 1, rectangle unchanged).

## Changed files (2752)

| Group | Files | What |
|---|---|---|
| Source fix | 3 | `struct.ts`, `serializers/mol/utils.js`, `ui/state/server/index.js` |
| New unit tests | 1 | `struct.test.ts` |
| Autotest fixtures & specs | 12 | see below |
| Regenerated expected data | 1019 | `.mol` 485, `.ket` 172, `.rdf` 122, `.rxn` 82, `.sdf` 63, `.cdxml` 45, `.cdx` 39, `.cml` 11 |
| Regenerated PNG baselines | 1717 | |

**Review the 16 code files; the other 2736 are mechanical** — the median rule shifts the scale factor for nearly every structure, so almost every stored geometry and screenshot moves. None are hand-edited.

The 12 autotest files:
- **Clipboard determinism** (`helpers.ts` + 5 fixtures) — new `resetClipboard()` called between tests. The clipboard survives across a Playwright worker, and Ketcher reads it in `isPasteContentAvailable()` to enable the context menu's *Paste* item, so any screenshot with a context menu depended on test order.
- **Un-skipped by this fix** — `monomer-creation.spec.ts` (case 30 fully; case 29 except *1. Amino Acid* / *6. CHEM*, still blocked by [Indigo #3292](https://github.com/epam/Indigo/issues/3292)) and `reagents-cdx-format.spec.ts` (a `test.fail()` deferred since 2.21.0-rc.1).
- **Adjusted for the new geometry** — the three `*-bugs.spec.ts` files and `copy-cut-paste.spec.ts`.

## Test plan
- [x] `struct.test.ts` — 14/14 pass
- [x] ESLint, Prettier and `tsc --noEmit` clean — re-verified after the master merge
- [ ] Full e2e — green before the master merge; the post-merge run has not completed, `build_ketcher` fails in *Install dependencies* on an unrelated apt 404 (`libperl5.32_5.32.1-4+deb11u5`) in the CI image, so the e2e parts are skipped
- [x] #5275 manual repro — corrupted ring (mean 2.19, median 1.00) plus text and shapes → Save as KET → paste back: ring side, rectangle and ellipse all at ratio 1.000000, position shift is pure translation. The mean rule would have scaled it to 45.6%.
- [x] #3406 manual repro — `S-Groups/Nucleotide components and superatoms for Clean Up test.ket` (6 superatoms; 42 bonds at ~1.0 plus 4 linkers at 3.4–4.2, mean 1.2476, median 1.0000) → open → save → clear → reopen → paste: extent ratio 1.000000 at every step. The mean rule would have scaled it to 80.2%.

## Check list
- [x] unit-tests written
- [x] e2e-tests written
- [ ] documentation updated

🤖 Generated with [Claude Code](https://claude.com/claude-code)


ketcher-autotests/tests/specs/User-Interface/Editing-Tools/Copy-Cut-Paste/copy-cut-paste.spec.ts
```diff
@@ -84,8 +84,11 @@ test.describe('Copy/Cut/Paste Actions', () => {
     Test case: EPMLSOPKET-1712
     Description: After the clicking the 'Cut' button, the selected object disappears.
     */
-    await openFileAndAddToCanvas(page, 'Molfiles-V2000/query-features.mol');
-    await getAtomLocator(page, { atomLabel: 'C', atomId: 41 }).click({
+    await openFileAndAddToCanvasAsNewProject(
+      page,
+      'Molfiles-V2000/query-features.mol',
+    );
+    await getAtomLocator(page, { atomLabel: 'C', atomId: 11 }).click({
```
la-nuta, line 89: ## What changed

In the "Cut" test (`EPMLSOPKET-1712`):

- Switched from `openFileAndAddToCanvas` to `openFileAndAddToCanvasAsNewProject`.
- Updated the target atom ID from `41` to `11`.

```diff
- await openFileAndAddToCanvas(page, 'Molfiles-V2000/query-features.mol');
- await getAtomLocator(page, { atomLabel: 'C', atomId: 41 }).click({
+ await openFileAndAddToCanvasAsNewProject(page, 'Molfiles-V2000/query-features.mol');
+ await getAtomLocator(page, { atomLabel: 'C', atomId: 11 }).click({
    force: true,
  });
```

## Why

`openFileAndAddToCanvas` adds the structure to the *existing* canvas (a
paste-style load), so atom IDs are offset by whatever is already present — the
target carbon ended up at ID `41`. `openFileAndAddToCanvasAsNewProject` clears
the canvas first, so IDs start from zero and the same atom is now `11`.

## Relationship to the rescale fix

This is a **test-hygiene** fix, not a behavior change caused by the
`rescale()` skip:

- The fixture is `Molfiles-V2000/query-features.mol` (V2000, **not** KET), so
  the KET-paste `rescale()` skip in `shared.ts` does not apply to it.
- The change only makes the test start from a clean, deterministic canvas
  state, removing the dependency on pre-existing atom IDs.

ketcher-autotests/tests/test-data/KET/modified-multitail-arrow-with-added-tails-expected.ket
```diff
@@ -217,7 +217,7 @@
             {
                 "label": "C",
                 "location": [
-                    10.334849152128566,
+                    10.334849152128568,
```
la-nuta: ## What changed

- Regenerated `KET/modified-multitail-arrow-with-added-tails-expected.ket` with updated coordinate values.

```diff
-                    10.334849152128566,
+                    10.334849152128568,
...
-                    -5.000033111149813,
+                    -5.000033111149811,
```

## Why

`verifyFileExport` compares the exported KET against the committed expected file. The diff is a sub-ULP (last bit) IEEE 754 precision difference — the same mathematical path now rounds slightly differently. The `rescale()` fix from PR #5275 introduces an extra scaling pass on coordinates, and rounding accumulates into the last digit of some values.

## Relationship to the rescale fix

This is a **test-data update**, not a behavior change visible to users:

- The coordinate values are functionally identical — the difference is in the 16th significant digit, well below any rendering or chemistry threshold.
- The expected file is regenerated by running the test once with `GENERATE_DATA=true`, the same mechanism used for all KET fixture updates.

packages/ketcher-react/src/script/ui/state/shared.ts
```diff
@@ -143,7 +143,16 @@ export function load(struct: string | Struct, options?) {
...
+      // KET is Ketcher's native format — coordinates are already in canvas
+      // scale, so rescaling a KET paste would shrink elements when bond
+      // lengths differ from 1 Å (e.g. corrupted molecules with moved atoms).
+      const pastedFormat =
+        typeof struct === 'string' ? identifyStructFormat(struct) : null;
+      if (
+        !preserveViewport &&
+        !hasMoleculeToMonomerConnections &&
+        !(isPaste && pastedFormat === SupportedFormat.ket)
+      ) {
```
svvald: The core of this issue is that `struct.rescale()` computes a scale factor based on average bond length and applies it to all canvas entities including non-chemical ones (texts, shapes, arrows). When a molecule is corrupted, the average bond length is changes hence the scale factor is wrong and is applied to everything

Current fix has several flaws:
1. It only guards the `load` function, however, `struct.rescale()` is called in other files and functions as well (`setMolecule`, `setFragment` etc.)
2. It only covers paste action and ket format for whatever reason which is not justified – the same issue will be reproduced if using e.g. MOL format and opening a file/loading a molecule with Ketcher API

Couple of ideas what might be done to make a proper generalized fix:
1. Do not take into account non-chemical entities when rescaling based on average bond length as it doesn't make much sense
2. Make `rescale` function less sensitive to outlier bond length values, for example, you can use median/mode bond length instead of average or you can calculate bond length variance and only apply scaling if it's within some threshold (which has to be determined heuristically)

svvald (changes requested):

AlexeyGirin:

AlexeyGirin:
Check the same bahaviour for other formats as well (mol v2000 and etc)
Rescale should work the same for each format
Rescale logic change (use median bond legth) should remain

PR #10371, opened 2026-07-02
#9549 Color Picker UI

### LINKED ISSUE #9549: Color Picker UI
## Color Picker UI 

In **Molecules** settings, under **Stereochemistry**, add a new color picker for:
- Absolute Center color
- AND Centers color
- OR Centers color

The color picker should match the design shown in the attached mockup, featuring:

- **Color Palette:**  
  A grid of preset color swatches is displayed at the top, allowing users to select from a range of commonly used colors.

- **Custom Colors Section:**  
  Below the palette, a "Custom Colors" area enables users to define their own color:
    - **Color Sliders:**  
      Two sliders are provided, one for selecting the hue and another for adjusting brightness or saturation.
    - **Color Preview:**  
      A square preview box shows the currently selected or custom color.
    - **HEX Input:**  
      Users can enter a specific HEX color code directly into the input field for color selection.  
      Only valid HEX codes from a predefined HEX code library will produce a color.  
      A mask restricts input to the correct length and symbols, preventing errors.
    - **Delete Button:**  
      An icon next to the HEX input allows users to clear the custom color selection.
    - **Close/Open Icon:**  
      The custom colors section is opened by clicking the “+” icon and closed by clicking the “X” icon.

- **Action Buttons:**
    - **Cancel:**  
      Discards any changes and closes the color picker.
    - **Apply:**  
      Confirms the selected color and applies it to the relevant stereochemistry center.

Image

Image

---

## Settings Presets

- A **“Reset to …”** menu provides **“Default”** and **“ACS”** options, as well as **“Save Settings”** and **“Load Settings”** actions.
    - Selecting **“Default”** resets all relevant settings to the Ketcher default preset.
    - Selecting **“ACS”** resets all relevant settings to the ACS preset.
- Preset and custom settings are saved in the browser cache.

Image

dmitrylabs:
image

What was done: 

1. Component was rewritten based on Figma designs https://www.figma.com/design/MZ28kJRelnXvBodXu54UKh/Macromolecules--Polymer-Editor--Playground?node-id=14665-317909&m=dev
2. Local storage settings are used for up to 10 custom colors saving (ketcher-opts)
3. 2 new claude code / github copilot compatible skills added: for adding new local storage settings properly; for decomposing large react component files.
4. react-colorful dependency is removed

## Check list
- [x] unit-tests written
- [ ] e2e-tests written
- [ ] documentation updated


packages/ketcher-react/src/script/ui/component/form/colorPicker/ColorPicker.tsx
```diff
@@ -14,95 +14,225 @@
...
+  const [lightness, setLightness] = useState(50);
+  const [hexInput, setHexInput] = useState('');
+  const [popupPosition, setPopupPosition] = useState<{
+    top?: number;
+    bottom?: number;
+    left: number;
+    visibility: 'hidden' | 'visible';
+  }>({ left: 0, visibility: 'hidden' });
+  const triggerRef = useRef<HTMLButtonElement>(null);
+  const popupRef = useRef<HTMLDivElement>(null);
```
svvald: This component is too bloated now as it handles several separate concerns – popup open/close/positioning, settings persistence, rendering the presets grid, rendering the custom color section

One way to de-clutter it would be to extract presets grid and custom color functionality into their separate components in use them here

packages/ketcher-react/src/script/ui/component/form/colorPicker/ColorPicker.tsx
```diff
@@ -14,95 +14,225 @@
...
+    const { h, l } = hexToHsl(hex);
+    setHue(h);
+    setLightness(l);
+    setHexInput(hex.replace('#', '').toUpperCase());
+  }, []);
+
+  // Sync internal state and compute popup position when popup opens
+  useEffect(() => {
+    if (isOpen) {
+      const initialColor = value || '#FF3232';
```
svvald: The same value is used as initial value in `useState` above, perhaps worth extracting a constant for it?

packages/ketcher-react/src/script/ui/component/form/colorPicker/ColorPicker.tsx
```diff
@@ -14,95 +14,225 @@
...
+        !popupRef.current.contains(e.target as Node) &&
+        triggerRef.current &&
+        !triggerRef.current.contains(e.target as Node)
+      ) {
+        setIsOpen(false);
       }
     };
+    document.addEventListener('mousedown', handleOutsideClick);
+    return () => document.removeEventListener('mousedown', handleOutsideClick);
+  }, [isOpen]);
```
svvald: We already have a MUI library in our app and I suppose we can use its primitives for handling this dropdown open/close/positioning rather than doing so from scratch. Please refer to `ModeControl` or `ElementWithPopover` components

packages/ketcher-react/src/script/ui/component/form/colorPicker/ColorPicker.tsx
```diff
@@ -14,95 +14,225 @@
...
 
-  const handlePaletteOpen = () => {
-    setIsPaletteOpen(true);
-  };
-  const handleColorChange = (color) => {
-    handleChange(color);
-  };
+  const handleHexInputChange = useCallback(
+    (e: React.ChangeEvent<HTMLInputElement>) => {
+      const raw = e.target.value.replace(/[^0-9a-fA-F]/g, '').slice(0, 6);
```
svvald: I suppose it belongs to utils rather than here

packages/ketcher-react/src/script/ui/component/form/colorPicker/ColorPicker.tsx
```diff
@@ -14,95 +14,225 @@
...
     if (e.key === 'Escape') {
       e.preventDefault();
       e.stopPropagation();
+      setIsOpen(false);
     }
-  };
+  }, []);
+
+  const hueBg =
+    'linear-gradient(to right, #ff0000, #ffff00, #00ff00, #00ffff, #0000ff, #ff00ff, #ff0000)';
```
svvald: It is a constant basically, why don't put it into css?

packages/ketcher-react/src/script/ui/component/form/colorPicker/ColorPicker.utils.ts
```diff
@@ -0,0 +1,107 @@
...
+  const normalizedColor = normalizeHexColor(color);
+
+  if (isPresetColor(normalizedColor)) {
+    return sanitizeCustomColors(colors);
+  }
+
+  return sanitizeCustomColors([normalizedColor, ...colors]);
+}
+
+export function hexToHsl(hex: string): { h: number; s: number; l: number } {
```
svvald, line 64: I get those are util functions which implementation is known but let's avoid shorthanded params and variables names

packages/ketcher-react/src/script/ui/component/form/colorPicker/ColorPicker.tsx
```diff
@@ -14,95 +14,225 @@
...
+      ) {
+        setIsOpen(false);
       }
     };
+    document.addEventListener('mousedown', handleOutsideClick);
+    return () => document.removeEventListener('mousedown', handleOutsideClick);
+  }, [isOpen]);
+
+  const applyColorFromHsl = useCallback((h: number, l: number) => {
+    const hex = hslToHex(h, 100, l);
```
svvald: I assume this is by design as we only have hue and lightness knobs, but not saturation yet I want to emphasize it – our current use flow is not uniform
We can specify some color using a hex value for which saturation is actually below 100, however once we adjust hue/lightness knobs even slightly, it will be reset to an absolutely another color with saturation 100 (which is especially visible for shades of grey, it will have lightness jumping from 0 to 100)
I just want you to double check and verify with the business that it's okay

packages/ketcher-react/src/script/ui/component/form/colorPicker/ColorPicker.tsx
```diff
@@ -14,95 +14,225 @@
...
+      const initialColor = value || '#FF3232';
+      applyHexColor(initialColor);
+      setIsCustomOpen(false);
+
+      const newColors = addCustomColor(
+        [...(settings?.colorPickerCustomColors ?? [])],
+        initialColor,
+      );
+      setCustomColors(newColors);
+      updateSettings({ colorPickerCustomColors: newColors });
```
svvald: This effect depends on `isOpen`, it will update settings each time dropdown is open even it's closed right after that with not actions performed. Ideally we should only update settings on applying the actions in dropdown

packages/ketcher-react/src/script/ui/component/form/colorPicker/ColorPicker.tsx
```diff
@@ -14,95 +14,225 @@
...
+    if (isOpen) {
+      const initialColor = value || '#FF3232';
+      applyHexColor(initialColor);
+      setIsCustomOpen(false);
+
+      const newColors = addCustomColor(
+        [...(settings?.colorPickerCustomColors ?? [])],
+        initialColor,
+      );
+      setCustomColors(newColors);
```
svvald: And essentially there is no much sense in having custom colors as a separate state – you have to always keep it in sync with settings while you only actually read them. An easier way would be just to derive it from settings

svvald (changes requested):

packages/ketcher-react/src/script/ui/component/form/colorPicker/ColorPicker.tsx
dmitrylabs: Extracted to separate sub-components

packages/ketcher-react/src/script/ui/component/form/colorPicker/ColorPicker.tsx
dmitrylabs: Added a constant for it

packages/ketcher-react/src/script/ui/component/form/colorPicker/ColorPicker.tsx
dmitrylabs: Added sanitizeHexInput util

packages/ketcher-react/src/script/ui/component/form/colorPicker/ColorPicker.tsx
dmitrylabs: Because we use different backgrounds for different ColorSlider's, and passing it as a param. For hue slider we use this value, for lightness slider we use dynamic (calculated) value.

packages/ketcher-react/src/script/ui/component/form/colorPicker/ColorPicker.utils.ts
dmitrylabs: We are using h, s, l here as part of HSL abbreviation, like r, g, b for RGB. Do you suggest to use full names like "saturation", "lightness", "red", "blue" etc. in such cases?

packages/ketcher-react/src/script/ui/component/form/colorPicker/ColorPicker.tsx
dmitrylabs: Good point, fixed

packages/ketcher-react/src/script/ui/component/form/colorPicker/ColorPicker.tsx
dmitrylabs: That's true, I verified it with designers — they agreed with our current non-ideal solution.

packages/ketcher-react/src/script/ui/component/form/colorPicker/ColorPicker.tsx
dmitrylabs: Good point, switched to material popover

packages/ketcher-react/src/script/ui/component/form/colorPicker/ColorPicker.tsx
dmitrylabs: It is used to ensure proper list of custom colors before the opening. If color picker contains color that are not in color swatcher or custom colors list - it adds it to the beginning of custom colors list (to make UI consistent). Currently we don't have internal custom colors state that might be applied later - it uses settings (local storage).

packages/ketcher-react/src/script/ui/component/form/colorPicker/ColorPicker.utils.ts
svvald: Yep, on the second thought in this particular case that might be not needed, let's keep it as is

packages/ketcher-react/src/script/ui/component/form/colorPicker/ColorPicker.tsx
```diff
@@ -14,176 +14,237 @@
...
 
 const ColorPicker = (props: Props) => {
-  const [isOpen, setIsOpen] = useState(false);
-  const [isPaletteOpen, setIsPaletteOpen] = useState(false);
   const { onChange, value } = props;
+  const { settings, updateSettings } = useSettings();
+  const [isOpen, setIsOpen] = useState(false);
+  const [isCustomOpen, setIsCustomOpen] = useState(false);
+  const [pendingColor, setPendingColor] = useState(value || DEFAULT_COLOR);
+  // eslint-disable-next-line react-hooks/exhaustive-deps
```
svvald: Is it needed here? I don't see any deps arrays

packages/ketcher-react/src/script/ui/component/form/colorPicker/ColorPicker.tsx
```diff
@@ -14,176 +14,237 @@
...
-      e.preventDefault();
-      e.stopPropagation();
-    }
-  };
+  const handleCancel = useCallback(() => {
+    setIsOpen(false);
+  }, []);
+
+  const handleKeyDown = useCallback(
+    (e: React.KeyboardEvent) => {
```
svvald: Let's remove `React.` prefix and import `KeyboardType` event directly from `react`. Please also check for this in other places

packages/ketcher-react/src/script/ui/component/form/colorPicker/ColorPicker.tsx
```diff
@@ -14,176 +14,237 @@
...
+    [handleToggleOpen],
+  );
 
   return (
     <div
       className={classes.colorPickerWrapper}
       data-testid={isOpen ? 'color-picker-field-open' : 'color-picker-field'}
       onClick={(e) => e.preventDefault()}
-      onKeyDown={handleWrapperKeyDown}
       role="none"
```
svvald, line 72: It was already here but I suppose we don't need both of these?

packages/ketcher-react/src/script/ui/component/form/colorPicker/ColorPicker.tsx
```diff
@@ -14,176 +14,237 @@
...
-    if (e.key === 'Enter' || e.key === ' ') {
-      e.preventDefault();
-      handleClick(e);
+    const newColors = customColors.filter(
+      (color) => color.toUpperCase() !== selectedCustomColor.toUpperCase(),
+    );
+    updateSettings({ colorPickerCustomColors: newColors });
+  }, [selectedCustomColor, customColors, updateSettings]);
+
+  const handleApply = useCallback(() => {
```
svvald: Please check `useCallback` usage here and in other places. If you pass it as handler to a native DOM element (`<div>`, `<button>` etc.), it won't bring any performance benefits and only will add unnecessary overhead. It only makes sense when passing props to memoized React components

packages/ketcher-react/src/script/ui/component/form/colorPicker/ColorSlider.tsx
```diff
@@ -0,0 +1,130 @@
...
+  ariaLabel,
+}: ColorSliderProps) {
+  const containerRef = useRef<HTMLDivElement>(null);
+  // Track container width as state so re-measuring on resize triggers re-render
+  const [containerWidth, setContainerWidth] = useState(0);
+  const dragRef = useRef<{ startX: number; startValue: number } | null>(null);
+
+  // Measure once after mount (before paint) and on resize
+  useLayoutEffect(() => {
+    if (!containerRef.current) return;
```
svvald: Stylistic nitpick but I'd prefer to have short-circuit on a newline for a better readability
```
if (...) {
return;
}
// newline here as well
```

Please also check it in other places

packages/ketcher-react/src/script/ui/component/form/colorPicker/ColorPicker.tsx
svvald: Correct me if I get you idea wrong but you can still derive it in render-time like the following:
```
const displayCustomColors = useMemo(
  () => addCustomColor([...persistedCustomColors], value || DEFAULT_COLOR),
  [persistedCustomColors, value],
);
```

There is no need to _write_ these values into local storage just on _opening_ the dropdown, it should only be done when _applying_ the changes

packages/ketcher-react/src/script/ui/component/form/colorPicker/ColorPicker.tsx
```diff
@@ -14,176 +14,237 @@
...
+    setLightness(l);
+    setHexInput(hex.replace('#', '').toUpperCase());
+  }, []);
+
+  // Sync internal state when popup opens
+  useEffect(() => {
+    if (isOpen) {
+      const initialColor = value || DEFAULT_COLOR;
+      applyHexColor(initialColor);
+      setIsCustomOpen(false);
```
svvald: And actually it seems that you don't need this hook at all, even if we reduce down to
```
useEffect(() => {
  if (isOpen) {
    applyHexColor(value || DEFAULT_COLOR);
    setIsCustomOpen(false);
  }
}, [isOpen]);
```

It is actually considered a React anti-pattern. You can read on it more [here](https://react.dev/learn/you-might-not-need-an-effect), but overall idea is you shouldn't use effects to reset/sync states. It makes code flow less straightforward and causes unnecessary re-renders

In this particular component I'd check if it's actually needed – I assume Popover unmounts its content when closed by default, so you won't have any stale states for reset

packages/ketcher-react/src/script/ui/component/form/colorPicker/ColorPicker.tsx
```diff
@@ -14,176 +14,237 @@
...
+  const [isOpen, setIsOpen] = useState(false);
+  const [isCustomOpen, setIsCustomOpen] = useState(false);
+  const [pendingColor, setPendingColor] = useState(value || DEFAULT_COLOR);
+  // eslint-disable-next-line react-hooks/exhaustive-deps
+  const customColors = [
+    ...(settings?.colorPickerCustomColors ?? DEFAULT_CUSTOM_COLORS),
+  ];
+  const [hue, setHue] = useState(0);
+  const [lightness, setLightness] = useState(50);
+  const [hexInput, setHexInput] = useState('');
```
svvald: You already did a good job with extracting some of the logic into children components but I think we could go further here – I think you don't actually have to keep those here
Create a wrapper component for color picker content which will gather these states and render the whole content inside – swatches, custom section, action buttons
It will be used here like that:
```
<Popover {...props} >
  <ColorPickerContent {...props} />
</Popover>
```
So essentially current component will only handle the popover open/closed states and CRUD operations with storage. All the "session"-wise state and handlers will be kept in content component thus separating the concerns nicely

svvald (changes requested):

packages/ketcher-react/src/script/ui/component/form/colorPicker/ColorPicker.tsx
dmitrylabs: I updated the line below (removed useless array extraction) and now it's not needed anymore

packages/ketcher-react/src/script/ui/component/form/colorPicker/ColorPicker.tsx
dmitrylabs: Fixed for all colorPicker files

packages/ketcher-react/src/script/ui/component/form/colorPicker/ColorPicker.tsx
dmitrylabs: Removed useless useCallback's from color picker components

packages/ketcher-react/src/script/ui/component/form/colorPicker/ColorSlider.tsx
dmitrylabs: Ok if it's a standard for this project, fixed

packages/ketcher-react/src/script/ui/component/form/colorPicker/ColorPicker.tsx
dmitrylabs: Moved this logic from useEffect to handleToggleOpen. This logic is needed to initialize the data for popup before it's opening. Maybe I'll additionally refactor it after extraction the popup to separate component.

packages/ketcher-react/src/script/ui/component/form/colorPicker/ColorPicker.tsx
dmitrylabs: Let's discuss this

packages/ketcher-react/src/script/ui/component/form/colorPicker/ColorPicker.tsx
dmitrylabs: Moved logic to ColorPickerContent

svvald (approved):

PR #10374, opened 2026-07-02
#10331 - Refactor: Functions should not have identical implementations

### LINKED ISSUE #10331: Refactor: Functions should not have identical implementations
**Problem:** 
Functions should not have identical implementations

**Why is this an issue?**
Two functions having the same implementation are suspicious. It might be that something else was intended. Or the duplication is intentional, which becomes a maintenance burden.

```
function calculateCode() {
  doTheThing();
  doOtherThing();
  return code;
}

function getName() {  // Noncompliant: duplicates calculateCode
  doTheThing();
  doOtherThing();
  return code;
}
```

If the identical logic is intentional, the code should be refactored to avoid duplication. For example, by having both functions call the same function or by having one implementation invoke the other.

```
function calculateCode() {
  doTheThing();
  doOtherThing();
  return code;
}

function getName() { // Intent is clear
  return calculateCode();
}
```

Exceptions

- Functions with fewer than 3 lines are ignored.
- This rule does not apply to function expressions and arrow functions because they don’t have explicit names and are often used in a way where refactoring is not applicable.

```
list.map((item) => ({
  name: item.name,
  address: item.address,
  country: item.country
}));
```

**Problem locations:**
packages/ketcher-core/src/application/editor/operations/drawingEntity/index.ts:79 - Update this function so that its implementation is not identical to the one on line 61.
packages/ketcher-core/src/application/editor/operations/modes/index.ts:32 - Update this function so that its implementation is not identical to the one on line 26.
packages/ketcher-core/src/application/render/renderers/PolymerBondRenderer/SnakeModePolymerBondRenderer.ts:966 - Update this function so that its implementation is not identical to the one on line 922.
packages/ketcher-core/src/application/render/renderers/sequence/SequenceRenderer.ts:1012 - Update this function so that its implementation is not identical to the one on line 970.
packages/ketcher-macromolecules/src/EditorEvents.tsx:156 - Update this function so that its implementation is not identical to the one on line 100.

mariam-khutuashvili:
Changes:

- ketcher-core/.../operations/drawingEntity/index.ts — `DrawingEntityMoveOperation.invertAfterAllOperations` now delegates to `executeAfterAllOperations` (identical redraw/move logic in both directions).
- ketcher-core/.../operations/modes/index.ts — `ReinitializeModeOperation.invert` now delegates to `execute` (both
  re-initialize the mode).
- ketcher-core/.../PolymerBondRenderer/SnakeModePolymerBondRenderer.ts — `moveSnakeBondStart` now delegates to `moveSnakeBondEnd` (a snake bond's path is recomputed wholesale, so which endpoint moved is irrelevant).
- ketcher-core/.../sequence/SequenceRenderer.ts — `previousFromCurrentEdittingMonomer` now delegates to the canonical `previousNode` getter.
- ketcher-macromolecules/src/EditorEvents.tsx — the two identical `selectTool` `handler` arrow functions were hoisted into a single `handleSelectTool` `useCallback`, referenced by both effects.

## Check list
- [ ] unit-tests written
- [ ] e2e-tests written
- [ ] documentation updated

### DIFF
--- a/packages/ketcher-core/src/application/render/renderers/PolymerBondRenderer/SnakeModePolymerBondRenderer.ts
+++ b/packages/ketcher-core/src/application/render/renderers/PolymerBondRenderer/SnakeModePolymerBondRenderer.ts
@@ -919,7 +919,7 @@ export class SnakeModePolymerBondRenderer extends BaseRenderer {
     }
   }
 
-  private moveSnakeBondEnd(): void {
+  private moveSnakeBond(): void {
     const startPosition = this.scaledPosition.startPosition;
     const endPosition = this.scaledPosition.endPosition;
     this.updateSnakeBondPath(startPosition, endPosition);
@@ -930,6 +930,10 @@ export class SnakeModePolymerBondRenderer extends BaseRenderer {
 
     this.hoverAreaElement.attr('d', this.path);
     this.selectionElement?.attr('d', this.path);
+  }
+
+  private moveSnakeBondEnd(): void {
+    this.moveSnakeBond();
   }
 
   private moveGraphBondEnd(): void {
@@ -964,16 +968,7 @@ export class SnakeModePolymerBondRenderer extends BaseRenderer {
   }
 
   private moveSnakeBondStart(): void {
-    const startPosition = this.scaledPosition.startPosition;
-    const endPosition = this.scaledPosition.endPosition;
-    this.updateSnakeBondPath(startPosition, endPosition);
-
-    assert(this.bodyElement);
-    assert(this.hoverAreaElement);
-    this.bodyElement.attr('d', this.path);
-
-    this.hoverAreaElement.attr('d', this.path);
-    this.selectionElement?.attr('d', this.path);
+    this.moveSnakeBond();
   }
 
   private moveGraphBondStart(): void {
--- a/packages/ketcher-core/src/application/render/renderers/sequence/SequenceRenderer.ts
+++ b/packages/ketcher-core/src/application/render/renderers/sequence/SequenceRenderer.ts
@@ -967,12 +967,6 @@ export class SequenceRenderer {
     return SequenceRenderer.getNodeByPointer(this.caretPosition);
   }
 
-  public static get previousFromCurrentEdittingMonomer() {
-    return SequenceRenderer.getNodeByPointer(
-      SequenceRenderer.previousCaretPosition,
-    );
-  }
-
   public static get currentChain() {
     return SequenceRenderer.chainsCollection.chains[
       SequenceRenderer.currentChainIndex
--- a/packages/ketcher-macromolecules/src/EditorEvents.tsx
+++ b/packages/ketcher-macromolecules/src/EditorEvents.tsx
@@ -152,43 +152,6 @@ export const EditorEvents = () => {
     [dispatchShowPreview],
   );
 
-  useEffect(() => {
-    const handler = ([toolName]: [string]) => {
-      if (toolName !== activeTool) {
-        dispatch(selectTool(toolName));
-      }
-    };
-
-    if (editor) {
-      editor.events.error.add((errorText) => {
-        dispatch(openErrorTooltip(errorText));
-      });
-      editor.events.openErrorModal.add(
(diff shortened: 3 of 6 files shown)


packages/ketcher-macromolecules/src/EditorEvents.tsx
```diff
@@ -180,12 +177,12 @@ export const EditorEvents = () => {
...
-      editor.events.selectTool.add(handler);
+      editor.events.selectTool.add(handleSelectTool);
     }
 
     return () => {
       dispatch(selectTool(null));
-      editor?.events.selectTool.remove(handler);
+      editor?.events.selectTool.remove(handleSelectTool);
     };
   }, [editor]);
```
svvald: Seems that this change might be not needed as this effect is almost an identical copy of the first one with `editor.events.openConfirmationDialog.add` missing. I suppose it could be safely deleted and `handleSelectTool` implementation being put back within an effect

packages/ketcher-core/src/application/render/renderers/sequence/SequenceRenderer.ts
```diff
@@ -968,9 +968,7 @@ export class SequenceRenderer {
   }
 
   public static get previousFromCurrentEdittingMonomer() {
```
svvald: Maybe we don't need this method at all then? I see it being used only once at `packages/ketcher-core/src/application/editor/modes/SequenceMode.ts:2808`. Suppose we can remove it and use `SequenceRenderer.previousNode` directly

packages/ketcher-core/src/application/render/renderers/PolymerBondRenderer/SnakeModePolymerBondRenderer.ts
```diff
@@ -964,16 +964,7 @@ export class SnakeModePolymerBondRenderer extends BaseRenderer {
...
-    const endPosition = this.scaledPosition.endPosition;
-    this.updateSnakeBondPath(startPosition, endPosition);
-
-    assert(this.bodyElement);
-    assert(this.hoverAreaElement);
-    this.bodyElement.attr('d', this.path);
-
-    this.hoverAreaElement.attr('d', this.path);
-    this.selectionElement?.attr('d', this.path);
+    this.moveSnakeBondEnd();
```
svvald: While the change per se is fine I suppose it might be slightly more confusing now. Let's create some generic method `moveSnakeBond` which will store the implementation and make both `moveSnakeBondStart` and `moveSnakeBondEnd` to call it so it will be aligned with graph logic

svvald (changes requested):

packages/ketcher-macromolecules/src/EditorEvents.tsx
mariam-khutuashvili: The second effect was a strict subset of the first (identical except the missing openConfirmationDialog.add), so it was only double-registering the handlers. I've deleted it entirely and put the handleSelectTool implementation back inline as handler within the first effect. This removes the duplication at its source, so the original Sonar finding is resolved by removal rather than extraction. Done in the latest commit.

packages/ketcher-core/src/application/render/renderers/sequence/SequenceRenderer.ts
mariam-khutuashvili: The getter was identical to previousNode and, confirmed, used in only that one place. I've removed previousFromCurrentEdittingMonomer and updated the call site in SequenceMode.ts to use SequenceRenderer.previousNode directly. Done in the latest commit.

packages/ketcher-core/src/application/render/renderers/PolymerBondRenderer/SnakeModePolymerBondRenderer.ts
mariam-khutuashvili: I've extracted a generic moveSnakeBond() that holds the shared implementation, and both moveSnakeBondStart() and moveSnakeBondEnd() now delegate to it, keeping the start/end structure symmetric with the graph path. Done in the latest commit.

svvald (approved):

PR #10383, opened 2026-07-03
#6438 - System shows phosphate as CHEM in antisense chain

### LINKED ISSUE #6438: System shows phosphate as CHEM in antisense chain
**Steps to Reproduce**
1. Go to Macro - Flex mode (clean canvas)
2. Load from HELM: `RNA1{r(A)p}|RNA2{p}|RNA3{r(A)}|CHEM1{[4aPEGMal]}|RNA4{p.r(U)}$RNA1,CHEM1,3:R2-1:R1|CHEM1,RNA3,1:R2-1:R1|RNA2,RNA4,1:R2-1:R1|RNA1,RNA4,2:pair-3:pair|RNA3,RNA2,2:pair-1:pair$$$V2.0`
Image
3. Switch to Sequence mode

**Actual behavior**
System shows phosphate as CHEM in antisense chain
Image

**Expected behavior**
System shows phosphate as it is
Image

**Versions**
- Ketcher Version 3.1.0-rc.1 Build at 2025-02-04; 21:27:02
- Indigo Version 1.29.0-rc.1.0-g2e30df8f7-wasm32-wasm-clang-19.0.0
- Chrome Version 133.0.6943.54 (Official Build) (64-bit)
- Win10

Issue found while testing: https://github.com/epam/ketcher/issues/5942

PaulinaKacz:
#6438 – System shows phosphate as CHEM in antisense chain

1. Problem

In **Sequence** mode, terminal phosphates of an antisense chain were displayed
with the default linker symbol `@` (i.e. "like CHEM") instead of `p`. Per the
specification:

- a terminal antisense phosphate (at the **start or end** of the chain) → `p`
- two adjacent terminal phosphates → `pp` (two separate `p`, one per column)
- every other phosphate → unchanged, still `@`

2. Key design decision: rendering layer, not model/editing

The first approach changed the **node structure** in
[Chain.ts](packages/ketcher-core/src/domain/entities/monomer-chains/Chain.ts) and
the insertion logic in
[SequenceMode.ts](packages/ketcher-core/src/application/editor/modes/SequenceMode.ts),
so that a terminal phosphate became a standalone node.

That caused **regressions in flex mode and during editing**: the
`SequenceViewModel` nodes drive both rendering **and** caret navigation.
Splitting the phosphate into separate cells shifted the insertion point (e.g.
adding `C` at antisense position 2 produced a different structure than on master
— a disappearing monomer / an extra phosphate), which changed the flex-mode
layout.

Therefore the final solution is **display-only**:

- [Chain.ts](packages/ketcher-core/src/domain/entities/monomer-chains/Chain.ts)
  and
  [SequenceMode.ts](packages/ketcher-core/src/application/editor/modes/SequenceMode.ts)
  remain **identical to master** (verified with `git diff`).
- The node structure, the monomer model, and the editing logic do not change →
  **flex mode and editing behave exactly as on master**.
- Only the **rendered symbol** of already-existing cells changes. No cell is
  added or removed, so caret navigation stays the same.

Deriving `p`/`pp` from the **monomer graph (R1/R2 neighbours)** rather than from
the sequence node structure is what allows the terminal state to be read
independently of the editing layer — the monomer graph is identical to master.

3. Changed files and rationale

### `packages/ketcher-core/src/domain/helpers/monomers.ts`
The heart of the decision logic — pure functions operating on the monomer graph:

- `isAntisenseTerminalPhosphateRun(monomers)` — returns `true` when a linker node
  consists **solely** of antisense phosphates and at least one side is a chain
  end (terminality checked via `getPreviousMonomerInChain` /
  `getNextMonomerInChain`).
- `getAntisenseTerminalPhosphateCount(monomers)` — the number of `p` glyphs to
  display. It accounts for the case where the inner neighbour is a phosphate
  **absorbed into a nucleotide** (Sugar + Base + Phosphate) — such a phosphate
  has no symbol of its own, so the terminal one must represent both (`+1`).
- `isPhosphateAbsorbedInNucleotide(phosphate)` — helper detecting an absorbed
  phosphate.

The logic lives here because it is the domain layer of pure functions — easy to
unit-test and independent of both the renderer and the node structure.

### `packages/ketcher-core/.../SequenceViewModel/SequenceViewModel.ts`
Builds, once per render, a `Map<item, { symbol, xOffset }>` via
`buildAntisensePhosphateSymbols()`:

- For a terminal phosphate run it **spreads** successive `p`s across its own
  linker cell and the neighbouring "gap" cells (`BackBone`/`Empty`) that
  **already exist** in the layout — producing `p p` (one per column) instead of a
  crammed `pp` in a single cell.
- It spreads toward the "interior" of the chain (the side where the antisense
  chain continues), computed by `collectGaps` / `continuesOnSide`.
- When there is **no** free gap cell to spread into (e.g. a nucleoside is
  directly adjacent), the overflowing phosphates are shifted into the **free
  outer margin** via `xOffset = -COLUMN_WIDTH * (anchorPhosphates - 1)`, so they
  do not overlap the neighbouring symbol.

This is the right layer because `SequenceViewModel` knows the whole cell layout
(including gaps) and lets us distribute the symbols without changing the number
of nodes.

### `packages/ketcher-core/.../sequence/ChemSequenceItemRenderer.ts`
Renderer for the linker/CHEM cell. `symbolToDisplay` returns `p`/`pp` from the
view-model map when the cell is a terminal antisense phosphate — otherwise the
default `@`. It also overrides:
- `symbolXOffset` — the margin shift for crammed phosphates,
- `symbolLetterSpacing` — 8px spacing for the multi-glyph `pp`, so each `p` lands
  on the 20px column grid (Courier New 20px has an advance of ~12px, 12 + 8 = 20).

### `packages/ketcher-core/.../sequence/BackBoneSequenceItemRenderer.ts` and `EmptySequenceItemRenderer.ts`
The "gap" cells (`-` and empty) must be able to accept one `p` from an adjacent
terminal run in order to produce `p p`. They check whether their antisense node
is a target in the map and, if so, display `p` instead of `-` / empty. Without
modifying these renderers, spreading `pp` across separate columns would not be
possible.

### `packages/ketcher-core/.../sequence/BaseSequenceItemRenderer.ts`
The base class. Adds two `protected` getters `symbolXOffset` and
`symbolLetterSpacing` (default `0`) and applies them when creating the `text`
element (`.attr('x', ...)`, `.attr('letter-spacing', ...)`). This is a generic
mechanism used by the renderers to shift/space symbols — also display-only.

### `packages/ketcher-core/__tests__/domain/entities/monomerChain.test.ts` (new)
Unit tests for the monomer-graph helpers, covering:
- a single terminal phosphate → `1` (`p`),
- two adjacent terminal phosphates → `2` (`pp`),
- a terminal phosphate next to an absorbed nucleotide phosphate → `2` (`pp`),
- a **sense** phosphate → not treated as an antisense terminal run,
- a non-terminal phosphate (neighbours on both sides) → not a terminal run,
- a run containing a non-phosphate → rejected.

### Autotest snapshots (`ketcher-autotests/**/*.png`)
Updated reference images for the Sequence-mode visual tests, reflecting the new
`p`/`pp` display for terminal antisense phosphates. The spec structure does not
change — only the baselines are regenerated.

4. Why this approach is safe

- **Flex mode and editing == master** — no changes to the model or editing logic
  (verified with `git diff master`).
- **No cells added/removed** — caret and navigation unchanged.
- **Decision from the monomer graph** — independent of the node structure, hence
  resilient to editing-layer differences.
- **Unit tests** for the helpers pass (6/6), no TypeScript errors in the changed
  files.

## Check list
- [x] unit-tests written
- [ ] e2e-tests written
- [ ] documentation updated


AlexeyGirin:

PR #10387, opened 2026-07-03
#7701 -  Sync mode button does not appear just after antisense creation

### LINKED ISSUE #7701: Sync mode button does not appear just after antisense creation
**Steps to Reproduce**
1. Go to macromolecules mode
2. Go to sequence mode
3. Enter "A" nucleotide
4. Select added nucleotide
5. Right click
6. Choose "Create RNA antisense"

**Actual behavior**
Sync button does not appear JUST AFTER  antisense creation. It appears only after some user actions with canvas, like hover sequence or pressing keyboard buttons or clicking on canvas.

**Expected behavior**
Sync button appeared JUST AFTER antisense creation
Image

sabaPailodze:
## Summary

The Sync mode button now appears immediately after antisense chain creation, instead of waiting for an unrelated user interaction like a hover or keypress, and stays correct after any canvas operation.

### Sync button re-render fix

**Problem:** The `hasAntisenseChains` selector read a live getter from `drawingEntitiesManager` rather than Redux state, so `useAppSelector` had no store change to react to after antisense creation — the button stayed hidden until some other action re-rendered the tree.

**Solution:** Store the real `hasAntisenseChains` boolean in the editor store and read it from there. Recompute it on the `renderComplete` event, which fires after every canvas render, so the button's visibility has explicit state and a single trigger that covers every operation (create, delete, edit, load/paste, undo/redo).

### Files Modified

**`ketcher-macromolecules`**
- `state/common/editorSlice.ts` — add a `hasAntisenseChains: boolean` state field and a `setHasAntisenseChains` reducer; selector now reads the stored boolean instead of the live drawing-manager getter
- `EditorEvents.tsx` — recompute `hasAntisenseChains` from the drawing manager on the `renderComplete` window event, so the value stays in sync after every canvas render

**`ketcher-autotests`**
- `antisense-chains.spec.ts` — unmark test 27 (`test.fail` → `test`, drop the `@IncorrectResultBecauseOfBug` tag and bug comment) now that the Sync button appears as expected

## Check list
- [ ] unit-tests written
- [ ] e2e-tests written
- [ ] documentation updated


packages/ketcher-macromolecules/src/EditorEvents.tsx
```diff
@@ -405,5 +406,17 @@ export const EditorEvents = () => {
...
+      dispatch(notifyAntisenseChainsCreated({}));
+    };
+
+    editor?.events.createAntisenseChain.add(handleAntisenseChainCreated);
+
+    return () => {
+      editor?.events.createAntisenseChain.remove(handleAntisenseChainCreated);
+    };
+  }, [editor, dispatch]);
+
```
AlexeyGirin, line 389: Please introduce more obvious fix - to see exactly what make button state change and what triggers that

AlexeyGirin (changes requested):

PR #10395, opened 2026-07-06
#7375 - Arrows appear outside visible canvas area when reopening saved file in macro mode

### LINKED ISSUE #7375: Arrows appear outside visible canvas area when reopening saved file in macro mode
**Description:**
When reopening a saved file that contains arrows in macro mode, the arrows are rendered far outside the visible canvas area. This creates the impression that the file is empty or failed to load, as no content is initially visible. Users must manually scroll or zoom out to locate the elements.

**Steps to Reproduce**
1. Open file in Macro mode [ketcher (16).zip](https://github.com/user-attachments/files/21033993/ketcher.16.zip)

**Actual behavior**
Arrows are placed far outside the visible region of the canvas, giving a false impression of an empty or broken file.

**Expected behavior**
All elements, including arrows, should be centered or brought into the visible canvas area upon loading the file.

**Screenshots**

https://github.com/user-attachments/assets/0b940aae-1562-4376-b332-b916cf8c3e33

**Desktop:**
 - OS: Windows 11
 - Browser Chrome
 Version 137.0.7151.104 (Official Build) (64-bit)

**Ketcher version** 
[Version 3.6.0-rc.1]
**Indigo version** 
[Version 1.34.0-rc.1]

This bug was found while testing: https://github.com/epam/ketcher/issues/7125

nataliiamelezhik:
…ng file in macro mode

## Check list
- [ ] unit-tests written
- [ ] e2e-tests written
- [ ] documentation updated


AlexeyGirin:

packages/ketcher-core/src/application/editor/Editor.ts
```diff
@@ -2456,14 +2456,23 @@ export class CoreEditor {
...
+      drawingEntitiesManager.rxnArrows.size > 0 ||
+      drawingEntitiesManager.multitailArrows.size > 0;
+
+    // Monomers are only auto-zoomed in flex/snake modes to preserve sequence
+    // layout behaviour. Arrows are always zoomed into view since they have no
+    // mode-specific layout and can otherwise end up outside the visible canvas.
+    const shouldZoomMonomers =
+      this.isCurrentModeWithAutozoom() && drawingEntitiesManager.hasMonomers;
+
+    if (!shouldZoomMonomers && !hasArrows) return;
```
AlexeyGirin: Use the same mechanizm as we use for molecules - may be it would be worth to use it

AlexeyGirin (changes requested):

packages/ketcher-core/src/application/editor/Editor.ts
nataliiamelezhik: Fixed

packages/ketcher-core/src/application/editor/Editor.ts
```diff
@@ -2456,14 +2456,23 @@ export class CoreEditor {
...
+    // layout behaviour. Arrows have no mode-specific layout and can otherwise
+    // end up outside the visible canvas, so they always participate.
+    const hasEntitiesToAutozoom =
+      (this.isCurrentModeWithAutozoom() &&
+        drawingEntitiesManager.hasMonomers) ||
+      drawingEntitiesManager.rxnArrows.size > 0 ||
+      drawingEntitiesManager.multitailArrows.size > 0;
+
+    if (!hasEntitiesToAutozoom) return;
     const structureBbox = getRenderedStructuresBbox();
```
AlexeyGirin, line 2476: Check how it works for molecules - it has some mechanizm that centers them
Add arrows to there

AlexeyGirin (changes requested):

packages/ketcher-core/src/application/editor/Editor.ts
nataliiamelezhik: The "Fixed" above addressed this by folding the separate shouldZoomMonomers/hasArrows checks into one hasEntitiesToAutozoom presence check, so arrows participate unconditionally rather than through their own flag.

For molecules: the bounding box used for centering (Struct.getCoordBoundingBox()) always includes arrows as part of computing the box — there's no separate arrows check at all, they're simply one of the element types folded into the same box as everything else. Molecules and monomers/arrows don't share the same underlying code (different data models — Struct/ReStruct for molecules vs DrawingEntity/drawingEntitiesManager for macro), but the current code follows the same principle: arrows are unconditionally part of the bbox that gets zoomed/centered to, same as in molecule mode.

PR #10401, opened 2026-07-07
#10326 - Refactor: replace esbuild with vite

### LINKED ISSUE #5565: vite build error
Steps to reproduce:

1. npm create vite@latest
2. choose React, typescript
3.  install libs "ketcher-react": "^2.24.0", "ketcher-standalone": "^2.24.0",
4.  invoke libs in component, for example: 
```
/* eslint-disable @typescript-eslint/no-explicit-any */
import { useEffect, useState } from "react";
import { Ketcher, StandaloneStructServiceProvider } from "ketcher-standalone";
import { Editor } from "ketcher-react";
import "ketcher-react/dist/index.css";
import "./App.css";

const structServiceProvider = new StandaloneStructServiceProvider();


function App() {
  const [isKetcherInitialized, setIsKetcherInitialized] = useState(false); // Track Ketcher initialization
  const [ketcherInstance, setKetcherInstance] = useState<Ketcher | null>(null); // Store Ketcher instance

  useEffect(() => {
    if (ketcherInstance && isKetcherInitialized) {
      try {
        ketcherInstance.setMolecule("");
      } catch (error) {
        console.error("Failed to set SMILES:", error);
      }
    }
  }, [isKetcherInitialized]);

  return (
    <div>
      <Editor
        staticResourcesUrl={""}
        structServiceProvider={structServiceProvider as any}
        errorHandler={(message: string | Error) => {
          console.error("Error in Ketcher:", message);
      
        }}
        onInit={(ketcher: Ketcher) => {
          // Assign Ketcher instance to window for KetcherLogger to use

          /* eslint-disable-next-line @typescript-eslint/ban-ts-comment */
          // @ts-ignore
          window.ketcher = ketcher;

          setKetcherInstance(ketcher); // Set Ketcher instance in state
          setIsKetcherInitialized(true); // Mark Ketcher as initialized
          try {
            ketcher.setMolecule(''); // Set the default SMILES after initialization
          } catch (error) {
            console.error("Error initializing Ketcher with SMILES:", error);
          }
        }}
      />
    </div>
  );
}

export default App;
```
5. Try to build: npm run build

Explanation:
After adding libs:
```
    "ketcher-react": "^2.24.0",
    "ketcher-standalone": "^2.24.0",
```

and trying to build my react ts project with vite, I get error:
```
thread '<unnamed>' panicked at /Users/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/swc_common-0.39.0/src/syntax_pos.rs:702:18:
width 3 given for non-narrow character
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
thread '<unnamed>' panicked at library/core/src/panicking.rs:221:5:
panic in a function that cannot unwind
stack backtrace:
   0:        0x1329fce37 - _napi_register_module_v1
   1:        0x1329899eb - <unknown>
   2:        0x1329d19f2 - _napi_register_module_v1
   3:        0x1329fe442 - _napi_register_module_v1
   4:        0x1329ff8d9 - _napi_register_module_v1
   5:        0x1329fe992 - _napi_register_module_v1
   6:        0x1329fe929 - _napi_register_module_v1
   7:        0x1329fe91c - _napi_register_module_v1
   8:        0x132b4c56c - _napi_register_module_v1
   9:        0x132b4c5e5 - _napi_register_module_v1
  10:        0x132b4c589 - _napi_register_module_v1
  11:        0x132981dfc - <unknown>
  12:        0x100616b40 - __ZN6v8impl12_GLOBAL__N_123FunctionCallbackWrapper6InvokeERKN2v820FunctionCallbackInfoINS2_5ValueEEE
thread caused non-unwinding panic. aborting.
/bin/sh: line 1: 13941 Abort trap: 6           vite build
```

As I understand, this is trouble : `width 3 given for non-narrow character`, chatGPT says:

> SWC is a Rust-based compiler that Vite uses for code transformation and minification during the build process. The error message:
> width 3 given for non-narrow character
> indicates that SWC is encountering a character that doesn't fit within its expected character width, possibly due to Unicode characters or certain syntax in the ketcher packages.
> 
> Given that the error appears after importing ketcher-react and ketcher-standalone, it's likely that:
> 
>     SWC cannot process some code within these packages.
>     The packages might contain non-standard syntax or characters that cause SWC to crash.

my dependencies:
```
    "react": "^18.2.0",
    "vite": "^5.3.2",
```

### LINKED ISSUE #10326: Refactor: Replace esbuild with vite
1. Replace esbuild builder with vite
2. Make `monomers.ket` separated media asset (and lazy loaded)
3. Eliminate reasons cause build warnings:
```
WARN  [plugin builtin:vite-reporter]                                                                12:34:48
(!) Some chunks are larger than 500 kB after minification. 
```

```
WARN  [PLUGIN_TIMINGS] Your build spent significant time in plugins. Here is a breakdown:           12:34:48
  - vite-plugin-svgr (54%)
  - vite-plugin-commonjs (11%)
  - replace (10%)
  - ketcher-package-scoped-aliases (9%)
  - replace (7%)
See https://rolldown.rs/options/checks#plugintimings for more details.
```

IMPORTANT: Do not touch chunk with `miew` (it slightly above 500 kB)

MariiaNebesnova:

Replaces the esbuild/Rollup library build with **Vite (Rolldown)** across all published
packages. Decision record, including the alternatives weighed and the accepted losses:
`.memory-bank/adr/2026-08-28-vite-for-library-builds.md`.

### ⚠️ Release owner decisions needed before merge

This PR currently **assumes** the three answers below. Please correct any that are wrong —
they are not settled.

1. **`4.0.0` ships as a real major.** Published output no longer downlevels to ES5: Babel's
   implicit `preset-env` pass is gone, so `ketcher-macromolecules/dist/index.js` went
   **3.56 MB → 602 KB**. That is a genuine browser-floor change, not just a size win.
2. **The browser-floor change is announced** in the changelog/README.
3. **`ketcher-standalone`'s known-wrong `main`/`module` is fixed in this major** rather than
   preserved again.

### What changed

**Build system.** Every published package builds with Vite. `ketcher-standalone` keeps its six
build variants (worker/wasm seams documented in `.memory-bank/modules/ketcher-standalone.md`).

**Lazy default monomer library.** The ~3.5 MB default monomers library is no longer bundled
synchronously — it is fetched on first entry into macromolecules mode, so micromolecules-only
sessions never download it. `CoreEditor.ensureDefaultMonomersLibraryLoaded()` is the idempotent
gate; every reader must await it, because an unloaded library reads as empty rather than
throwing.

**Merge repair.** Merge commit `21e59ce41` resolved 13 conflicted manifests by taking the branch
side wholesale, silently reverting tooling that had landed on master (stylelint 17, Babel 8,
jest-dom 7, Node 24.20, indigo-ketcher 1.48). Two later "CI fix" commits compounded it by pinning
`demo`/`example-ssr` deps *backwards* to match a stale lockfile. A second master merge,
`a6bb077062`, then repeated the same resolution — its manifests are byte-identical to the branch
side and diverge from master on all 13 files. All of it is restored, and the lockfile is
regenerated from clean.

**If you merge master into this branch again, do not hand-resolve the manifests.** Diff the
result against *both* parents first. Taking the branch side wholesale is what went wrong twice,
and it is invisible in a normal PR diff.

Two consequences worth calling out:
- The branch had added an inline `"stylelint"` block to the root `package.json`, which **shadows
  `.stylelintrc.json` entirely** — `stylelint --print-config` returned `{selector-class-pattern:
  null}`, i.e. lint was effectively off repo-wide. Removed.
- The autotest Docker base had been reverted `node:24-trixie-slim` → `node:24.14.1-bullseye-slim`,
  a **different fontconfig** than the one master's screenshot baselines were captured in. Restored.

### Two pins that must not be "cleaned up"

**`ketcher-macromolecules` stays on `@babel/core@^7.29.7`.** Not a botched revert.
`@emotion/babel-plugin@11` calls `path.hoist()`, which Babel 8 removed from `@babel/traverse`.
The Vite config forces that plugin onto every file (the old Rollup config did not), so Babel 8
hard-fails the build. `ketcher-react` *is* on Babel 8, because its Vite build uses no Babel at all.

**`vite` is exact-pinned at `8.0.16` in `ketcher-standalone`.** Its `.wasm` handling depends on
Vite's internal `?no-inline` check being evaluated *ahead of* the `if (build.lib) return true`
branch in the asset plugin. A patch bump could silently re-inline ~16 MB into the two `binaryWasm`
variants.

### Monomer library guards

An audit of all **16** synchronous read sites of `monomersLibrary` / `monomersLibraryParsedJson`
found 12 already guarded — some by explicit awaits, the rest by the macro-mode entry gate in
`ketcher-react/src/Editor.tsx`, which awaits the load before switching. Two real gaps fixed here:

- **`standaloneStructService.convert()`** read the library with no guard, while
  `remoteStructService.convert()` had one. Standalone is the default mode and the mode the popup
  e2e suite runs in, so any convert before macro mode was ever opened handed the worker an empty
  library.
- **The monomer creation wizard** is reachable directly from molecules mode, so its
  symbol / HELM alias / BILN alias / RNA-preset-code uniqueness checks could run against an empty
  library and silently pass — saving a monomer whose identifier collides with a real library entry.
  Fixed by re-validating when the library resolves, and by awaiting the load at the single submit
  entry point both save paths funnel through.

### Accepted losses

CSS sourcemaps and `index.js.map`. Rationale in the ADR.

### Verification

| Check | Result |
|---|---|
| `npm ci` from clean | exit 0 |
| `npm audit --all --audit-level=critical` | passes — 5 high, 0 critical |
| `npm test` (4 packages) | exit 0 — 1439 passed, 38 snapshots, 0 failed |
| `npm run build:packages` | exit 0, all 9 builds |
| `npm run build:example` | exit 0 |
| `example-ssr` build (real-consumer check) | exit 0 — compiled, 4/4 pages prerendered |
| Manual macro-mode smoke test, **production** `example` build | pass (below) |

The smoke test was run against the served production standalone build, because a clean build is
not proof of runtime init order:
- library **not** fetched on cold load in molecules mode (0 of 67 requests)
- fetched on first macro-mode entry: 200, 3,548,976 bytes, **337** monomer cards rendered
- no console errors or unhandled rejections at any point
- from a **cold** page in molecules mode, the wizard correctly rejects `A6OH` (a real CHEM entry
  in the shipped library) with "The code must be unique amongst peptide, RNA, or CHEM monomers",
  and accepts a non-colliding code

### The screenshot regression, and why it was not a Vite bug

The dominant e2e failure was ~570 screenshots whose pixel deltas fell only on glyphs: identical
box geometry, identical typeface, text baselines ~1px high.

`example/src/index.css` does `@import 'normalize.css'`. Under the previous build that did **not**
mean the normalize.css package. react-scripts runs `postcss-normalize`
(`react-scripts/config/webpack.config.js:159`), which rewrites that import to the
browserslist-tailored `@csstools/normalize.css`. Every selector in that file is wrapped in
`:where(...)`, so the reset carries **zero specificity**; for form controls it sets only
`margin: 0`, and it has no `sub, sup` rule at all. Form controls and `<sub>` therefore rendered
with user-agent metrics — and that is what every committed baseline was captured against.

`vite build` has no such plugin, so the same import resolved to normalize.css 8.x, whose plain
type selectors *do* compete:

```css
button, input, optgroup, select, textarea {
  font-family: inherit; font-size: 100%; line-height: 1.15; margin: 0 }
sub, sup { vertical-align: baseline; font-size: 75%; line-height: 0 }
sub { bottom: -0.25em }
```

That shifted text baselines in every dialog containing a button, input or select.

Fixed by importing `@csstools/normalize.css` directly rather than reinstating the plugin: Vite
inlines `@import` before user PostCSS plugins run, so `postcss-normalize` never sees the import
to rewrite. Verified against a master build — both bundles carry the same 22 `:where()` rules,
none missing and none extra, with `optgroup` and `sub, sup` absent from both.

This also removed two compensating hacks that `327ce88bc7` ("upd styles to match snapshots")
added for the same symptom and that the revert in `eabcc2f1ef` missed:

- a global `button, input, optgroup, select, textarea { line-height: normal; font-size: revert }`
  override in both example apps
- `FormulaInput`'s `.subscript` rule, which overrode all four of normalize's `<sub>` properties
  with hard-coded values including `font-size: 11.6667px`

`FormulaInput.tsx`, `FormulaInput.module.less`, `example/src/index.css` and
`example-ssr/src/app/globals.css` are byte-identical to master again. **No baseline was
regenerated or accepted in this PR.**

Two earlier theories about this were wrong and are recorded here so they are not re-tried: the
`Dockerfile-ketcher` fontconfig base cannot affect CI screenshots (the screenshot job runs in the
pinned `mcr.microsoft.com/playwright:v1.62.1-jammy`, and that Dockerfile is referenced only by
`ketcher-autotests/docker-compose.yml`), and `packages/ketcher-react/dist/index.css` is not the
relevant artifact (both master and this branch alias it to `src/index.less`, so the example
compiles the LESS from source and never loads it).

### Requirement 3 is deferred to #11927

Issue #10326 requirement 3 asks to eliminate the build warnings. **Both still fire**, so this is
explicitly *not* done here:

| plugin | #10326 | current |
| --- | --- | --- |
| `vite-plugin-svgr` | 54% | **51%** |
| `ketcher-package-scoped-aliases` | 9% | **24%** |
| `vite-plugin-raw` | - | **11%** |

The chunk-size warning also still fires. It was previously "eliminated" by raising
`chunkSizeWarningLimit` from 500 to 1400 kB, but the offender is `ketcher-standalone` at
**15,492 kB** — 15.1 MB of one base64 literal that decodes to `\x00asm\x01\x00\x00\x00`, the
inlined Indigo WASM. `maxSize` splits only at module boundaries, so an indivisible module escapes
the cap. `vendor-miew` is 1267 kB and sits *under* the limit, warning about nothing.

The comment justifying that limit claimed it "cannot mask an unnoticed regression" and that a
firing warning means vendor-miew grew. Both are false on every build. This PR corrects the
comment only — no behaviour change — and defers the actual work to #11927.

### Still outstanding

Playwright e2e is still running on this head; no claim is made here about its result. Remaining
failures will be triaged into screenshot-only / real bugs / pre-existing flakes, with baselines
inspected individually rather than bulk-regenerated.

## Check list
- [ ] unit-tests written — see note
- [ ] e2e-tests written — suite running on this head, see above
- [ ] documentation updated — see note

**Note on unit tests.** The existing 1439 unit tests all pass, but the two monomer-library
guards in this PR are **not** pinned by new unit tests. `ketcher-standalone` has no unit test
suite at all, and no component-level test infrastructure exists for the monomer creation wizard
(its co-located tests cover pure helpers only). Both fixes were verified by the manual
production-build smoke test above instead. Standing up component-test infrastructure for the
wizard is worth doing, but it is larger than this PR.

**Note on documentation.** The ADR and the `ketcher-standalone` module docs are written. Still to
do before archiving, per `CLAUDE.md`: drop "Migration in progress" from
`.memory-bank/architecture.md`, and add a memory-bank entry for the **lazy monomer library** —
its observable behaviour (fetch on first macro-mode entry, error modal on failure) currently
exists only in commit messages.


AlexeyGirin:

example-ssr/src/app/page.tsx
```diff
@@ -1,6 +1,13 @@
...
+
+import dynamic from 'next/dynamic';
+
 import 'ketcher-react/dist/index.css';
 
-import { EditorComponent } from './editor';
+const EditorComponent = dynamic(
+  () => import('./editor').then((module) => module.EditorComponent),
+  { ssr: false },
+);
```
AlexeyGirin: Remove dynamic import as per customer request

example/src/vite/MiewReactRuntime.jsx
```diff
@@ -0,0 +1,149 @@
...
+  }, [onInit, options]);
+
+  return React.createElement(
+    'div',
+    { ref: containerRef, style: { width: '100%', height: '100%' } },
+    'Viewer',
+  );
+};
+
+export default MiewViewer;
```
AlexeyGirin: Try to load dynamicly in one string

example/vite.config.js
```diff
@@ -238,6 +514,20 @@ export default defineConfig({
...
+      '@mui/system/createStyled',
+      '@mui/system/RtlProvider',
+      '@mui/system/styled',
+      '@mui/system/styleFunctionSx',
+      '@mui/system/useTheme',
+      '@mui/system/useThemeProps',
+      '@mui/system/useThemeWithoutDefault',
+      '@mui/utils',
+      'prop-types',
+      'react-is',
```
AlexeyGirin: Try to remove since we have automatic chank configuration

example/vite.config.js
```diff
@@ -347,6 +610,10 @@ export default defineConfig({
           '../packages/ketcher-macromolecules/src/index.tsx',
         ),
       },
+      {
+        find: 'miew-react',
+        replacement: resolve(__dirname, 'src/vite/MiewReactRuntime.jsx'),
```
AlexeyGirin: Rollback as well

ketcher-autotests/tests/pages/common/CommonTopRightToolbar.ts
```diff
@@ -106,6 +106,7 @@ export const CommonTopRightToolbar = (page: Page) => {
       options: {
         disableChainLengthRuler?: boolean;
         disableAutozoom?: boolean;
+        enableFlexMode?: boolean;
```
AlexeyGirin: Please rollback that changes in that file

packages/ketcher-core/src/application/editor/actions/bond.ts
```diff
@@ -316,11 +316,7 @@ export function fromBondFlipping(restruct: ReStruct, id: number): Action {
   return action;
 }
 
-const plainBondTypes = [
-  Bond.PATTERN.TYPE.SINGLE,
-  Bond.PATTERN.TYPE.DOUBLE,
-  Bond.PATTERN.TYPE.TRIPLE,
-];
+const plainBondTypes: BondAttributes['type'][] = [1, 2, 3];
```
AlexeyGirin: Rollback if possible

packages/ketcher-core/src/application/editor/operations/sgroup/index.ts
```diff
@@ -33,8 +33,10 @@ type Data = {
   oldSgroup?: SGroup;
 };
 
+const SUP_SGROUP_TYPE = 'SUP';
+
 const SGROUP_TYPE_MAPPING: Record<string, string> = {
-  nucleotideComponent: SGroup.TYPES.SUP,
+  nucleotideComponent: SUP_SGROUP_TYPE,
```
AlexeyGirin: Please rollback that changes in that file

packages/ketcher-core/src/application/editor/operations/sgroup/index.ts
```diff
@@ -71,9 +73,9 @@ class SGroupCreate extends BaseOperation {
     let sgroup: SGroup;
 
     if (oldSgroup && oldSgroup instanceof MonomerMicromolecule) {
-      sgroup = new MonomerMicromolecule(SGroup.TYPES.SUP, oldSgroup.monomer);
+      sgroup = new MonomerMicromolecule(SUP_SGROUP_TYPE, oldSgroup.monomer);
     } else if (this.monomer) {
-      sgroup = new MonomerMicromolecule(SGroup.TYPES.SUP, this.monomer);
+      sgroup = new MonomerMicromolecule(SUP_SGROUP_TYPE, this.monomer);
```
AlexeyGirin: Please rollback that changes in that file

packages/ketcher-core/src/application/editor/editorEvents.ts
```diff
@@ -197,9 +206,9 @@ export const renderersEvents: ToolEventHandlerName[] = [
 ];
 
 const selectTools = [
-  ToolName.selectRectangle,
-  ToolName.selectLasso,
-  ToolName.selectStructure,
+  SELECT_RECTANGLE_TOOL_NAME,
+  SELECT_LASSO_TOOL_NAME,
+  SELECT_STRUCTURE_TOOL_NAME,
```
AlexeyGirin: Please rollback that changes in that file

packages/ketcher-core/src/application/editor/editorEvents.ts
```diff
@@ -328,7 +337,7 @@ export const hotkeysConfiguration = {
   hand: {
     shortcut: 'Mod+Alt+h',
     handler: (editor: CoreEditor) => {
-      editor.events.selectTool.dispatch([ToolName.hand]);
+      editor.events.selectTool.dispatch([HAND_TOOL_NAME]);
```
AlexeyGirin: Please rollback that changes in that file

packages/ketcher-core/src/application/formatters/formatProperties.ts
```diff
@@ -14,178 +14,204 @@
...
     'RDF V2000',
-    ChemicalMimeType.RDF,
+    CHEMICAL_MIME_TYPE.RDF,
     ['.rdf'],
     true,
   ),
   rdfV3000: new SupportedFormatProperties(
     'RDF V3000',
-    ChemicalMimeType.RDF,
+    CHEMICAL_MIME_TYPE.RDF,
```
AlexeyGirin: Please rollback that changes in that file

packages/ketcher-core/src/application/render/renderers/BondPathRenderer/constants.ts
```diff
@@ -19,11 +19,11 @@ export const StereoBondWidth = 6;
...
-  [BondType.DoubleAromatic]: '4 4 1 4',
-  [BondType.Any]: '6',
-  [BondType.Hydrogen]: '3',
+export const BondDashArrayMap: Partial<Record<BondType, string>> = {
+  4: '6',
+  5: '6',
+  6: '4 4 1 4',
+  7: '4 4 1 4',
+  8: '6',
+  10: '3',
```
AlexeyGirin: Please rollback that changes in that file

packages/ketcher-core/src/application/render/renderers/ChemRenderer.ts
```diff
@@ -2,13 +2,17 @@ import type { Selection } from 'd3';
...
 import { BaseMonomerRenderer } from 'application/render/renderers/BaseMonomerRenderer';
 import { MONOMER_SYMBOLS_IDS } from 'application/render/renderers/constants';
-import { KetMonomerClass } from 'application/formatters/types/ket';
 import { isMonomerSgroupWithAttachmentPoints } from '../../../utilities/monomers';
 
-const CHEM_HOVERED_ELEMENT_ID = MONOMER_SYMBOLS_IDS[KetMonomerClass.CHEM].hover;
-const CHEM_SYMBOL_ELEMENT_ID = MONOMER_SYMBOLS_IDS[KetMonomerClass.CHEM].body;
+const CHEM_SYMBOL_ELEMENTS_IDS = MONOMER_SYMBOLS_IDS.CHEM;
+if (!CHEM_SYMBOL_ELEMENTS_IDS) {
+  throw new Error('Missing monomer symbol ids for CHEM');
```
AlexeyGirin: Change to KertcherLogger

packages/ketcher-core/src/application/render/renderers/ChemRenderer.ts
```diff
@@ -2,13 +2,17 @@ import type { Selection } from 'd3';
 import type { Chem } from 'domain/entities/Chem';
 import { BaseMonomerRenderer } from 'application/render/renderers/BaseMonomerRenderer';
 import { MONOMER_SYMBOLS_IDS } from 'application/render/renderers/constants';
-import { KetMonomerClass } from 'application/formatters/types/ket';
 import { isMonomerSgroupWithAttachmentPoints } from '../../../utilities/monomers';
 
-const CHEM_HOVERED_ELEMENT_ID = MONOMER_SYMBOLS_IDS[KetMonomerClass.CHEM].hover;
-const CHEM_SYMBOL_ELEMENT_ID = MONOMER_SYMBOLS_IDS[KetMonomerClass.CHEM].body;
```
AlexeyGirin, line 9: Please rollback that changes in that file

packages/ketcher-core/src/application/render/renderers/constants.ts
```diff
@@ -8,36 +8,46 @@ export const BAD_VALENCE_LINE_OFFSET = 2;
...
     hover: '#phosphate-selection',
     body: '#phosphate',
     variant: '#phosphate-variant',
     autochainPreview: '#phosphate-autochain-preview',
   },
-  [KetMonomerClass.RNA]: {
+  RNA: {
     hover: '#nucleotide-hover',
     body: '#nucleotide',
     autochainPreview: '#nucleotide-autochain-preview',
```
AlexeyGirin: Please rollback that changes in that file

packages/ketcher-core/src/application/render/renderers/PeptideRenderer.ts
```diff
@@ -6,15 +6,17 @@ import {
...
+const PEPTIDE_SYMBOL_ELEMENTS_IDS = MONOMER_SYMBOLS_IDS.AminoAcid;
+if (!PEPTIDE_SYMBOL_ELEMENTS_IDS) {
+  throw new Error('Missing monomer symbol ids for AminoAcid');
+}
+
+const PEPTIDE_HOVERED_ELEMENT_ID = PEPTIDE_SYMBOL_ELEMENTS_IDS.hover;
+const PEPTIDE_SYMBOL_ELEMENT_ID = PEPTIDE_SYMBOL_ELEMENTS_IDS.body;
 const PEPTIDE_AUTOCHAIN_PREVIEW_ELEMENT_ID =
-  MONOMER_SYMBOLS_IDS[KetMonomerClass.AminoAcid].autochainPreview;
+  PEPTIDE_SYMBOL_ELEMENTS_IDS.autochainPreview;
```
AlexeyGirin, line 22: Please rollback that changes in that file

packages/ketcher-core/src/application/render/renderers/PhosphateRenderer.ts
```diff
@@ -5,15 +5,17 @@ import {
...
+const PHOSPHATE_SYMBOL_ELEMENTS_IDS = MONOMER_SYMBOLS_IDS.Phosphate;
+if (!PHOSPHATE_SYMBOL_ELEMENTS_IDS) {
+  throw new Error('Missing monomer symbol ids for Phosphate');
+}
+
+const PHOSPHATE_HOVERED_ELEMENT_ID = PHOSPHATE_SYMBOL_ELEMENTS_IDS.hover;
+const PHOSPHATE_SYMBOL_ELEMENT_ID = PHOSPHATE_SYMBOL_ELEMENTS_IDS.body;
 const PHOSPHATE_AUTOCHAIN_PREVIEW_ELEMENT_ID =
-  MONOMER_SYMBOLS_IDS[KetMonomerClass.Phosphate].autochainPreview;
+  PHOSPHATE_SYMBOL_ELEMENTS_IDS.autochainPreview;
```
AlexeyGirin: Please rollback that changes in that file

packages/ketcher-core/src/application/render/renderers/RNABaseRenderer.ts
```diff
@@ -5,14 +5,16 @@ import {
...
+const RNABASE_SYMBOL_ELEMENTS_IDS = MONOMER_SYMBOLS_IDS.Base;
+if (!RNABASE_SYMBOL_ELEMENTS_IDS) {
+  throw new Error('Missing monomer symbol ids for Base');
+}
+
+const RNABASE_HOVERED_ELEMENT_ID = RNABASE_SYMBOL_ELEMENTS_IDS.hover;
+const RNABASE_SYMBOL_ELEMENT_ID = RNABASE_SYMBOL_ELEMENTS_IDS.body;
 const RNABASE_AUTOCHAIN_PREVIEW_ELEMENT_ID =
-  MONOMER_SYMBOLS_IDS[KetMonomerClass.Base].autochainPreview;
+  RNABASE_SYMBOL_ELEMENTS_IDS.autochainPreview;
```
AlexeyGirin, line 21: Please rollback that changes in that file

packages/ketcher-core/src/application/render/renderers/SugarRenderer.ts
```diff
@@ -5,14 +5,17 @@ import {
...
+const SUGAR_SYMBOL_ELEMENTS_IDS = MONOMER_SYMBOLS_IDS.Sugar;
+if (!SUGAR_SYMBOL_ELEMENTS_IDS) {
+  throw new Error('Missing monomer symbol ids for Sugar');
+}
+
+const SUGAR_HOVERED_ELEMENT_ID = SUGAR_SYMBOL_ELEMENTS_IDS.hover;
+const SUGAR_SYMBOL_ELEMENT_ID = SUGAR_SYMBOL_ELEMENTS_IDS.body;
 const SUGAR_AUTOCHAIN_PREVIEW_ELEMENT_ID =
-  MONOMER_SYMBOLS_IDS[KetMonomerClass.Sugar].autochainPreview;
+  SUGAR_SYMBOL_ELEMENTS_IDS.autochainPreview;
```
AlexeyGirin, line 16: Please rollback that changes in that file

packages/ketcher-core/src/application/render/renderers/UnsplitNucleotideRenderer.ts
```diff
@@ -6,14 +6,16 @@ import {
...
+const NUCLEOTIDE_SYMBOL_ELEMENTS_IDS = MONOMER_SYMBOLS_IDS.RNA;
+if (!NUCLEOTIDE_SYMBOL_ELEMENTS_IDS) {
+  throw new Error('Missing monomer symbol ids for RNA');
+}
+
+const NUCLEOTIDE_HOVERED_ELEMENT_ID = NUCLEOTIDE_SYMBOL_ELEMENTS_IDS.hover;
+const NUCLEOTIDE_SYMBOL_ELEMENT_ID = NUCLEOTIDE_SYMBOL_ELEMENTS_IDS.body;
 const NUCLEOTIDE_AUTOCHAIN_PREVIEW_ELEMENT_ID =
-  MONOMER_SYMBOLS_IDS[KetMonomerClass.RNA].autochainPreview;
+  NUCLEOTIDE_SYMBOL_ELEMENTS_IDS.autochainPreview;
```
AlexeyGirin, line 21: Please rollback that changes in that file

packages/ketcher-core/src/application/render/restruct/resgroup.ts
```diff
@@ -45,10 +45,10 @@ interface SGroupdrawBracketsOptions {
...
 }
 
-export const SUPERATOM_CLASS_TEXT = {
-  [SUPERATOM_CLASS.BASE]: 'Base',
-  [SUPERATOM_CLASS.SUGAR]: 'Sugar',
-  [SUPERATOM_CLASS.PHOSPHATE]: 'Phosphate',
+export const SUPERATOM_CLASS_TEXT: Record<SUPERATOM_CLASS, string> = {
+  BASE: 'Base',
+  SUGAR: 'Sugar',
+  PHOSPHATE: 'Phosphate',
```
AlexeyGirin: Please rollback that changes in that file

packages/ketcher-core/src/application/render/raphael-ext.ts
```diff
@@ -17,10 +17,7 @@
 // Single entry point to Raphaël library
 
 import { Vec2 } from 'domain/entities/vec2';
-
-// eslint-disable-next-line @typescript-eslint/no-explicit-any
-const raphaelModule: any =
-  typeof window !== 'undefined' ? require('raphael') : undefined;
+import raphaelModule from 'raphael';
```
AlexeyGirin, line 29: Please rollback that changes in that file

packages/ketcher-core/src/application/render/raphael-ext.ts
```diff
@@ -32,7 +29,8 @@ function resolveRaphael(): any {
 
   return typeof raphaelModule === 'function'
     ? raphaelModule
-    : raphaelModule.default;
+    : // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      (raphaelModule as any).default;
 }
```
AlexeyGirin, line 40: Please rollback that changes in that file

packages/ketcher-core/src/application/ketcher.ts
```diff
@@ -84,10 +84,13 @@ const allowedApiSettings = {
...
+const MONOMER_LIBRARY_MIME_TYPE =
+  'chemical/x-monomer-library' as ChemicalMimeType.MonomerLibrary;
+
 const MONOMER_LIBRARY_FORMAT_OPTIONS = {
-  inputFormat: ChemicalMimeType.MonomerLibrary,
-  outputFormat: ChemicalMimeType.MonomerLibrary,
-  outputContentType: ChemicalMimeType.MonomerLibrary,
+  inputFormat: MONOMER_LIBRARY_MIME_TYPE,
+  outputFormat: MONOMER_LIBRARY_MIME_TYPE,
+  outputContentType: MONOMER_LIBRARY_MIME_TYPE,
```
AlexeyGirin: Please rollback that changes in that file

packages/ketcher-core/src/domain/helpers/monomers.ts
```diff
@@ -432,7 +441,7 @@ export {
 
 export function isRnaBaseApplicableForAntisense(monomer?: BaseMonomer) {
   return (
-    isMonomerOfClass(monomer, KetMonomerClass.RNA) ||
+    isMonomerOfClass(monomer, KET_MONOMER_CLASS.RNA) ||
     (isRnaBaseOrAmbiguousRnaBase(monomer) &&
       Boolean(getSugarFromRnaBase(monomer)))
```
AlexeyGirin, line 531: Please rollback that changes in that file

packages/ketcher-core/src/domain/serializers/mol/common.ts
```diff
@@ -85,7 +85,7 @@ function parseRxn(
 /* Prepare For Saving */
 const prepareForSaving: Record<string, (sgroup: SGroup, mol: Struct) => void> =
   {
-    MUL: SGroup.prepareMulForSaving,
+    MUL: (sgroup, mol) => SGroup.prepareMulForSaving(sgroup, mol),
```
AlexeyGirin: Please rollback that change

packages/ketcher-core/src/domain/serializers/mol/utils.js
```diff
@@ -74,34 +73,34 @@ function partitionLineFixed(
...
+    2: 4,
+    3: 6,
   },
   bondTopologyMap: {
-    0: Bond.PATTERN.TOPOLOGY.EITHER,
-    1: Bond.PATTERN.TOPOLOGY.RING,
-    2: Bond.PATTERN.TOPOLOGY.CHAIN,
+    0: 0,
+    1: 1,
+    2: 2,
```
AlexeyGirin: Please rollback that changes in that file

packages/ketcher-macromolecules/src/Editor.tsx
```diff
@@ -239,6 +239,14 @@ function Editor({
     };
   }, [dispatch]);
 
+  useEffect(() => {
+    if (!editor) {
+      return;
+    }
+
+    editor.ensureDefaultMonomersLibraryLoaded();
```
AlexeyGirin: We need library in Micro as well, so rollback this

AlexeyGirin (changes requested):

AlexeyGirin:

AlexeyGirin:
@copilot please make changed based on review comments

packages/ketcher-core/src/application/render/renderers/AmbiguousMonomerRenderer.ts
```diff
@@ -20,15 +20,18 @@ type PreviewAttachmentPointParams = {
...
     variant?: string;
+    autochainPreview: string;
   };
 
   constructor(public monomer: AmbiguousMonomer, scale?: number) {
     const monomerClass = AmbiguousMonomer.getMonomerClass(monomer.monomers);
     const monomerSymbolElementsIds = MONOMER_SYMBOLS_IDS[monomerClass];
+    if (!monomerSymbolElementsIds) {
+      throw new Error(`Missing monomer symbol ids for ${monomerClass}`);
+    }
```
AlexeyGirin: @copilot Use `KetcherLogger` and `return` instead of `Error`

AlexeyGirin: @copilot please make fix based on provided comment

AlexeyGirin (changes requested):

AlexeyGirin:

AlexeyGirin:

AlexeyGirin:

nickkurkan:
## Plan for the remaining migration

Rationale and trade-offs are recorded in [`.memory-bank/adr/2026-08-28-vite-for-library-builds.md`](../blob/10326-refactor-replace-esbuild-with-vite-2/.memory-bank/adr/2026-08-28-vite-for-library-builds.md), added in this branch. Short version below so it's reviewable without opening the ADR.

**Scope.** `demo` + the four library packages. `example-ssr` stays on Next.js (it's an SSR app with its own bundler, not a Vite target). `ketcher-autotests` has no bundler.

**Hard constraint: the published contract is frozen.** Same file names, formats, and `main`/`module`/`types`/`exports` entries — including `ketcher-standalone`'s `main`/`module` fields, which are known to be stale (they point at `dist/index.js` while the build emits `dist/main.js`). Its `exports` map is correct, so modern resolvers are unaffected; "fixing" them could change resolution for consumers currently working by accident. Left for a deliberate version bump.

### Sequence

- [x] **Prep** (`caf054d5c7`) — no bundler changes. Shared build constants extracted to `build-config/`; `PolymerBond` imported from `ketcher-core`'s public entry instead of through its `dist/` tree; `ketcher-macromolecules` now built *before* `ketcher-react` (react's build reads macromolecules' built CSS, but `build:packages` built react first); `shx` hoisted to root; `Dockerfile-ketcher` bumped to node 24.14.1 to match `.nvmrc`/`engines`/CI.
- [ ] `demo` — CRA → Vite, minimal-change (keep `REACT_APP_` env prefix via `envPrefix`, keep `build/` as `outDir`)
- [ ] `ketcher-core`
- [ ] `ketcher-macromolecules`
- [ ] `ketcher-react`
- [ ] `ketcher-standalone` — last, and droppable (see below)
- [ ] `example` cleanup — drop `react-scripts` / `react-app-rewired` / `config/webpack.config.js`

### Notes for reviewers

**`preserveModules` was verified before committing to this.** Vite 8 is Rolldown-only — there is no config to fall back to Rollup — and `ketcher-core` publishes per-file unbundled output. A spike built core with `build.lib` + `rolldownOptions.output.preserveModules` and diffed against the Rollup baseline: ~550 modules map 1:1 by path in both `es` and `cjs`, imports and banner intact. It works.

Three differences from Rollup that the configs must set explicitly rather than inherit:
| Difference | Handling |
| --- | --- |
| Rolldown minifies library output by default | `build.minify: false` — publishing minified library code breaks downstream stack traces |
| `events` resolved to a local polyfill and bundled (Rollup left it external) | explicit Node-builtin `external` list |
| `rollup-plugin-string` has no equivalent, `.ket` imports fail to parse | small inline raw-text transform in `build-config/` |

**`ketcher-standalone` may stay on Rollup permanently.** It runs six builds to emit six output dirs behind four `exports` subpaths, and depends on `rollup-plugin-web-worker-loader` and `@surma/rollup-plugin-off-main-thread`. Agreed abort criterion: **if the web-worker output can't be reproduced, it stays on Rollup.** Two toolchains beats degrading the package that standalone mode's WASM loading depends on.

**Acceptance per package:** build → `dist/` diffed against the Rollup baseline → `example-ssr` builds and renders → Playwright green. `example-ssr` matters here because it's the only target that resolves the packages from `dist` through their `exports` maps — `example` aliases them to source, so it never exercises published output.

Babel is dropped along with the `@babel/runtime` dependency, except in `ketcher-macromolecules`, which keeps Emotion's plugin for stable class names (the suite is screenshot-based).

nickkurkan:
## Migration complete — status update

All planned work has landed. Updated checklist against the [earlier plan](#issuecomment-5452933401):

- [x] **Prep** (`caf054d5c7`) — shared constants to `build-config/`, `PolymerBond` public-entry import, build order fixed, `shx` hoisted, Dockerfile → node 24.14.1
- [x] **`demo`** (`e7e14222a5`) — CRA → Vite, minimal-change
- [x] **`ketcher-core`** (`fec1d845e5`) — per-file `preserveModules` output preserved
- [x] **`ketcher-macromolecules`** (`38bf23024c`)
- [x] **`ketcher-react`** (`908791e094`)
- [ ] **`ketcher-standalone`** — **stays on Rollup 2**, see below
- [x] **`example` cleanup** (`a9b6889d24`) — webpack config and 8 dead deps removed
- [x] Version bump to `4.0.0-rc.1` (`60aa588644`) — **needs release-owner sign-off**, see below
- [x] Code-review fixes (`4e89950e33`), `example-ssr` build fix (`469ef0fabb`)

### Two things reviewers should look at first

**1. `ketcher-standalone` stays on Rollup 2 — deliberately.** The migration was attempted and the pre-agreed abort criterion fired. `rollup-plugin-web-worker-loader` bundles the Indigo worker by running a *nested Rollup build inside its own `load` hook*, seeded with the outer build's entire plugin list. Under Vite 8 the outer list holds Rolldown-oriented plugins, and re-invoking them through real Rollup 2 collides the two module-resolution contracts — `Error: Missing field 'moduleType'`. Getting past it needs either a fork of the plugin or dropping the `web-worker:` protocol for Vite's `?worker&inline` (a source change to suit the bundler). Neither was taken. **The remaining Rollup config is intentional, not unfinished.**

**2. Published packages now ship modern syntax — this is the breaking change.** Dropping Babel dropped `@babel/preset-env`'s ES5 downleveling with it. `ketcher-core`'s `Editor.modern.js` went from 18 downlevel helpers and 1 raw `class` to 0 and 6; `ketcher-macromolecules`' `dist/index.js` fell from 3.56 MB to 602 KB. Every build and test in this repo stays green through that change — only a consumer on an older browser breaks, at runtime. Hence the major bump to `4.0.0-rc.1`, which **needs the release owner's agreement** since 3.19 appears to be in flight.

### Verification status

| Gate | Status |
| --- | --- |
| Builds (all packages + both apps) | green |
| `dist/` diffed against the Rollup baseline, per package | every difference explained |
| Type declarations | all present; `ketcher-macromolecules` now resolves 24 aliases the old build shipped unresolved |
| `tsc --noEmit`, all workspaces | green |
| Unit tests | green (pre-existing macromolecules failures unchanged, verified against baseline) |
| `example-ssr` builds and genuinely prerenders | green |
| **Playwright** | **not run — see below** |

**The e2e suite has not been run against this branch.** That is the one gate still open, and it matters most here: `ketcher-react`'s CSS-Modules class-name hashes changed (self-consistent between JS and CSS, and the few plain-class selectors the suite uses are unaffected — but that is reasoning, not evidence).

### Things noted and deliberately not fixed

- `ketcher-standalone`'s `main`/`module` fields remain wrong (they point at `dist/index.js`; the build emits `dist/main.js`). Its `exports` map is correct so modern resolvers are unaffected. Left as-is under the frozen-contract rule; worth a deliberate fix in a version bump.
- `ketcher-macromolecules` no longer emits `dist/index.css.map` — `cssCodeSplit: false` is required for single-file CSS and Vite 8.0.16 has no sourcemap path in that mode. Affects devtools CSS debugging only.
- `example`'s `build:*:analyze` scripts are now inert — bundle visualisation came from the deleted webpack config.

Full rationale, including the `preserveModules` spike that de-risked `ketcher-core` before any of this was committed, is in `.memory-bank/adr/2026-08-28-vite-for-library-builds.md`.

nickkurkan:
## Verification results — migration complete

Local verification of the Vite migration, re-run end-to-end after `ketcher-standalone` landed. Both trees were built back-to-back and serially on one machine so the numbers are comparable. Baseline is `caf054d5c` (the prep commit on this branch, before any migration work); HEAD is `4b7efd52a4`.

**Update from the previous version of this comment:** `ketcher-standalone` is now on Vite as well — no Rollup config remains in the repo. It had been acting as the *control* for these measurements, so that role is gone; see [Measurement caveat](#measurement-caveat) below.

### Build time (full package build, incl. `tsc` declaration emit)

| target | before | after | delta |
| --- | ---: | ---: | ---: |
| ketcher-core | 10.8s | 5.2s | **−52%** |
| ketcher-standalone (6 variants) | 20.2s | 4.0s | **−80%** |
| ketcher-macromolecules | 32.4s | 4.5s | **−86%** |
| ketcher-react | 10.8s | 3.4s | **−69%** |
| demo (CRA → Vite) | 18.7s | 1.2s | **−94%** |
| **total** | **92.9s** | **18.3s** | **−80%** |

`npm run build:packages` (which parallelises standalone + macromolecules) now completes in **~14s**.

### Published size

| package | total before | total after | JS before | JS after | files |
| --- | ---: | ---: | ---: | ---: | ---: |
| ketcher-core | 24.37 MB | 29.79 MB **(+22%)** | 16.01 MB | 13.11 MB (−18%) | 2540 → 2612 |
| ketcher-standalone | 67.19 MB | 69.49 MB (+3%) | 66.51 MB | 50.23 MB (−24%) | 84 → 84 |
| ketcher-macromolecules | 9.95 MB | 3.32 MB (−67%) | 6.77 MB | 1.10 MB (−84%) | 324 → 287 |
| ketcher-react | 30.20 MB | 15.52 MB (−49%) | 11.83 MB | 5.51 MB (−53%) | 1180 → 538 |
| demo | 30.00 MB | 23.95 MB (−20%) | 29.82 MB | 23.80 MB (−20%) | 15 → 11 |

### Three size numbers that need explaining

**macromolecules −84% JS is a behavioural change, not an optimisation.** The Rollup baseline bundled everything in — `index.modern.js` was 3.5 MB with **zero** external imports (MUI, emotion, d3, lodash, react-redux all inlined). The Vite build externalises them: 557 KB with **24** externals.

That changes what consumers must resolve, so all 16 root packages were checked against the manifest — every one is declared in `dependencies` or `peerDependencies`. Safe, and arguably a real improvement: consumers no longer receive a second copy of MUI and React welded into the package. `ketcher-react`'s −49% is largely downstream of the same effect, since it bundles the macromolecules chunk.

**standalone's +3% total hides a −24% JS drop, because the baseline was shipping a broken package.** At `caf054d5c` the two `binaryWasm` variants shipped **no `.wasm` file at all** — the `rollup-plugin-copy` glob pointed at the workspace root while `indigo-ketcher` installs into the package, so it matched nothing, silently, for three weeks. Both variants were unloadable. Fixed in `964c0f36` (details in that commit; it is a master bug, not a migration one).

So the comparison is JS 66.51 → 50.23 MB (−24%, chiefly from dropping Babel's downlevelling) *plus* 18.45 MB of `.wasm` that should always have been there. Vite ships 18.45 MB rather than the 36.89 MB the fixed Rollup config would have, because the glob copied *both* binaries into *both* fetch directories while Vite emits only the one each variant actually references.

**ketcher-core is the one target that genuinely grew, +22%, and it is entirely sourcemaps** — 7.64 MB → 15.96 MB (+109%) while its actual JS *shrank* 18%.

> **Correction.** An earlier version of this comment attributed this to Rolldown "embedding sources more redundantly" and called it ~8 MB per release "for no functional gain", flagged as a follow-up. That was wrong, and the follow-up should not be actioned. Having now diffed the maps on both sides, the growth is Rolldown *fixing* a defect.

Both builds emit exactly **978 maps**. The difference is what is inside them:

| | Rollup 2 (`caf054d5c`) | Vite 8 (`4b7efd52a4`) |
| --- | ---: | ---: |
| maps with `sourcesContent` | **8 of 978** | **978 of 978** |
| `sourcesContent` bytes | 0.38 MB | 12.44 MB |
| `mappings` bytes | 6.21 MB | 2.33 MB |
| total | 7.64 MB | 15.96 MB |

The Rollup build wrote `"sourcesContent": [null]` for 970 of 978 maps — the map named its source (`../../src/domain/AttachmentPoint.ts`) but did not embed it. `package.json` has `files: ["dist"]`, so **`src/` is not in the published tarball** (verified against a published `ketcher-core` tarball: `dist`, `package.json`, `README.md`, and nothing else). Those maps therefore resolved to a path that does not exist for any consumer. They worked only inside this monorepo, where `src/` happens to sit next to `dist/`. The 8 exceptions are all plain `.js` files, which skipped the TS transform that was dropping the content.

Rolldown embeds the source in every map, so they now work from any consumer's `node_modules`. The +8 MB buys sourcemaps that actually resolve — which is the entire point of shipping them.

Two second-order notes: `mappings` *fell* 62%, because Babel's downlevelling produced far denser position data than the modern output needs; and the `sourcesContent` total (12.44 MB) is roughly 2× `src/` on disk (6.67 MB) because each source is embedded once per output format, in `.js.map` and again in `.modern.js.map` — inherent to publishing dual CJS/ESM with `preserveModules`.

If tarball size ever does need to come down here, the lever is `sourcemapExcludeSources` — but that would re-break consumer debugging, so it is a deliberate trade, not a cleanup.

<a name="measurement-caveat"></a>
### Measurement caveat

The earlier version of this comment used `ketcher-standalone` as a control, since it was staying on Rollup: its −5% was the noise floor. That control no longer exists. The BASE row above was re-measured from scratch for this run rather than reused, and BASE/HEAD ran back-to-back, but there is now no in-run estimate of variance. The build-time deltas are large enough (−52% to −94%) that this does not change any conclusion; treat the individual percentages as ±5% rather than exact.

### Published contract diff

Built both trees and compared `dist/` file inventories:

| package | runtime files missing at HEAD | declarations missing |
| --- | ---: | ---: |
| ketcher-core | 0 (+72 added) | 0 |
| ketcher-react | 0 | 0 |
| ketcher-macromolecules | 0 | 0 |
| ketcher-standalone | 0 | 0 |

Nothing in the published contract disappeared. Two ketcher-react files that first appeared missing were content-hashed chunks (2 removed / 2 added, new hash). Two declaration groups did go away, both verified harmless: `.test.d.ts` (test declarations that were previously being published) and the duplicate `dist/cjs/**/*.d.ts` tree, which no entry point references.

For `ketcher-standalone`, all six variants load in Node and export the same five symbols as the Rollup baseline. Its inventory differs in exactly two ways, both intended:

- the six `index.js.map` files are gone — the Rollup ones had empty `sources` and `mappings`, and the new `index.js` carries no `sourceMappingURL`;
- the two fetch variants gained `assets/` entries (worker chunk, its map, and the `.wasm`), which nothing in the `exports` map or any consumer references.

One more finding worth recording: Vite emits a sourcemap for an inlined worker even though the worker is embedded as a Blob, so the map can never be fetched — the `//# sourceMappingURL` resolves against a `blob:` origin. That was **50 MB, 42% of the package**, for four maps whose `.js` is not even written to disk. Dropped in `4b7efd52a4`; the sourcemap set now matches the Rollup baseline exactly, plus the two fetch variants' real ones.

Caveat on method: an inventory check cannot see content-level changes such as the externalisation described above. It is a necessary check, not a sufficient one.

### E2E status — incomplete

Ran the suite locally in Docker (Colima VM, Linux, to match the `-chromium-linux.png` baselines).

- Smoke run: **22/22 passed in 22.6s**. This confirms the editor boots and the dynamic `monomers.ket` import still resolves — the specific path the `build.lib` / `modulePreload: false` handling exists to protect.
- Full suite: started, **stopped at ~530 / 20,150 (2.6%) with 0 failures**, at a measured ETA of ~6h.

Two reasons the local full run was not worth completing as configured:

1. The container was not CPU-bound (6 workers, ~1.2 cores drawn) — tests are wait-bound, so worker count is mismatched to the workload.
2. `.env` sets `CI_ENVIRONMENT=true`, which enables `retries: 2`. Every failure runs three times, and this branch carries ~3,945 **pre-existing** screenshot mismatches from before any migration work — so the expensive part of the suite had not even been reached.

Also note the local VM is aarch64 while CI is amd64, and Playwright's `{platform}` token is only `linux` — it does not encode arch. Screenshot comparisons locally are therefore not authoritative. **CI still owns the e2e verdict.**

### Other local checks (all green at `4b7efd52a4`)

`build:packages` · `example-ssr` build + prerender · `example` standalone build · `demo` build · `test:types` across all workspaces · unit tests for core, react, standalone.

`ketcher-macromolecules` has 11 Emotion snapshot failures. These reproduce unchanged on `964c0f36` and predate this work.

### Prior CI bisect, for the record

| commit | screenshot mismatches | `waitForFunction` hangs | shard time |
| --- | ---: | ---: | ---: |
| `caf054d5c` (pre-migration) | 3,945 | 15 | 8–17 min |
| `908791e094` (3 libs migrated) | 3,939 | 15 | 8–17 min |
| `469ef0fabb` | 0 | 1,812 | up to 2h, 2 shards cancelled |

This exonerated the library migrations and isolated the regression to the `events` dependency, fixed in `53ad1d736d`. The ~3,945 screenshot mismatches are pre-existing branch drift, unrelated to this work and not addressed here.

### Still open

- Full CI e2e green on `4b7efd52a4` — the remaining gate.
- The `4.0.0-rc.1` bump needs release-owner agreement (3.19 appears to be in flight).
- Pre-existing screenshot drift on this branch.

nickkurkan:
## Open questions

Everything still unresolved on this PR, grouped by who has to decide. Nothing here is a task I can close on my own — each needs a call from a maintainer or release owner, or a CI run.

### 1. Blocking — needs a decision

**1.1 · The `4.0.0-rc.1` version bump.** All four packages are set to `4.0.0-rc.1` on this branch; `master` is on `3.19.0-rc.1`. The major is not cosmetic — see 1.2. **Needs the release owner's agreement**, and 3.19 appears to be in flight, so the two need sequencing.

**1.2 · Shipping modern syntax is a breaking change.** The Rollup builds ran `@babel/preset-env` with no explicit target, downlevelling toward ES5. Dropping Babel drops that: `ketcher-core`'s `Editor.modern.js` went from 18 downlevel helpers and 1 raw `class` to 0 and 6.

This is the failure mode the frozen contract exists to catch — the build stays green, tests pass, `example-ssr` renders, and only a consumer on an older browser breaks, at runtime, silently. The ADR handles it as a major bump rather than hiding it behind build config. **Someone needs to confirm that is acceptable for this library's consumers**, or state a browser floor so a `build.target` can be set instead.

**1.3 · Merge conflicts with `master`.** The PR is currently `CONFLICTING`; the branch is 17 commits behind. The overlap is narrow but lands exactly where this work sits:

| file | `master` changed | this branch changed |
| --- | --- | --- |
| all four `.babelrc` | dropped `allowDeclareFields: true` from the TS preset | removed the `@babel/plugin-transform-runtime` plugin; deleted standalone's entirely |
| all four `package.json` | version → `3.19.0-rc.1` | version → `4.0.0-rc.1`, plus dependency removals |
| `package-lock.json` | — | — |

The `.babelrc` conflict needs care: **both sides changed the same two files for unrelated reasons.** `master` removing `allowDeclareFields` is a semantic change to `declare` class fields; this branch's edit is about dropping a Babel plugin that no longer has a declared dependency. A resolution that just takes one side will silently lose the other.

**1.4 · Full CI e2e is the remaining gate, and the last three commits are not pushed yet.** `964c0f36`, `7dd32b86` and `4b7efd52` are local only, so CI has not seen the `ketcher-standalone` migration at all. Everything below about local verification is a proxy, not a substitute.

### 2. Needs a product / consumer call

**2.1 · `ketcher-macromolecules` now externalises 24 dependencies it previously bundled.** The Rollup baseline inlined MUI, emotion, d3, lodash and react-redux — `index.modern.js` was 3.5 MB with **zero** external imports. The Vite build emits 557 KB with 24. All 16 root packages were verified declared in `dependencies`/`peerDependencies`, so resolution is sound, and consumers no longer receive a second copy of MUI and React welded in.

But it does change what a consumer's bundler must resolve, and anyone relying on the package being self-contained will notice. **Is that acceptable, or should these be re-bundled?**

**2.2 · Known-wrong package metadata is preserved deliberately, and it is worse than "unused".** `ketcher-standalone`'s `module` field points at `dist/index.modern.js`, which **does not exist**; `main` points at `dist/index.js`, which exists but is the empty stub entry and exports nothing. Its `exports` map is correct, so modern resolvers are fine and only tooling that ignores `exports` is affected — which is why it was left alone rather than "fixed" mid-migration and changing resolution for consumers who currently work by accident. **Worth fixing in the same major bump, given one is already on the table?**

### 3. Known, accepted, not addressed here

**3.1 · `ketcher-core` sourcemaps are +109%.** Now understood and *correct* — Rollup was writing `sourcesContent: null` for 970 of 978 maps, pointing at a `src/` that is not in the published tarball. See the correction in the verification comment. Recorded here only so nobody re-opens it as a size regression.

**3.2 · CSS sourcemaps are lost for `ketcher-macromolecules`.** Producing a single `dist/index.css` requires `build.cssCodeSplit: false`, and in that path Vite 8.0.16 emits the extracted CSS via a plain `emitFile` with no sourcemap generation — no configuration produces `index.css.map`. Affects devtools CSS debugging only; not part of the published contract.

**3.3 · ~3,945 pre-existing screenshot mismatches.** Present at `caf054d5c`, before any migration work; the CI bisect in the verification comment confirms they are branch drift, not caused by this PR. **They will make the e2e run in 1.4 hard to read** — a green/red verdict needs to be taken against the pre-migration baseline, not against zero.

**3.4 · Local e2e can never be authoritative here.** The local VM is aarch64, CI is amd64, and Playwright's `{platform}` token only encodes `linux` — not arch. Local screenshot comparisons are indicative at best.

**3.5 · `ketcher-macromolecules` has 11 Emotion snapshot failures.** Reproduce unchanged on `964c0f36`; predate this work. Not investigated.

### 4. Smaller things I decided unilaterally — flagging in case you disagree

- **Renamed two worker shims** (`useWasmLoader` → `useViteInlineWorker`, `useOffMainThreadPlugin` → `useNativeWorkerUrl`). Both were named after Rollup plugins that no longer exist. Rationale in `7dd32b86`; happy to revert if you would rather keep the diff minimal.
- **Dropped `@babel/plugin-transform-runtime` from `ketcher-core` and `ketcher-react`'s `.babelrc`** rather than re-declaring the dependency. Removing `@babel/*` from `ketcher-standalone` exposed that both were resolving it through hoisting without declaring it. Their `.babelrc` is jest-only now that builds use Vite, where transform-runtime does nothing useful. See 1.3 — this is the file `master` also touched.
- **Each `binaryWasm` variant now ships only the `.wasm` it uses**, not both. The old copy glob shipped both into both directories (36.89 MB vs 18.45 MB). This was never load-bearing — the glob matched zero files for three weeks — but it is a change from stated intent.

AlexeyGirin:

nickkurkan:
## Review of `cbc05aa1cb`: not ready to merge yet

I compared what this branch publishes (`npm run build:packages` + `npm pack`) with npm `3.18.0` (current latest) and `3.20.0-rc.1` (last build from master). The build change itself is well documented, but three things block the merge:

- two problems in the published packages that no test catches,
- CI is red, and the failures do not happen on master,
- the reason given for 4.0.0 is no longer true.

### Blockers (in the published packages, not caught by any test)

**B1. Wrong CSS colors and a broken grid in `ketcher-react/dist/index.css` and `ketcher-macromolecules/dist/index.css`.**
Vite uses the package-local `less@3.13.1` (both packages declare `"less": "^3.12.2"`). Rollup used the root `less@4.6.4`. Less 3 treats `/` as division:
- `rgb(94 94 94 / 80%)` becomes `#5e5e03`. The overlay turns olive and is no longer transparent.
- `grid-column: 2 / 3` becomes `.666667`, which is invalid CSS (`Recognize.module.less`).

This affects 15 color expressions plus the grid rule. `example` compiles the Less with root Less 4, so the e2e suite cannot see it. Only npm consumers get the broken CSS.
Fix: move both packages to Less 4, or set `css.preprocessorOptions.less.math: 'parens-division'`.

**B2. The `binaryWasm` / `binaryWasmNoRender` builds do not work in a consumer's build.**
The output is `new Worker(new URL("" + new URL("assets/indigoWorker-….js", import.meta.url).href, …))`. A consumer's bundler does not recognize this as a worker, so it copies the file as-is and never emits the `.wasm`, which then returns 404 at runtime. I confirmed this with a real Vite consumer build; webpack was not tested. These builds were already broken on master in `3.20.0-rc.1`, but `3.18.0` worked. `example` only uses the default inline build, so nothing covers this. I suggest adding a CI check that builds a consumer against the packed `dist`.

### Probable regression (found by reading the code, not reproduced)

`Ketcher.updateMonomersLibrary()` and `replaceMonomersLibrary()` (`ketcher.ts:912`, `:946`) do not wait for the default monomer library. If one of them runs before the default library has loaded, the later default load overwrites the consumer's library (`Editor.ts` `setMonomersLibrary`). A replace is lost completely. An update survives only if it was saved with `shouldPersist`. Both methods should await `ensureDefaultMonomersLibraryLoaded()`, with a unit test for the ordering. The two existing guards (standalone `convert()` and the monomer creation wizard) also have no unit tests.

### CI

- **Build job.** CI tests the PR merged with master. Master added `reportUnusedDisableDirectives: 'error'`, and this branch upgrades `globals` to 16.5, which defines `ClipboardItem`. Together they turn two `eslint-disable no-undef` comments into errors: `ketcher-core/src/utilities/clipboardUtils.ts` and `ketcher-react/src/script/ui/action/copyImageToClipboard.js`. Deleting both comments fixes it.
- **Playwright.** About 1,404 tests fail in 6 of the 11 parts. Most are small screenshot differences, but about 265 are assertion failures and about 100 are timeouts or click failures. Master's latest full run (Sep 22) passed, so these failures are new. The "~3,945 existing mismatches" in the PR body was measured against a branch commit, not master.

### The version question

The ADR and PR body say 4.0.0 is needed because the output is no longer compiled down to ES5. But master's `3.20.0-rc.1` already ships modern syntax (native `class`, `?.`, `async`), because Babel 8 has different default targets, and this branch does not change the syntax level compared with master. A major version is still justified, by the breaking changes below. The ADR and changelog should give those as the reason instead.

Other points where the docs and the code disagree:
- **`ketcher-standalone` `main`/`module`.** The ADR says they are left alone on purpose; the PR body says they are fixed in this major. They are not fixed: `main` points to an empty file, and `module` points to a file that is not in the package.
- **"Micro-only sessions never download the library".** `getKet`, `setMolecule`, `addFragment` and both `convert()` calls now wait for the library, so any API user loads and parses about 3.5 MB without opening macromolecules mode. Also, only `example` fetches it as a separate asset; for npm consumers it is a code-split JS chunk.

### Breaking changes compared with 3.18.0 that should go in the changelog

| Change | Who it affects |
| --- | --- |
| The monomer library is empty until it loads. `CoreEditor` no longer loads it in its constructor, so a synchronous read of `monomersLibrary` / `monomersLibraryParsedJson` returns empty, and `updateMonomersLibrary` fires later. | Anyone who reads the library directly |
| `binaryWasm*` asset layout: files are in `assets/` with hashed names, and each folder has one `.wasm` instead of both. | Anyone copying these files by a hard-coded path |
| Inline builds now use a module worker created from a Blob URL (before: a classic worker from base64). | Browsers without module workers (Firefox < 114, Safari < 15), strict `worker-src` CSP rules |
| CSS Modules class names changed format (`X-module_root__hash` → `_root_hash_N`). Global class names are unchanged. | Consumers who style Ketcher's internal classes |
| No `__esModule` marker in `ketcher-standalone/dist/cjs/main.js` (core's fix was not applied here). | Default-import interop in CJS code |
| `ketcher-react/dist/cjs/**/*.d.ts` (507 files) and CSS source maps are no longer shipped. | Deep imports of the CJS types |
| The CSS minifier drops some vendor prefixes and writes colors as `#rrggbbaa`. | Old browsers only |

Not breaking: `exports`, `main`, `module`, `types`, `sideEffects`, peer dependencies and `engines` match master's build. All of core's named exports are still there (559 → 562), deep imports resolve, and the `.d.ts` files have no leftover path aliases. Removing `@babel/runtime` is safe. Minor: `ketcher-react` now leaks the bundler helper `__toESM` as a CJS export. It is harmless, but it should be removed.

### Smaller points

- `ketcher-react`'s build runs `tsc --noCheck`, so type errors only surface in `test:types`, not in the build.
- `vite` is pinned to exactly `8.0.16` in all four packages, not only in standalone as the PR body says. Please record why.
- Issue #10326 requirement 3 (remove the build warnings) is not met: the chunk-size warning still fires because of the 15.9 MB inlined-WASM standalone chunk. It is deferred to #11927. Please say that explicitly in the PR.
- The branch is 20 commits behind master. `git merge-tree` shows a clean merge, but after the two earlier bad merges (`21e59ce41`, `a6bb077062`), please compare the 13 manifests with both parents.
- @AlexeyGirin's 29 threads are outdated but still unresolved. The one that asks to revert the `ketcher.ts` changes conflicts with the lazy-library design and needs an explicit reply. My four questions from Sep 1 (version bump, syntax level, 24 externalized dependencies, standalone metadata) also have no reply yet.

### Suggested order before merging

1. Fix the Less version (B1), then check `dist/index.css` for `#5e5e03`.
2. Fix the `binaryWasm` worker output (B2), and add a check that builds a consumer against `dist`.
3. Make `updateMonomersLibrary` / `replaceMonomersLibrary` wait for the default library, and add a test for the ordering.
4. Restore `__esModule` in standalone CJS, and decide what to do with its `main`/`module` fields.
5. Rewrite the 4.0.0 reasoning in the ADR and changelog around the breaking changes above, and get the release owner's sign-off.
6. Merge master, delete the two ESLint comments, then sort the Playwright failures into screenshot-only, real bugs and flaky tests.
7. Reply to the open review threads.

nickkurkan:
## Summary

```text
Before: Rollup 2 (4 libraries) + CRA (`demo`) + Vite (`example`)
After:  Vite 8 / Rolldown (4 libraries + `demo` + `example`)
Keep:   Next.js (`example-ssr`); no bundler in `ketcher-autotests`
```

Type declarations remain `tsc`-emitted. The default monomer library is code-split and lazy-loaded, but not only on entry to macromolecules mode: `getKet`, `setMolecule`, `addFragment`, and `convert()` also trigger loading. Synchronous library reads are empty until it resolves; `updateMonomersLibrary`/`replaceMonomersLibrary` now await that load to avoid a late default load overwriting consumer changes.

Follow-up to the [Sep 23 review](https://github.com/epam/ketcher/pull/10401#issuecomment-5793898049): both UI packages now use Less 4.2+, the malformed Less 3 CSS output (`#5e5e03` / `grid-column: 0.666667`) is absent from the built package CSS, and the standalone CJS marker plus `main`/`module` entries are corrected. The packed standalone consumer check now passes in Vite and webpack 5.

## Evidence

**Build time** — two serial runs per clean worktree; `npm ci` completed first (install excluded). macOS arm64, Node 24.21.0 / npm 12.1.0. Baseline is the PR base `75bbb98` (`3.20.0-rc.1`); migrated build is PR head `07612be` (`4.0.0-rc.1`).

| Command | Baseline | Migrated | Change |
|---|---:|---:|---:|
| `npm run build` (packages + standalone/remote `example`) | 101.5 s (92.4–110.7) | 22.0 s (20.7–23.2) | **−78.4%, 4.6× faster** |
| `npm run build -w demo` | 20.0 s (17.7–22.2) | 1.17 s (1.16–1.19) | **−94.1%, 17× faster** |

These are local wall-clock timings, not CI timings; they supersede the earlier per-package timing table, which used different commands and endpoints.

**Published package size** — `npm pack --dry-run` compressed tarball size:

| Package | `3.20.0-rc.1` | `4.0.0-rc.1` | Change |
|---|---:|---:|---:|
| `ketcher-core` | 1.82 MiB | 2.84 MiB | +56% |
| `ketcher-react` | 2.09 MiB | 2.96 MiB | +42% |
| `ketcher-macromolecules` | 0.71 MiB | 0.66 MiB | −7% |
| `ketcher-standalone` | 32.67 MiB | 20.95 MiB | −36% |
| **Total** | **37.28 MiB** | **27.41 MiB** | **−9.87 MiB / −26.5%** |

Unpacked package contents fall from 139.3 MiB to 118.5 MiB (−14.9%). Standalone is the main saving: each URL-WASM variant now ships only the binary it uses, rather than duplicate binaries. Core/react are not smaller: the core `.map` payload is 5.42 MB → 16.76 MB, with maps embedding source content increasing from 8/978 to 978/978. That increases published size but makes consumer sourcemaps usable outside the monorepo; it is not runtime bundle growth.

On current head, CI build, CodeQL/JS analysis, all 10 full-scale Playwright shards, and the popup suite pass. The standalone external-consumer check passes for both Vite and webpack 5 (each emits the expected worker and `.wasm` files). The critical-only npm audit gate passes, with 3 high findings still reported. See [PR checks](https://github.com/epam/ketcher/pull/10401/checks).

## Merge Danger

**Door:** Two-way before release. **Blast Radius:** High — published package consumers.

Consumer-visible changes need 4.0.0 release notes and release-owner sign-off: asynchronous monomer-library readiness; `ketcher-standalone` root entry points now return real exports; hashed worker/`.wasm` paths under `assets/`; module workers created from Blob URLs (CSP must permit `worker-src blob:`); and externalized `ketcher-macromolecules` dependencies. The syntax level is already the same as current master, so the major-version rationale should be these behavior/packaging changes, not a new syntax floor.

Two remaining gates/notes: GitHub still reports Sonar **QUEUED** and the PR review decision **CHANGES_REQUESTED**, so the overall PR is currently blocked despite the green build and Playwright runs. Also, issue #10326 requirement 3 (remove build warnings) is not complete: the large-chunk warning remains for the inline WASM build and is deferred to #11927. Keep Vite pinned at 8.0.16 unless the `?no-inline` behavior and both consumer builds are revalidated.

PR #10402, opened 2026-07-07
#10363 - Refactor: Explicit Any type clean up (part 5.4)

### LINKED ISSUE #10363: Refactor: Explicit Any type clean up (part 5.4)
**Problem:** 
Unspecified any.

**Why is this an issue?**

**How can it be fixed?**

**Problem locations:**
packages/ketcher-core/src/utilities/SettingsManager.ts:37

LashaDevLasha:

## Check list
- [ ] unit-tests written
- [ ] e2e-tests written
- [ ] documentation updated

### DIFF
--- a/packages/ketcher-core/src/utilities/SettingsManager.ts
+++ b/packages/ketcher-core/src/utilities/SettingsManager.ts
@@ -32,9 +32,19 @@ const DefaultEditorLineLength: EditorLineLength = {
 
 export const SetEditorLineLengthAction = 'SetEditorLineLength';
 
+/**
+ * Shape of the selection tool persisted in local storage.
+ * Mirrors the `ActionObj` used by the UI reducer that assigns it,
+ * but is kept minimal here to avoid coupling `ketcher-core` to
+ * UI-layer action types.
+ */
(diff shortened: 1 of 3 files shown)


packages/ketcher-core/src/utilities/SettingsManager.ts
```diff
@@ -32,9 +32,19 @@ const DefaultEditorLineLength: EditorLineLength = {
...
+/**
+ * Shape of the selection tool persisted in local storage.
+ * Mirrors the `ActionObj` used by the UI reducer that assigns it,
+ * but is kept minimal here to avoid coupling `ketcher-core` to
+ * UI-layer action types.
+ */
+export interface PersistedSelectionTool {
+  tool: string;
+  opts?: unknown;
+}
```
MariiaNebesnova, line 44: as it goes from the further logic, for this type `tool` is always equal to `select`. would it be better to rework this type like following: `export type PersistedSelectionTool = {
  tool: 'select';
  opts?: unknown;
};` ?

MariiaNebesnova (approved):

Issue #9887, opened 2026-04-27
Copy paste does not work in Safari


rrodionov91:
**Steps to Reproduce**
1. Go to Molecules mode
2. Draw any structure
3. Select it
4. Press Ctrl + C
5. Press Ctrl + V

**Actual behavior**
Structure is not copied and pasted

**Expected behavior**
Structure is copied and pasted

**Desktop (please complete the following information):**
 - Safari  v26.2 (21623.1.14.11.9)

**Ketcher version** [e.g. v2.4.2]. 
v3.15.0-rc.2

AlexeyGirin:
It works according @rrodionov91 

- Ketcher Version 3.16.0-rc.2 Build at 2026-05-26; 16:42:16
- Indigo Version 1.44.0-rc.5.0-g22e1cef65-wasm32-wasm-clang-23.0.0
- Chrome Version 148.0.7778.179 (Official Build) (64-bit)
- Win11

PR #10403, opened 2026-07-07
#9887 – Fix Paste does not work in Safari

chamakarama:

Paste (Ctrl/Cmd+V) did not work in Safari. Copy and cut were fixed for Safari in #9887, but the paste handler still called `navigator.clipboard.read()` unconditionally whenever the Clipboard API was present. Safari exposes that API, so it took the async path — but Safari rejects `navigator.clipboard.read()` when it is invoked from within a `paste` event (its user-activation model is stricter than Chromium/Gecko), and the rejected promise had no handler, so paste silently did nothing.

This routes Safari through the **existing** synchronous fallback that copy/cut already use — reading from the paste event's `clipboardData` via `legacyPaste` / `onLegacyPaste` — by gating the async path on `isClipboardAPIAvailable() && !isSafariBrowser()`. Chrome and Firefox are unchanged and continue to use the async Clipboard API.

```diff
-        if (isClipboardAPIAvailable()) {
+        // Safari rejects navigator.clipboard.read() in this context, so fall
+        // back to the synchronous paste event data (same approach as copy/cut).
+        if (isClipboardAPIAvailable() && !isSafariBrowser()) {
           navigator.clipboard.read().then((data: ClipboardItem[]) => {
```

Verified manually: paste now works in Safari, and paste in Chrome and Firefox continues to work.

Completes #9887, which fixed copy/cut for Safari but left the paste path unchanged.

> Note (out of scope): the macromolecules editor (`packages/ketcher-core/src/application/editor/modes/BaseMode.ts`) has the same `navigator.clipboard.read()` pattern and may need the equivalent fallback; happy to follow up in a separate PR.

## Check list
- [x] unit-tests written
- [ ] e2e-tests written — **N/A**: the behavior is Safari-specific and the e2e suite runs on Chromium only, so it cannot exercise this path. A unit test covers both the Safari (legacy fallback) and non-Safari (async Clipboard API) paste branches.
- [ ] documentation updated — N/A
- [x] base branch (master) is correct
- [ ] task status changed to "Code review" — N/A (external contribution)
- [ ] reviewers are notified about the pull request — N/A (external contribution)

### DIFF
--- a/packages/ketcher-react/src/script/ui/component/cliparea/cliparea.test.tsx
+++ b/packages/ketcher-react/src/script/ui/component/cliparea/cliparea.test.tsx
@@ -1,0 +1,92 @@
+import type { ComponentProps } from 'react';
+import { render } from '@testing-library/react';
+import ClipArea from './cliparea';
+
+// A Safari UA string (contains "Safari" but not "Chrome"/"Android").
+const SAFARI_USER_AGENT =
+  'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/605.1.15 ' +
+  '(KHTML, like Gecko) Version/17.0 Safari/605.1.15';
+// A Chrome UA string (contains "Chrome" so it is not detected as Safari).
+const CHROME_USER_AGENT =
+  'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 ' +
+  '(KHTML, like Gecko) Chrome/120.0.0.0 Safari/537.36';
+
+const setUserAgent = (userAgent: string) => {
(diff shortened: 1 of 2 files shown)


rrodionov91:
Hi @chamakarama 
Strange, I'm trying now on Safari and copy paste works for me

chamakarama:
Closing this — after further investigation it's unnecessary.

The Safari clipboard problem I was chasing reproduces on older builds (e.g. the v3.12.0 standalone demo), but it was already resolved upstream by #9887, which routes copy/cut through a synchronous path in Safari. On current master I verified that copy **and** paste work correctly in Safari, so this change to the paste path fixes nothing that is actually broken. Apologies for the noise.

PR #10409, opened 2026-07-08
#4858 - Non-user friendly way of exiting RNA builder

### LINKED ISSUE #4858: Non-user friendly way of exiting RNA builder
**Steps to Reproduce**
1. Go to macromolecules sequence mode.
2. Enter any RNA sequence.
3. Select a nucleotide.
4. Select modify in RNA builder.
5. Click away or escape.

**Actual behavior**

Clicking away doesn't allow the user to exit the RNA builder mode.
The behavior is the same for pressing esc, but esc unselects the monomer(s).
The only way to exit the mode is to click cancel in the RNA builder.

https://github.com/epam/ketcher/assets/169887211/0634307d-5539-45c4-a98c-7232c633eaa5

**Expected behavior**

While in the RNA builder mode the user should be able to exit it by clicking away or pressing escape.
All the changes made in the RNA builder should be discarded.

**Ketcher version v2.22.0**

sabaPailodze:
## Summary
Let users leave RNA Builder edit mode ("Modify in RNA Builder" flow) by clicking away or pressing Escape, discarding pending changes. Before, only Cancel worked — clicking away did nothing and Escape needed two presses.

## Changes

**Click away to exit**
Problem: `SequenceMode.mousedown` returned early in RNA Builder mode, so canvas clicks did nothing.
Solution: Add a `cancelSequenceEditInRNABuilderMode` event — the core dispatches it on canvas mousedown, the RNA Builder panel subscribes and runs `onCancel` (exit + discard).

**Single Escape to exit**
Problem: `ContextMenu`'s capture-phase Escape handler swallowed the first Escape because the context-menu-active flag stayed on after selecting the menu item.
Solution: Clear that flag when entering RNA Builder, so the first Escape reaches the exit handler.

## Files Modified

**ketcher-core**
- `editorEvents.ts` — add `cancelSequenceEditInRNABuilderMode` event
- `modes/SequenceMode.ts` — dispatch it on canvas mousedown in RNA Builder mode

**ketcher-macromolecules**
- `RnaEditorExpanded.tsx` — subscribe to the event and run `onCancel`
- `SequenceItemContextMenu.tsx` — clear context-menu-active flag when entering RNA Builder
- `RnaEditorExpanded.test.tsx` — add the new event to the mock editor

## Check list
- [ ] unit-tests written
- [ ] e2e-tests written
- [ ] documentation updated


packages/ketcher-macromolecules/src/components/contextMenu/SequenceItemContextMenu/SequenceItemContextMenu.tsx
```diff
@@ -241,6 +242,7 @@ export const SequenceItemContextMenu = ({
 
     switch (true) {
       case menuItemId === SequenceItemContextMenuNames.modifyInRnaBuilder:
+        dispatch(setContextMenuActive(false));
```
AlexeyGirin, line 245: Probably this row is not needed and everything would work without it

AlexeyGirin (dismissed):

packages/ketcher-macromolecules/src/components/contextMenu/SequenceItemContextMenu/SequenceItemContextMenu.tsx
sabaPailodze: This line isn't redundant, I added it because there was a bug where ESC needed 2 presses to exit. ContextMenu has a capture phase Escape handler that swallows Escape while isContextMenuActive is true (it just turns the flag off instead of exiting), and clicking the menu item doesn't clear that flag because react-contexify calls stopPropagation() on the click. So without this line the first Escape gets eaten and only the second one exits. Setting setContextMenuActive(false) turns the flag off up front so it exits in one press. @rrodionov91 @AlexeyGirin

svvald (approved):

PR #10416, opened 2026-07-09
#6142 - An error occurred while saving nucleotides with different connections

### LINKED ISSUE #6142: An error occurred while saving nucleotides with different connections
**Steps to Reproduce**
1. Open Macromolecules mode - Flex mode
2. Add schema from file 
[Schema nucleotide with different monomers.ket.zip](https://github.com/user-attachments/files/16249880/Schema.nucleotide.with.different.monomers.ket.zip)
or create manually
Снимок экрана 2024-07-16 в 15 53 23
3. Open Molecules mode
4. Click on Save

**Expected behavior**
File can be saved in the MolfileV2000 format

**Actual behavior**
An error occurred
Снимок экрана 2024-07-16 в 15 50 59

**Environment details:**
 Ketcher Version 2.23.0-rc.2 Build at 2024-07-10; 19:39:22
Indigo Version 1.22.0-rc.2.0-ga21fb5d8f-wasm32-wasm-clang-19.0.0
Chrome Версия 126.0.6478.127 (Официальная сборка), (arm64)
MacOs Sonoma 14.5
Issue found for https://github.com/epam/ketcher/issues/4382

aslmbk:
### Problem
Saving a large macromolecule schema as **MDL Molfile V2000** failed with a
generic "An error occurred" toast instead of producing a file (issue #6142,
Priority: High). Reproduced from 2.23 up to 3.6.

### Root cause
After switching to Molecules mode, the schema expands to a micromolecule of
2250 atoms / 2387 bonds / 102 SUP superatom sgroups. The JS `MolSerializer`
writes SUP member indices with `paddedNum(idx, 3)`, which throws
`Error('number does not fit')` once any 1-based index reaches 1000. The MDL
Molfile V2000 format physically cannot encode more than 999 atoms/bonds, so
the correct resolution is to emit V3000 — which Indigo's `molfile-saving-mode: 'auto'`
already does.

### Fix
- `MolfileV2000Formatter` now accepts an optional fallback formatter and
  delegates to it when the struct exceeds the V2000 limit
  (`atoms.size > 999 || bonds.size > 999`); otherwise it keeps the fast JS path.
- `FormatterFactory` wires a `ServerFormatter(..., SupportedFormat.molAuto)`
  as that fallback for the non-query `mol` case. This covers both the Save
  dialog and `getMolfile('v2000')`.
- Added a shared `exceedsMolfileV2000Limit` helper / `MOLFILE_V2000_ATOM_BOND_LIMIT`
  constant in `ketcher-core`.
- The Save dialog's Warnings tab now shows a non-blocking notice that the
  structure exceeds the V2000 limit and will be saved as V3000
  (via `couldBeSaved`).

Structures that fit V2000 are unchanged (fast JS path preserved).

### Tests
- Unit (`ketcher-core`): formatter delegates at ≥1000 atoms/bonds, keeps JS
  path at ≤999 (boundary matrix); helper boundary tests; a regression test
  that the raw `MolSerializer` still throws on oversized SUP (the fix is
  routing, not silencing the serializer).
- Unit (`ketcher-react`): `couldBeSaved` warns about the V3000 upgrade for
  oversized `mol` saves and stays silent for small structures.
- E2E: load the issue schema, switch to Molecules, save as MDL Molfile V2000 —
  no error, output is valid V3000, warning shown.

## Check list
- [x] unit-tests written
- [ ] e2e-tests written
- [ ] documentation updated


packages/ketcher-core/src/application/formatters/formatterFactory.ts
```diff
@@ -90,6 +90,12 @@ export class FormatterFactory {
         } else {
           formatter = new MolfileV2000Formatter(
             new MolSerializer(molSerializerOptions),
+            new ServerFormatter(
+              this.#structService,
+              new KetSerializer(),
+              SupportedFormat.molAuto,
+              structServiceOptions,
+            ),
```
AlexeyGirin: Instead of passing v3000 formatter ro v2000 formatter, try to define here and create right (one) formatter instead of two formatters

AlexeyGirin (dismissed):

AlexeyGirin:

packages/ketcher-core/src/application/formatters/formatterFactory.ts
aslmbk: done

rrodionov91 (approved):

PR #10448, opened 2026-07-13
#5168 – Elements changes their positions after switching from Micro to Macro and back

### LINKED ISSUE #5168: Elements changes their positions after switching from Micro to Macro and back
**Steps to Reproduce:**
1. Add different elements to Canvas or load it from .ket file:
[test1.zip](https://github.com/user-attachments/files/16378682/test1.zip)
image
2. Switch to Macro mode
3. Switch back to Micro mode

**Actual behavior:**
- Elements are moved to left side, shapes, text and reaction tools are at the same place
image

**Expected behavior:**
- All entities are in the same place as it was at step 1

- OS: Windows 11 Enterprise
- Browser: Chrome 126.0.6478.128 (Official Build) (64-bit)

**Ketcher version:**
- [Version 2.23.0-rc.3] Build at 2024-07-14; 20:34:10

**Indigo version:**
- Version 1.22.0-rc.3.0-gbb5c51fc2-wasm32-wasm-clang-19.0.0

**Additional context**
- Detected during preview tests for Image feature - https://github.com/epam/ketcher/issues/4897, this issue is reproducible for images as well, as for now, this feature is not in the 2.23.0-rc.3. The fix should be verified for images as well.
- The recording is attached: 

https://github.com/user-attachments/assets/f710f544-df6e-48d6-bea7-4d22fa927591

vitaepam:
To fix #5168 
Added one core helper in the drawing-entity manager that applies an offset to both visible drawing entities and hidden micromolecule entities retained for later macro→micro conversion.
Reused that helper from the shared KET paste repositioning path and the macromolecules open flow.

Also found an issue that when switching between Sequence layout mode and Snake or Flex layout mode in Macro view we loose S-group brackets.

https://github.com/user-attachments/assets/d311fba7-422f-4405-b5dd-47c567f3a482

Fixed this issue by adding a unified persistent-renderer restore hook

## Check list
- [ ] unit-tests written
- [ ] e2e-tests written
- [ ] documentation updated

### DIFF
--- a/packages/ketcher-core/src/domain/entities/DrawingEntitiesManager.ts
+++ b/packages/ketcher-core/src/domain/entities/DrawingEntitiesManager.ts
@@ -2788,6 +2788,14 @@ export class DrawingEntitiesManager {
     SequenceRenderer.clear();
   }
 
+  public restoreSGroupRenderers() {
+    const editor = provideEditorInstance();
+
+    this.sgroups.forEach((sgroup) => {
+      editor.renderersContainer.addSGroup(sgroup);
+    });
+  }
+
   public applyFlexLayoutMode(needRedrawBonds = false) {
     const editor = provideEditorInstance();
     const command = new Command();
--- a/packages/ketcher-core/src/domain/entities/DrawingEntitiesManager.ts
+++ b/packages/ketcher-core/src/domain/entities/DrawingEntitiesManager.ts
@@ -600,6 +600,33 @@ export class DrawingEntitiesManager {
     return drawingEntity;
   }
 
+  public moveAllEntitiesWithHiddenMicromolecules(offset: Vec2) {
+    this.allEntities.forEach(([, drawingEntity]) => {
+      this.moveDrawingEntityModelChange(drawingEntity, offset);
+    });
+
+    this.micromoleculesHiddenEntities.simpleObjects.forEach((simpleObject) => {
+      simpleObject.pos = simpleObject.pos.map((position) =>
+        position.add(offset),
+      );
+    });
+
+    this.micromoleculesHiddenEntities.texts.forEach((text) => {
+      text.position = text.position.add(offset);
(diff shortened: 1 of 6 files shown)


packages/ketcher-core/src/domain/entities/DrawingEntitiesManager.ts
```diff
@@ -2763,6 +2788,14 @@ export class DrawingEntitiesManager {
...
   }
 
+  public restoreSGroupRenderers() {
+    const editor = provideEditorInstance();
+
+    this.sgroups.forEach((sgroup) => {
+      editor.renderersContainer.addSGroup(sgroup);
+    });
+  }
+
```
AlexeyGirin, line 2798: Check if it can be fixed that way

packages/ketcher-macromolecules/src/components/modal/Open/Open.tsx
```diff
@@ -147,6 +147,42 @@ const peptideLettersFormatOptions: Array<Option> = [
...
+    },
+  );
+
+  drawingEntitiesManager.micromoleculesHiddenEntities.multitailArrows.forEach(
+    (multitailArrow) => {
+      multitailArrow.move(offset);
+    },
+  );
+};
+
```
AlexeyGirin, line 185: Remove it. Use exited drawing manager from core

AlexeyGirin (changes requested):

packages/ketcher-macromolecules/src/components/modal/Open/Open.tsx
vitaepam: Done

packages/ketcher-core/src/domain/entities/DrawingEntitiesManager.ts
vitaepam: Done, moved SGroup renderer restoration to shared mode reinit flow

PR #10450, opened 2026-07-13
#7126 - Merging all chains on the canvas to one doesn't trigger recalculation in Calculate Properties window

### LINKED ISSUE #7126: Merging all chains on the canvas to one doesn't trigger recalculation in Calculate Properties window
**Steps to Reproduce**
1. Go to Macromolecules mode - Flex
2. Load from HELM: `PEPTIDE1{O}|PEPTIDE2{U}$$$$V2.0`
Image
3. Open Calculate properties (press `Alt+C`)
It is empty - this is ok
Image
4. Connect peptides to one chain with bond

**Actual behavior**
Calculate values window remains empty
Image

**Expected behavior**
Calculate values window got filled with data
Image

as per [requirement](https://github.com/epam/ketcher/issues/5727):

> 4. If only one chain is present on canvas, and the "Calculate Properties" window is opened, Ketcher should send the whole canvas to Indigo.

**Environment details:**
- Ketcher Version 3.4.0-rc.3 Build at 2025-05-15; 19:51:50
- Indigo Version 1.32.0-rc.2.0-g94ab270bd-wasm32-wasm-clang-19.0.0
- Chrome Version 136.0.7103.93 (Official Build) (64-bit)
- Win10

Found while testing: https://github.com/epam/ketcher/issues/5727

AvetBadalyan:

The "Calculate Properties" panel only recalculated in response to the editor's `selectEntities` event. Actions like deleting, flipping, or selecting entities happen to re-fire that event as a side effect, so recalculation looked automatic for them — but finishing a bond via the Single Bond tool never touches selection. As a result, merging two chains into one left the panel showing stale ("No Data Available") data until the user manually closed and reopened it.

**Root cause:** recalculation was tied to a selection event rather than a model-change event, so any model mutation that doesn't alter selection was missed.

**Fix:** introduce a dedicated `modelChange` editor event and dispatch it from the central history choke point (`EditorHistory` — `update`, `undo`, `redo`), right next to the existing `changeEvent` dispatch. This way every model mutation — including bond creation that merges chains — signals the panel, regardless of selection. In `update`, the event is fired only when the command actually changed something (`command.operations.length > 0`) to avoid needless recalculation on no-op commands; in `undo`/`redo` it is dispatched after the model has been reverted/re-applied so subscribers observe the up-to-date structure.

The "Calculate Properties" panel now subscribes to both `selectEntities` (selection-driven recalculation) and `modelChange` (structure-driven recalculation), with the existing 500 ms debounce coalescing overlapping dispatches into a single recalculation.

The three temporary `selectEntities` dispatches previously added to `Bond.ts` (`mouseUpAttachmentPoint`, `mouseUpMonomer`, `handleBondCreation`) are removed, since bond finalization already routes through `EditorHistory.update` and is now covered by `modelChange`.

## Check list
- [ ] unit-tests written
- [ ] e2e-tests written
- [ ] documentation updated

### DIFF
--- a/packages/ketcher-core/src/application/editor/tools/Bond.ts
+++ b/packages/ketcher-core/src/application/editor/tools/Bond.ts
@@ -360,6 +360,11 @@ class PolymerBond implements BaseTool {
         this.bondRenderer.polymerBond,
       );
       this.bondRenderer = undefined;
+      this.editor.events.selectEntities.dispatch(
+        this.editor.drawingEntitiesManager.selectedEntities.map(
+          (entity) => entity[1],
+        ),
+      );
       event.stopPropagation();
     }
   }
@@ -492,6 +497,11 @@ class PolymerBond implements BaseTool {
       );
       this.bondRenderer = undefined;
       this.history.update(modelChanges);
+      this.editor.events.selectEntities.dispatch(
+        this.editor.drawingEntitiesManager.selectedEntities.map(
+          (entity) => entity[1],
+        ),
+      );
       event.stopPropagation();
     }
   }
@@ -593,6 +603,11 @@ class PolymerBond implements BaseTool {
       this.bondRenderer.polymerBond,
     );
     this.bondRenderer = undefined;
+    this.editor.events.selectEntities.dispatch(
+      this.editor.drawingEntitiesManager.selectedEntities.map(
+        (entity) => entity[1],
+      ),
+    );
   };
 
   public handleBondCreationCancellation = (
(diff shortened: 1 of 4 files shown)


packages/ketcher-core/src/application/editor/tools/Bond.ts
```diff
@@ -360,6 +360,11 @@ class PolymerBond implements BaseTool {
         this.bondRenderer.polymerBond,
       );
       this.bondRenderer = undefined;
+      this.editor.events.selectEntities.dispatch(
+        this.editor.drawingEntitiesManager.selectedEntities.map(
+          (entity) => entity[1],
+        ),
+      );
```
AlexeyGirin: Make macromolecules window event that fired on any model change - and recalculate based on that event

Talk with Roman on details

AlexeyGirin (dismissed):

packages/ketcher-core/src/application/editor/tools/Bond.ts
AvetBadalyan: re-check please

svvald (approved):

PR #10476, opened 2026-07-15
#6125 - R2 should be NH2 for Phosphorodiamidate

### LINKED ISSUE #6125: R2 should be NH2 for Phosphorodiamidate
**Steps to Reproduce**
1. Go to Macro - Snake mode
2. Load from HELM using paste from clipboard: `RNA1{[m2np]}$$$$V2.0`
image
3. Hover mouse over

**Actual behavior**
R2 is N
image

**Expected behavior**
R2 is NH2 (two hydrogens since N has valence = 3 and one single bond connected)
image

**Versions**
- Ketcher Version 2.28.0-rc.1 Build at 2024-12-03; 17:04:52
- Indigo Version 1.27.0-rc.1.0-g8d656a341-wasm32-wasm-clang-19.0.0
- Chrome Version 131.0.6778.109 (Official Build) (64-bit)
- Win10

PaulinaKacz:
R2 should be NH2 for Phosphorodiamidate

Problem
For the Phosphorodiamidate monomer, the leaving group assigned to R2 was displayed/stored as the bare atom (e.g. N) instead of accounting for implicit hydrogens, so N was shown instead of NH2.

Solution

- Added a helper function getLeavingGroupLabelWithHydrogens in [monomerTemplateUtils.ts:166](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html), which counts the number of bonds the leaving group atom has within the monomer template structure, calculates the correct valence (atom.calcValence), and based on that appends the implicit hydrogens (atom.implicitH) to the label (e.g. N + 1 bond → NH2, O + 1 bond → OH).

- Updated fillStructRgLabelsByMonomerTemplate so that instead of directly assigning leavingGroupAtom.label to MonomerCaps, it uses the new helper function to build the label including hydrogens.

- Extended the LeavingGroup type in [monomers.ts:101](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) to allow an arbitrary string (not just 'O' | 'OH' | 'H'), since leaving group labels can now take different values depending on the atom and hydrogen count (e.g. NH2).

## Check list
- [ ] unit-tests written
- [ ] e2e-tests written
- [ ] documentation updated


packages/ketcher-core/src/domain/serializers/ket/fromKet/monomerTemplateUtils.ts
```diff
@@ -166,6 +166,38 @@ export function convertMonomerTemplateToStruct(
...
+  struct: Struct,
+  atomId: number,
+  atom: Atom,
+): string {
+  let bondsCount = 0;
+  struct.bonds.forEach((bond) => {
+    if (bond.begin === atomId || bond.end === atomId) {
+      bondsCount++;
+    }
+  });
```
svvald: This is not quite correct way to calculate connections as it doesn't account for bond type, please check `Struct.calcConn` function
Perhaps we could even extract it into some reusable functionality

packages/ketcher-core/src/domain/types/monomers.ts
```diff
@@ -98,7 +98,7 @@ export const attachmentPointNames = [
   'R8',
 ];
 
-export type LeavingGroup = 'O' | 'OH' | 'H';
+export type LeavingGroup = 'O' | 'OH' | 'H' | (string & NonNullable<unknown>);
```
svvald: Essentially this type is now being extended up to just `string`. Given we do not rely on leaving groups being only O or OH we can safely remove it and use string instead

packages/ketcher-core/src/domain/serializers/ket/fromKet/monomerTemplateUtils.ts
```diff
@@ -166,6 +166,38 @@ export function convertMonomerTemplateToStruct(
...
+
+  const hasValidValence = atom.calcValence(bondsCount);
```
svvald: `calcValence` mutates an atom inside which could lead to unexpected side effects
Perhaps `struct.calImplicitHydrogen` would be a better choice here

packages/ketcher-core/src/domain/serializers/ket/fromKet/monomerTemplateUtils.ts
```diff
@@ -166,6 +166,38 @@ export function convertMonomerTemplateToStruct(
...
 }
 
+/**
+ * Builds a display label for a leaving group atom, appending implicit
+ * hydrogens computed from the atom's standard valence and the number of
+ * bonds it has within the monomer template structure (e.g. an "N" atom
+ * with a single bond becomes "NH2", an "O" atom with a single bond
+ * becomes "OH").
+ */
+function getLeavingGroupLabelWithHydrogens(
```
svvald: Perhaps it's worth covering this function with unit tests

packages/ketcher-core/src/domain/types/monomers.ts
svvald: And the only usage is at `helpers/hydrateLeavingGroup.ts`, it could be safely removed now as well

svvald (changes requested):

packages/ketcher-core/src/domain/types/monomers.ts
```diff
@@ -98,7 +98,7 @@ export const attachmentPointNames = [
   'R8',
 ];
 
-export type LeavingGroup = 'O' | 'OH' | 'H';
+export type LeavingGroup = string;
```
AlexeyGirin: Please remove that type completelly

AlexeyGirin (dismissed):

packages/ketcher-macromolecules/src/helpers/hydrateLeavingGroup.ts
```diff
@@ -1,6 +1,4 @@
-import { LeavingGroup } from 'ketcher-core';
-
-const hydrateLeavingGroup = (leavingGroup: LeavingGroup) => {
+const hydrateLeavingGroup = (leavingGroup: string) => {
   return leavingGroup === 'O' ? 'OH' : leavingGroup;
```
svvald, line 2: It think we actually don't need this utility function now at all as `MonomerCaps` now carries the correct label

packages/ketcher-core/src/domain/serializers/ket/fromKet/monomerTemplateUtils.ts
```diff
@@ -166,6 +166,32 @@ export function convertMonomerTemplateToStruct(
...
+ * hydrogens computed from the atom's standard valence and the number of
+ * bonds it has within the monomer template structure (e.g. an "N" atom
+ * with a single bond becomes "NH2", an "O" atom with a single bond
+ * becomes "OH").
+ * @internal Exported for testing purposes
+ */
+export function getLeavingGroupLabelWithHydrogens(
+  struct: Struct,
+  atomId: number,
+  atom: Atom,
```
svvald, line 180: Perhaps atom is redundant here as you already pass `atomId` and can get it directly from `struct` thus no mismatch will happen

packages/ketcher-core/src/domain/serializers/ket/fromKet/monomerTemplateUtils.ts
```diff
@@ -166,6 +166,32 @@ export function convertMonomerTemplateToStruct(
...
+): string {
+  struct.calcImplicitHydrogen(atomId);
+  const implicitHydrogenCount = atom.implicitH ?? 0;
```
svvald, line 183: `?? 0` is redundant as `implicitH` declared as `number` and assigned to 0 as a fallback in atom constructor

svvald (changes requested):

AlexeyGirin:

svvald (approved):

PR #10478, opened 2026-07-15
5329 - Prevent Structure Library crash for R-Group templates

### LINKED ISSUE #5329: Application crashes when user open Structure Library with saved User Template of arrow and and element with R-Group Fragment
**Steps to Reproduce**
1. Add arrow and chemical element with R-Group Fragment to canvas:
image
2. Click on Save As...
3. Save to Templates
4. Click on Structure Library

**Actual behavior**
App is crashed, white screen, errors are in console:
image

**Expected behavior**
User can add template with arrows and element with R-Group fragment to Canvas

**Environment details:**
- Ketcher [Version 2.24.0-rc.3] Build at 2024-08-20; 13:16:59
- Indigo Version 1.23.0-rc.2.0-g6a2a7526a-wasm32-wasm-clang-19.0.0
- OS: Windows 11 Enterprise
- Browser: Chrome, Version 127.0.6533.100 (Official Build) (64-bit)

**Additional context**
- User can save element with R-Group Fragment separately as a user template, it reproduces only when both arrow and element with R-Group Fragment are saved
- Issue is detected during the investigation of Multi-Tailed Arrow Tool feature - https://github.com/epam/ketcher/issues/5055.

Nollili:
SDF serialization is now deferred until the user clicks Save to SDF, allowing affected templates to be opened and added to the canvas normally.

Added unit tests covering deferred serialization and the Structure Library integration.

### DIFF
--- a/packages/ketcher-react/src/script/ui/component/view/SaveButton.test.tsx
+++ b/packages/ketcher-react/src/script/ui/component/view/SaveButton.test.tsx
@@ -14,7 +14,7 @@
  * limitations under the License.
  ***************************************************************************/
 
-import { render, screen } from '@testing-library/react';
+import { fireEvent, render, screen } from '@testing-library/react';
 import { SaveButton } from './savebutton';
 
 jest.mock('../../../../hooks', () => {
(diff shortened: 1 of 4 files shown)


svvald (approved):
Current fix addresses the immediate issue – the application crash when trying to open Templates Library with incorrect saved template entry, now the whole library is not serialized eagerly upon dialog open which fixed the crash.

However, it could use one or several follow-up fixes/improvements. 
1. `SaveButton` within `FooterContent` of `TemplateDialog` isn't provided any `onError` handlers thus if user clicks to `Save to SDF`, it will just silently fail. Ideally, a warning/error message should be shown.
2. A further improvement after #1 – with existing approach saving an incorrect structure as a template will "contaminate" the whole group as currently it is serialized as a whole and an error is thrown for the whole group. An individual serialization of each item with skipping incorrect structures + showing a warning and returning an SDF for correct items only will provide a better UX.
3. Even further improvement after #2 – incorrect structures might be validated upon saving and not allowed to be saved or warned about inability to be exported as SDF later. This probably has to be aligned with the BA

PR #10499, opened 2026-07-17
#5331 - Attempt to replace monomer on inpropper one from the library to the sequence causes wrong error message

### LINKED ISSUE #5331: Attempt to replace (in view mode) monomer on inpropper one from the library to the sequence causes wrong error message
**Steps to Reproduce**
1. Go to Macromolecules mode - Sequence mode
2. Load from file: [Attempt to replace monomer on inpropper one from the library to the sequence causes wrong error message.zip](https://github.com/user-attachments/files/16687747/Attempt.to.replace.monomer.on.inpropper.one.from.the.library.to.the.sequence.causes.wrong.error.message.zip)
image
3. Click on the last `A` peptide to select it
image
4. Click on `DACys` peptide (C section, it has no R1 connection point)

**Actual behavior**
System shows error message: `It is impossible to merge fragments. Attachment point to establish bonds are not available.`
image

**Expected behavior**
Error message text should be `The monomer lacks R1 connection point and cannot be inserted at current position`
image

As per [requirement](https://github.com/epam/ketcher/issues/4878): 

> 3. Error message text should be `The monomer lacks R1 connection point and cannot be inserted at current position`

**Versions**
- Ketcher Version 2.24.0-rc.3 Build at 2024-08-20; 13:16:59
- Indigo Toolkit Version 1.23.0-rc.2.0-g6a2a7526a-wasm32-wasm-clang-19.0.0
- Chrome Version 127.0.6533.100 (Official Build) (64-bit)
- Win10

Related issue - https://github.com/epam/ketcher/issues/4878

HelenaIsh:
When replacing a selected monomer in Sequence mode with a monomer from the
library that lacks the connection point required to keep the backbone bond
(e.g. selecting the last `A` and picking `DACys`, which has no R1), the editor
used to show a generic error: "It is impossible to merge fragments. Attachment
point to establish bonds are not available."

Now the same specific message used in edit mode is shown for the selection-based
replacement path:
`The monomer lacks R1 attachment point and cannot be inserted at current position`
(R1 or R2 depending on which backbone connection point is missing).

Changes:
- Added `getMissingBackboneAttachmentPointForSelections` in `SequenceMode` to
  detect which backbone attachment point (R1/R2) the new monomer is missing.
- `insertMonomerFromLibrary` now passes the specific message to the error modal.
- Updated existing replacement e2e tests (Cases 7–16) to expect the specific
  message and added a regression test.

### Note on wording ("attachment point" vs "connection point")
The original requirement phrases the message as "...lacks R1 **connection
point**...", but the current codebase consistently uses "...lacks R1
**attachment point**..." for this message (three places in `SequenceMode.ts`).
I kept "attachment point" to stay consistent with the existing code and the
edit-mode message. If we prefer the exact wording from the requirement, it
should be changed in all three places in `SequenceMode.ts` plus the test helper
so everything stays in sync.
## Check list
- [ ] unit-tests written
- [x] e2e-tests written
- [ ] documentation updated

### DIFF
--- a/packages/ketcher-core/src/application/editor/modes/SequenceMode.ts
+++ b/packages/ketcher-core/src/application/editor/modes/SequenceMode.ts
@@ -2237,6 +2237,61 @@ export class SequenceMode extends BaseMode {
     );
   }
 
+  private getMissingBackboneAttachmentPointForSelections(
+    selections: TwoStrandedNodesSelection,
+    monomerItem: MonomerItemType,
+  ): AttachmentPointName | null {
+    if (!monomerItem.attachmentPoints) {
+      return null;
+    }
+
+    const newMonomerAttachmentPoints =
+      BaseMonomer.getAttachmentPointDictFromMonomerDefinition(
+        monomerItem.attachmentPoints,
+      );
+
+    for (const selectionRange of selections) {
+      for (const nodeSelection of selectionRange) {
+        const senseNode = nodeSelection.node.senseNode;
+        if (!senseNode) {
+          continue;
+        }
+
+        const backboneBonds: [
+          AttachmentPointName,
+          PolymerBond | MonomerToAtomBond | null | undefined,
+        ][] = [
+          [
+            AttachmentPointName.R1,
+            senseNode.firstMonomerInNode.attachmentPointsToBonds.R1,
+          ],
+          [
+            AttachmentPointName.R2,
+            senseNode.lastMonomerInNode.attachmentPointsToBonds.R2,
+          ],
+        ];
+
+        for (const [attachmentPointName, bond] of backboneBonds) {
+          const isBackboneBond =
+            bond &&
+            (bond instanceof MonomerToAtomBond ||
+              bond.isBackBoneChainConnection);
+
+          if (
+            isBackboneBond &&
+            !newMonomerAttachmentPoints.attachmentPointsList.includes(
+              attachmentPointName,
+            )
+          ) {
+            return attachmentPointName;
+          }
+        }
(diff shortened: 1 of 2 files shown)


packages/ketcher-core/src/application/editor/modes/SequenceMode.ts
```diff
@@ -2358,7 +2413,15 @@ export class SequenceMode extends BaseMode {
...
-        this.showMergeWarningModal();
+        const missingAttachmentPoint =
+          this.getMissingBackboneAttachmentPointForSelections(
+            selections,
+            monomerItem,
+          );
+        const message =
+          missingAttachmentPoint &&
+          `The monomer lacks ${missingAttachmentPoint} attachment point and cannot be inserted at current position`;
+        this.showMergeWarningModal(message);
```
AlexeyGirin: Extend `selectionsCantPreserveConnectionsWithMonomer` function to return invalid attachment points
(and rename it to more relevant name)

AlexeyGirin (dismissed):

packages/ketcher-core/src/application/editor/modes/SequenceMode.ts
HelenaIsh: Good point! Both `selectionsCantPreserveConnectionsWithMonomer` and `getMissingBackboneAttachmentPointForSelections` were iterating over the same `selections`, so I merged them into `getFirstMissingAttachmentPoint` that returns `AttachmentPointName | null` in a single pass

svvald (approved):

Issue #4927, opened 2024-06-28
Able to change attachment point label to an atom if it is selected first


Zhirnoff:
**Steps to Reproduce**
1. Add Benzene ring on canvas
2. Add attachment point on Benzene ring by right-click menu
3. Select with Rectangle selection R1 label
4. Press on Nitrogen atom in atom toolbar

**Actual behavior**
Able to change attachment point label to an atom if it is selected first

**Expected behavior**
Unble to change attachment point label to an atom

**Video**

https://github.com/epam/ketcher/assets/62464241/b184540d-ac70-4c01-b033-bb04cad8e37b

**Desktop:**
 - OS: Windows 11
 - Browser Chrome
 - Version 124.0.6367.79 (Official Build) (64-bit)

**Ketcher version** 
[Version 2.22.0-rc.7]
**Indigo version** 
[Version 1.21.0-rc.2]

Issue found while testing https://github.com/epam/ketcher/issues/4530

AnnaKosol:
A modal window with information about the changes is displayed.

https://github.com/epam/ketcher/assets/174325443/d4746a4e-f234-4bb3-b402-a7eb7a53cf53


**Versions:**
Ketcher Version 2.23.0-rc.1 Build at 2024-07-02; 09:03:38
Indigo Version 1.21.0-rc.1.0-g9194599b2-x86_64-linux-gnu-11.2.1
MacOs Sonoma version 14.5
Chrome Version 126.0.6478.127 (Официальная сборка), (arm64)

Zhirnoff:
Moved to 2.23.0-rc.2 after discussion with developer @Guch1g0v because he has a pull request ready

Zhirnoff:
Tested. Bug fixed.

https://github.com/epam/ketcher/assets/62464241/43590f9c-9753-4800-8f50-5a730a045f0d



**Desktop:**
 - OS: Windows 11
 - Browser Chrome
 - Version 124.0.6367.79 (Official Build) (64-bit)

**Ketcher version** 
[Version 2.23.0-rc.2]
**Indigo version** 
[Version 1.22.0-rc.2]

PR #10508, opened 2026-07-17
#6743 – Able to move the leaving groups (R-atoms) of expanded monomers

### LINKED ISSUE #6743: Able to move the leaving groups (R-atoms) of expanded monomers
**Steps to Reproduce**

**Actual behavior**

**Expected behavior**

**Screenshots**

**Smartphone (please complete the following information):**

To determine it, hover over the header of the Ketcher tab in a browser. 

**Test case**
Add test case number (in EPMLSOPKET-* format) if bug was found during test case run.

**Additional context**

AvetBadalyan:
Dragging an atom in Molecules mode already moves the whole related sgroup correctly
(via `fromMultipleMove`/`getRelSGroupsBySelection` in ketcher-core, which is unconditional
and doesn't check leaving-group status). However, after dragging a leaving-group (R#) atom
of an *expanded* monomer, `editor.selection()` was left containing only that one atom
(`{"atoms":[2]}`), instead of the whole structure (`{"atoms":[0,1,2,3,4,5,6],"bonds":[...]}`)
as happens when dragging any other atom (e.g. nitrogen). Verified this discrepancy directly
in a real browser by introspecting the app's internal selection state before/after a drag,
toggling the fix on and off.

Practical impact: after dragging R1, only R1 remains "selected" — so a follow-up Delete,
arrow-key nudge, or re-drag would only affect that one atom, not the structure, even though
it visually looks like everything moved together. That's the real substance of "R-groups
should behave like any other atom" from the issue.

Root cause: `getGroupIdsFromItemArrays` (shared helper used by `select.ts`'s mousedown
handler to decide whether a click should auto-expand into "select the whole enclosing
group") explicitly excludes leaving-group atoms via `Atom.isSuperatomLeavingGroupAtom`.
That exclusion was added for a narrower, different purpose (#4927 — prevent the
atom-relabeling tool from accidentally relabeling a whole group when clicking an R# atom
to rename it), but it leaked into the shared selection-click logic used by the Select/Move
tool too.

Fix: added an opt-in `includeExpandedLeavingGroupAtoms` option to `getGroupIdsFromItemArrays`
(default `false`, so the #4927 fix and every other call site are untouched), and enabled it
only in `select.ts`'s mousedown handler — so a leaving-group atom is only treated as an
ordinary atom of the structure for selection purposes once its owning sgroup is actually
expanded (`sgroup.isExpanded()`), matching the ticket's exact scenario.

Verified in a real browser (`dev:standalone`): expanded an Alanine monomer, dragged the R1
(leaving-group) atom, and confirmed the resulting selection now includes the whole structure
— reproduced the bug on the unpatched code first (selection stayed `{"atoms":[2]}`), then
confirmed the fix.

## Minor observations (not blockers):

Lasso-selection (`onSelectionEnd` in `select.helpers.ts:216-218`) still uses the old
(no-options) call. If a user lasso-selects just an R1 atom and drags, it could exhibit the
same independent-move bug — untested by you since it's not in the ticket's repro steps, but
worth confirming intentionally out of scope.

## Check list
- [ ] unit-tests written
- [ ] e2e-tests written
- [ ] documentation updated

### DIFF
--- a/packages/ketcher-react/src/script/editor/tool/helper/getGroupIdsFromItems.ts
+++ b/packages/ketcher-react/src/script/editor/tool/helper/getGroupIdsFromItems.ts
@@ -5,14 +5,32 @@ type Items = {
   bonds?: number[];
 };
 
-function getGroupIdsFromItemArrays(struct: Struct, items?: Items): number[] {
+type GetGroupIdsFromItemArraysOptions = {
+  // Once a superatom is expanded, its leaving-group (R#) atoms become
+  // ordinary atoms of the structure and should be treated the same as any
+  // other atom when resolving which group a click belongs to (#6743).
+  includeExpandedLeavingGroupAtoms?: boolean;
+};
+
+function getGroupIdsFromItemArrays(
+  struct: Struct,
+  items?: Items,
+  options?: GetGroupIdsFromItemArraysOptions,
+): number[] {
   if (!struct.sgroups.size) return [];
 
   const groupsIds = new Set<number>();
   if (items?.atoms) {
     items.atoms
       .filter((atomId) => {
-        return !Atom.isSuperatomLeavingGroupAtom(struct, atomId);
+        if (!Atom.isSuperatomLeavingGroupAtom(struct, atomId)) {
+          return true;
+        }
+
+        return (
+          options?.includeExpandedLeavingGroupAtoms &&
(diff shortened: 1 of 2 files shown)


AlexeyGirin:
Bug is already fixed

PR #10513, opened 2026-07-17
#10443 - Refactor: Explicit Any type clean up from packages\ketcher-core\src\application\editor\operations\bond\BondAttr.ts file

### LINKED ISSUE #10443: Refactor: Explicit/implicit Any type clean up from `packages\ketcher-core\src\application\editor\operations\bond\BondAttr.ts` file
**Problem:** 
Unspecified any.

**Why is this an issue?**

**How can it be fixed?**

**Problem locations:**
packages\ketcher-core\src\application\editor\operations\bond\BondAttr.ts

elenamkoyan:
Fixed all any types in packages/ketcher-core/src/application/editor/operations/bond/BondAttr.ts:

Changes Made:

  1. bid: any → number: 
    - Bond ID is always a numeric identifier
  2. Constructor parameters typed:
    - bondId?: any → bondId?: number
    - attribute?: any → attribute?: BondAttributeName
    - value?: any → value?: unknown
    - Uses unknown since bond attributes have heterogeneous value types
  3. Created BondAttributeName union type 
    - Lists all modifiable bond properties: stereo, customQuery, len, sb, sa, angle, isPreview, etc.
    - Includes | string to allow dynamic attribute names from Object.keys() patterns
  4. Enhanced constructor logic 
    - Added guards to ensure data is only set when both required parameters are defined
  5. Used appropriate type assertion 
    - (bond as unknown as Record<BondAttributeName, unknown>) to safely handle readonly
  property assignment

## Check list
- [ ] unit-tests written
- [ ] e2e-tests written
- [ ] documentation updated

### DIFF
--- a/packages/ketcher-core/src/application/editor/actions/atomMerge.ts
+++ b/packages/ketcher-core/src/application/editor/actions/atomMerge.ts
@@ -65,7 +65,13 @@ export function fromAtomMerge(
       // replace old bond with new bond
       const attrs = Bond.getAttrHash(bond);
       Object.keys(attrs).forEach((key) => {
-        action.addOp(new BondAttr(mergeBondId, key, attrs[key]));
+        action.addOp(
+          new BondAttr(
+            mergeBondId,
+            key as keyof typeof attrs,
+            attrs[key as keyof typeof attrs],
+          ),
+        );
       });
     }
 
--- a/packages/ketcher-core/src/application/editor/actions/atomMerge.ts
+++ b/packages/ketcher-core/src/application/editor/actions/atomMerge.ts
@@ -65,13 +65,8 @@ export function fromAtomMerge(
       // replace old bond with new bond
       const attrs = Bond.getAttrHash(bond);
       Object.keys(attrs).forEach((key) => {
-        action.addOp(
-          new BondAttr(
-            mergeBondId,
-            key as keyof typeof attrs,
-            attrs[key as keyof typeof attrs],
-          ),
-        );
+        const attrKey = key as keyof typeof attrs;
+        action.addOp(new BondAttr(mergeBondId, attrKey, attrs[attrKey]));
       });
     }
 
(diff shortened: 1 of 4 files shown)


packages/ketcher-core/src/application/editor/operations/bond/BondAttr.ts
```diff
@@ -18,10 +18,28 @@ import { BaseOperation } from '../BaseOperation';
...
+  | 'isPreview'
+  | 'beginSuperatomAttachmentPointNumber'
+  | 'endSuperatomAttachmentPointNumber'
+  | 'cip'
+  | 'hb1'
+  | 'hb2'
+  | 'type'
+  | 'topology'
+  | 'reactingCenterStatus'
+  | string;
```
MariiaNebesnova: please investigate if it can be replaced with `type BondAttributeName = keyof BondAttributes;` to avoid the code duplication. Or smth like 
``type BondAttributeName =
  keyof BondAttributes
  | 'len'
  | 'sb'
  | 'sa'
  | 'angle'
  | 'hb1'
  | 'hb2';`` if the logic requires more values than there are in `BondAttributes`. 

Please avoid `| string` if possible, since it makes the whole type unsafe

packages/ketcher-core/src/application/editor/actions/atomMerge.ts
```diff
@@ -65,7 +65,13 @@ export function fromAtomMerge(
       // replace old bond with new bond
       const attrs = Bond.getAttrHash(bond);
       Object.keys(attrs).forEach((key) => {
-        action.addOp(new BondAttr(mergeBondId, key, attrs[key]));
+        action.addOp(
+          new BondAttr(
+            mergeBondId,
+            key as keyof typeof attrs,
+            attrs[key as keyof typeof attrs],
```
MariiaNebesnova: If you define a return type at `Bond.getAttrHash` it will be a clearer solution than this double typecast. It caused because `attrs` don't have a proper type. Please, try to rework it this way

AlexeyGirin: @copilot please make a fix to respect review comment

MariiaNebesnova (approved):

PR #10576, opened 2026-07-17
#10434 – Refactor: Explicit Any type clean up from packages\ketcher-core\src\application\render\restruct\reatom.ts file

### LINKED ISSUE #10434: Refactor: Explicit/implicit Any type clean up from `packages\ketcher-core\src\application\render\restruct\reatom.ts` file
**Problem:** 
Unspecified any.

**Why is this an issue?**

**How can it be fixed?**

**Problem locations:**
packages\ketcher-core\src\application\render\restruct\reatom.ts

OnurKucukoz:

- Removed explicit `any` usages from `packages/ketcher-core/src/application/render/restruct/reatom.ts`.
- Replaced them with explicit types to improve type safety.
- Updated the related render option typings where required.
- Added safe fallback values to preserve the existing behavior while satisfying TypeScript.

## Check list
- [ ] unit-tests written
- [ ] e2e-tests written
- [ ] documentation updated

### DIFF
--- a/packages/ketcher-core/src/application/render/restruct/reatom.ts
+++ b/packages/ketcher-core/src/application/render/restruct/reatom.ts
@@ -38,6 +38,7 @@ import draw from '../draw';
 import util from '../util';
 import { toFixed } from 'utilities';
 import type {
+  RelativeBox,
   RenderOptions,
   RenderOptionStyles,
 } from 'application/render/render.types';
@@ -54,11 +55,12 @@ import { ShowHydrogenLabels } from './showHydrogenLabels';
 interface ElemAttr {
   text: string;
   path: Element | RaphaelSet;
-  rbb: { x: number; y: number; width: number; height: number };
+  rbb: RelativeBox;
   background?: Element;
 }
 
 const StereoLabelMinOpacity = 0.3;
+const DEFAULT_STEREO_COLOR = '#000';
 const MAX_LABEL_LENGTH = 8;
 
 export enum ShowHydrogenLabelNames {
@@ -556,21 +558,21 @@ class ReAtom extends ReObject {
     }
 
     if (options.showAtomIds) {
-      index = {} as ElemAttr;
-      index.text = aid.toString();
+      const text = aid.toString();
       let idPos = this.hydrogenOnTheLeft
         ? Vec2.lc(ps, 1, new Vec2({ x: -2, y: 0, z: 0 }), 6)
         : Vec2.lc(ps, 1, new Vec2({ x: 2, y: 0, z: 0 }), 6);
       if (this.showLabel) {
         idPos = Vec2.lc(idPos, 1, new Vec2({ x: 1, y: -3, z: 0 }), 6);
       }
-      index.path = render.paper.text(idPos.x, idPos.y, index.text).attr({
+      const path = render.paper.text(idPos.x, idPos.y, text).attr({
         font: options.font,
         'font-size': options.fontszsubInPx,
         fill: '#070',
       });
-      index.rbb = util.relBox(index.path.getBBox());
-      draw.recenterText(index.path, index.rbb);
+      const rbb = util.relBox(path.getBBox());
+      draw.recenterText(path, rbb);
+      index = { text, path, rbb };
       restruct.addReObjectPath(LayerMap.indices, this.visel, index.path, ps);
     }
 
@@ -629,7 +631,6 @@ class ReAtom extends ReObject {
         !shouldHideHydrogenInPreview
       ) {
         const data = showHydrogen(this, render, implh, {
-          hydrogen: {} as ElemAttr,
           hydroIndex,
           rightMargin,
           leftMargin,
@@ -1074,7 +1075,9 @@ class ReAtom extends ReObject {
         // of just created text
         // text -> tspan
         const color = getStereoAtomColor(render.options, stereoLabel);
-        aamPath.node.childNodes[0].setAttribute('fill', color);
+        if (color !== undefined) {
+          aamPath.node.childNodes[0].setAttribute('fill', color);
+        }
         const opacity = getStereoAtomOpacity(render.options, stereoLabel);
         aamPath.node.childNodes[0].setAttribute('fill-opacity', opacity);
       }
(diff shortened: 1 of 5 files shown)


packages/ketcher-core/src/application/render/restruct/reatom.ts
```diff
@@ -2056,15 +2080,18 @@ export function checkIsSmartPropertiesExist(atom) {
...
+) {
   let queryAttrsText = '';
   const nonQueryAttributes = ['charge', 'explicitValence', 'isotope'];
 
   const addSemicolon = () => {
     if (queryAttrsText.length > 0) queryAttrsText += ';';
   };
   const patterns: {
-    [key: string]: (value: string, atom) => string;
+    [key: string]: (value: string, atom: unknown) => string;
```
svvald: I suppose `atom` type here and `atom` type in the whole `getAtomCustomQuery` function definition should be the same, as that's the same value that's being passed into the pattern later on

packages/ketcher-core/src/application/render/restruct/reatom.ts
```diff
@@ -2056,15 +2080,18 @@ export function checkIsSmartPropertiesExist(atom) {
   return smartsSpecificProperties.some((name) => atom.queryProperties?.[name]);
 }
 
-export function getAtomCustomQuery(atom, includeOnlyQueryAttributes?: boolean) {
+export function getAtomCustomQuery(
+  atom: Record<string, unknown>,
```
svvald: Is it a correct type here, can't we use `Atom` type?

packages/ketcher-core/src/application/render/restruct/reatom.ts
```diff
@@ -2101,7 +2128,7 @@ export function getAtomCustomQuery(atom, includeOnlyQueryAttributes?: boolean) {
 
     const value = atom[propertyName];
     if (propertyName in atom && value !== null) {
-      const attrText = patterns[propertyName](value, atom);
+      const attrText = patterns[propertyName](value as unknown as string, atom);
```
svvald: I'd try to avoid using `as` casting, I suppose it will be resolved if you use a proper type for `atom`

packages/ketcher-core/src/application/render/restruct/reatom.ts
```diff
@@ -2140,7 +2167,12 @@ function getQueryAttrsText(atom): string {
   return queryAttrsText;
 }
 
-function pathAndRBoxTranslate(path, rbb, x, y) {
+function pathAndRBoxTranslate(
+  path: Element | RaphaelSet,
+  rbb: { x: number; y: number; width: number; height: number },
```
svvald: The very same type is used at `ElemAttr` in this file and in couple more files. Perhaps we could extract it as a shared type definition and import it where needed?

packages/ketcher-core/src/application/render/restruct/reatom.ts
```diff
@@ -1329,23 +1334,26 @@ function getStereoAtomColor(options, stereoLabel) {
...
-      return options.colorOfAndCenters;
+      return options.colorOfAndCenters ?? '#000';
     case StereoLabel.Or:
-      return options.colorOfOrCenters;
+      return options.colorOfOrCenters ?? '#000';
     case StereoLabel.Abs:
-      return options.colorOfAbsoluteCenters;
+      return options.colorOfAbsoluteCenters ?? '#000';
     default:
       return '#000';
```
svvald: Perhaps extract this value as a constant for default color?

packages/ketcher-core/src/application/render/restruct/reatom.ts
```diff
@@ -1483,9 +1497,9 @@ function shouldHydrogenBeOnLeft(struct, atom) {
 
   if (visibleNeighbors.length === 1) {
     const neighbor = visibleNeighbors[0];
-    const neighborDirection = struct.halfBonds.get(neighbor).dir;
+    const neighborHalfBond = struct.halfBonds.get(neighbor);
 
-    return neighborDirection.x > 0;
+    return (neighborHalfBond?.dir.x ?? 0) > 0;
```
svvald: Let's extract it into a variable and use it here for better readability

svvald: Also please check other cases, there are few in this pr

packages/ketcher-core/src/application/render/restruct/reatom.ts
```diff
@@ -1544,14 +1561,10 @@ function buildLabel(
     usageInMacromolecule,
   } = options;
   // eslint-disable-line max-statements
-  const label: any = {
-    text: getLabelText(atom.a, atomId, sgroup, options),
-  };
+  const label = {} as ElemAttr;
```
svvald: I'd suggest to avoid using `as` casting as much as possible, in this particular case I assume we can gather `text`, `path` and other needed properties along the whole function and then just create an object `label = { text, path ... }` in the very end rather than mutating it all over the place

svvald: And same for many other cases of `as` casting for empty objects

svvald (changes requested):

AlexeyGirin:
@copilot please make fixes based on provided review comments

packages/ketcher-core/src/application/render/restruct/reatom.ts
```diff
@@ -2072,72 +2214,44 @@ export function checkIsSmartPropertiesExist(atom) {
...
-      if (attrText) {
-        addSemicolon();
-      }
-      queryAttrsText += attrText;
+    const value = getValue(atom);
+    if (value === null || value === undefined) {
+      continue;
     }
+
+    const normalizedValue = typeof value === 'boolean' ? Number(value) : value;
```
svvald: That's a strange coercion of boolean value to numeric, is it intentional? I can not find anything remotely similar in the previous code

AlexeyGirin: @copilot please comment on that and make correction if it was done not intentional

MariiaNebesnova (approved):

svvald (approved):

PR #10646, opened 2026-07-20
  #10557 – Refactor: Explicit/implicit Any type clean up from packages\ketcher-react\src\script\ui\utils\index.ts file

### LINKED ISSUE #10557: Refactor: Explicit/implicit Any type clean up from `packages\ketcher-react\src\script\ui\utils\index.ts` file
**Problem:** 
Unspecified any.

**Why is this an issue?**

**How can it be fixed?**

**Problem locations:**
packages\ketcher-react\src\script\ui\utils\index.ts

dan-iliescu:

  Removed all explicit and implicit any types from
  packages/ketcher-react/src/script/ui/utils/index.ts,
  replacing them with precise types (reusing existing
  ones where available):

  - filterLib / filterFGLib — the two explicit (item:
  any) predicates now use the existing exported Template
  type. Also typed the lib/filter params, the reduce
  accumulators, and the return value (Record<string, 
  Template[]>). Replaced the .filter(Boolean) as 
  string[] cast with a field is string type guard.
  - getSelectOptionsFromSchema — the untyped schema
  param now uses a EnumSchemaLike type plus runtime
  Array.isArray guards, so it safely accepts every
  caller shape (form SchemaProperty, jsonschema Schema,
  plain records, undefined) without weakening the return
  type. Enum values are coerced with String(value),
  which is sound since Option.value is typed string (and
  enum values are already strings at runtime).
  - memoizedDebounce — added a generic <TArgs extends 
  unknown[]> signature, typed lastArgs, and converted
  the returned function expression to an arrow function.
  - greekify — typed GREEK_SIMBOLS as Record<string, 
  string> to remove the implicit-any index access.

  One downstream change: tightening filterLib's return
  type surfaced that TemplateTable's templates prop was
  mutable while the shared EMPTY_TEMPLATES sentinel is
  readonly. Widened the prop to ReadonlyArray<Template>
  — the component only reads the array, and mutable
  arrays remain assignable, so no caller is affected.

  No behavior change; this is a type-safety-only
  refactor.

  Verification: tsc --noEmit, eslint, prettier --check,
  and test:circ (no circular dependency — the import 
  type is erased) all pass on the changed files.

  Check list

  - [ ] unit-tests written
  - [ ] e2e-tests written
  - [ ] documentation updated
  - [x] PR name follows the pattern #1234 – issue name

  ---
  A few notes on the unchecked boxes so you can tick
  them as you go:

  - unit/e2e tests — none written; this is a pure type
  refactor with no behavior change, and the file has no
  existing unit tests. Reasonable to leave unchecked (or
  note "N/A — type-only refactor" if reviewers expect a
  reason).
  - documentation — N/A.
  - PR linked / base branch / task status / reviewers —
  these are GitHub/board actions to do when you open the
  PR. Base branch should be master (matches where the
  branch was cut).

### DIFF
--- a/packages/ketcher-react/src/script/ui/component/form/Select/Select.tsx
+++ b/packages/ketcher-react/src/script/ui/component/form/Select/Select.tsx
@@ -63,7 +63,9 @@ const Select = ({
   error,
   title,
 }: Props) => {
-  const currentValue = options?.find((option) => option.value === value);
+  const currentValue = options?.find(
+    (option) => String(option.value) === String(value),
+  );
   const isFullscreen = !!document.fullscreenElement;
   const portalContainer = isFullscreen
     ? document.querySelector('#root')
--- a/packages/ketcher-react/src/script/ui/utils/index.ts
+++ b/packages/ketcher-react/src/script/ui/utils/index.ts
@@ -16,8 +16,9 @@
 import _ from 'lodash';
 import { escapeRegExp, filter as _filter, flow, reduce } from 'lodash/fp';
 import type { Option } from '../component/form/Select';
+import type { Template } from '../dialog/template/TemplateTable';
 
-const GREEK_SIMBOLS = {
+const GREEK_SIMBOLS: Record<string, string> = {
   Alpha: 'A',
   alpha: 'α',
   Beta: 'B',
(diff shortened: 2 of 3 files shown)


packages/ketcher-react/src/script/ui/utils/index.ts
```diff
@@ -31,55 +32,77 @@ const greekRe = new RegExp(
...
+ * Loose JSON-schema shape consumed by {@link getSelectOptionsFromSchema}.
+ * Callers pass a variety of schema-ish objects (form `SchemaProperty`,
+ * jsonschema `Schema`, plain records), so `enum`/`enumNames` are validated at
+ * runtime rather than relied on structurally.
+ */
+type EnumSchemaLike =
+  | { enum?: unknown; enumNames?: unknown }
+  | Record<string, unknown>;
+
 export function greekify(str: string): string {
```
svvald, line 48: Is it needed though? `Record<string, unknown>` already covers `{ enum?: unknown; enumNames?: unknown }` itself

packages/ketcher-react/src/script/ui/utils/index.ts
```diff
@@ -31,55 +32,77 @@ const greekRe = new RegExp(
...
   )(lib);
 }
 
-export const getSelectOptionsFromSchema = (schema): Array<Option> => {
-  return schema.enum.reduce((options, value, index) => {
+export const getSelectOptionsFromSchema = (
+  schema?: EnumSchemaLike | null,
+): Array<Option> => {
+  const values = Array.isArray(schema?.enum) ? schema.enum : [];
+  const names = Array.isArray(schema?.enumNames) ? schema.enumNames : undefined;
```
svvald, line 97: Is `undefined` as a fallback expected here rather than `[]`?

packages/ketcher-react/src/script/ui/utils/index.ts
```diff
@@ -31,55 +32,77 @@ const greekRe = new RegExp(
...
   const re = new RegExp(escapeRegExp(greekify(trimmedFilter)), 'i');
   return flow(
     _filter(
-      (item: any) =>
+      (item: Template) =>
         !trimmedFilter ||
         re.test(greekify(item.struct.name)) ||
         re.test(greekify(item.props.group)) ||
-        (item.props.abbreviation && re.test(greekify(item.props.abbreviation))),
+        (!!item.props.abbreviation &&
```
svvald, line 61: I think `!!` is redundant here but if it yields some errors without it, let's stick to `Boolean()` coercion instead

svvald (changes requested):

packages/ketcher-react/src/script/ui/utils/index.ts
dan-iliescu: I think the named type helps with autocomplete and Record type would help with future proofing, like maybe using other new fields.

AlexeyGirin:
@copilot please make fix based on provided review comments

packages/ketcher-react/src/script/ui/utils/index.ts
dan-iliescu: Yes, undefined is on purpose. names is optional here, some schemas don't have enumNames at all. Since I only read it as names?.[index], undefined just means "no labels" and it falls back to String(value). 

[] would do the same thing but felt a bit misleading, like the labels are there but empty. Happy to switch it to [] if you'd rather keep both fallbacks consistent though.

packages/ketcher-react/src/script/ui/utils/index.ts
dan-iliescu: Unfortunately without it, the && returns item.props.abbreviation's value (a string | undefined) when the left side is falsy, so the callback ends up returning string | boolean instead of a clean boolean, which trips the type. That's why I added the coercion. Happy to swap !! for Boolean(item.props.abbreviation) though if that reads cleaner, no problem

packages/ketcher-react/src/script/ui/utils/index.ts
```diff
@@ -31,55 +32,77 @@ const greekRe = new RegExp(
...
+
+/**
+ * Loose JSON-schema shape consumed by {@link getSelectOptionsFromSchema}.
+ * Callers pass a variety of schema-ish objects (form `SchemaProperty`,
+ * jsonschema `Schema`, plain records), so `enum`/`enumNames` are validated at
+ * runtime rather than relied on structurally.
+ */
+type EnumSchemaLike =
+  | { enum?: unknown; enumNames?: unknown }
+  | Record<string, unknown>;
```
la-nuta: Fails `prettier/prettier` on master (prettier 3 since #10806):

```
46:3  error  Replace `|·{·enum?:·unknown;·enumNames?:·unknown·}⏎·` with `{·enum?:·unknown;·enumNames?:·unknown·}`
```

Collapsing to the structural member fixes it and settles the thread above. (Collapsing to `Record<string, unknown>` instead breaks tsc — `Schema`/`SchemaProperty` have no index signature.) I built this variant: eslint and `tsc -p ketcher-react` clean.

```suggestion
type EnumSchemaLike = { enum?: unknown; enumNames?: unknown };
```

packages/ketcher-react/src/script/ui/component/form/Select/Select.tsx
```diff
@@ -63,7 +63,9 @@ const Select = ({
   error,
   title,
 }: Props) => {
-  const currentValue = options?.find((option) => option.value === value);
+  const currentValue = options?.find(
+    (option) => String(option.value) === String(value),
+  );
```
la-nuta: Non-blocking. This was added in `645e6ad` to compensate for `value: String(value)`, which `71e3036` then reverted — utils now emits the same raw values master does, so this shared component sees identical inputs and the change no longer has a reason.

```suggestion
  const currentValue = options?.find((option) => option.value === value);
```

la-nuta (dismissed):
The deliverable is master's line 1 of `ui/utils/index.ts`, and it isn't in this diff: #10806 added `/* eslint-disable @typescript-eslint/no-explicit-any */` after this branch's merge base, so it survives the merge and is dead once this lands:

```
1:1  error  Unused eslint-disable directive (no problems were reported from '@typescript-eslint/no-explicit-any')
```

Please delete it once you've picked up master (merge or rebase, whichever you prefer). The inline note below is the other thing that fails lint on current master.

AlexeyGirin:
@copilot please make fix based on review comment

la-nuta (approved):
Both asks are addressed in `19b21e7`. The line-1 `@typescript-eslint/no-explicit-any` disable is gone and the file lints clean on the post-merge shape (`--report-unused-disable-directives`, rc=0); the union is collapsed to `EnumSchemaLike`, so `prettier --check` passes. `Select.tsx` is back to byte-identical with master.

PR #10652, opened 2026-07-20
#10568 Refactor: Non-null assertion clean up from \packages\ketcher-core\src\application\render\renderStruct.ts` file

### LINKED ISSUE #10568: Refactor: Non-null assertion clean up from `packages\ketcher-core\src\application\render\renderStruct.ts` file
**Problem:** 
Non-null assertion.

**Why is this an issue?**
The `non-null assertion` in TypeScript, that is, `value!`, isn't bad in itself, but rather because it disables the check exactly where the compiler was trying to warn you about the risk. After the `!` , TypeScript takes your word for it, and if the invariant is violated, an error will occur at runtime, usually in the form of "Cannot read properties of undefined."

**Main disadvantages:**

1. It makes types less fair. Code appears safe, although the real guarantee exists only in the author's mind.
2. It doesn't handle refactoring well. Today, an invariant is correct, but tomorrow, adjacent logic changes, and `!` becomes a silent crash.
3. It breaks local readability. The reader can't understand why something is definitely not `null` or `undefined`.
4. It masks a weak type model. `!` often appears where it would be better to clarify the type, add a guard, or rewrite the API.

**How can it be fixed?**
Is it always possible to avoid it? No, it's not always wise. But in most cases, it's possible and worthwhile.

1. Early exit:
````ts
const atom = restruct.atoms.get(id);
if (!atom) return null;
return atom.a.pp;
````

2. Explicit assertion function:
````ts
function assertDefined<T>(value: T | null | undefined): asserts value is T {
  if (value == null) throw new Error('Expected value to be defined');
}
````

3. A more precise type or a different API contract to ensure undefined never appears in normal flow.
4. A local code refactoring to ensure that verification and usage are close together, rather than separated.

When is `!` usually acceptable:

1. There's a strong domain invariant, and the violation really means a broken data structure, not a user scenario.
2. TypeScript can't infer what's already guaranteed by an external system or lifecycle.
3. The alternative makes the code noticeably worse, and the space is well-isolated.

A quick rule of thumb: `!` isn't forbidden, but it's a "debt of trust." If it can be replaced with a check or a more precise contract without introducing unnecessary noise into the code, it's better to do so. If it's not allowed, only leave it in places where the invariant is truly strong and clear to the reader.

**Problem locations:**
packages\ketcher-core\src\application\render\renderStruct.ts

### PR DESCRIPTION
**Pull request by Makarov-Ivan**
Sgroup attachment points always reference an existing atom in the same struct, so the lookup can never actually be undefined at runtime. Replace the non-null assertion with a guard clause that throws a descriptive error if that invariant is ever violated, keeping the same behavior for valid input while documenting the assumption instead of silently asserting it.

Fixes #10568

## Check list
- [ ] unit-tests written
- [ ] e2e-tests written
- [ ] documentation updated

### CODE REVIEW

File: packages/ketcher-core/src/application/render/renderStruct.ts [hunk 1]
```diff
@@ -165,8 +165,15 @@ function convertAllSGroupAttachmentPointsToRGroupAttachmentPoints(
...
-      const attachmentPointAtom = struct.atoms.get(attachmentPoint.atomId)!;
+      const attachmentPointAtom = struct.atoms.get(attachmentPoint.atomId);
+      if (!attachmentPointAtom) {
+        // An sgroup attachment point must always reference an existing atom
+        // in the same struct; a missing atom indicates a programming error
+        // (e.g. the atom was removed without updating the attachment point).
+        throw new Error(
+          `Atom with id ${attachmentPoint.atomId} not found in struct while converting sgroup attachment points`,
+        );
+      }
```
svvald: Let's wait until https://github.com/epam/ketcher/pull/10660 is merged and use `getOrThrow` utility here

svvald (changes requested):

packages/ketcher-core/src/application/render/renderStruct.ts
AlexeyGirin: @copilot please make fix based on comment since #10660 has been merged

AlexeyGirin:

svvald (approved):

PR #10660, opened 2026-07-20
#10559 Refactor: Non-null assertion clean up from \packages\ketcher-react\src\script\editor\shared\closest.ts` file

### LINKED ISSUE #10559: Refactor: Non-null assertion clean up from `packages\ketcher-react\src\script\editor\shared\closest.ts` file
**Problem:** 
Non-null assertion.

**Why is this an issue?**
The `non-null assertion` in TypeScript, that is, `value!`, isn't bad in itself, but rather because it disables the check exactly where the compiler was trying to warn you about the risk. After the `!` , TypeScript takes your word for it, and if the invariant is violated, an error will occur at runtime, usually in the form of "Cannot read properties of undefined."

**Main disadvantages:**

1. It makes types less fair. Code appears safe, although the real guarantee exists only in the author's mind.
2. It doesn't handle refactoring well. Today, an invariant is correct, but tomorrow, adjacent logic changes, and `!` becomes a silent crash.
3. It breaks local readability. The reader can't understand why something is definitely not `null` or `undefined`.
4. It masks a weak type model. `!` often appears where it would be better to clarify the type, add a guard, or rewrite the API.

**How can it be fixed?**
Is it always possible to avoid it? No, it's not always wise. But in most cases, it's possible and worthwhile.

1. Early exit:
````ts
const atom = restruct.atoms.get(id);
if (!atom) return null;
return atom.a.pp;
````

2. Explicit assertion function:
````ts
function assertDefined<T>(value: T | null | undefined): asserts value is T {
  if (value == null) throw new Error('Expected value to be defined');
}
````

3. A more precise type or a different API contract to ensure undefined never appears in normal flow.
4. A local code refactoring to ensure that verification and usage are close together, rather than separated.

When is `!` usually acceptable:

1. There's a strong domain invariant, and the violation really means a broken data structure, not a user scenario.
2. TypeScript can't infer what's already guaranteed by an external system or lifecycle.
3. The alternative makes the code noticeably worse, and the space is well-isolated.

A quick rule of thumb: `!` isn't forbidden, but it's a "debt of trust." If it can be replaced with a check or a more precise contract without introducing unnecessary noise into the code, it's better to do so. If it's not allowed, only leave it in places where the invariant is truly strong and clear to the reader.

**Problem locations:**
packages\ketcher-react\src\script\editor\shared\closest.ts

### PR DESCRIPTION
**Pull request by Makarov-Ivan**
All assertions guarded lookups that are true structural invariants (a bond's begin/end atoms, its half-bond, an ReSGroup's item). Replaced them with explicit guard clauses that throw a descriptive error if the invariant is ever violated, which both narrows the type via control flow and documents the assumption instead of silently asserting non-null.

Fixes #10559

## Check list
- [ ] unit-tests written
- [ ] e2e-tests written
- [ ] documentation updated

### DIFF
--- a/packages/ketcher-react/src/script/editor/shared/closest.ts
+++ b/packages/ketcher-react/src/script/editor/shared/closest.ts
@@ -23,6 +23,9 @@ import {
   IMAGE_KEY,
   MULTITAIL_ARROW_KEY,
   FunctionalGroup,
+  getOrThrow,
+  atomsForBondNotFoundMessage,
+  entityNotFoundMessage,
 } from 'ketcher-core';
 import type { ClosestItem, ClosestItemWithMap } from './closest.types';
 
@@ -189,15 +192,16 @@ function findClosestBond(
       return;
     }
 
-    const beginAtom = restruct.atoms.get(bond.b.begin);
-    const endAtom = restruct.atoms.get(bond.b.end);
-    if (!beginAtom || !endAtom) {
-      // A bond's begin/end atoms are always present in the same restruct;
-      // their absence would indicate a corrupted structure.
-      throw new Error(
-        `Atom(s) for bond ${bid} not found: begin=${bond.b.begin}, end=${bond.b.end}`,
-      );
-    }
+    const beginAtom = getOrThrow(
+      restruct.atoms,
+      bond.b.begin,
+      atomsForBondNotFoundMessage(bid, bond.b.begin, bond.b.end),
+    );
+    const endAtom = getOrThrow(
+      restruct.atoms,
+      bond.b.end,
+      atomsForBondNotFoundMessage(bid, bond.b.begin, bond.b.end),
+    );
     const p1 = beginAtom.a.pp;
     const p2 = endAtom.a.pp;
 
@@ -215,12 +219,11 @@ function findClosestBond(
       closestBondCenter = bid;
     }
 
-    const hb = restruct.molecule.halfBonds.get(bond.b.hb1 as number);
-    if (!hb) {
-      // Every bond has a corresponding half-bond created alongside it;
-      // its absence would indicate a corrupted structure.
-      throw new Error(`Half-bond ${bond.b.hb1} for bond ${bid} not found`);
-    }
+    const hb = getOrThrow(
+      restruct.molecule.halfBonds,
+      bond.b.hb1 as number,
+      `Half-bond ${bond.b.hb1} for bond ${bid} not found`,
+    );
     const dir = hb.dir;
     const norm = hb.norm;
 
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: packages/ketcher-react/src/script/editor/shared/closest.ts [hunk 1]
```diff
@@ -189,10 +189,17 @@ function findClosestBond(
...
-    const p2 = restruct.atoms.get(bond.b.end)!.a.pp;
+    const beginAtom = restruct.atoms.get(bond.b.begin);
+    const endAtom = restruct.atoms.get(bond.b.end);
+    if (!beginAtom || !endAtom) {
+      // A bond's begin/end atoms are always present in the same restruct;
+      // their absence would indicate a corrupted structure.
+      throw new Error(
+        `Atom(s) for bond ${bid} not found: begin=${bond.b.begin}, end=${bond.b.end}`,
+      );
+    }
```
svvald: We have such Map get checks all over the code and most of them use `assert` library. I'd stick with that but also introduce a more uniform approach for further refactoring:
1. Currently we do not enforce any message to be passed into assert which leading to unclear error messages especially on production build
2. We have to repeat these check over and over again

In this pull-request let's introduce some `getOrThrow` (name is just an idea, you can come up with a better one) utility:
```
import assert from 'assert';

export function getOrThrow<K, V>(
  map: Map<K, V>,
  key: K,
  message: string,
): V {
  const value = map.get(key);
  assert(value !== undefined, message);
  return value;
}
```
Let's use it in this file and then will try to cover other assert cases later on.

Also maybe a good idea would be to create some set of pre-defined error messages (or error message construction functions if data is dynamic) to no duplicate and put the same string literal/string template again and again

svvald (changes requested):

packages/ketcher-react/src/script/editor/shared/closest.ts
Makarov-Ivan: all done as proposed

packages/ketcher-react/src/script/editor/shared/closest.ts
```diff
@@ -510,8 +536,12 @@ function findClosestFG(restruct: ReStruct, pos: Vec2, skip) {
       const cursorPosition = new Vec2(x, y);
 
       const dist = Vec2.dist(rectangleCenter, cursorPosition);
-      // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-      const { id } = reSGroup.item!;
+      const sGroupItem = reSGroup.item;
+      if (!sGroupItem) {
+        // A ReSGroup is always constructed with its underlying SGroup item.
+        throw new Error(`ReSGroup ${reSGroupId} has no item`);
+      }
```
svvald: Let's use `assert` here with this error message as well

svvald (changes requested):

AlexeyGirin:
@copilot please make fixes based on provided review comments

svvald (approved):

Issue #7455, opened 2025-07-25
Support monomer replacement with drag-and-drop from library


anna-stankovic:
## Background

This is an improvement on the drag and drop functionality #6404 that will allow users to replace a monomer (preset) in structures by dragging and dropping onto them.

Related issues:
Monomer insertion, #7925 
Bond establishment, #7926 

## Requirements

$\color{Red}{\textbf{Monomers replacing monomers}}$

1. When the user is dragging a single monomer from the library, and the distance between the mouse cursor and the monomer center is [TBD] px the monomer on canvas should change its appearance (see mockups).

2. While the monomer on canvas has a different appearance, on mouse drop, the monomer from the library should replace the monomer on canvas.

3. All possible bonds of the original monomer should be established with the new monomer.

3.1. If the dragged monomer lacks an Rn AP that the original monomer had a bond with, the user should get a warning message in a modal window. 
The title should be: Deletion of bonds
The text should be: Some bonds will get deleted suring replacement. Do you with to proceed.
The options should be: Cancel (default) and Yes.
The modal window should be the same one used in sequence mode:
<img width="243" height="84" alt="Image" src="https://github.com/user-attachments/assets/ee181259-e8a3-4b45-be8d-50a8fbfc5ff3" />
If Yes is chosen, the replacement happens, and some bonds get deleted.

> The monomer can never replace a preset -- dropping an amino-acid onto preset component will just replace that component:

> <img width="367" height="322" alt="Image" src="https://github.com/user-attachments/assets/be4fab2a-fac3-45c9-8177-b841da1a6b7d" />

> :warning: Not a mockup, for illustration purposes only! :warning:

$\color{Red}{\textbf{Presets replacing presets}}$

4. When the user is dragging a preset from the library, and the distance between the mouse cursor and a center of any of the monomers in the preset of the same geometry (see req. 6) is [TBD] px away, the preset on canvas should change its appearance (see mockups).

5. While the preset on canvas has a different appearance, on mouse drop, the preset from the library should replace the preset on canvas.

6. Presets of the same geometry are defined as presets that contain the same components, and the same position of the phosphate.

7. All possible bonds of the original sugar/phosphate/base should be established with the new sugar/phosphate/base.

7.1. If one of the monomers doesn't have appropriate APs, the warning message from req. 3.1. should be shown.

> <img width="1217" height="529" alt="Image" src="https://github.com/user-attachments/assets/dfa59485-9b6f-4c9c-82ce-001b9bf3857d" />

[...]

PR #10667, opened 2026-07-21
#7455 - Support monomer replacement with drag-and-drop from library

vanoprenko:
## Summary

- Enable users to replace an existing canvas monomer by dragging a library monomer onto it, preserving all bond connections
- Add visual feedback (blue/red highlight circle) during drag to indicate valid/invalid replacement targets
- Validate attachment point compatibility before allowing replacement (replacement monomer must have all currently-bonded attachment points of the target)

## Details

- Hit-testing uses 0.5 angstrom radius to find closest monomer under cursor during drag
- Replacement is executed via the existing `replaceMonomer` command, supporting undo/redo out of the box
- Original placement behavior (drop on empty canvas) is preserved
- Highlight cleanup happens automatically on drag end and monomer removal

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/packages/ketcher-core/src/application/editor/Editor.ts
+++ b/packages/ketcher-core/src/application/editor/Editor.ts
@@ -1187,7 +1187,31 @@ export class CoreEditor {
     );
 
     this.events.setLibraryItemDragState.add((state: LibraryItemDragState) => {
+      // Clear previous replacement target highlight
+      if (this.libraryItemDragState?.hoveredMonomer) {
+        const previousRenderer =
+          this.libraryItemDragState.hoveredMonomer.renderer;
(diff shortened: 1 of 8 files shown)


vanoprenko:
AI-generated implementation -- can be used as a basis for the final solution.

Known issues:

- Dropping area is currently detected  based on predefined radius around the center of the target monomer (0.5 angstroms). Should be migrated to existing approach used for highlighting a monomer on mouse hover (using the monomer shape area)
- Dragging presets is not supported
- Scaling issue (may be caused by unrelated changes in master). All elements in macromolecules mode (toolbars, library) shrank in size

PR #10685, opened 2026-07-22
#3840 - Buttons are collapsed in the top toolbar because hidden buttons are not taken into account

### LINKED ISSUE #3840: Buttons are collapsed in the top toolbar because hidden buttons are not taken into account
**Steps to Reproduce**
1. Configure the url with some hiddenControls in the parameters. For example: https://lifescience.opensource.epam.com/KetcherDemo/index.html?api_path=/v2&hiddenControls=text,layout,arom,dearom,check,analyse,recognize,miew,enhanced-stereo,settings,help,about 
2. Reduce the window width below 690px
3. Notice that the `Calculate CIP` button is collapsed under the `Clean` button while there is a lot of empty space left. 

**Actual behavior**
In the following screenshot, notice that the stereo button is in a sub-menu while there is lots of place to accommodate for it in the main toolbar
image

This is controlled by this line of code in TopToolbar.tsx component: https://github.com/epam/ketcher/blob/b70f39a78424af3faf9a82010b6a5b68b1ab372c/packages/ketcher-react/src/script/ui/views/toolbars/TopToolbar/TopToolbar.tsx#L205
A `const collapseLimit = 650` is used (L68), but the hidden buttons in the TopToolbar are not accounted for 
 

**Expected behavior**
The hidden buttons should be taken into account when computing the available width to decide if the group of buttons should collapse

**Screenshots**
see above

 - OS: MacOS
 - Browser Chrome
 - Version 118

**Ketcher version**
2.15

**Test case**
Add test case number (in EPMLSOPKET-* format) if bug was found during test case run.

**Additional context**

armankarapetyan-svg:

Updated TopToolbar.tsx so toolbar collapse logic dynamically accounts for hidden controls.

https://github.com/user-attachments/assets/648535a1-93cf-4b77-b7da-cc4cea204c92

Updated collapse threshold calculation (TopToolbar.tsx) to subtract the width of hidden controls:
* Collapse Limit = Base Limit + Custom Buttons Width - Hidden Controls Width
Controls like Calculate CIP now remain visible as long as real space allows, rather than collapsing based on a static baseline.

## Check list
- [ ] unit-tests written
- [ ] e2e-tests written
- [ ] documentation updated

### DIFF
--- a/packages/ketcher-react/src/script/ui/views/toolbars/TopToolbar/TopToolbar.tsx
+++ b/packages/ketcher-react/src/script/ui/views/toolbars/TopToolbar/TopToolbar.tsx
@@ -74,6 +74,30 @@ export interface PanelProps {
 
 const collapseLimit = 650;
 const CUSTOM_BUTTON_ADDITIONAL_WIDTH = 40;
+const DEFAULT_COLLAPSE_BUTTON_WIDTH = 24;
+
+const COLLAPSE_WIDTH_BY_BUTTON: Record<string, number> = {
+  clear: DEFAULT_COLLAPSE_BUTTON_WIDTH,
+  open: DEFAULT_COLLAPSE_BUTTON_WIDTH,
+  save: DEFAULT_COLLAPSE_BUTTON_WIDTH,
+  copies: DEFAULT_COLLAPSE_BUTTON_WIDTH,
+  paste: DEFAULT_COLLAPSE_BUTTON_WIDTH,
+  cut: DEFAULT_COLLAPSE_BUTTON_WIDTH,
+  undo: DEFAULT_COLLAPSE_BUTTON_WIDTH,
+  redo: DEFAULT_COLLAPSE_BUTTON_WIDTH,
+  arom: DEFAULT_COLLAPSE_BUTTON_WIDTH,
+  dearom: DEFAULT_COLLAPSE_BUTTON_WIDTH,
+  layout: DEFAULT_COLLAPSE_BUTTON_WIDTH,
+  clean: DEFAULT_COLLAPSE_BUTTON_WIDTH,
+  cip: DEFAULT_COLLAPSE_BUTTON_WIDTH,
+  check: DEFAULT_COLLAPSE_BUTTON_WIDTH,
+  analyse: DEFAULT_COLLAPSE_BUTTON_WIDTH,
+  miew: DEFAULT_COLLAPSE_BUTTON_WIDTH,
+  settings: DEFAULT_COLLAPSE_BUTTON_WIDTH,
+  help: DEFAULT_COLLAPSE_BUTTON_WIDTH,
+  about: DEFAULT_COLLAPSE_BUTTON_WIDTH,
+  fullscreen: DEFAULT_COLLAPSE_BUTTON_WIDTH,
+};
 
 const ControlsPanel = styled('div')`
   display: flex;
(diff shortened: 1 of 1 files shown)


packages/ketcher-react/src/script/ui/views/toolbars/TopToolbar/TopToolbar.tsx
```diff
@@ -74,6 +74,30 @@ export interface PanelProps {
 
 const collapseLimit = 650;
 const CUSTOM_BUTTON_ADDITIONAL_WIDTH = 40;
+const DEFAULT_COLLAPSE_BUTTON_WIDTH = 24;
```
svvald, line 77: This hardcoded value won't work for all breakpoints, according to `TopToolbarIconButton.ts` it could be either 24/32/40px depending on the screen size

packages/ketcher-react/src/script/ui/views/toolbars/TopToolbar/TopToolbar.tsx
```diff
@@ -74,6 +74,30 @@ export interface PanelProps {
...
+  clean: DEFAULT_COLLAPSE_BUTTON_WIDTH,
+  cip: DEFAULT_COLLAPSE_BUTTON_WIDTH,
+  check: DEFAULT_COLLAPSE_BUTTON_WIDTH,
+  analyse: DEFAULT_COLLAPSE_BUTTON_WIDTH,
+  miew: DEFAULT_COLLAPSE_BUTTON_WIDTH,
+  settings: DEFAULT_COLLAPSE_BUTTON_WIDTH,
+  help: DEFAULT_COLLAPSE_BUTTON_WIDTH,
+  about: DEFAULT_COLLAPSE_BUTTON_WIDTH,
+  fullscreen: DEFAULT_COLLAPSE_BUTTON_WIDTH,
+};
```
svvald, line 100: This list is quite verbose and yet it's not complete, some top toolbar tools are missing here such as `explicit-hydrogens`

Overall you don't need a map here, you can just use a Set of collapsible buttons and the use it below as
```
hiddenControlsWidth = [...new Set(hiddenButtons)].filter(id => YourCreatedSetHere.has(id)).length * DEFAULT_COLLAPSE_BUTTON_WIDTH
```

svvald (commented):
Overall I think this approach is pretty fragile and being based on many shaky assumptions – fixed button width, list of collapsible buttons which we have to update if any new tool will appear

I'd prefer an approach which would track the actual toolbar's unclamped content with the resize observer and compare it against available width

Here's the one of the ideas Claude has suggested to me that you could explore:

### The perfect approach: measure, don't guess

Instead of maintaining a table of pretend pixel widths, render the toolbar's *fully-expanded* content once, off-screen, and let the browser tell you its real width — then compare that to the actual available width to decide `isCollapsed`. No per-button constants, nothing to keep in sync when a button, icon, breakpoint, or CSS gap changes.

Concretely, in `TopToolbar.tsx`:

1. **Add a hidden mirror node** containing the same collapsible groups (`ExternalFuncControls` with `isCollapsed={false}`, `CustomButtons` with `isCollapsed={false}`), rendered with `position: absolute; visibility: hidden; pointer-events: none;` (kept out of flow, but still laid out so it has a real width) — plus the always-visible groups if you want one number for the whole bar, or just the collapsible slice if you want to isolate it.
2. **Observe that mirror** with the same `useResizeObserver` hook already used for the container — it only needs to re-measure when `hiddenButtons`/`customButtons`/`disabledButtons` change, not on every scroll frame.
3. **Derive `isCollapsed`** as `containerWidth < mirrorNaturalWidth` (plus the fixed width of the non-collapsible siblings if you split the mirror).
4. **Delete** `collapseLimit`, `CUSTOM_BUTTON_ADDITIONAL_WIDTH`, `COLLAPSE_WIDTH_BY_BUTTON`, `DEFAULT_COLLAPSE_BUTTON_WIDTH` entirely.

svvald (changes requested):

packages/ketcher-react/src/script/ui/views/toolbars/TopToolbar/TopToolbar.tsx
armankarapetyan-svg: Resolved with the last commit

packages/ketcher-react/src/script/ui/views/toolbars/TopToolbar/TopToolbar.tsx
armankarapetyan-svg: Resolved with the last commit

PR #10687, opened 2026-07-22
#2947 - After entering zero or deleting the previous value, zero is not deleted in the "Repeat count" field 

### LINKED ISSUE #2947: After entering zero or deleting the previous value, zero is not deleted in the "Repeat count" field
**Steps to Reproduce**
1.	Launch "Ketcher"
2.	Select any structure from "Templates toolbar"
3.	Click on "S-group" button
4.	Click on the selected structure on the canvas
5.	In the window that appears, in the drop-down list "Type" select "Multiple group"
6.	Enter 0 in the field "Repeat count" 
7.	Delete 0 in the field "Repeat count" 

**Actual behavior**
Zero is displayed in the field "Repeat count" and the field itself glows red

**Expected behavior**
The field "Repeat count" is empty and is lit in gray, the "Apply" button is not active

**Screenshots**
Bug EPMLSOPKET-16893
Bug EPMLSOPKET-16893 SA

 - Browser: Сhrome

**Ketcher version** 2.13.0-rc.1 (Remote, Standalone)

**Test case**
EPMLSOPKET-16893, «S-Group tool – Check validations on limitation»

armankarapetyan-svg:

https://github.com/user-attachments/assets/68955eb2-899d-478c-888c-5341e21ca58c

Fix S-Group Repeat count snapping to 0 when cleared

- Updated Input.tsx to treat empty fields as undefined instead of coercing to 0.
- Updated form.tsx to omit undefined values from form state.
- Added regression unit tests in sgroup.test.tsx.

## Check list
- [ ] unit-tests written
- [ ] e2e-tests written
- [ ] documentation updated

### DIFF
--- a/packages/ketcher-react/src/script/ui/component/form/Input/Input.tsx
+++ b/packages/ketcher-react/src/script/ui/component/form/Input/Input.tsx
@@ -89,7 +89,12 @@ GenericInput.val = function (ev, schema) {
   const value = isNumber ? input.value.replace(/,/g, '.') : input.value;
 
   if (isInteger) {
-    return Number(value) || 0;
+    if (value.trim() === '') {
+      return undefined;
+    }
+
+    const parsedValue = Number(value);
+    return Number.isNaN(parsedValue) ? value : parsedValue;
   }
 
   // When the value can be a float the validation is passed to the parent component
--- a/packages/ketcher-react/src/script/ui/component/form/Input/Input.tsx
+++ b/packages/ketcher-react/src/script/ui/component/form/Input/Input.tsx
@@ -88,14 +88,14 @@ GenericInput.val = function (ev, schema) {
 
   const value = isNumber ? input.value.replace(/,/g, '.') : input.value;
 
-  if (isInteger) {
-    if (value.trim() === '') {
-      return undefined;
-    }
-
-    const parsedValue = Number(value);
-    return Number.isNaN(parsedValue) ? value : parsedValue;
+  if (isNumber && value.trim() === '') {
+    return;
   }
+
+if (isInteger) {
+  const parsedValue = Number(value);
+  return Number.isNaN(parsedValue) ? value : parsedValue;
+}
 
   // When the value can be a float the validation is passed to the parent component
   // because it's more complicated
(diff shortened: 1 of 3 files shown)


packages/ketcher-react/src/script/ui/component/form/form/form.tsx
```diff
@@ -187,12 +187,19 @@ class Form extends Component<FormProps> {
...
         this.updateState(newState);
         if (onChange) onChange(val);
       },
       onExtraChange: (val: unknown) => {
-        const newState = { ...this.props.result, [extraName ?? '']: val };
+        const extraFieldName = extraName ?? '';
+        const newState =
+          val === undefined
+            ? omit(this.props.result, [extraFieldName])
+            : { ...this.props.result, [extraFieldName]: val };
```
svvald: It's duplicated twice now, already enough to have an util function

packages/ketcher-react/src/script/ui/component/form/Input/Input.tsx
```diff
@@ -89,7 +89,12 @@ GenericInput.val = function (ev, schema) {
   const value = isNumber ? input.value.replace(/,/g, '.') : input.value;
 
   if (isInteger) {
-    return Number(value) || 0;
+    if (value.trim() === '') {
+      return undefined;
+    }
```
svvald: Is it correct though to put it inside `isInteger` check only? I think we should have a common check first like
```
if (isNumber && value.trim() === '') {
    return;
}
```

svvald (changes requested):
Currently these changes may lead to unexpected changes - this input logic is used by many form fields in our app, if some of them are not marked as `required`, clearing them won't make the form invalid and will allow to commit the empty/undefined value which might break the existing functionality if value consumer relies on only getting the numeric value

@AlexeyGirin FYI, I'm not sure how to tackle that, do we have EACH form/form field validation covered by autotests to safely merge it??

packages/ketcher-react/src/script/ui/component/form/form/form.tsx
armankarapetyan-svg: Good point, agreed. I removed the duplication by extracting a small helper to build the next form state (undefined -> omit field, otherwise set value), and reused it in both onChange and onExtraChange

PR #10688, opened 2026-07-22
#4451 – Macro: Unable to move sequence (on Flex mode) by holding ALT+grabbing single monomer is impossible

### LINKED ISSUE #4451: Macro: Unable to move sequence (on Flex mode) by holding Ctrl+grabbing single monomer is impossible
**Steps to Reproduce**

1. Switch to Macro (Flex mode)
3. Hold Ctrl button and drag any monomer (by pressing and holding left mouse button) and move it
image

**Actual behavior**
Nothing happens.
image

**Expected behavior**
System switches to Select Rectangle mode.
All monomers (and bonds) become selected while user grabbing it.
Sequence follows mouse cursor while user holding right mouse button
image

**Versions:**
Windows 10

AvetBadalyan:
`SelectBase.mousedownEntity()` branches on `shiftKey`/`modKey`(Ctrl/Cmd)/`altKey` to decide
what a click selects. The "select the whole chain" gesture was only wired up for
`BaseSequenceItemRenderer` (Sequence mode, via Ctrl/Cmd) — Flex mode's monomer renderers are
plain `BaseRenderer` and never satisfied that guard, so holding a modifier and dragging a
Flex-mode monomer fell through every branch in the if/else chain. Critically,
`startMoveIfNeeded()` (which sets `this.mode = 'moving'`, required for the drag to move
anything at all) is only called in the first branch, so the tool's mode stayed `'standby'`
— matching the reported "nothing happens."

Fix: added a new branch for `altKey` on a plain monomer renderer (excluding
`BaseSequenceItemRenderer`, so Sequence mode's existing Ctrl/Cmd gesture stays untouched),
using `getAllConnectedMonomersRecursively` — a graph-traversal utility already present in
`domain/helpers/monomers.ts` for exactly this purpose but never called anywhere in the
codebase. Selects the whole connected chain (monomers + bonds between them, mirroring the
existing Sequence-mode branch's pattern) and starts the move, so the chain now follows the
cursor while Alt is held.

Note: the issue was originally reported/implemented against Ctrl+drag; the task description
was corrected to Alt+drag after the fix was first written, so the modifier key was switched
from `modKey` (Ctrl/Cmd) to `altKey`. Sequence mode's own Ctrl/Cmd "select whole chain"
gesture is a separate, pre-existing feature and is unaffected either way.

Verified in a real browser using the repro `.ket` file: reproduced the bug on the unpatched
code (Alt+drag on a Flex-mode monomer did nothing), then confirmed the fix — all monomers
and bonds in the chain (including a second chain linked via a hydrogen bond) highlight as
selected and move together as a group while Alt+dragging.

## Check list
- [ ] unit-tests written
- [ ] e2e-tests written
- [ ] documentation updated


packages/ketcher-core/src/application/editor/tools/select/SelectBase.ts
```diff
@@ -345,6 +342,38 @@ abstract class SelectBase implements BaseTool {
           drawingEntities,
         ),
       );
+    } else if (
+      altKey &&
+      !(renderer instanceof BaseSequenceItemRenderer) &&
```
rrodionov91: Please use ```this.editor.mode.modeName !== 'sequence-layout-mode'``` instead of ```renderer instanceof BaseSequenceItemRenderer```

rrodionov91 (dismissed):

packages/ketcher-core/src/application/editor/tools/select/SelectBase.ts
AvetBadalyan: re-check please

svvald (approved):

PR #10689, opened 2026-07-22
3485 - Using touchpad to zoom is too sharp

### LINKED ISSUE #3485: Using touchpad to zoom is too sharp
If zooming in or zooming out using the touchpad, the movements look very sharp

https://github.com/epam/ketcher/assets/139227567/e4731cec-1ace-4e32-b594-bf58354f1531

_Originally posted by @ViktoriaTkacheva in https://github.com/epam/ketcher/issues/3412#issuecomment-1767202512_

antoniabajic-epam:
## Summary
After drawing some stuff on the canvas, using the touchpad/trackpad to zoom in/out and the zooming is too sharp

## Changes

**StructEditor.tsx**
Problem: The zooming is too sharp
Solution: Make the zooming not that sharp

## Files Modified

**ketcher-react**
- `StructEditor.tsx` — add smoother zoom on touchpad

https://github.com/user-attachments/assets/b9ca124b-cb0a-4a7d-beb3-8e0c1c62465c

## Check list
- [ ] unit-tests written
- [ ] e2e-tests written
- [ ] documentation updated


svvald (commented):
While this approach is sound and addresses the original issue, currently there is some divergence between Micro mode and Macro mode zoom functionalities – Macro mode is not step-based and it works smoothly for both mouse and touchpad as is.

Could you please explore the way it implemented in Macro mode and try to unify it with the Micro mode zoom? Macro mode uses D3.js for rendering and zoom is also a part of it and Micro mode uses Raphael for rendering but maybe we could still use just D3.js zoom functionality separately and make it work with Micro mode? If it's not possible, maybe we could derive some calculations from there?

antoniabajic-epam:
> While this approach is sound and addresses the original issue, currently there is some divergence between Micro mode and Macro mode zoom functionalities – Macro mode is not step-based and it works smoothly for both mouse and touchpad as is.
> 
> Could you please explore the way it implemented in Macro mode and try to unify it with the Micro mode zoom? Macro mode uses D3.js for rendering and zoom is also a part of it and Micro mode uses Raphael for rendering but maybe we could still use just D3.js zoom functionality separately and make it work with Micro mode? If it's not possible, maybe we could derive some calculations from there?

@svvald I tested the suggested approach, but it introduced a regression: the zoom indicator is now stuck at 100% and fails to reflect any zoom-in or zoom-out actions. It looks like the state is no longer reacting to the scale changes. I'll need to dig deeper into and find best solution

PR #10693, opened 2026-07-23
#10551 Refactor: Explicit/implicit Any type clean up from packages\ketcher-react\src\script\editor\tool\sgroup.ts file

### LINKED ISSUE #10551: Refactor: Explicit/implicit Any type clean up from `packages\ketcher-react\src\script\editor\tool\sgroup.ts` file
**Problem:** 
Unspecified any.

**Why is this an issue?**

**How can it be fixed?**

**Problem locations:**
packages\ketcher-react\src\script\editor\tool\sgroup.ts

LashaDevLasha:
Replace all `any` types with proper TypeScript types: Atom, Bond, Struct, ReStruct, ClosestItemWithMap, Pool<FunctionalGroup>, and a new SGroupDialogResult interface for the s-group dialog result. Also fixes implicit any on event params and lassoHelper.running/end call-site arg that was silently ignored at runtime.

## Check list
- [ ] unit-tests written
- [ ] e2e-tests written
- [ ] documentation updated

### DIFF
--- a/packages/ketcher-react/src/script/editor/tool/sgroup.ts
+++ b/packages/ketcher-react/src/script/editor/tool/sgroup.ts
@@ -16,6 +16,10 @@
 
 import {
   type Pool,
+  type Atom,
+  type Bond,
+  type Struct,
+  type ReStruct,
   Pile,
   SgContexts,
   checkOverlapping,
@@ -34,6 +38,12 @@ import { selMerge } from './select/select.helpers';
 import type { Editor, Selection } from '../Editor';
 import type { Tool } from './Tool';
 import { filterNotPartOfSuperatomWithoutLabel } from './helper/filterNotInCollapsedSGroup';
+import type { ClosestItemWithMap } from '../shared/closest.types';
+
+interface SGroupDialogResult {
+  type?: string;
+  attrs: { context?: string; expanded?: boolean; [key: string]: unknown };
+}
 
 const searchMaps = [
   'atoms',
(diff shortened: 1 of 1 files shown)


packages/ketcher-react/src/script/editor/tool/sgroup.ts
```diff
@@ -271,34 +281,41 @@ class SGroupTool implements Tool {
     }
   }
 
-  mousemove(event) {
-    if (this.lassoHelper.running(event)) {
+  mousemove(event: MouseEvent) {
```
MariiaNebesnova: do we really need to define `event` as a `MouseEvent`? Why not `PointerEvent` and get rid of type cast at line 291?

LashaDevLasha: Without the explicit type, `event` could be treated as `any` if TypeScript cannot infer its type from the context. In that case, we would lose type checking for the event object. The type was added to make the expected event type clear and help TypeScript catch possible issues. If this type is not needed here or TypeScript already infers it correctly, please let me know and it can be removed

MariiaNebesnova: you're right, type is needed here. My question is why exactly `MouseEvent` instead of `PointerEvent`?

LashaDevLasha: yeah you are right, Updated

MariiaNebesnova (approved):

PR #10713, opened 2026-07-27
#10543 – Refactor: Explicit/implicit Any type clean up from packages\ketcher-macromolecules\src\components\modal\Open\Open.tsx file

AvetBadalyan:

`Open.tsx` had three callbacks relying on implicit `any` parameters (no type annotation, so TypeScript couldn't infer anything from context): `showParsingError`, the module-level `errorHandler`, and the file-load `onLoad` callback.

**Fix:**
- `showParsingError(stringError)` → typed `stringError: string`, matching `normalizeError(error).message: string` (its only non-literal call site).
- `errorHandler(error)` → typed `error: string`, matching the already-existing `OpenOptionsProps.errorHandler: (err: string) => void` declaration in `OpenOptions.tsx` — reused rather than re-declared.
- `onLoad(fileContent)` → typed `fileContent: string`, with the actual type-safety enforced at the source: `fileOpener.ts`'s `throughFileReader` now has an explicit `Promise<string>` return type and validates its `FileReader.result` (typed `string | ArrayBuffer | null`) with a real `typeof content !== 'string'` runtime check before resolving, rejecting otherwise. This replaces an earlier version of this fix that instead typed `onLoad`'s parameter as `unknown` and cast it with `as string` at the point of use — flagged in review (see below) as not a good idea, since the cast just moved the type-safety gap to the call site instead of closing it at the source.

**Review feedback addressed:** per @MariiaNebesnova's comment on the `onLoad` line, moved the type check into `throughFileReader` itself rather than casting `unknown` to `string` downstream, so `FileOpener`'s type (`(file: File) => Promise<string>`) is accurate end-to-end and `onLoad` receives a properly-typed `string` with no cast.

No behavior changes — type annotations only, plus the new (previously implicit) runtime guard in `throughFileReader`.

**Verified:**
- `tsc --noEmit` and `eslint` clean on both `Open.tsx` and `fileOpener.ts`.
- Confirmed `FileOpener`/`fileOpener()` have exactly one consumer (`Open.tsx`), so tightening the type doesn't affect anything else.
- Full `ketcher-macromolecules` Jest suite passes.
- Live in browser: used the actual "browse from file" flow (not just paste) to open a real `.ket` file through the modified `throughFileReader` code path — content loaded correctly, no console/page errors.

## Check list
- [ ] unit-tests written
- [ ] e2e-tests written
- [ ] documentation updated


packages/ketcher-macromolecules/src/components/modal/Open/Open.tsx
```diff
@@ -357,8 +357,8 @@ const Open = ({ isModalOpen, onClose }: RequiredModalProps) => {
       windowContext.isKetcherFullscreenBeforeFilePicker = false;
     }
 
-    const onLoad = (fileContent) => {
-      setStructStr(fileContent);
+    const onLoad = (fileContent: unknown) => {
```
MariiaNebesnova: not a good idea to define `unknown` here because the further logic expects `string`. Instead of it please try to define proper styles for `FileOpener`. Probably it will require some additional type checks inside the `throughFileReader` or a generic type

AvetBadalyan: please re-check

MariiaNebesnova (approved):

PR #10715, opened 2026-07-27
3483- False selection of several atoms at the same time in the Extended Table

### LINKED ISSUE #3483: False selection of several atoms at the same time in the Extended Table
**Steps to Reproduce**
1.	Launch "Ketcher"
2.	Click on the Extended Table
3.	Select any atom and, holding Shift, select another one
4.	Click on the "Add" button
5.	Click on the canvas

**Actual behavior**
The Extended Table provides a misleading possibility of selecting two atoms at once. When adding, the last atom selected is displayed on the canvas

**Expected behavior**
In the Extended Table it should be possible to select only one atom

**Screenshots**

https://github.com/epam/ketcher/assets/139227567/092ebac3-8331-4ae8-bfbb-125f4fa83099

 - Browser: Сhrome
 - Version: (117.0.5938.152) x64

**Ketcher version** 2.15.0-rc.4

antoniabajic-epam:
## Summary
The Extended Table provides a misleading possibility of selecting two atoms at once. When adding, the last atom selected is displayed on the canvas

## Changes

**ButtonGenSet.tsx**
Problem: The Extended Table provides a misleading possibility of selecting two atoms at once. When adding, the last atom selected is displayed on the canvas
Solution: In the Extended Table it should be possible to select only one atom, added handle click when selecting atom

## Files Modified

**ketcher-react**
- `GenericGroups/components/ButtonGenSet/ButtonGenSet.tsx` — dded handle click when selecting atom, only one atom is possible to select

## Check list
- [ ] unit-tests written
- [ ] e2e-tests written
- [ ] documentation updated


packages/ketcher-react/src/script/ui/views/modal/components/ExtendedTable/components/GenericGroups/components/ButtonGenSet/ButtonGenSet.tsx
```diff
@@ -37,9 +37,19 @@ const ButtonGenSet = ({
...
+  const handleClick = (event: React.MouseEvent) => {
+    // Prevent Shift+Click from selecting multiple atoms
+    // Extended Table should only allow single atom selection
+    if (event.shiftKey) {
+      event.preventDefault();
+      return;
+    }
+    onAtomSelect(button.label);
+  };
+
```
svvald: It only addresses the symptom, not the real issue – our controls are pure buttons, not checkboxes also `selected` state in `ExtendedTable` is also a single string property, not a collection so multiple selection in fact is impossible (what exactly reflected when we try to add something on canvas - there is only one atom there)

What actually happens here is that we have identical styles for `selected` and `focused` states and when clicking an atom with Shift button pressed, `selected` property is changed but focus remains on the previously selected one. Just gating the Shift press is not a good solution - first, it disrupts the user experience, second, it also occurs with Ctrl press as well

The better approach would be just `blur` the event target

svvald (changes requested):

svvald (approved):

PR #10721, opened 2026-07-28
#7140 - Update the layout for CHEM-containing sequences

### LINKED ISSUE #7140: Update the layout for CHEM-containing sequences
**Background**

This is a ticket complementary to a [previous ticket](https://github.com/epam/ketcher/issues/7057).
In the previous ticket logic for adding CHEMs in sequence mode was changed.
In this ticket layout for both snake and sequence mode is addressed to better reflect changes.

**Requirements**

1. Terminal CHEMs (the ones in the beginning or the end of the chain connected only via one AP) should be in line with the chain regardless of the APs used.

<details>
<summary>Expand for a further explanation</summary>

Currently if the terminal CHEM is connected with an R2-R1 connection it will stay in the same line, but other combinations of APs (R1-R1, R1-R3...) result in the CHEM in a different line.

</details>

2. Internal CHEMs who use only two APs should be in the line with the chain on the left of it.

<details>
<summary>Expand for a further explanation</summary>

Image

</details>

3. CHEMs that use more than two APs should be layouted using the previous logic.

HelenaIsh:

Fixes how CHEM monomers are positioned within a chain:
1. Terminal CHEMs (start/end of a chain, connected via a single attachment point) are now laid out inline with the chain, regardless of which APs are used.
2. Internal CHEMs using exactly two APs are placed inline, to the left of the chain.
3. CHEMs using more than two APs keep the previous layout logic.

### Combined Layout Comparison

| Scenario | Mode | Master — before | This branch — after |
|---|---|---|---|
| **Terminal CHEM at the end of the chain, R2–R1**<br><br>Chain: `A—A—CHEM`<br>Used CHEM APs: `1` | **Sequence** | image | image |
| | **Snake** | image | image |
| **Terminal CHEM at the end of the chain, R2–R2**<br><br>Chain: `A—A—CHEM`<br>Used CHEM APs: `1` | **Sequence** | image | image |
| | **Snake** | image | image |
| **Terminal CHEM at the end of the chain, R2–R3**<br><br>Chain: `A—A—CHEM`<br>Used CHEM APs: `1` | **Sequence** | image | image |
| | **Snake** | image | image |
| **Terminal CHEM at the beginning of the chain, R1-R1**<br><br>Chain: `CHEM—A`<br>Used CHEM APs: `1` | **Sequence** | image | image |
| | **Snake** | image | image |
| **Internal CHEM with two used APs**<br><br>Chain: `A—A—CHEM—A—A`<br>Used CHEM APs: `R1` and `R2` | **Sequence** | image | image |
| | **Snake** | image | image |
| **Branched CHEM with more than two used APs**<br><br>Used CHEM APs: `R1`, `R2`, and `R3` | **Sequence** | image | image |
| | **Snake** | image | image |

## Check list
- [x] unit-tests written
- [ ] e2e-tests written
- [ ] documentation updated


packages/ketcher-core/src/domain/helpers/monomers.ts
```diff
@@ -159,7 +159,64 @@ export function isMonomerConnectedToR2RnaBase(monomer?: BaseMonomer) {
...
 }
 
+export function isChemMonomer(monomer?: BaseMonomer): boolean {
+  return isMonomerOfClass(monomer, KetMonomerClass.CHEM);
+}
+
+export function isLinearChem(monomer?: BaseMonomer): boolean {
+  return (
+    isChemMonomer(monomer) &&
+    (monomer as BaseMonomer).usedAttachmentPointsNamesList.length <= 2
```
svvald: I assume `as` casting here is only to address `undefined` case? If so, I think we could make a short-circuit return?
```
if (!monomer) {
  return false;
}
```

You won't need `as` casting afterwards, also `isChemMonomer` definition will always accept `BaseMonomer` rather than having it optional

packages/ketcher-core/src/domain/helpers/monomers.ts
```diff
@@ -183,6 +247,20 @@ export function getNextMonomerInChain(
...
 
+  if (isLinearChem(monomer)) {
+    const nextMonomer = getOrientedChemNeighbors(monomer).next;
+
+    if (
+      !nextMonomer ||
+      nextMonomer === firstMonomer ||
+      isMonomerConnectedToR2RnaBase(nextMonomer)
+    ) {
+      return undefined;
```
svvald: Nitpick: you can just have plain `return`, there is no much sense in explicit `return undefined` usually

svvald (approved):

packages/ketcher-core/src/domain/helpers/monomers.ts
HelenaIsh: yes, it was for `undefined` case. I fixed it

packages/ketcher-core/src/domain/helpers/monomers.ts
HelenaIsh: fixed

PR #10722, opened 2026-07-28
#8185 - Throw error when SDF preset is missing mandatory groupClass

AnzhelaMartsinovich:
## Summary
- Fixes epam/Indigo#3822: loading a preset (monomer group template) via `ketcher.updateMonomersLibrary(sdf, {format: 'sdf'})` without a `<groupClass>` field silently disappeared, with no error in the console and nothing added to the library.
- Root cause: Indigo's SDF→KET conversion drops the `monomerGroupTemplate` record entirely when `<groupClass>` is missing/invalid ([epam/Indigo#3161](https://github.com/epam/Indigo/issues/3161)), instead of surfacing an error. The resulting KET JSON never contains the group template, so the existing KET-level validation in `Editor.ts` has nothing to check.
- Added `assertValidMonomerGroupTemplatesInSdf` in `ketcher-core/src/application/ketcher.ts`, which parses the raw SDF data fields (`<type>`, `<groupClass>`, `<groupName>`) before the Indigo conversion runs, and throws a `MonomerLibraryConvertError` with a clear message when a `monomerGroupTemplate` record is missing a valid `<groupClass>` (must equal `RNA`).
- Made `class` mandatory (was optional) on `IKetMonomerGroupTemplate` in `formatters/types/ket.ts`, matching the Indigo spec requirement that `groupClass` is mandatory.

## Check list
- [ ] unit-tests written
- [ ] e2e-tests written
- [ ] documentation updated


svvald (commented):
We shouldn't parse and validate SDF on Ketcher side, it's Indigo's concern which should return an error if provided SDF is invalid

Issue #5942, opened 2024-11-07
Update the representation of sense and antisense chains in sequence mode (1/2)


ljubica-milovic:
## Background
This ticket represents an evolution of a previous one ([#5925](https://github.com/epam/ketcher/issues/5925)) addressing the layout of sense and antisense chains in sequence mode. Much of the logic is shared between this ticket, and the ticket describing the layout in snake mode ([#5712](https://github.com/epam/ketcher/issues/5712)).

The other part of this functionality is covered in the following ticket: [#6254](https://github.com/epam/ketcher/issues/6254).

## Summary of differences between editing both chains and editing one chain

![image](https://github.com/user-attachments/assets/1bbfdcc1-4afa-4d7c-92d7-6efcf049f5ee)

### Chains that can not be represented one bellow another

1.1. If hydrogen bonds connect only monomers within one chain, that chain should be represented in sequence mode just like it would be without hydrogen bonds.
> ![image](https://github.com/user-attachments/assets/38c287dc-be45-4337-bba7-29dcb5065379)
> Should in sequence mode be:
> AAACCCCCCU UU

1.2. If there is a circular hydrogen bond connection between three or more chains, those chains should be represented in sequence mode just like they would be without hydrogen bonds.
> ![image](https://github.com/user-attachments/assets/2e0655b7-0215-4caa-8a92-bac160888a7f)
> Should in sequence mode be:
> AA
> UU
> CC

---

### Chains placement and formatting

2. If two or more chains are connected via H-bonds and do not satisfy requirement 2, in sequence mode they should be represented in such a way that:

2.1. The chain with more monomers is placed on top, or if there is a tie the, one with a higher center on canvas in snake/flex modes.
> Same logic as requirement 2.2. from [#5712](https://github.com/epam/ketcher/issues/5712).

2.2. Multiple chains can be placed on a single line if both of those chains are connected via H-bonds to one chain.
> Same logic as requirement 2.8. from [#5712](https://github.com/epam/ketcher/issues/5712).

2.3. The chain(s) placed on the bottom should have their numbering reversed.
> ![image](https://github.com/user-attachments/assets/36be2e78-f496-44ce-a024-7d7cf7a9125a)

2.4. After every tenth symbol in one of the chains should come a break, and a number indicator should be placed above/bellow appropriate monomers/symbols.
> The maximum amount of monomer symbols before a break is ten for one of the chains (the other one can have ten or less).
[...]

PR #10729, opened 2026-07-29
#6463 - Cursor position is wrong after adding peptide after - symbol

### LINKED ISSUE #6463: Cursor position is wrong after adding peptide after `-` symbol
**Steps to Reproduce**
1. Go to Macro - Sequence mode (clean canvas)
2. Load from HELM: `RNA1{R(A)P.R(G)}|RNA2{R(U)}|RNA3{R(C)}$RNA1,RNA2,2:pair-2:pair|RNA1,RNA3,5:pair-2:pair$$$V2.0`
Image
3. Click between `-` and `G` symbols to switch to Edit mode
Image
4. Switch editor to `PEP` mode and press `C` key

**Actual behavior**
Cursor jumps over `G` symbol to the right
Image

**Expected behavior**
Cursor remain after `C` symbol and before `G` one
Image

**Versions**
- Ketcher Version 3.1.0-rc.2 Build at 2025-02-07; 15:09:23
- Indigo Version 1.29.0-rc.1.0-g2e30df8f7-wasm32-wasm-clang-19.0.0
- Chrome Version 133.0.6943.54 (Official Build) (64-bit)
- Win10

Issue found while testing: https://github.com/epam/ketcher/issues/5942

AvetBadalyan:

Repro: in Sequence mode, load a HELM structure where one chain (RNA1: A-P-G) is hydrogen-bonded to two different antisense chains (RNA2 paired with A, RNA3 paired with G). Between A and G, the view renders a "-" placeholder — a `BackBoneSequenceNode` inserted wherever a chain's antisense partner changes ("#5942" introduced this). Click between "-" and G to place the edit caret there, switch to PEPTIDE mode, and type "C". Expected: cursor ends up right after the new C and before G. Actual: cursor jumps past G.

**Root cause — two compounding issues, both from the sense/antisense sync-editing logic introduced in #5942:**

1. Sync-edit mode mirrors typed monomers onto the antisense strand. It didn't check whether the monomer being typed is a peptide — peptides can't be hydrogen-bond partners, so this created a dangling, bondless "mirror" peptide monomer on the antisense strand, perturbing the rebuilt view model.
2. Separately, the caret is advanced via `SequenceRenderer.moveCaretForward()`, which blindly adds 1 to the previous numeric column index. But typing a real monomer at a position occupied by the "-" backbone placeholder *replaces* that column rather than adding a new one (net column count unchanged), so the blind `+1` overshoots by exactly one position regardless of issue #1.

**Fix (`SequenceMode.ts`):**
- Added a `sequenceTypeEnterMode !== SequenceType.PEPTIDE` guard to the antisense-mirroring condition, preventing the dangling phantom monomer. This mirrors the same peptide-mode gating pattern already used elsewhere in this file (e.g. the phosphate-insertion branch a few hundred lines down).
- For the sense-edit case, replaced the blind `moveCaretForward()` call with identity-based positioning — `SequenceRenderer.setCaretPositionNextToMonomer(senseNodeToConnect.lastMonomerInNode)` — which locates the newly-inserted monomer in the freshly-rebuilt view model and positions the caret relative to it, rather than assuming a fixed column delta. This is an existing helper/pattern already used identically by `insertMonomerFromLibrary` and `createRnaPresetNode` elsewhere in the same file; it just hadn't been applied to the typed-letter path yet.

**Note (unrelated, separate commit):** this branch also includes a one-line fix to `ketcher-autotests/tests/fixtures/commonPageObjectFixtures.ts` (`Record<string, never>` → `Record<never, never>`). It's a pre-existing `tsc --noEmit` breakage in that workspace caused by the Playwright 1.59 dependency bump (unrelated to the #6463 change itself), left in because it blocks the pre-push hook for every branch until fixed. Kept as its own commit for clarity.

## Check list
- [ ] unit-tests written
- [ ] e2e-tests written
- [ ] documentation updated


svvald (approved):

AlexeyGirin:
Regression in tests

PR #10750, opened 2026-07-30
#7076 – The chemical elements are arranged in reverse order

### LINKED ISSUE #7076: The chemical elements are arranged in reverse order
**Steps to Reproduce**
1. Micro mode
2. Open KET file [ketcher (57).zip](https://github.com/user-attachments/files/20105805/ketcher.57.zip) 

Image

3. Switch to Macro mode 

Image

**Actual behavior**
The chemical elements are arranged in reverse order.

**Expected behavior**
The chemical elements should maintain the same order in Macro mode as they do in Micro mode.

Ketcher version
• Version 3.4.0-rc.1 Build at 2025-05-01; 08:56:06
• [Indigo Toolkit](http://lifescience.opensource.epam.com/indigo/) Version 1.32.0-rc.1.0-g65e26e370-wasm32-wasm-clang-19.0.0
• Chrome Version 136.0.7103.48 (Official Build) (64-bit)
• Win11

maksimdev:
## How did you fix the issue?

Loading a KET file with several free-floating, unbonded single-atom "molecules" (e.g. Cl, O, H, O) rendered correctly in Micro mode but appeared reversed in Macro mode.

This was not an ordering/parsing bug — Ket parsing and the Micro→Macro conversion preserve atom order faithfully in both modes. The real defect is in `AtomRenderer.shouldHydrogenBeOnLeft` (`packages/ketcher-core/src/application/render/renderers/AtomRenderer.ts`):

- `viewModel.atomsToHalfEdges` is a `Map` populated only while iterating bonds, so a bondless atom has **no entry at all** — the lookup returns `undefined`, not `[]`.
- The guard was `atomHaldEdges?.length === 0`, which is `false` when the value is `undefined`, so bondless `leftH` elements (Cl, O, F, S, Se, Br, I) fell through to "hydrogen on the right", producing `ClH`/`OH2` instead of `HCl`/`H2O`. Visually this reads as the row being reversed.
- Fixed by changing the guard to `!atomHaldEdges?.length`, treating "no map entry" and "empty array" identically — matching the legacy Micro-mode renderer (`restruct/reatom.ts`), which never hit this bug because its neighbor lookup always returns a real (possibly empty) array.

Also fixed a secondary defect found during verification: a lone unbonded `H` atom rendered as `HH` (`H2` with the hydrogen count), because `labelLength`/`appendLabel` always appended an implicit-hydrogen suffix when `hydrogenAmount > 0`, without checking whether the atom's own label is already a hydrogen isotope (`H`/`D`/`T`). Added an `isHydrogenIsotopeLabel` guard to prevent that.

## Check list
- [ ] unit-tests written
- [ ] e2e-tests written
- [ ] documentation updated


packages/ketcher-core/src/application/render/renderers/AtomRenderer.ts
```diff
@@ -309,6 +309,15 @@ export class AtomRenderer extends BaseRenderer {
     return isGenericAtom(this.atom.label);
   }
 
+  // A label that is itself a hydrogen isotope must not get an implicit-hydrogen suffix appended.
+  private get isHydrogenIsotopeLabel() {
+    return (
+      this.atom.label === AtomLabel.H ||
```
AlexeyGirin: Remove that or make sure that H is shown as H2 on Macro
Same for D and T - shoud be shown on Macro DH and TH

AlexeyGirin (dismissed):

svvald (approved):

PR #10760, opened 2026-07-31
4444: fix it is possible to remove selection in Modify in RNA builder…

### LINKED ISSUE #4444: Macro: It is possible to remove selection while in Modify in RNA Builder mode
**Steps to Reproduce**

1. Switch to Macro - Sequence mode
3. Select few nucleotides and try to modify them
image
4. Press Esc button

**Actual behavior**
Selection disappears
image

**Expected behavior**
Selection remain in place
image

As per [requirement](https://github.com/epam/ketcher/issues/3824):
> 9. Do not allow an user to unselect nucleotides if a sequence in "Modify in RNA builder" mode, while an user in "Modify in RNA builder" mode.

**Versions:**
Windows 10

oguzhanduran:
… mode

## Check list
- [ ] unit-tests written
- [ ] e2e-tests written
- [ ] documentation updated


packages/ketcher-macromolecules/src/components/monomerLibrary/RnaBuilder/RnaEditor/RnaEditorExpanded/helpers/resetRnaBuilder.ts
```diff
@@ -18,10 +18,15 @@ export const resetRnaBuilder = (dispatch: Dispatch<AnyAction>) => {
 export const resetRnaBuilderAfterSequenceUpdate = (
   dispatch: Dispatch<AnyAction>,
   editor: CoreEditor | undefined,
+  keepSelection = false,
```
svvald, line 21: Naming/polarity inconsistency: this is `keepSelection` here but the same concept is `needToRemoveSelection` in `Editor.ts`/`SequenceMode.ts`, and this function inverts it at the call site (`!keepSelection`). A future edit to one layer without the other (e.g. a new caller of `turnOffSequenceEditInRNABuilderMode` elsewhere) can silently flip behavior with no type error to catch it. Consider using one polarity end-to-end, or an options object (e.g. `{ preserveSelection: true }`) so call sites are self-documenting.

packages/ketcher-macromolecules/src/components/monomerLibrary/RnaBuilder/RnaEditor/RnaEditorExpanded/helpers/resetRnaBuilder.ts
```diff
@@ -18,10 +18,15 @@ export const resetRnaBuilder = (dispatch: Dispatch<AnyAction>) => {
...
 ) => {
   resetRnaBuilderCommon(dispatch);
   dispatch(setSequenceSelection([]));
```
svvald, line 24: This unconditionally clears the redux `sequenceSelection` regardless of `keepSelection`, right above two calls that are now gated on it. If intentional (this drives the RNA-Builder panel form state, distinct from the canvas selection the `keepSelection` param protects), a short comment here would prevent the next reader from assuming it's a bug.

packages/ketcher-macromolecules/src/components/monomerLibrary/RnaBuilder/RnaEditor/RnaEditorExpanded/helpers/resetRnaBuilder.ts
```diff
@@ -18,10 +18,15 @@ export const resetRnaBuilder = (dispatch: Dispatch<AnyAction>) => {
...
-  editor?.events.turnOffSequenceEditInRNABuilderMode.dispatch();
+  editor?.events.turnOffSequenceEditInRNABuilderMode.dispatch(!keepSelection);
   if (editor?.mode?.modeName === 'sequence-layout-mode')
-    (editor.mode as unknown as { turnOffEditMode(): void }).turnOffEditMode();
+    (
+      editor.mode as unknown as {
+        turnOffEditMode(needToRemoveSelection?: boolean): void;
```
svvald, line 29: Since this PR already added `turnOffSequenceEditInRNABuilderMode` to the `SequenceMode` public interface (`types/sequenceMode.ts`), it might be worth doing the same for `turnOffEditMode` here instead of extending this untyped `as unknown as {...}` cast with the new parameter — the cast currently has no compile-time check against the real `SequenceMode.turnOffEditMode` signature, so the two can drift silently.

svvald (commented):
Ran a 4-angle review (code-quality, performance, documentation-accuracy, security) on top of the noted concern about this PR's footprint (a boolean threaded through several layers).

**On the "flags/params footprint" concern:** it's real but bounded and traceable — `needToRemoveSelection` flows `Editor.ts` → `SequenceMode.ts` → `BaseMode.initialize` in one path and `resetRnaBuilder.ts` (as the inverted `keepSelection`) in the other, mirroring the existing layered mode architecture rather than sprawling arbitrarily. The two root causes are correctly diagnosed and the fix is targeted, not a bandaid: `ContextMenu.tsx`'s capture-phase Escape listener and a second, independent `exit` hotkey listener registered on `document` in `Editor.ts` (`setupHotKeysEvents`) both needed to be neutralized, which is exactly what `setContextMenuActive(false)` and `event.stopImmediatePropagation()` do. Both new inline comments were verified accurate against the actual listener code.

**No performance or security issues found.** The widened `useEffect` dependency array is a genuine stale-closure fix (previously Enter could act on a stale RNA-builder-mode flag), not just cleanup — worth calling out as a real improvement, not only a bugfix side effect.

See inline comments for the naming/consistency nits below — none are blocking, but worth a quick pass before merge.

AlexeyGirin:

svvald (approved):

PR #10859, opened 2026-08-07
#7780 - Selected selection tool in Macro mode is not saved to Local storage and resets to Rectangle tool after reload 

### LINKED ISSUE #7780: Selected selection tool in Macro mode is not saved to Local storage and resets to Rectangle tool after reload
**Description:**
In Macro mode, the currently selected selection tool (e.g., Lasso or Fragment selection) is not persisted in Local storage. After reloading Ketcher, the selection tool always resets to Rectangle tool. In contrast, in Micro mode the selected selection tool is correctly stored in Local storage and restored after reload. This creates inconsistency between modes.

**Steps to Reproduce**
1. Open Ketcher in Macro mode.
2. Change the selection tool from Rectangle to Lasso or Fragment.
3. Reload the page.
4. Observe the active selection tool after reload.
5. Repeat the same steps in Micro mode.

**Actual behavior**
In Macro mode, the tool resets to Rectangle after reload.
In Micro mode, the last used selection tool is correctly restored.

**Expected behavior**
The selected selection tool should be saved in Local storage in Macro mode.
After reload, Ketcher should restore the last used selection tool (Lasso, Fragment, etc.) consistently across both modes.

**Video**

https://github.com/user-attachments/assets/3197c537-5b86-419a-a930-5f622cd4c3ec

**Desktop:**
 - OS: Windows 11
 - Browser Chrome
 Version 138.0.7204.101 (Official Build) (64-bit)

**Ketcher version** 
[Version 3.8.0-rc.1]
**Indigo version** 
[Version 1.36.0-rc.1]

This bug was found while testing: https://github.com/epam/ketcher/issues/6358

AnzhelaMartsinovich:

Related Issue - [7780](https://github.com/epam/ketcher/issues/7780)

Solution

- Added local storage functionality for the selection tool in Macro mode
- Tool selection now persists across page reloads in both Macro and Micro modes
- Ensures consistent behavior between modes

Testing

- Switch to Macro mode
- Select Lasso or Fragment tool
- Reload the page
- Verify that the previously selected tool remains active

## Check list
- [ ] unit-tests written
- [ ] e2e-tests written
- [ ] documentation updated


packages/ketcher-macromolecules/src/components/menu/subMenu/SubMenu.tsx
```diff
@@ -111,6 +113,14 @@ const SubMenu = ({
           activeItemName: activeOption,
         }),
       );
+
+      if (subMenuId === SELECT_SUBMENU_ID) {
+        const toolType = activeOption.replace('select-', '');
+        SettingsManager.selectionTool = {
+          tool: 'select',
+          opts: toolType,
+        };
```
svvald: With this approach both Micro and Macro mode share the same local storage entry. However, there is an issue – for Micro mode selecting a `Structure Selection` tool will write `fragment` in storage while the same tool in Macro mode writes `structure`. In subsequent reload in will crash the application

So we have to either make tool names uniform or use separate storage entries for each mode (I think this option is better yet having identical names for the same tools wouldn't hurt anyway)

svvald: Also there is a `Fragment selection` tool in Micro mode which is not present in Macro mode but luckily it doesn't write anything to storage now (apparently a bug), however, it would be nice to have an crash-stable approach in case of such inconsistencies

packages/ketcher-macromolecules/src/EditorEvents.tsx
```diff
@@ -106,41 +107,63 @@ export const EditorEvents = () => {
...
 
     return () => {
       dispatch(selectTool(null));
+      editor?.events.error.remove(errorHandler);
+      editor?.events.openErrorModal.remove(errorModalHandler);
+      editor?.events.openMonomerConnectionModal.remove(
+        monomerConnectionModalHandler,
+      );
+      editor?.events.openConfirmationDialog.remove(confirmationDialogHandler);
       editor?.events.selectTool.remove(handler);
```
svvald: Perhaps it's better to rename just `handler` to `selectToolHandler`?

packages/ketcher-macromolecules/src/EditorEvents.tsx
```diff
@@ -106,41 +107,63 @@ export const EditorEvents = () => {
...
       editor.events.selectTool.add(handler);
+
+      // Initialize with saved selection tool or default to rectangle
+      const savedSelectionTool = SettingsManager.selectionTool;
+      const initialTool = savedSelectionTool?.opts
+        ? `select-${savedSelectionTool.opts}`
+        : 'select-rectangle';
+
+      dispatch(selectTool(initialTool));
+      editor.events.selectTool.dispatch([initialTool]);
```
svvald: I suppose there is no much sense in calling both?

packages/ketcher-macromolecules/src/components/menu/subMenu/SubMenu.tsx
```diff
@@ -111,6 +113,14 @@ const SubMenu = ({
...
+      }
```
svvald: Overall this existing effect already violates React best practices and yields the ESLint warning, let's address it and make a proper event handler function instead

svvald (changes requested):

packages/ketcher-macromolecules/src/components/menu/subMenu/SubMenu.tsx
```diff
@@ -100,17 +121,6 @@ const SubMenu = ({
...
-    if (subMenuId && activeOption && activeOption !== lastActiveOption) {
-      dispatch(
-        setSelectedMenuGroupItem({
-          groupName: subMenuId,
-          activeItemName: activeOption,
-        }),
-      );
-    }
-  }, [dispatch, subMenuId, activeOption, lastActiveOption]);
-
```
svvald, line 114: It warms my heart when we remove those nasty effects 🔥

AlexeyGirin:

packages/ketcher-macromolecules/src/components/menu/subMenu/SubMenu.tsx
```diff
@@ -101,17 +122,6 @@ const SubMenu = ({
...
-  useEffect(() => {
-    if (subMenuId && activeOption && activeOption !== lastActiveOption) {
-      dispatch(
-        setSelectedMenuGroupItem({
-          groupName: subMenuId,
-          activeItemName: activeOption,
-        }),
-      );
-    }
-  }, [dispatch, subMenuId, activeOption, lastActiveOption]);
```
svvald, line 113: However, I got a nice finding from AI that with doing so we might introduce regression issues as just onClick handler in SubMenu is not enough for this state to be synced as it only fires when the user clicks an item inside the SubMenu dropdown. Tool changes from other paths (page-load restore, Shift+Tab cycling, programmatic selectTool dispatch) no longer update selectedMenuGroupItems[SELECT_SUBMENU_ID]

I had a chat with Claude and that's what it proposed me, can you take a look and implement this proposal if it's valid:

```
What activeOption already is: SubMenu's isActive comes from useMenuContext(), which Menu.tsx builds as activeMenuItems.includes(itemKey), and activeMenuItems = [activeTool] (LeftMenuComponent.tsx:27), where activeTool is state.editor.activeTool — set by exactly one reducer, selectTool (editorSlice.ts:128-130). So activeOption in SubMenu was never anything but "does state.editor.activeTool equal one of my options" — it's fully derived, not real local state.

What selectedMenuGroupItems[SELECT_SUBMENU_ID] is actually for: not rendering — it's a memory of the last selection-tool used, read by TopMenuComponent.tsx:111 and EditorEvents.tsx to restore a selection tool after the app forces something else (exiting sequence-edit mode, clear, restore-on-load). That's legitimately state (you can't derive "what was active before" from "what's active now"), but it doesn't need its own component-level effect to stay in sync — it needs to be written at the one place tool changes actually happen.

And that one place already exists: every tool change in the app funnels through editor.events.selectTool.dispatch(...) → the single selectToolHandler subscriber in EditorEvents.tsx → dispatch(selectTool(toolName)). I checked every call site (editorEvents.ts core cycling/Shift+Tab, LeftMenuComponent, TopMenuComponent, the new load-time restore) — they all go through that event, with a couple of call sites also directly dispatch(selectTool(...))ing the same value defensively. So this handler already sees 100% of tool changes, from every path.

Proposal:

1. Redux: extend the selectTool reducer itself (editorSlice.ts:128-130) to also update the "last selection tool" when the incoming tool is a selection variant:
selectTool: (state, action: PayloadAction<string>) => {
  state.activeTool = action.payload;
  if (action.payload?.startsWith('select-')) {
    state.selectedMenuGroupItems[SELECT_SUBMENU_ID] = action.payload;
  }
},
This is pure, synchronous, colocated with the state it affects, and fires for literally every dispatch(selectTool(...)) regardless of caller — no event wiring, no effect, no missed path.

2. Persistence: move the SettingsManager.setSelectionTool('macro', ...) call the PR added in SubMenu's onClick into selectToolHandler in EditorEvents.tsx (guarded the same way, toolName.startsWith('select-')). That's a real side effect, so it belongs in the actual event-handler callback (which the react-you-might-not-need-an-effect rule is fine with — it's not reacting to a derived render value, it's reacting to an actual event), not in a reducer and not in one UI component's click handler.
3. SubMenu.tsx: delete the handleMenuItemClick dispatch/persistence logic entirely (revert to a plain onClick?.() passthrough). SubMenu goes back to being purely presentational, reading activeOption/lastActiveOption from Redux but never writing them — which is what should have replaced the old useEffect, rather than a new, narrower write path.

Net effect: no useEffect anywhere for this, SELECT_SUBMENU_ID tracking is correct for every path (click, Shift+Tab, load-restore, programmatic), and SettingsManager persistence is correct too, since both are driven off the same single event rather than off SubMenu's render.
```

svvald (changes requested):

svvald (approved):

PR #10861, opened 2026-08-07
#10407 - Field values get can only be selected and moved in the case of absolute and relative data S-groups in the Macromolecules mode

### LINKED ISSUE #10407: Field values get can only be selected  and moved in the case of absolute and relative data S-groups in the Macromolecules mode
**Steps to Reproduce**
1.  Clear Canvas 
2. Molecules mode > Open KET file [S-groups_7352r.zip](https://github.com/user-attachments/files/29746110/S-groups_7352r.zip)
3. Macromolecules mode > Open the same  KET file [S-groups_7352r.zip](https://github.com/user-attachments/files/29746110/S-groups_7352r.zip)
4. Select the field values in the case of absolute and relative data S-groups  Macromolecules and  Molecules mode 

**Actual behavior**
Field values cannot be selected in the case of absolute and relative data S-groups in the Macromolecules mode 

https://github.com/user-attachments/assets/80b2e1d2-8002-47a7-a596-365bb2a2550c

**Expected behavior**
According to https://github.com/epam/ketcher/issues/7352

- Field values get can only be selected in the case of absolute and relative data S-groups in the any mode
- Field values should be able to be moved independently form the structure in case of absolute and relative data S-groups

https://github.com/user-attachments/assets/485b5f4d-1c85-472f-8c52-7421e09750ba

**Environment details:**
• [Version 3.18.0-rc.1](https://lifescience.opensource.epam.com/ketcher/index.html) Build at 2026-07-02; 16:03:43
• [Indigo Toolkit](http://lifescience.opensource.epam.com/indigo/) Version 1.46.0-rc.1.0-gf3408986b-wasm32-wasm-clang-23.0.0
• Chrome Version 147.0.7727.102 (Official Build) (64-bit)
• Win11

Nollili:
Data S-group field values were rendered as non-selectable overlays in Macromolecules mode, so absolute and relative values could not be selected or moved independently from their structures.

This change:
Adds eligible Data S-groups to the selectable drawing entities.
Makes absolute and relative field values selectable and independently movable.
Adds selection highlighting and movement rendering for field values.
Preserves the existing behavior for attached values, which remain non-selectable.
Adds unit tests covering selection, movement, and attached-value behavior.

Closes #10407


packages/ketcher-core/src/application/render/renderers/SGroupRenderer.ts
```diff
@@ -517,11 +533,21 @@ export class SGroupRenderer extends BaseRenderer {
...
+        .attr('fill', valueBackgroundColor)
+        .attr('stroke', valueBackgroundColor);
+    });
   }
 
   public moveSelection(): void {
-    // S-groups are rendered as non-selectable macro overlays.
+    this.remove();
+    this.show();
   }
```
dorogusha, line 551: `moveSelection()` now tears down and rebuilds the S-group's DOM (label/value groups, plus `getBBox()` calls in `appendValue`) on every drag frame via `remove()` + `show()`. Other renderers in this codebase (atoms, monomers) update a `transform` on the existing node during move instead of recreating it. When multiple data S-groups are selected and dragged together, this remove+rebuild-per-frame could cause visible jank that scales with the number of selected S-groups. Worth considering a cheaper transform-based update here instead.

packages/ketcher-core/src/domain/entities/DrawingEntitiesManager.ts
```diff
@@ -343,6 +343,9 @@ export class DrawingEntitiesManager {
       ...(this.multitailArrows as Map<number, DrawingEntity>),
       ...(this.rxnPluses as Map<number, DrawingEntity>),
       ...(this.stereoFlags as Map<number, DrawingEntity>),
+      ...([...this.sgroups].filter(
+        ([, sgroup]) => sgroup.isSelectableDataSGroup,
+      ) as [number, DrawingEntity][]),
```
dorogusha, line 348: `allEntities` is uncached and is read from `selectIfLocatedInRectangle` / `selectIfLocatedInPolygon`, which run on every mousemove during rubber-band selection. This added `[...this.sgroups].filter(...)` + spread performs an O(sgroups) allocation on every such call, scaling with document size. Might be worth avoiding the per-access filter here (e.g. maintaining/caching a selectable-sgroups view) rather than recomputing it on every getter read.

packages/ketcher-core/src/domain/entities/SGroupDrawingEntity.ts
```diff
@@ -17,9 +17,25 @@ export class SGroupDrawingEntity extends DrawingEntity {
...
+    return this.sgroup.type === SGroup.TYPES.DAT && !this.sgroup.data.attached;
+  }
+
+  public moveRelative(position: Vec2): void {
+    super.moveRelative(position);
+
+    if (this.isSelectableDataSGroup) {
+      this.sgroup.pp?.add_(position);
+    }
+  }
```
dorogusha, line 37: `moveRelative` updates both the inherited `_position` (via `super.moveRelative`) and `sgroup.pp` separately, but `center` for a selectable data S-group only reads `sgroup.pp` (line 20) — so `_position` ends up maintained but never consumed for this entity type. Not a bug today, but having two positions tracked without explanation could be confusing/easy to desync in a future change. A short comment on why both are updated (or dropping the now-unused one) would help future readers.

dorogusha (changes requested):

packages/ketcher-core/src/application/render/renderers/SGroupRenderer.ts
Nollili: Updated, `moveSelection()` now keeps the existing value SVG group and updates its `transform` based on the movement offset, avoiding DOM reconstruction and repeated `getBBox()` calls during dragging. Also extended the unit test to verify that the same SVG node is retained after movement.

packages/ketcher-core/src/domain/entities/DrawingEntitiesManager.ts
Nollili: Updated, selectable Data S-groups are now maintained in a dedicated map when S-groups are added, restored, or deleted. allEntities reads that map directly, so rubber-band selection no longer filters and allocates over all S-groups on every access.

packages/ketcher-core/src/domain/entities/SGroupDrawingEntity.ts
Nollili: Updated, for selectable Data S-groups, moveRelative() now updates only sgroup.pp and returns without modifying the inherited _position. This leaves a single source of truth matching the center getter.

dorogusha (approved):

Issue #3508, opened 2023-10-25
When adding list of atoms (at least 3), some edges of the template "Cyclooctane" move to the side


urszula-gawlowska:
**Steps to Reproduce**
1. Launch "Ketcher"
2. Draw the Template "Cyclooctane" on the canvas
3. Click on the lower left atom
4. Click on the "Edit" button in the context menu 
5. Change atom type to the list
6. Add at least 3 atoms
7. Click on the "Apply" button in the context menu 
8. Verify the integrity of the Template "Cyclooctane"

**Actual behavior**
Some edges of the Template "Cyclooctane" move to the side. The Template is broken

**Expected behavior**
The edges of the Template "Cyclooctane" do not move to the side. The Template remains unbroken

**Screenshots**
![egde_move_to_the_side_bug](https://github.com/epam/ketcher/assets/145990256/fce9bf35-8949-48f0-abc2-3d9971a9c958)

**Desktop**
- OS: Windows 11 Pro
- Browser: chrome version 118.0.5993.89

**Ketcher version** 
v2.16.0 

**Additional context**
Bug similar to: #3424

leralumea:
The issue is reproduced.

Ketcher Version 3.18.0-rc.1 Build at 2026-07-02; 16:03:43
Indigo Version 1.46.0-rc.1.0-gf3408986b-wasm32-wasm-clang-23.0.0
Chrome Version 150.0.7871.115 (Official Build) (64-bit)
Windows 11

<img width="1917" height="905" alt="Image" src="https://github.com/user-attachments/assets/6ffe78e5-c041-45de-8230-0487903eee52" />
<img width="1919" height="902" alt="Image" src="https://github.com/user-attachments/assets/358831ea-f479-4549-8a94-6540494d72a6" />

d-mrzv:
Issue is fixed
Verified on:
Version 3.19.0-rc.1 Build at 2026-08-03; 17:04:29
Indigo Toolkit Version 1.46.0-rc.1.0-gf3408986b-x86_64-linux-gnu-11.2.1
Chrome Version 151.0.7922.170 (Official Build) (arm64)
Macos 26.5.2 (25F84)

<img width="469" height="208" alt="Image" src="https://github.com/user-attachments/assets/9a8b8091-1315-4067-9d01-031dc9ff3c71" />

PR #11084, opened 2026-08-11
#3268 – Incorrect display of Attachment points along with Atom Properties

### LINKED ISSUE #3268: Incorrect display of Attachment points along with Atom Properties
**Steps to Reproduce**
1.	Launch "Ketcher"
2.	Draw any Template (like Benzene) on the canvas
3.	For each atom of the Template, add some properties in the "Query properties" section of the context menu
4.	For each atom of the Template, add one Attachment point in the "Edit attachment point" section of the context menu
5.	Verify the location and integrity of the Attachment points and the labels with the Properties

**Actual behavior**
a) A part of the Attachment point is not displayed
b) The Attachment point is turned to the other side (not every time it is reproduced)
c) The Attachment point overlaps the label with the Properties
d) The Attachment point number also overlaps the label with the Properties

**Expected behavior**
a) The whole Attachment point is displayed
b) The Attachment point is not turned to the other side
c) The Attachment point does not overlap the label with the Properties
d) The Attachment point number does not overlap the label with the Properties

**Screenshots**
image

 - Browser: Сhrome

**Ketcher version** 2.15.0-rc.1

**Additional context**
If we swap step 3 and step 4, i.e. add Attachment points first and then some properties, we still get the same actual result

OnurKucukoz:
### How the feature works? / How did you fix the issue?

Fixed attachment-point rendering when an atom also displays query-property labels.

Previously, atom/query-property labels and attachment points could be placed in
the same open sector, causing the attachment-point stem, wave, or number to
overlap the property label. An earlier approach moved the attachment-point
endpoint outward, which made the attachment-point bond visually too long.

The updated solution keeps the attachment-point endpoint and number at their
nominal positions. When a property label intersects the nominal attachment-point
area, the label is displaced laterally. The stem start is also capped to preserve
a visible, outward-facing stem.

Closes #3268

### Verification
Added focused unit regression coverage for nominal, moderate, and extreme
attachment-point geometry.
Verified that query-property labels and attachment-point numbers do not
overlap under deterministic rendered geometry.
Loaded the SMILES from #3268 in standalone mode and confirmed that all six
attachment points remain visible without elongated bonds or property-label
overlap.

### Screenshot

Screenshot 2026-08-20 095212

## Check list
- [x] unit-tests written
- [ ] e2e-tests written
- [ ] documentation updated

### DIFF
--- a/packages/ketcher-core/__tests__/application/render/restruct/rergroupAttachmentPoint.test.ts
+++ b/packages/ketcher-core/__tests__/application/render/restruct/rergroupAttachmentPoint.test.ts
@@ -1,5 +1,7 @@
 import { Render, ReStruct } from 'application/render';
+import draw from 'application/render/draw';
 import type { RenderOptions } from 'application/render/render.types';
+import { Box2Abs } from 'domain/entities/box2Abs';
 import {
   AttachmentPoints,
   Atom,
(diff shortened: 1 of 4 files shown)


svvald (changes requested):
The provided output is not correct – with this approach attachment point bonds look visually longer, which shouldn't happen

A similar issue was addressed recently and fixed correctly – could you explore it please, perhaps they might have a shared solution?
https://github.com/epam/ketcher/issues/3508, PR: https://github.com/epam/ketcher/pull/10699

PR #11210, opened 2026-08-14
#7129 - Calculate Properties window clears on mouse release when macrom…

### LINKED ISSUE #7129: Calculate Properties window clears on mouse release when macromolecules are connected via a microstructure without attachment points
**Description:**
When selecting macromolecules that are connected through a microstructure without attachment points (e.g., benzene ring), the Calculate Properties window briefly shows valid data only while the mouse is held down. Once the mouse button is released, the data disappears from the window, leaving it empty.

**Steps to Reproduce**
1. Add a nucleoside and a peptide to the canvas.
2. Connect them via a benzene ring that does not have attachment points.
3. Select the entire structure by clicking and holding the mouse.
4. Observe the Calculate Properties window while holding the mouse.
5. Release the mouse button.

**Test file:** 

[microstructure-without-attachment-points.zip](https://github.com/user-attachments/files/20244397/microstructure-without-attachment-points.zip)

**Actual behavior**
While holding the mouse selection, valid data appears in the Calculate Properties window. As soon as the mouse is released, the data vanishes and the window is cleared.

**Expected behavior**
The Calculate Properties window should persistently display calculated values for the selected structure regardless of mouse state, as long as the selection remains active.

**Video**

https://github.com/user-attachments/assets/b2835893-e838-4a96-9e48-8134dc0f0eaa

**Desktop:**
 - OS: Windows 11
 - Browser Chrome
 - Version 133.0.6943.127 (Official Build) (64-bit)

**Ketcher version** 
[Version 3.4.0-rc.3]
**Indigo version** 
[Version 1.32.0-rc.2]

This bug was found while testing: https://github.com/epam/Indigo/issues/2903
Related feature: https://github.com/epam/ketcher/issues/5727

crossaba:
…olecules are connected via a microstructure without attachment points

Problem

System clean up calculation when user finishes selection (mouse up) - The Calculate Properties window was incorrectly showing or clearing based on:
1. Macromolecules connected via microstructures (like benzene rings) without attachment points
2. Whether connecting microstructures were included/excluded from selection
3. Multiple unconnected monomers being selected together

Root Cause

The getAllConnectedMonomersRecursively function only checked direct PolymerBond and HydrogenBond connections between monomers. It completely ignored:
- MonomerToAtomBond connections (monomer → atom in microstructure)
- The atoms and bonds within microstructures themselves
- Selection context (traversed entire canvas instead of just selected entities)

Solution Overview

Two files modified:

1. packages/ketcher-core/src/domain/helpers/monomers.ts
  - Added imports for Atom and Bond types
  - Enhanced getAllConnectedMonomersRecursively with BFS traversal through microstructures
  - Added optional allowedMonomers parameter for selection-context awareness
2. packages/ketcher-macromolecules/src/hooks/useRecalculateMacromoleculeProperties.ts
  - Created allowedMonomers set from selected entities
  - Passed it to connectivity check function

Key Implementation Details

Function signature change:
export function getAllConnectedMonomersRecursively(
  monomer: BaseMonomer,
  allowedMonomers?: Set<BaseMonomer>,  // NEW: selection context
): BaseMonomer[]

Multiple safety guards implemented:
1. ✅ Microstructure entry guard: Only enter BFS if atom belongs to microstructure (isMicromoleculeFragment)
2. ✅ Selection guard: Only traverse microstructures in the allowedMonomers set
3. ✅ Self-reference guard: Don't traverse current monomer's own structure
4. ✅ Owner exclusion: Don't count microstructure owner as a connected monomer
5. ✅ Found monomer guard: Only add monomers that are in allowedMonomers set
6. ✅ Atom restriction: Only traverse atoms within microstructures
7. ✅ Loop prevention: Track visited atoms and monomers

BFS Algorithm:
- When MonomerToAtomBond found → check guards → enter BFS through microstructure
- Use queue to systematically visit all atoms in the microstructure
- For each atom: check MonomerToAtomBond (find monomers) and Bond (traverse to next atom)
- Only traverse/add entities within the selection context

All Test Cases Now Working                                                                                                                                                                                                                                                                                        
                                                                                                                                                                                                                                                                                                                    
  ┌─────┬─────────────────────┬─────────────────┬──────────────────┬────────┐                                                                                                                                                                                                                                       
  │  #  │      Scenario       │    Selection    │     Expected     │ Status │                                                                                                                                                                                                                                       
  ├─────┼─────────────────────┼─────────────────┼──────────────────┼────────┤                                                                                                                                                                                                                                       
  │ 1   │ A → Benzene → B     │ All 3           │ Show properties  │ ✅     │                                                                                                                                                                                                                                       
  ├─────┼─────────────────────┼─────────────────┼──────────────────┼────────┤                                                                                                                                                                                                                                       
  │ 2   │ A → Benzene → B     │ A + B only      │ Clear properties │ ✅     │                                                                                                                                                                                                                                       
  ├─────┼─────────────────────┼─────────────────┼──────────────────┼────────┤                                                                                                                                                                                                                                       
  │ 3   │ A → B direct        │ A + B           │ Show properties  │ ✅     │                                                                                                                                                                                                                                       
  ├─────┼─────────────────────┼─────────────────┼──────────────────┼────────┤                                                                                                                                                                                                                                       
  │ 4   │ A, B unconnected    │ A + B           │ Clear properties │ ✅     │                                                                                                                                                                                                                                       
  ├─────┼─────────────────────┼─────────────────┼──────────────────┼────────┤                                                                                                                                                                                                                                       
  │ 5   │ A → Benzene → B → C │ A + C + Benzene │ Clear properties │ ✅     │                                                                                                                                                                                                                                       
  ├─────┼─────────────────────┼─────────────────┼──────────────────┼────────┤                                                                                                                                                                                                                                       
  │ 6   │ A → Benzene → B → C │ All 4           │ Show properties  │ ✅     │                                                                                                                                                                                                                                       
  └─────┴─────────────────────┴─────────────────┴──────────────────┴────────┘    

## Check list
- [ ] unit-tests written
- [x] e2e-tests written
- [ ] documentation updated


packages/ketcher-core/src/domain/helpers/monomers.ts
```diff
@@ -534,26 +535,83 @@ export function isRnaBaseApplicableForAntisense(monomer?: BaseMonomer) {
 
 export function getAllConnectedMonomersRecursively(
   monomer: BaseMonomer,
+  allowedMonomers?: Set<BaseMonomer>,
```
svvald: `allowedMonomer` name is not quite clear, what makes them "allowed"?
Maybe something like `traversableMonomers` or something else will work better?

packages/ketcher-core/src/domain/helpers/monomers.ts
```diff
@@ -534,26 +535,83 @@ export function isRnaBaseApplicableForAntisense(monomer?: BaseMonomer) {
...
+        if (
+          atomOwnerMonomer.monomerItem.props.isMicromoleculeFragment &&
+          atomOwnerMonomer !== currentMonomer &&
+          (!allowedMonomers || allowedMonomers.has(atomOwnerMonomer))
+        ) {
+          // Use BFS to traverse all connected atoms in the microstructure
+          const atomQueue: Atom[] = [bond.atom];
+
+          while (atomQueue.length > 0) {
+            const currentAtom = atomQueue.shift();
```
svvald, line 577: Is there any reason we use `shift` here while for monomer traversal in the outer loop we use `pop`?

packages/ketcher-core/src/domain/helpers/monomers.ts
```diff
@@ -534,26 +535,83 @@ export function isRnaBaseApplicableForAntisense(monomer?: BaseMonomer) {
...
+                // Found a monomer connection
+                const connectedMonomer = atomBond.monomer;
+                // Only add if it's not the current monomer we're exploring from,
+                // not the microstructure owner, not already visited,
+                // and is in the allowed set (if provided)
+                if (
+                  connectedMonomer !== currentMonomer &&
+                  connectedMonomer !== atomOwnerMonomer &&
+                  !visitedMonomers.has(connectedMonomer) &&
+                  (!allowedMonomers || allowedMonomers.has(connectedMonomer))
```
svvald: This check is already used twice – here and on lines 569-570, maybe make it slightly more readable by hoisting it, introducing some small helper?

packages/ketcher-core/src/domain/helpers/monomers.ts
```diff
@@ -534,26 +535,83 @@ export function isRnaBaseApplicableForAntisense(monomer?: BaseMonomer) {
...
+                // Found a bond to another atom - only traverse if it also belongs to a microstructure
+                const otherAtom =
+                  atomBond.firstAtom === currentAtom
+                    ? atomBond.secondAtom
+                    : atomBond.firstAtom;
+                if (
+                  otherAtom &&
+                  !visitedAtoms.has(otherAtom) &&
+                  otherAtom.monomer.monomerItem.props.isMicromoleculeFragment
+                ) {
```
svvald, line 610: Isn't `!allowedMonomers || allowedMonomers.has(otherAtom.monomer)` missing here?

packages/ketcher-core/src/domain/helpers/monomers.ts
```diff
@@ -534,26 +535,83 @@ export function isRnaBaseApplicableForAntisense(monomer?: BaseMonomer) {
 
 export function getAllConnectedMonomersRecursively(
```
svvald, line 535: This function is also used at `SelectBase` for Alt+drag functionality which selects the whole chain. Without modifying that call site it changes the behavior:

https://github.com/user-attachments/assets/368ec9ab-1a00-4256-8f54-62f7edb8b7f9
https://github.com/user-attachments/assets/9495c6d6-0373-4ef0-9b33-cb44e2b230b4

Original is not quite correct from my perspective as well – I'd expect selecting the whole chain (perhaps a follow-up ticket is needed @AlexeyGirin) but moving the monomers while keeping molecule intact seems like a regression issue to me

packages/ketcher-core/src/domain/helpers/monomers.ts
```diff
@@ -534,26 +535,83 @@ export function isRnaBaseApplicableForAntisense(monomer?: BaseMonomer) {
...
+              if (atomBond instanceof MonomerToAtomBond) {
+                // Found a monomer connection
+                const connectedMonomer = atomBond.monomer;
+                // Only add if it's not the current monomer we're exploring from,
+                // not the microstructure owner, not already visited,
+                // and is in the allowed set (if provided)
+                if (
+                  connectedMonomer !== currentMonomer &&
+                  connectedMonomer !== atomOwnerMonomer &&
+                  !visitedMonomers.has(connectedMonomer) &&
```
svvald: `connectedMonomer !== currentMonomer` looks redundant – `currentMonomer` was already added to `visitedMonomers` on line 552 and `!visitedMonomers.has(connectedMonomer)` covers it

packages/ketcher-core/src/domain/helpers/monomers.ts
```diff
@@ -534,26 +535,83 @@ export function isRnaBaseApplicableForAntisense(monomer?: BaseMonomer) {
...
+          monomerStack.push(anotherMonomer);
+        }
+      } else if (bond instanceof MonomerToAtomBond) {
+        // Handle connections through microstructure atoms
+        // Only traverse atoms that belong to a microstructure (isMicromoleculeFragment)
+        // to avoid connecting unrelated monomers
+        const atomOwnerMonomer = bond.atom.monomer;
+        if (
+          atomOwnerMonomer.monomerItem.props.isMicromoleculeFragment &&
+          atomOwnerMonomer !== currentMonomer &&
```
svvald: I'm not sure if this check is needed – monomers never hold `MonomerToAtomBond` bonds to their own atoms

packages/ketcher-core/src/domain/helpers/monomers.ts
```diff
@@ -534,26 +535,83 @@ export function isRnaBaseApplicableForAntisense(monomer?: BaseMonomer) {
...
 
     currentMonomer.forEachBond((bond) => {
       if (bond instanceof PolymerBond || bond instanceof HydrogenBond) {
         const anotherMonomer = bond.getAnotherMonomer(currentMonomer);
-        if (anotherMonomer && !visited.has(anotherMonomer)) {
-          stack.push(anotherMonomer);
+        if (anotherMonomer && !visitedMonomers.has(anotherMonomer)) {
+          monomerStack.push(anotherMonomer);
+        }
+      } else if (bond instanceof MonomerToAtomBond) {
```
svvald, line 564: Nitpick – this new condition has several levels of nesting and overall inflates `getAllConnectedMonomersRecursively` function and increases its cognitive complexity. Perhaps we could extract micromolecule traversal into separate function here?

packages/ketcher-core/src/domain/helpers/monomers.ts
```diff
@@ -534,26 +535,83 @@ export function isRnaBaseApplicableForAntisense(monomer?: BaseMonomer) {
...
+                // and is in the allowed set (if provided)
+                if (
+                  connectedMonomer !== currentMonomer &&
+                  connectedMonomer !== atomOwnerMonomer &&
+                  !visitedMonomers.has(connectedMonomer) &&
+                  (!allowedMonomers || allowedMonomers.has(connectedMonomer))
+                ) {
+                  monomerStack.push(connectedMonomer);
+                }
+              } else if (atomBond instanceof Bond) {
```
svvald: Atom bonds are either `MonomerToAtomBond` or `Bond` type, I suppose you don't need to check for `instanceof Bond` here, just `else` would be enough

This will also allow to remove `Bond` value import on line 7 which would be better

svvald (changes requested):

packages/ketcher-core/src/domain/helpers/monomers.ts
crossaba: Line 546 (outer loop): monomerStack.pop() - DFS (Depth-First Search)                                                                                                                                                                                                                                              
  Line 575 (inner loop): atomQueue.shift() - BFS (Breadth-First Search)                                                                                                                                                                                                                                             
                                                                                                                                                                                                                                                                                                                    
  Why This Matters                                                                                                                                                                                                                                                                                                  
                                                                                                                                                                                                                                                                                                                    
  Outer loop (Monomer traversal): Uses DFS via pop()                                                                                                                                                                                                                                                                
  - Traverses monomers depth-first                                                                                                                                                                                                                                                                                  
  - Order doesn't matter much for collecting all connected monomers                                                                                                                                                                                                                                                 
  - pop() is O(1) - more efficient than shift()                                                                                                                                                                                                                                                                     
                                                                                                                                                                                                                                                                                                                    
  Inner loop (Atom traversal within microstructure): Uses BFS via shift()                                                                                                                                                                                                                                           
  - Traverses atoms within a microstructure breadth-first                                                                                                                                                                                                                                                           
  - The comment explicitly says: // Use BFS to traverse all connected atoms in the microstructure (line 571)                                                                                                                                                                                                        
  - BFS ensures you explore the microstructure level-by-level                                                                                                                                                                                                                                                       
  - This is important for finding all MonomerToAtomBond connections systematically                                                                                                                                                                                                                                  
                                                                                                                                                                                                                                                                                                                    
  Why BFS for Atoms?                                                                                                                                                                                                                                                                                                
                                                                                                                                                                                                                                                                                                                    
  When traversing atoms in a microstructure fragment, BFS is better because:                                                                                                                                                                                                                                        
  1. Systematic exploration: Finds all immediate atom neighbors before going deeper                                                                                                                                                                                                                                 
  2. Shortest path: If there are multiple paths to connected monomers, BFS finds the shortest                                                                                                                                                                                                                       
  3. Level-by-level processing: More predictable behavior when exploring molecular structures                                                                                                                                                                                                                       
                                                                                                                                                                                                                                                                                                                    
  Performance Note                                                                                                                                                                                                                                                                                                  
                                                                                                                                                                                                                                                                                                                    
  While shift() is O(n) and pop() is O(1), the inner loop (atom traversal) typically processes small microstructure fragments, so the performance difference is negligible compared to the algorithmic benefits of BFS.                                                                                             
                                                                                                                                                                                                                                                                                                                    
  Summary                                                                                                                                                                                                                                                                                                           
                                                                                                                                                                                                                                                                                                                    
  This is intentional mixed-mode traversal:                                                                                                                                                                                                                                                                         
  - DFS for monomers (performance, order doesn't matter)                                                                                                                                                                                                                                                            
  - BFS for atoms within microstructures (systematic exploration, explicitly documented)

packages/ketcher-core/src/domain/helpers/monomers.ts
crossaba: Yep, thank you. Added.

packages/ketcher-core/src/domain/helpers/monomers.ts
crossaba: Renamed

packages/ketcher-core/src/domain/helpers/monomers.ts
crossaba: Good catch! Removed

packages/ketcher-core/src/domain/helpers/monomers.ts
crossaba: You are right. Removed

packages/ketcher-core/src/domain/helpers/monomers.ts
crossaba: Done

packages/ketcher-core/src/domain/helpers/monomers.ts
crossaba: Added helper function isMonomerTraversable(candidateMonomer: BaseMonomer) that returns true if:
  - No traversableMonomers set was provided (traverse all), OR
  - The candidate monomer is in the traversableMonomers set

packages/ketcher-core/src/domain/helpers/monomers.ts
crossaba: I will talk with Alexey. I checked master: this behaviour was added before my changes. I think we could fix it in other PR.

crossaba: Current behaviour is consistent with molecular mode, so it is not regression. Confirmed with Alexey.

svvald (approved):

PR #11298, opened 2026-08-17
#4445 – Macro: Context menu opened outside selection have to remove selection in the same manner as it works in Micro mode

### LINKED ISSUE #4445: Selection and context menu have to be corrected to "standard behaviour" way in the same manner as it works in Micro mode
**Steps to Reproduce**

1. Switch to Macro - Sequence mode
3. Select one or few nucleotides 
4. Click outside the selection 

**Actual behavior**
Selection remain in place
image

**Expected behavior**
Selection disappears (in the same manner as in it happens in Micro mode)
image

**Versions:**
Windows 10

OnurKucukoz:
When the context menu was opened on an empty area of the canvas in Macro Sequence mode, the existing nucleotide selection was passed to the context menu and remained visible.

The selection is now cleared before opening the canvas context menu. Both drawing entities and empty/backbone sequence nodes are unselected, and the context menu receives an empty selection.

Right-clicking a selected nucleotide still preserves the current selection.

### Result after the fix

Right-clicking an empty area of the canvas now clears the current selection in all macromolecule layout modes:

Sequence mode
Flex mode
Snake mode

Right-clicking a selected entity continues to preserve the current selection.

Manually verified in all three layout modes.
Screenshot 2026-08-17 113134

Validation:

Manually verified empty-canvas right-click behavior in Sequence, Flex, and Snake modes
Added unit regression tests for all three layout modes
Existing and new `Editor.test.ts` tests passed: 45/45
TypeScript check passed
ESLint check passed

Closes #4445

## Check list
- [x] unit-tests written
- [ ] e2e-tests written
- [ ] documentation updated

### DIFF
--- a/packages/ketcher-core/src/application/editor/Editor.ts
+++ b/packages/ketcher-core/src/application/editor/Editor.ts
@@ -1153,10 +1153,15 @@ export class CoreEditor {
         ]);
       } else if (isClickOnCanvas) {
         if (this.mode.modeName === 'sequence-layout-mode') {
-          this.events.rightClickCanvasSequence.dispatch([
-            event,
-            sequenceSelections,
-          ]);
+          const modelChanges =
+            this.drawingEntitiesManager.unselectAllDrawingEntities();
+
+          modelChanges.merge(
+            SequenceRenderer.unselectEmptyAndBackboneSequenceNodes(),
+          );
+
+          this.renderersContainer.update(modelChanges);
+          this.events.rightClickCanvasSequence.dispatch([event, []]);
         } else {
           this.events.rightClickCanvas.dispatch([event, selectedMonomers]);
         }
--- a/packages/ketcher-core/src/application/editor/Editor.ts
+++ b/packages/ketcher-core/src/application/editor/Editor.ts
@@ -1074,20 +1074,7 @@ export class CoreEditor {
         return;
       }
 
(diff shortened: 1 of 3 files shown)


packages/ketcher-core/src/application/editor/Editor.ts
```diff
@@ -1153,10 +1153,15 @@ export class CoreEditor {
         ]);
       } else if (isClickOnCanvas) {
         if (this.mode.modeName === 'sequence-layout-mode') {
-          this.events.rightClickCanvasSequence.dispatch([
-            event,
-            sequenceSelections,
-          ]);
+          const modelChanges =
+            this.drawingEntitiesManager.unselectAllDrawingEntities();
+
```
svvald: This change only covers sequence mode – and the original ticket reports it for sequence mode only, but the same behaviour is observed in flex/snake mode as well, don't we want to make it mode-agnostic?

@AlexeyGirin

AlexeyGirin: > This change only covers sequence mode – and the original ticket reports it for sequence mode only, but the same behaviour is observed in flex/snake mode as well, don't we want to make it mode-agnostic?
> 
> @AlexeyGirin

Agree. Let's make same behavior in each canvas mode

PR #11300, opened 2026-08-17
#11101 - Refactor: set-state-in-effect anti-pattern elimination from packages\ketcher-react\src\script\ui\views\modal\components\process\Check\Check.tsx file

### LINKED ISSUE #11101: Refactor: set-state-in-effect anti-pattern elimination from `packages\ketcher-react\src\script\ui\views\modal\components\process\Check\Check.tsx` file
**Problem:** 
1. Duplicate source of truth
2. Extra render cycle:
- Render
- Effect runs
- setState runs
- Second render

**Why is this an issue?**

**How can it be fixed?**

```ts
function Component({data}) {

  useEffect(() => {
  }, [data]);
}

function Component() {

  useEffect(() => {
  }, []);
}

  useEffect(() => {
  }, [rawData]);
}

  useEffect(() => {
  }, [selectedId, items]);
}
```

**Examples of correct code:**
```ts
function Tooltip() {
  const ref = useRef(null);

  useLayoutEffect(() => {
    setTooltipHeight(height);
  }, []);
}

// ✅ Calculate during render
}
```

**Problem locations:**
packages\ketcher-react\src\script\ui\views\modal\components\process\Check\Check.tsx:231

LashaDevLasha:
Removes the react-hooks/set-state-in-effect anti-pattern from packages/ketcher-react/src/script/ui/views/modal/components/process/Check/Check.tsx.

The mount effect previously called handleCheck(), which synchronously called setIsStructureChecking(false) inside a useEffect — the pattern the rule flags. The fix inlines only the async part of handleCheck directly into the effect (the setIsStructureChecking(false) reset is not needed on mount since the state already starts as false).

Also removes the now-dead file-level /* eslint-disable react-hooks/set-state-in-effect */ (added to master after this branch was cut) as part of the rebase.

## Check list
- [ ] unit-tests written
- [ ] e2e-tests written
- [ ] documentation updated

### DIFF
--- a/packages/ketcher-react/src/script/ui/views/modal/components/process/Check/Check.tsx
+++ b/packages/ketcher-react/src/script/ui/views/modal/components/process/Check/Check.tsx
@@ -1,4 +1,3 @@
-/* eslint-disable react-hooks/set-state-in-effect */
 /****************************************************************************
  * Copyright 2021 EPAM Systems
  *
@@ -229,7 +228,13 @@ const CheckDialog: FC<CheckDialogProps> = (props) => {
   const handleSettingsChange = () => setIsCheckedWithNewSettings(false);
 
   useEffect(() => {
-    handleCheck();
+    // Intentionally omits setIsStructureChecking(false): on mount the loading
+    // state starts as false, so there is nothing to reset.
+    onCheck(result.checkOptions).then(() => {
+      setIsStructureChecking(true);
+      setLastCheckDate(new Date());
+      setIsCheckedWithNewSettings(true);
+    });
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, []);
 
--- a/packages/ketcher-react/src/script/ui/views/modal/components/process/Check/Check.tsx
+++ b/packages/ketcher-react/src/script/ui/views/modal/components/process/Check/Check.tsx
@@ -228,7 +228,11 @@ const CheckDialog: FC<CheckDialogProps> = (props) => {
   const handleSettingsChange = () => setIsCheckedWithNewSettings(false);
 
   useEffect(() => {
-    handleCheck();
+    onCheck(result.checkOptions).then(() => {
+      setIsStructureChecking(true);
+      setLastCheckDate(new Date());
+      setIsCheckedWithNewSettings(true);
+    });
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, []);
 


packages/ketcher-react/src/script/ui/views/modal/components/process/Check/Check.tsx
```diff
@@ -228,7 +228,11 @@ const CheckDialog: FC<CheckDialogProps> = (props) => {
   const handleSettingsChange = () => setIsCheckedWithNewSettings(false);
 
   useEffect(() => {
-    handleCheck();
+    onCheck(result.checkOptions).then(() => {
```
la-nuta: This is now a copy of `handleCheck` minus one line, with nothing explaining the difference. Anyone later adding a reset to `handleCheck` silently won't get it on the mount path — a bug that only reproduces on first open. Extract one `runCheck` helper and call it from both; the effect stays rule-compliant.

Related, and closer to what the issue is after: neither call site guards the async writes. Close the dialog mid-request and three setters run on an unmounted component; click Check twice with different options and an out-of-order response pairs an older `lastCheckDate` with `isCheckedWithNewSettings = true`. A `cancelled` flag or request-id ref fixes both — fine as a follow-up.

packages/ketcher-react/src/script/ui/views/modal/components/process/Check/Check.tsx
```diff
@@ -228,7 +228,11 @@ const CheckDialog: FC<CheckDialogProps> = (props) => {
...
+      setIsStructureChecking(true);
+      setLastCheckDate(new Date());
+      setIsCheckedWithNewSettings(true);
+    });
     // eslint-disable-next-line react-hooks/exhaustive-deps
```
la-nuta: nit: unused directive — `exhaustive-deps` is inert in `eslint-plugin-react-hooks@7.1.1`. Pre-existing and repo-wide, so only worth removing since you're here.

packages/ketcher-react/src/script/ui/views/modal/components/process/Check/Check.tsx
la-nuta: This is now a copy of `handleCheck` minus `setIsStructureChecking(false)`, with nothing recording why they differ. The next person to add a reset to `handleCheck` — `setLastCheckDate(null)`, an abort, a `moleculeErrors` clear — silently won't get it on the mount path, and the bug will only reproduce on first open of the dialog, never via the Check button.

One shared helper removes the duplication and still satisfies the rule:

```ts
const runCheck = useCallback(() => {
  return onCheck(result.checkOptions).then(() => {
    setIsStructureChecking(true);
    setLastCheckDate(new Date());
    setIsCheckedWithNewSettings(true);
  });
}, [onCheck, result.checkOptions]);

const handleCheck = () => {
  setIsStructureChecking(false);
  runCheck();
};
```

Then the effect body is just `runCheck();`.

Optional follow-up, pre-existing rather than introduced here: neither call site guards the async writes. Close the dialog before the server responds and three setters run on an unmounted component; click Check twice with different options and an out-of-order response pairs the older `lastCheckDate` with `isCheckedWithNewSettings === true`, so the button reads "already checked with these settings" against errors from a different option set. A `cancelled` flag in the effect cleanup, or a request-id ref checked in the `.then`, closes both.

la-nuta: This is now a copy of `handleCheck` minus `setIsStructureChecking(false)`, with nothing recording why they differ. The next person to add a reset to `handleCheck` — `setLastCheckDate(null)`, an abort, a `moleculeErrors` clear — silently won't get it on the mount path, and it'll only reproduce on first open of the dialog, never via the Check button.

If you want the duplication gone without trading one suppression for another, the ref guard is the only shape I could get to lint clean (verified against the repo config with `set-state-in-effect` live and no disables anywhere in the file):

```ts
const runCheck = useCallback(() => {
  return onCheck(result.checkOptions).then(() => {
    setIsStructureChecking(true);
    setLastCheckDate(new Date());
    setIsCheckedWithNewSettings(true);
  });
}, [onCheck, result.checkOptions]);

const handleCheck = () => {
  setIsStructureChecking(false);
  runCheck();
};

const hasRunInitialCheck = useRef(false);

useEffect(() => {
  if (hasRunInitialCheck.current) {
    return;
  }
  hasRunInitialCheck.current = true;
  runCheck();
}, [runCheck]);
```

Two things to be aware of before adopting it:

- Extracting `runCheck` on its own does **not** let you drop the disable. With `}, []);` the rule just renames the missing dependency: `242:6 error React Hook useEffect has a missing dependency: 'runCheck'`. And `}, [runCheck])` without the guard re-runs the check on every option toggle, because `runCheck` depends on `result.checkOptions` and a toggle builds a new array (`Input.tsx:330-345` → `form.tsx:172-176` → `UPDATE_FORM`). The guard is what makes `[runCheck]` safe.
- It trades a disable comment for a ref. Keeping the current `handleCheck()`-style helper *plus* the existing disable is a perfectly defensible alternative — just not "extract a helper and the disable goes away", which is what I'd have implied otherwise.

Separately, and pre-existing rather than introduced here — but this PR duplicates it into a second call site, so it's worth fixing while you're in the file. The container defines `onCheck: (opts) => dispatch(check(opts)).catch(ownProps.onCancel)`. `.catch()` returns a **resolved** promise, so on a failed check the `.then()` still runs: `onCancel()` closes the modal, then `setIsStructureChecking(true)` / `setLastCheckDate(new Date())` / `setIsCheckedWithNewSettings(true)` fire on an unmounting component, recording a successful check that never happened. Rethrowing after `onCancel`, or moving the success writes into an `onFulfilled` argument rather than a chained `.then()`, fixes it.

la-nuta: Non-blocking: this is now `handleCheck` copied minus `setIsStructureChecking(false)`. It works, but if someone later adds a reset to `handleCheck` — an abort, `setLastCheckDate(null)`, a `moleculeErrors` clear — the mount path silently won't get it, and that only shows up on first open of the dialog. A one-line comment saying the omission is deliberate would be enough.

If you'd rather remove the duplication outright, heads-up that extracting a `runCheck` helper on its own doesn't let you drop the disable — with `}, []);` the rule just renames the missing dep to `runCheck`. The one shape I got to lint clean with no disables at all is a ref guard:

```ts
const hasRunInitialCheck = useRef(false);

useEffect(() => {
  if (hasRunInitialCheck.current) return;
  hasRunInitialCheck.current = true;
  runCheck();
}, [runCheck]);
```

That trades a comment for a ref, so honestly either way is fine — just flagging it so the choice is deliberate.

While you're in here, one pre-existing thing this now has two copies of: `onCheck` is `(opts) => dispatch(check(opts)).catch(ownProps.onCancel)`, and `.catch()` returns a *resolved* promise — so on a failed check the `.then()` still runs and records a success that never happened. Happy for that to be a separate PR if you'd rather keep this one tight.

la-nuta, line 231: Non-blocking: this is `handleCheck` copied minus `setIsStructureChecking(false)`. Fine as-is, but a one-line comment saying the omission is deliberate would help — otherwise a reset added to `handleCheck` later silently won't reach the mount path.

(If you're tempted to extract a shared helper instead: that alone doesn't let you drop the disable — the rule just renames the missing dep to `runCheck`. A `useRef` guard with `}, [runCheck])` is the only clean version, and it trades the comment for a ref, so either way is reasonable.)

la-nuta (commented):
Nice catch that inlining the body is what actually clears the rule here — the synchronous `setIsStructureChecking(false)` was the only thing tripping it. Two things and I think this is good to go.

**1. There's a bonus win available: the file-level `/* eslint-disable react-hooks/set-state-in-effect */` can now be deleted.** It's on line 1 on master (added by #10806 after this branch was cut), so it isn't in the diff, but it comes along on merge. Your change makes it dead — linting the merged shape says so directly:

```
1:1  warning  Unused eslint-disable directive (no problems were reported from 'react-hooks/set-state-in-effect')
```

So once you've picked up master — merge or rebase, whichever you prefer — dropping that line would mean the file is genuinely clean rather than still blanket-suppressed. That feels like the real prize for #11101.

**2. The PR description looks like it came from a sibling PR** — it lists `usePortalOpening`, `TemplateDialog`, `Settings`, `InfoPanel` and "six instances", none of which are in this diff. Worth fixing before squash-merge so it doesn't land in master's history.

One small inline note below, non-blocking.

packages/ketcher-react/src/script/ui/views/modal/components/process/Check/Check.tsx
LashaDevLasha: Done — added a comment on the setIsStructureChecking(false) omission: initial state is already false, so no reset needed on mount.

la-nuta (approved):
Deleting line 1 is the part that actually closes #11101 — thanks. I confirmed the file is genuinely guarded now rather than just quieter: re-injecting a synchronous `setIsStructureChecking(false)` into the effect reproduces `234:5 error Calling setState synchronously within an effect can trigger cascading renders`. And keeping the `exhaustive-deps` disable was the right call — removing it errors on `onCheck` and `result.checkOptions`, and listing those would re-fire the check on every option toggle.

Approving. Non-blocking: the PR description still describes a different changeset (`usePortalOpening`, `TemplateDialog`, `Settings`, `InfoPanel`, "six instances") — worth replacing before squash-merge so master's history matches what landed.

packages/ketcher-react/src/script/ui/views/modal/components/process/Check/Check.tsx
```diff
@@ -229,7 +228,13 @@ const CheckDialog: FC<CheckDialogProps> = (props) => {
   const handleSettingsChange = () => setIsCheckedWithNewSettings(false);
 
   useEffect(() => {
-    handleCheck();
```
AlexeyGirin, line 232: Keep it as it is, but make suppression local

AlexeyGirin (dismissed):

rrodionov91 (approved):

PR #11306, opened 2026-08-17
#7367 - Only bonds remain on canvas after applying 3D Viewer to structures marked as nucleotide components

### LINKED ISSUE #7367: Only bonds remain on canvas after applying 3D Viewer to structures marked as nucleotide components
**Description:**
When using the 3D Viewer on structures marked as nucleotide components (e.g. Base, Sugar, Phosphate), and clicking Apply in the 3D Viewer window, the canvas incorrectly updates: all atoms and labels disappear, and only bonds remain.
This behavior is inconsistent with how S-groups behave: upon applying 3D, expanded structures are correctly collapsed into their abbreviated (contracted) form.

**Steps to Reproduce**
1. Open structure [3-benzene-rings.zip](https://github.com/user-attachments/files/21018400/3-benzene-rings.zip)
2. Click the 3D Viewer button on the top panel.
3. In the opened viewer, click Apply.
4. Observe the canvas result.

**Actual behavior**
The canvas shows only the bonds - all atoms and labels are lost, resulting in broken visual representation.
**Expected behavior**
Marked nucleotide structures should behave like S-groups - they should collapse into their abbreviated labels on the canvas after applying 3D.

**Screenshots**

https://github.com/user-attachments/assets/c43222ed-317c-4531-bda2-e8774967d402

**Desktop:**
 - OS: Windows 11
 - Browser Chrome
 Version 137.0.7151.104 (Official Build) (64-bit)

**Ketcher version** 
[Version 3.6.0-rc.1]
**Indigo version** 
[Version 1.34.0-rc.1]

This bug was found while testing: https://github.com/epam/ketcher/issues/7164

HelenaIsh:

Fixed an issue where applying 3D changes to structures containing `Sugar`,
`Base`, or `Phosphate` groups could remove atom and group labels, leaving only
bonds on the canvas.

### Root cause

Apply previously replaced the Ketcher structure with CML returned by Miew.
CML preserves atoms, bonds, and coordinates, but not all Ketcher-specific
S-group metadata required for rendering abbreviated groups.

### What changed

- Preserve the original Ketcher structure and apply only atom coordinates from
  Miew.
- Preserve S-group metadata, including type, class, name, atom membership,
  brackets, and attachment data.
- Keep the existing behavior of collapsing expanded `SUP` groups after Apply.
- Resolve empty nucleotide group names using their class labels.
- Reset internal initial-selection state on the cloned structure before loading
  it, allowing repeated Apply operations.

| Before 3D | After 3D |
| --- | --- |
| image | image |
| image | image |

## Check list
- [x] unit-tests written
- [ ] e2e-tests written
- [ ] documentation updated


packages/ketcher-core/src/application/render/restruct/resgroup.ts
```diff
@@ -55,6 +55,19 @@ export const SUPERATOM_CLASS_TEXT = {
   [SUPERATOM_CLASS.PHOSPHATE]: 'Phosphate',
 };
 
+export function getSuperatomLabel(sgroup: {
+  data: { name?: string | null; class?: string | null };
```
svvald: There is no sense in this ad-hoc type here, either `SGroup` or `SgroupData` from core domain package should work

packages/ketcher-core/src/application/render/restruct/resgroup.ts
```diff
@@ -55,6 +55,19 @@ export const SUPERATOM_CLASS_TEXT = {
...
+}): string {
+  const superatomClass = sgroup.data?.class as
+    | keyof typeof SUPERATOM_CLASS_TEXT
+    | undefined;
```
svvald: While we're here, perhaps we could introduce a type guard like that:
```
const isSuperatomClass = (value?: string): value is SUPERATOM_CLASS =>
  Boolean(value) && value in SUPERATOM_CLASS_TEXT;
```

This will also help to avoid additional `|| ''` below

packages/ketcher-react/src/script/ui/views/modal/components/process/Miew/miewStructMerge.test.ts
```diff
@@ -3,16 +3,34 @@ import {
...
+};
+function resolveSuperatomLabel(sgroup: {
+  data: { name?: string | null; class?: string | null };
+}): string {
+  const cls = sgroup.data?.class ?? undefined;
+  return (
+    sgroup.data?.name?.trim() || (cls ? SUPERATOM_CLASS_TEXT[cls] : '') || ''
+  );
+}
+
```
svvald: We don't need to introduce a custom implementation for the test and import the actual implementation from working code

packages/ketcher-react/src/script/ui/views/modal/components/process/Miew/miewStructMerge.test.ts
```diff
@@ -3,16 +3,34 @@ import {
...
   mergeMetaObjects,
   needsMetaPreservation,
+  needsStructurePreservation,
 } from './miewStructMerge';
 
+const SUPERATOM_CLASS_TEXT: Record<string, string> = {
+  SUGAR: 'Sugar',
+  BASE: 'Base',
+  PHOSPHATE: 'Phosphate',
+};
```
svvald: And you won't need to re-introduce this map as well

packages/ketcher-react/src/script/ui/views/modal/components/process/Miew/miewStructMerge.ts
```diff
@@ -35,6 +35,34 @@ export function needsMetaPreservation(struct: Struct): boolean {
...
+  );
+
+  let index = 0;
+  original.atoms.forEach((atom) => {
+    const newPosition = resultPositions[index];
+    if (newPosition) {
+      atom.pp = new Vec2(newPosition.x, newPosition.y, newPosition.z);
+    }
+    index += 1;
+  });
```
svvald: I'm not sure that the index-based approach is 100% stable, for additional safety I'd propose to at least check for atom label equality per index and return false on mismatch

packages/ketcher-react/src/script/ui/views/modal/components/process/Miew/miewStructMerge.ts
```diff
@@ -35,6 +35,34 @@ export function needsMetaPreservation(struct: Struct): boolean {
...
+export function needsStructurePreservation(struct: Struct): boolean {
+  return struct.sgroups.size > 0 || struct.rgroups.size > 0;
+}
+
+export function mergeCoordinatesFromResult(
+  original: Struct,
+  result: Struct,
+): boolean {
+  if (original.atoms.size !== result.atoms.size) {
+    return false;
```
svvald, line 58: For each `false` return I'd propose not to fail silently but to log some warning message

packages/ketcher-react/src/script/ui/views/modal/components/process/Miew/Miew.tsx
```diff
@@ -185,6 +188,26 @@ const MiewDialog = ({
       return;
     }
 
+    if (needsStructurePreservation(struct)) {
+      try {
+        const result = await parseStruct(cmlStruct, server, serverSettings);
```
svvald: This parsing is done here and if we perform an early return, it will do that again (line 217), I'd suggest hoisting the struct parsing above

packages/ketcher-react/src/script/ui/views/modal/components/process/Miew/Miew.tsx
svvald: This existing branch called `rescale`, I do not see the reason why we shouldn't do that in the new branch, either include it or specify the reason for that

packages/ketcher-react/src/script/ui/views/modal/components/process/Miew/Miew.tsx
```diff
@@ -185,6 +188,26 @@ const MiewDialog = ({
...
+        if (mergeCoordinatesFromResult(preserved, result)) {
+          collapseExpandedSuperatoms(preserved);
+          dispatch(
+            load(preserved, { preserveViewport: true, skipCenter: true }),
+          );
+          return;
+        }
+      } catch (e) {
+        KetcherLogger.error('Miew.tsx::MiewDialog::exportCML', e);
+        return;
```
svvald: This catch doesn't have any visual error indication – the modal stays open and nothing happens on canvas – either let it fall through the following path or surface the error to the user

packages/ketcher-react/src/script/ui/views/modal/components/process/Miew/Miew.tsx
```diff
@@ -185,6 +188,26 @@ const MiewDialog = ({
...
+        preserved.enableInitiallySelected();
+        if (mergeCoordinatesFromResult(preserved, result)) {
+          collapseExpandedSuperatoms(preserved);
+          dispatch(
+            load(preserved, { preserveViewport: true, skipCenter: true }),
+          );
+          return;
+        }
+      } catch (e) {
+        KetcherLogger.error('Miew.tsx::MiewDialog::exportCML', e);
```
svvald: Also this logger string ideally should be different from the on on line 223

packages/ketcher-core/src/application/render/restruct/resgroup.ts
```diff
@@ -55,6 +55,19 @@ export const SUPERATOM_CLASS_TEXT = {
   [SUPERATOM_CLASS.PHOSPHATE]: 'Phosphate',
 };
 
+export function getSuperatomLabel(sgroup: {
```
svvald: Also perhaps it's not a good place for this functionality – I'd rather make it a getter of `Sgroup` entity class

packages/ketcher-react/src/script/ui/views/modal/components/process/Miew/miewStructMerge.ts
```diff
@@ -35,6 +35,34 @@ export function needsMetaPreservation(struct: Struct): boolean {
...
+  }
+
+  const resultPositions = Array.from(result.atoms.values()).map(
+    (atom) => atom.pp,
+  );
```
svvald: I don't think you need this intermediate array, you can get a corresponding atom from `result.atoms` pool while iterating over `original.atoms` directly

packages/ketcher-react/src/script/ui/views/modal/components/process/Miew/miewStructMerge.ts
```diff
@@ -35,6 +35,34 @@ export function needsMetaPreservation(struct: Struct): boolean {
   );
 }
 
+export function needsStructurePreservation(struct: Struct): boolean {
+  return struct.sgroups.size > 0 || struct.rgroups.size > 0;
+}
+
+export function mergeCoordinatesFromResult(
+  original: Struct,
+  result: Struct,
```
svvald: Nitpick – the other functions in this file use `result – original` order of params, here you have the opposite, let's align with the rest

packages/ketcher-react/src/script/ui/views/modal/components/process/Miew/Miew.tsx
```diff
@@ -185,6 +188,26 @@ const MiewDialog = ({
...
+      try {
+        const result = await parseStruct(cmlStruct, server, serverSettings);
+        alignToCentroid(result, struct);
+
+        const preserved = struct.clone();
+        preserved.enableInitiallySelected();
+        if (mergeCoordinatesFromResult(preserved, result)) {
+          collapseExpandedSuperatoms(preserved);
+          dispatch(
+            load(preserved, { preserveViewport: true, skipCenter: true }),
```
svvald: `struct` can contain monomers transformations (rotation, flipping), which are already applied on initial load and calling the `load` on `preserved` being a `struct` clone might apply them again thus having an incorrect result on canvas – please check if it works correctly

packages/ketcher-react/src/script/ui/views/modal/components/process/Miew/miewStructMerge.ts
```diff
@@ -64,6 +92,14 @@ export function alignToCentroid(result: Struct, original: Struct): void {
   });
 }
 
+export function collapseExpandedSuperatoms(struct: Struct): void {
+  struct.sgroups.forEach((sgroup) => {
+    if (sgroup.type === SGroup.TYPES.SUP) {
+      sgroup.data.expanded = false;
+    }
+  });
```
svvald, line 136: This only works for regular s-groups but not for monomers which are wrappers around `SUP` s-groups:
1. Internal `monomerItem`'s `expanded` property is not updated this way
2. It doesn't handle the monomer position before and after contraction 

Either skip monomers and leave them expanded (but then check that the original case works fine for them) or apply the changes for monomers properly

svvald (changes requested):

packages/ketcher-core/src/application/render/restruct/resgroup.ts
HelenaIsh: not relevant after extracting getSuperatomLabel into SGroup getter

packages/ketcher-core/src/application/render/restruct/resgroup.ts
HelenaIsh: added a type guard

packages/ketcher-react/src/script/ui/views/modal/components/process/Miew/miewStructMerge.test.ts
HelenaIsh: you are absolutely right, fixed it

packages/ketcher-react/src/script/ui/views/modal/components/process/Miew/miewStructMerge.test.ts
HelenaIsh: sure, removed

packages/ketcher-react/src/script/ui/views/modal/components/process/Miew/miewStructMerge.ts
HelenaIsh: added label check

packages/ketcher-react/src/script/ui/views/modal/components/process/Miew/miewStructMerge.ts
HelenaIsh: added KetcherLogger

packages/ketcher-react/src/script/ui/views/modal/components/process/Miew/Miew.tsx
HelenaIsh: keep only one parsing

packages/ketcher-react/src/script/ui/views/modal/components/process/Miew/Miew.tsx
HelenaIsh: not relevant as we keep only one branch with parser

packages/ketcher-react/src/script/ui/views/modal/components/process/Miew/Miew.tsx
HelenaIsh: added showSnackbarNotification

packages/ketcher-react/src/script/ui/views/modal/components/process/Miew/Miew.tsx
HelenaIsh: fixed

packages/ketcher-core/src/application/render/restruct/resgroup.ts
HelenaIsh: moved to Sgroup

packages/ketcher-react/src/script/ui/views/modal/components/process/Miew/miewStructMerge.ts
HelenaIsh: fixed

packages/ketcher-react/src/script/ui/views/modal/components/process/Miew/miewStructMerge.ts
HelenaIsh: fixed

packages/ketcher-react/src/script/ui/views/modal/components/process/Miew/Miew.tsx
HelenaIsh: I checked manually by rotating and flipping monomers - then apply 3D - it works fine

packages/ketcher-react/src/script/ui/views/modal/components/process/Miew/miewStructMerge.ts
HelenaIsh: fixed

packages/ketcher-react/src/script/ui/views/modal/components/process/Miew/miewStructMerge.ts
```diff
@@ -64,6 +122,17 @@ export function alignToCentroid(result: Struct, original: Struct): void {
   });
 }
 
+export function collapseExpandedSuperatoms(struct: Struct): void {
+  struct.sgroups.forEach((sgroup) => {
+    if (sgroup.type === SGroup.TYPES.SUP) {
+      if (sgroup instanceof MonomerMicromolecule && sgroup.monomer) {
```
svvald: Is `sgroup.monomer` check actually needed?

packages/ketcher-react/src/script/ui/views/modal/components/process/Miew/miewStructMerge.ts
```diff
@@ -64,6 +122,17 @@ export function alignToCentroid(result: Struct, original: Struct): void {
...
+        sgroup.monomer.monomerItem.expanded = false;
```
svvald: Monomer item sourced from RTK store and is immer-frozen so bare assignment might throw an error in runtime. Please is `Object.isFrozen` guard – the same as `SGroupAttr` or `SGroupCreate` use

svvald (changes requested):

packages/ketcher-react/src/script/ui/views/modal/components/process/Miew/miewStructMerge.ts
HelenaIsh: I removed it - it's redundant as we check for MonomerMicromolecule instance

packages/ketcher-react/src/script/ui/views/modal/components/process/Miew/miewStructMerge.ts
HelenaIsh: added Object.isFrozen check

HelenaIsh: I also decided to extract this logic from three places into one

svvald (approved):

PR #11307, opened 2026-08-17
#11106 - Refactor: set-state-in-effect anti-pattern elimination from packages\ketcher-react\src\script\ui\views\components\MonomerCreationWizard\MonomerCreationWizard.tsx file

### LINKED ISSUE #11106: Refactor: set-state-in-effect anti-pattern elimination from `packages\ketcher-react\src\script\ui\views\components\MonomerCreationWizard\MonomerCreationWizard.tsx` file
**Problem:** 
1. Duplicate source of truth
2. Extra render cycle:
- Render
- Effect runs
- setState runs
- Second render

**Why is this an issue?**

**How can it be fixed?**

```ts
function Component({data}) {

  useEffect(() => {
  }, [data]);
}

function Component() {

  useEffect(() => {
  }, []);
}

  useEffect(() => {
  }, [rawData]);
}

  useEffect(() => {
  }, [selectedId, items]);
}
```

**Examples of correct code:**
```ts
function Tooltip() {
  const ref = useRef(null);

  useLayoutEffect(() => {
    setTooltipHeight(height);
  }, []);
}

// ✅ Calculate during render
}
```

**Problem locations:**
packages\ketcher-react\src\script\ui\views\components\MonomerCreationWizard\MonomerCreationWizard.tsx:1139
packages\ketcher-react\src\script\ui\views\components\MonomerCreationWizard\MonomerCreationWizard.tsx:1201

ALexanderSPB:

## Check list
- [ ] unit-tests written
- [ ] e2e-tests written
- [ ] documentation updated

### DIFF
--- a/packages/ketcher-react/src/script/ui/views/components/MonomerCreationWizard/MonomerCreationWizard.tsx
+++ b/packages/ketcher-react/src/script/ui/views/components/MonomerCreationWizard/MonomerCreationWizard.tsx
@@ -100,26 +100,43 @@ const initialWizardState: WizardState = getInitialWizardState();
 
 /**
  * Builds initial wizard state seeded with values from an existing monomer
- * being edited. When `initialValues` is undefined returns the default empty
- * wizard state.
+ * being edited, and with the "default attachment points" notification when
+ * the editor auto-assigned them. Both inputs are fixed for the lifetime of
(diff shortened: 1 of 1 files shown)


AlexeyGirin:
Duplicate

PR #11309, opened 2026-08-18
 #11309 Refactor: react-hooks/exhaustive-deps rule violation clean up from packages\ketcher-macromolecules\src\components\monomerLibrary\RnaBuilder\RnaEditor\RnaEditor.tsx file

### LINKED ISSUE #11288: Refactor: `react-hooks/exhaustive-deps` rule violation clean up from `packages\ketcher-macromolecules\src\components\monomerLibrary\RnaBuilder\RnaEditor\RnaEditor.tsx` file

Example:
```ts

useEffect(() => {
  loadUser(userId);
}, []); // userId is missing
```

**Why is this a problem?**

This can cause:
- stale state or props;
- effects using old data;

**How is it usually fixed?**
```ts
useEffect(() => {
  loadUser(userId);
}, [userId]);
```

For a memoized callback:
```ts
}, [documentId, content]);
```

For a memoized value:
```ts
  [items, category],
);
```

**Problem locations:**
packages\ketcher-macromolecules\src\components\monomerLibrary\RnaBuilder\RnaEditor\RnaEditor.tsx

LashaDevLasha:
…ustive-deps rule

## Check list
- [ ] unit-tests written
- [ ] e2e-tests written
- [ ] documentation updated

### DIFF
--- a/packages/ketcher-macromolecules/src/components/monomerLibrary/RnaBuilder/RnaEditor/RnaEditor.tsx
+++ b/packages/ketcher-macromolecules/src/components/monomerLibrary/RnaBuilder/RnaEditor/RnaEditor.tsx
@@ -45,14 +45,17 @@ export const RnaEditor = ({ duplicatePreset }) => {
   const [expanded, setExpanded] = useState(false);
 
   useEffect(() => {
-    if (activePreset) {
-      if (activePreset.name || isEditMode) setExpanded(true);
-      return;
+    if (activePreset?.name || isEditMode) {
+      setExpanded(true);
     }
+  }, [activePreset, isEditMode]);
 
-    dispatch(createNewPreset());
-    dispatch(setActiveRnaBuilderItem(RnaBuilderPresetsItem.Presets));
-  }, [activePreset, isEditMode, dispatch]);
+  useEffect(() => {
+    if (!activePreset) {
+      dispatch(createNewPreset());
+      dispatch(setActiveRnaBuilderItem(RnaBuilderPresetsItem.Presets));
+    }
+  }, [activePreset, dispatch]);
 
   useEffect(() => {
     dispatch(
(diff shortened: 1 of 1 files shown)


packages/ketcher-macromolecules/src/components/monomerLibrary/RnaBuilder/RnaEditor/RnaEditor.tsx
```diff
@@ -52,13 +52,13 @@ export const RnaEditor = ({ duplicatePreset }) => {
 
     dispatch(createNewPreset());
     dispatch(setActiveRnaBuilderItem(RnaBuilderPresetsItem.Presets));
-  }, [activePreset]);
+  }, [activePreset, isEditMode, dispatch]);
```
svvald: We shouldn't just mindlessly add missing values to dependency array – if something was missing there, apparently the ideas was to trigger an effect on particular reactive properties change which was a useEffect misuse in the first place

Here in particular we also violate set-state-in-effect rule even though eslint doesn't highlight it for some reason, I suppose `if (activePreset)` part could be safely extracted from the effect thus only making an effect to bootstrap an empty preset

packages/ketcher-macromolecules/src/components/monomerLibrary/RnaBuilder/RnaEditor/RnaEditor.tsx
```diff
@@ -52,13 +52,13 @@ export const RnaEditor = ({ duplicatePreset }) => {
...
 
   useEffect(() => {
     dispatch(
       recalculateRnaBuilderValidations({ rnaPreset: activePreset, isEditMode }),
     );
-  }, [isEditMode]);
+  }, [isEditMode, dispatch, activePreset]);
```
svvald, line 67: This effect is probably fine now, I'd rethink whether we even need to trigger some dispatch actions in effect, perhaps there is more straightforward way to do that, but it's up for the further refactoring

svvald (changes requested):

svvald (approved):

PR #11317, opened 2026-08-18
#11291 Refactor: react-hooks/exhaustive-deps rule violation clean up from packages\ketcher-macromolecules\src\EditorEvents.tsx file

### LINKED ISSUE #11291: Refactor: `react-hooks/exhaustive-deps` rule violation clean up from `packages\ketcher-macromolecules\src\EditorEvents.tsx` file

Example:
```ts

useEffect(() => {
  loadUser(userId);
}, []); // userId is missing
```

**Why is this a problem?**

This can cause:
- stale state or props;
- effects using old data;

**How is it usually fixed?**
```ts
useEffect(() => {
  loadUser(userId);
}, [userId]);
```

For a memoized callback:
```ts
}, [documentId, content]);
```

For a memoized value:
```ts
  [items, category],
);
```

**Problem locations:**
packages\ketcher-macromolecules\src\EditorEvents.tsx

LashaDevLasha:
…deps rule

## Check list
- [ ] unit-tests written
- [ ] e2e-tests written
- [ ] documentation updated


packages/ketcher-macromolecules/src/EditorEvents.tsx
```diff
@@ -144,15 +142,15 @@ export const EditorEvents = () => {
...
     (payload) => dispatch(showPreview(payload)),
     [dispatch],
   );
 
-  const debouncedShowPreview = useCallback(
-    debounce((p) => dispatchShowPreview(p), 500),
+  const debouncedShowPreview = useMemo(
+    () => debounce((p) => dispatchShowPreview(p), 500),
     [dispatchShowPreview],
   );
```
svvald: Same comments as in https://github.com/epam/ketcher/pull/11313#pullrequestreview-4971383250:

Inline `dispatchShowPreview`, add a cleanup effect

svvald: And I think we have the same issue in couple of files thus it's going to be duplicated, maybe we could introduce some reusable hook `useDebouncedShowPreview` that you could use here and in all other refactoring tasks

packages/ketcher-macromolecules/src/EditorEvents.tsx
svvald: All these handlers are added to editor events subscriptions but they are never cleaned up, let's create named event handlers and clean them up in useEffect's return

svvald, line 342: This event handler is also not cleaned up, let's do that while we're here

packages/ketcher-macromolecules/src/EditorEvents.tsx
```diff
@@ -392,13 +390,19 @@ export const EditorEvents = () => {
...
+    handleClosePreview,
+    handleOpenAtomLabelTooltip,
+  ]);
 
   useEffect(() => {
     if (!hasAtLeastOneAntisense) {
       editor?.events.resetSequenceEditMode.dispatch();
     }
-  }, [hasAtLeastOneAntisense]);
+  }, [hasAtLeastOneAntisense, editor?.events.resetSequenceEditMode]);
```
svvald: Can we use just `editor` in effect deps? While `editor` is stable, `resetSequenceEditMode` might change on editor initialization thus calling this effect one additional time unnecesarily

svvald (changes requested):

svvald (approved):

PR #11318, opened 2026-08-18
#3269 - When click on the dividing lines, the context menu disappears

### LINKED ISSUE #3269: When click on the dividing lines, the context menu disappears
**Steps to Reproduce**
1.	Launch "Ketcher"
2.	Add Chain structure to the canvas
3.	Right-click on any bond of this structure
4.	Click on any dividing line in the context menu

**Actual behavior**
The context menu disappears

**Expected behavior**
The context menu is displayed on the canvas

**Screenshots**
2023-09-05_16h31_36

 - Browser: Сhrome

**Ketcher version** 2.15.0-rc.1

PaulinaKacz:
Fix: Context menu closes when clicking on separators or padding                                                                                
   
**Problem**                                                                                                                                                      
The context menu in the macromolecules editor closes unexpectedly when clicking on:
  - Separator lines between menu items
  - Padding/empty space within the menu container
  - Any area inside the menu that isn't a clickable menu item

**Root Cause**
The handleContextMenuClose event listener in ContextMenu.tsx was closing the menu on ANY click event, without checking if the click occurred inside the menu itself. This caused clicks on non-interactive elements (separators, padding) to bubble up and trigger the close handler.

**Solution**
Added a capture-phase event listener that intercepts clicks before react-contexify's own handlers:

  const preventCloseOnMenuClick = (e: MouseEvent) => {
    const clickedElement = e.target;
    if (!(clickedElement instanceof Element)) return;

    const isClickInsideMenu = clickedElement.closest('.contexify');
    const isClickOnMenuItem = clickedElement.closest('.contexify_item');

    if (isClickInsideMenu && !isClickOnMenuItem) {
      e.stopPropagation();
    }
  };

**Why this approach:**
  1. Capture phase is essential: The listener must run in capture phase (addEventListener(..., true)) to intercept events               BEFORE react-contexify's bubble-phase listener that closes the menu on outside clicks.
  2. Selective propagation stopping: We only stop propagation for clicks inside the menu (.contexify) but NOT on menu items (.contexify_item).
  This ensures:
    - Separators and padding don't close the menu
    - Menu items still work correctly (execute actions and close menu)
  3. Both click and mousedown: Handle both events because react-contexify uses both for different purposes.
  4. Type safety: Proper type guard with instanceof Element ensures safe use of DOM methods.

## Check list
- [ ] unit-tests written
- [ ] e2e-tests written
- [ ] documentation updated


packages/ketcher-macromolecules/src/components/contextMenu/ContextMenu.tsx
```diff
@@ -140,13 +158,21 @@ export const ContextMenu = ({ id, handleMenuChange, menuItems }: MenuProps) => {
       }
       dispatch(setContextMenuActive(false));
     };
+
+    // Add listener in capture phase (runs before react-contexify's listeners)
+    document.addEventListener('click', preventCloseOnMenuClick, true);
+    document.addEventListener('mousedown', preventCloseOnMenuClick, true);
+
```
svvald, line 165: There is no need to attach these event handlers for document, you can just provide the following function as `onClick` property for `StyledMenu`

```
if (!(event.target as Element).closest('.contexify_item')) {
      event.stopPropagation();
    }
```

svvald (changes requested):
Also please note that your fix only covers macro mode, there was an attempt to fix it for micro mode: https://github.com/epam/ketcher/pull/10481/changes which didn't work – the ticket was reopened for the same issue in micro, not macro

I'd also advise to use the same approach with providing a property for menu wrapper in micro mode and dispose of `MenuSeparator` wrapper

Also, there is a `Divider` component in `HighlightColor` submenu which is just styled `div` so the proposed fix won't work for it, I'd suggest to swap it to `react-contexify`'s `Separator` instead

packages/ketcher-macromolecules/src/components/contextMenu/ContextMenu.tsx
PaulinaKacz: After implementing the component-level approach, I found it wasn't scalable:

For now it works for both macro and micro mode.

  **Problem**: ketcher-react has 8 different Menu components, and the fix needs to work across BOTH packages (ketcher-react and
  ketcher-macromolecules). Adding onClickCapture to each Menu would mean:
  - Duplicating code 8+ times
  - Easy to miss when adding new menus
  - Inconsistent if we forget one

  **Solution**: The global document listener approach is more appropriate here because:
  1. ✅ **Single source of truth**: One listener handles ALL context menus
  2. ✅ **Consistent with react-contexify**: The library itself uses global listeners
  3. ✅ **Scalable**: Works automatically for any new menu added

  I've kept the original global approach but added the Divider→Separator fix you mentioned. 
  
  Do you want me to delete this global approach and add onClickCapture to each Menu?

PR #11319, opened 2026-08-18
#11292 Refactor: react-hooks/exhaustive-deps rule violation clean up from packages\ketcher-macromolecules\src\Editor.tsx file

### LINKED ISSUE #11292: Refactor: `react-hooks/exhaustive-deps` rule violation clean up from `packages\ketcher-macromolecules\src\Editor.tsx` file

Example:
```ts

useEffect(() => {
  loadUser(userId);
}, []); // userId is missing
```

**Why is this a problem?**

This can cause:
- stale state or props;
- effects using old data;

**How is it usually fixed?**
```ts
useEffect(() => {
  loadUser(userId);
}, [userId]);
```

For a memoized callback:
```ts
}, [documentId, content]);
```

For a memoized value:
```ts
  [items, category],
);
```

**Problem locations:**
packages\ketcher-macromolecules\src\Editor.tsx

LashaDevLasha:
…ive-deps rule

## Check list
- [ ] unit-tests written
- [ ] e2e-tests written
- [ ] documentation updated

### DIFF
--- a/packages/ketcher-macromolecules/src/Editor.tsx
+++ b/packages/ketcher-macromolecules/src/Editor.tsx
@@ -213,6 +213,19 @@ function Editor({
     id: CONTEXT_MENU_ID.FOR_SELECTED_MONOMERS,
   });
 
+  const showSequenceContextMenuRef = useRef(showSequenceContextMenu);
+  useEffect(() => {
+    showSequenceContextMenuRef.current = showSequenceContextMenu;
+  }, [showSequenceContextMenu]);
+
+  const showSelectedMonomersContextMenuRef = useRef(
+    showSelectedMonomersContextMenu,
+  );
+  useEffect(() => {
+    showSelectedMonomersContextMenuRef.current =
+      showSelectedMonomersContextMenu;
+  }, [showSelectedMonomersContextMenu]);
+
   const editorInitParamsRef = useRef({
     ketcherId,
     theme,
@@ -262,13 +275,13 @@ function Editor({
     const onRightClickSequence = ([event, selections]: [
       PointerEvent,
       NodeSelection[][],
-    ]) => {
+    ]): void => {
       setSelections(selections);
       setContextMenuEvent(event);
       updatePasteAvailability();
       window.dispatchEvent(new Event('hidePreview'));
       dispatch(setContextMenuActive(true));
-      showSequenceContextMenu({
+      showSequenceContextMenuRef.current({
         event,
         props: {
           sequenceItemRenderer: event.target.__data__,
@@ -281,7 +294,7 @@ function Editor({
     ]): void => {
       setContextMenuEvent(event);
       setSelectedMonomers([]);
-      showSelectedMonomersContextMenu({
+      showSelectedMonomersContextMenuRef.current({
         event,
         props: {
           polymerBondRenderer,
@@ -291,11 +304,11 @@ function Editor({
     const onRightClickSelectedMonomers = ([event, selectedMonomers]: [
       PointerEvent,
       BaseMonomer[],
-    ]) => {
+    ]): void => {
       setSelectedMonomers(selectedMonomers);
       setContextMenuEvent(event);
       updatePasteAvailability();
-      showSelectedMonomersContextMenu({
+      showSelectedMonomersContextMenuRef.current({
         event,
         props: { selectedMonomers },
       });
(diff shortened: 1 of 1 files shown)


dorogusha (commented):
Nice cleanup overall — extracting the inline listeners into named handlers and adding `.remove()` in the effect's cleanup function is a real fix, not just cosmetic: previously the listeners were only ever `.add()`ed, so they'd accumulate every time `editor`'s identity changed.

Two things worth a look before merging:

- The file still has `/* eslint-disable react-hooks/exhaustive-deps */` at line 1 (untouched by this PR), so this fix isn't actually enforced by ESLint/CI going forward. Worth removing it (or narrowing it to whatever specific hook, if any, still needs it) now that the effects here are meant to be exhaustive-deps compliant — otherwise a future edit can silently reintroduce a violation with nothing to catch it.
- Per the branch's commit history, the first attempt at this exact fix (`830916149f`) added `theme`/`ketcherId`/etc. directly to the `createEditor` effect's deps, which caused a destroy/create loop that froze the app (`theme` is recreated every render via `merge(defaultTheme, theme)`). That was caught and fixed in the very next commit via the `editorInitParamsRef` pattern — good catch, but it means this PR is more than a mechanical lint cleanup, and there's no test guarding against that regression being reintroduced. Might be worth calling that out in the PR description and adding a small regression test.

packages/ketcher-macromolecules/src/Editor.tsx
```diff
@@ -255,64 +274,94 @@ function Editor({
...
+        onRightClickCanvasSequence,
+      );
+      editor?.events.toggleMacromoleculesPropertiesVisibility.remove(
+        onToggleMacromoleculesPropertiesVisibility,
+      );
+    };
+  }, [
+    editor,
+    dispatch,
+    showSequenceContextMenu,
```
dorogusha: `showSequenceContextMenu` / `showSelectedMonomersContextMenu` come from `useContextMenu()` (react-contexify). That hook returns a brand-new object with brand-new closures on every render — there's no `useMemo`/`useCallback` inside react-contexify's implementation. Adding them to this effect's deps means all 6 listeners are now torn down and re-registered on **every render** of `Editor`, not just when `editor` changes as before. Since these same handlers call `setSelections`/`setContextMenuEvent`/`setSelectedMonomers`/`setIsPasteAvailable`, every right-click causes a re-render that re-triggers this churn (each `.add`/`.remove` on `Subscription` allocates new arrays).

Since these `show` functions don't actually capture any reactive value that could go stale, consider applying the same ref-capture pattern used for `editorInitParamsRef` above (stash the latest `show` fn in a ref via a layout effect or a plain assignment during render, and call `ref.current(...)` from the handlers) so they don't need to be in this deps array at all.

packages/ketcher-macromolecules/src/Editor.tsx
```diff
@@ -243,7 +259,10 @@ function Editor({
   useMacromoleculesHotkeys();
 
   useEffect(() => {
-    editor?.events.rightClickSequence.add(([event, selections]) => {
+    const onRightClickSequence = ([event, selections]: [
```
dorogusha, line 275: Minor: `onRightClickSequence`'s destructured params aren't explicitly typed (`[event, selections]`), unlike the other extracted handlers right below it (e.g. `onRightClickPolymerBond` has an explicit `[PointerEvent, DeprecatedFlexModeOrSnakeModePolymerBondRenderer]` tuple type). Worth aligning for consistency.

LashaDevLasha: Done — both showSequenceContextMenu and showSelectedMonomersContextMenu are now captured in refs (each updated via useEffect) and called as ref.current(...) inside the handlers, so neither appears in the main effect's deps. The effect now only re-subscribes when editor, dispatch, or updatePasteAvailability changes. Also added explicit : void return types to all five right-click handlers for consistency.

dorogusha (approved):

packages/ketcher-macromolecules/src/Editor.tsx
```diff
@@ -214,15 +213,44 @@ function Editor({
...
+
+  useEffect(() => {
+    const {
+      ketcherId: initKetcherId,
+      theme: initTheme,
+      monomersLibraryUpdate: initMonomersLibraryUpdate,
+      monomersLibraryReplace: initMonomersLibraryReplace,
+      onInit: initOnInit,
+    } = editorInitParamsRef.current;
+
```
AlexeyGirin, line 245: Check if we can revert it (and supperess rule with comment)

AlexeyGirin (changes requested):

packages/ketcher-macromolecules/src/Editor.tsx
LashaDevLasha: Checked — we can’t drop the refs without either recreating the editor / rebinding listeners, or suppressing exhaustive-deps. Kept the refs so the rule stays clean; can switch to suppress if you prefer.

PR #11320, opened 2026-08-18
#11110 Refactor: eliminate setState in MeasureInput during render to sync ex…

### LINKED ISSUE #11110: Refactor: set-state-in-effect anti-pattern elimination from `packages\ketcher-react\src\script\ui\component\form\MeasureInput\measure-input.tsx` file
**Problem:** 
1. Duplicate source of truth
2. Extra render cycle:
- Render
- Effect runs
- setState runs
- Second render

**Why is this an issue?**

**How can it be fixed?**

```ts
function Component({data}) {

  useEffect(() => {
  }, [data]);
}

function Component() {

  useEffect(() => {
  }, []);
}

  useEffect(() => {
  }, [rawData]);
}

  useEffect(() => {
  }, [selectedId, items]);
}
```

**Examples of correct code:**
```ts
function Tooltip() {
  const ref = useRef(null);

  useLayoutEffect(() => {
    setTooltipHeight(height);
  }, []);
}

// ✅ Calculate during render
}
```

**Problem locations:**
packages\ketcher-react\src\script\ui\component\form\MeasureInput\measure-input.tsx:123

Nyildizzz:

## Problem

setState was called inside a useEffect to sync the prop into local state:

```
// ❌ Old code
useEffect(() => {
  setInternalValue(String(value)); // causes extra render cycle
}, [value]);
```

Every prop change triggered 2 renders:
1. Render with new prop
2. Effect runs → setState → render again

---
Fix

Replaced the useEffect with in-render state adjustment:

// ✅ New code
const [prevPropValue, setPrevPropValue] = useState(value);

```
if (prevPropValue !== value) {
  setPrevPropValue(value);
  setInternalValue(String(value)); // during render, not after
}
```

When React sees a setState call during render, it discards the current render output and immediately re-renders with the updated state — the intermediate state never reaches the DOM, so no flicker and no extra paint.

---
```
Why the second useEffect stays

useEffect(() => {
  if (internalValue !== String(value)) {
    onChange(parseFloat(internalValue));
  }
}, [internalValue]);
```

The Input component captures its onChange handler via constructor, so the closure never sees updated MeasureInput state. The effect is the only reliable way to propagate user-typed changes back to the parent — this one is legitimate.

## Check list
- [ ] unit-tests written
- [ ] e2e-tests written
- [ ] documentation updated

### DIFF
--- a/packages/ketcher-react/src/script/ui/component/form/MeasureInput/measure-input.tsx
+++ b/packages/ketcher-react/src/script/ui/component/form/MeasureInput/measure-input.tsx
@@ -105,28 +105,25 @@ const MeasureInput = ({
   ...rest
 }: MeasureInputProps) => {
   const [internalValue, setInternalValue] = useState(String(value));
+  const [prevPropValue, setPrevPropValue] = useState(value);
   const {
     anchorEl,
     handleOpen: handlePopoverOpen,
     handleClose: handlePopoverClose,
   } = usePopoverAnchor();
 
-  // NOTE: onChange handler in the Input comopnent (packages/ketcher-react/src/script/ui/component/form/Input/Input.tsx)
-  // is mapped to the internal function via constructor
-  // therefore the referencies to the MeasureInput's state are not updated
-  // so we need to sync the props and the internal value through useEffects and use callbacks with
-  // previous state to have the latest value
+  // Sync external value changes during render instead of in a useEffect to avoid
+  // the extra render cycle. React re-renders immediately when setState is called
+  // during render, skipping the intermediate paint.
+  if (prevPropValue !== value) {
+    setPrevPropValue(value);
+    setInternalValue(String(value));
+  }
 
-  useEffect(() => {
-    setInternalValue((prevValue) => {
-      if (prevValue !== String(value)) {
-        return String(value);
-      }
-
-      return prevValue;
-    });
-  }, [value]);
-
+  // NOTE: onChange handler in the Input component (packages/ketcher-react/src/script/ui/component/form/Input/Input.tsx)
+  // is mapped to the internal function via constructor, so its closure does not
+  // capture updated MeasureInput state. Propagate internal changes via effect
+  // with a functional-update callback to always read the latest value.
   useEffect(() => {
     if (internalValue !== String(value)) {
       const isNewInternalValueValid = !isNaN(parseFloat(internalValue));
--- a/packages/ketcher-react/src/script/ui/component/form/MeasureInput/measure-input.tsx
+++ b/packages/ketcher-react/src/script/ui/component/form/MeasureInput/measure-input.tsx
@@ -104,33 +104,27 @@ const MeasureInput = ({
   className,
   ...rest
 }: MeasureInputProps) => {
-  const [internalValue, setInternalValue] = useState(String(value));
-  const [prevPropValue, setPrevPropValue] = useState(value);
+  const stringifiedValue = String(value);
+  const [internalValue, setInternalValue] = useState(stringifiedValue);
+  const [prevPropValue, setPrevPropValue] = useState(stringifiedValue);
   const {
     anchorEl,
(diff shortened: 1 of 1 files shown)


packages/ketcher-react/src/script/ui/component/form/MeasureInput/measure-input.tsx
```diff
@@ -105,28 +105,25 @@ const MeasureInput = ({
...
-        return String(value);
-      }
-
-      return prevValue;
-    });
-  }, [value]);
+  // Sync external value changes during render instead of in a useEffect to avoid
+  // the extra render cycle. React re-renders immediately when setState is called
+  // during render, skipping the intermediate paint.
+  if (prevPropValue !== value) {
```
svvald: This may cause `Too many re-renders` error if both values are `NaN` as `Nan !== NaN` is `true`

An advice here - put `String(value)` on top of the component and use this stringified value where possible: an initial value for both `internalValue` and `prevPropValue` states, in comparison here and next effect on line 128 and as set value for the following setState calls

packages/ketcher-react/src/script/ui/component/form/MeasureInput/measure-input.tsx
```diff
@@ -105,28 +105,25 @@ const MeasureInput = ({
...
-      if (prevValue !== String(value)) {
-        return String(value);
-      }
-
-      return prevValue;
-    });
-  }, [value]);
+  // Sync external value changes during render instead of in a useEffect to avoid
+  // the extra render cycle. React re-renders immediately when setState is called
+  // during render, skipping the intermediate paint.
```
svvald: The code is self-explanatory, no need in obvious comments

packages/ketcher-react/src/script/ui/component/form/MeasureInput/measure-input.tsx
svvald, line 130: Having this effect is still an anti pattern as it violates the same approach but apparently it can not be resolved on this component level rather than on `Input` level

Not a blocker now but I'd ensure there is a follow-up ticket on that one as well

packages/ketcher-react/src/script/ui/component/form/MeasureInput/measure-input.tsx
```diff
@@ -105,28 +105,25 @@ const MeasureInput = ({
...
+  if (prevPropValue !== value) {
+    setPrevPropValue(value);
+    setInternalValue(String(value));
+  }
 
+  // NOTE: onChange handler in the Input component (packages/ketcher-react/src/script/ui/component/form/Input/Input.tsx)
+  // is mapped to the internal function via constructor, so its closure does not
+  // capture updated MeasureInput state. Propagate internal changes via effect
+  // with a functional-update callback to always read the latest value.
```
svvald: This comment was related to the removed effect and now is not relevant, I'd either remove it or use the different wording for why it's needed now, something like:
```
Input binds onChange once in its constructor, so handleChange is stuck with the first render's closure and cannot call the current onChange. This effect is re-created every render, so it always holds the latest onChange/value — hence the deliberate single-dep list.
```

packages/ketcher-react/src/script/ui/component/form/MeasureInput/measure-input.tsx
```diff
@@ -105,28 +105,25 @@ const MeasureInput = ({
...
   useEffect(() => {
     if (internalValue !== String(value)) {
       const isNewInternalValueValid = !isNaN(parseFloat(internalValue));
```
svvald: It was already here but `isNewInternalValueValid` flag is redundant as non-NaN check is performed on line 151 already

Could be kept for additional strictness though but removing it would make code slightly cleaner

svvald (changes requested):

svvald (approved):

PR #11322, opened 2026-08-18
#5474 - Ambiguous monomer are not scaled in expanded Select Connection Points dialog

### LINKED ISSUE #5474: Ambiguous monomer are not scaled in expanded Select Connection Points dialog
**Steps to Reproduce**
1. Toggle to Macro - Flex
2. Load using paste from clipboard way following HELM: `PEPTIDE1{(A:40,C:80)}|CHEM1{[SS3]}$$$$V2.0`
image
3. Start to connect them with bond using center to center way
(Select Connection Points dialog appears - ok)
4. Press expand window button
image

**Actual behavior**
Ambiguous peptide remain unscaled
image

**Expected behavior**
Ambiguous peptide scaled to window size
image

**Versions**
- Ketcher Version 2.25.0-rc.1 Build at 2024-09-03; 15:31:03
- Indigo Toolkit Version 1.24.0-rc.1.0-gb8f099851-wasm32-wasm-clang-19.0.0
- Chrome Version 128.0.6613.114 (Official Build) (64-bit)
- Win10

Reference issues: https://github.com/epam/ketcher/issues/4554

dorogusha:
…n Points dialog

The ambiguous monomer preview inside the "Select Attachment Points" dialog stayed pinned to its collapsed 150×150 size when the dialog was expanded, instead of scaling up to fill the larger window.

## Root cause

`MonomerMiniature` (used to render `AmbiguousMonomer` previews in this dialog) drew the monomer once via `AmbiguousMonomerRenderer.showExternal(...)` inside a `useLayoutEffect` whose dependency array only included `selectedAttachmentPoint` / `connectedAttachmentPoints` — not `expanded`. Toggling the dialog's expand button resized the surrounding container (CSS), but nothing ever re-triggered a redraw of the SVG content, so the rendered monomer kept its original small size.

The sibling (non-ambiguous) code path doesn't have this problem: `StyledStructRender` re-renders via its own `update` prop, which is already wired to `expanded`.

## Fix

- `useLayoutEffect` now also depends on `expanded` (and `monomer`), and clears the SVG before each redraw (the renderer only appends, never clears).
- The preview is always drawn into a fixed 150×150 reference box (unchanged geometry — body and R1/R2 attachment points keep their original relative positions).
- The finished SVG is then scaled up via CSS `transform: scale(...)` with `transform-origin: center`, where the scale factor is `min(containerWidth, containerHeight) / 150`, measured from the actual (already-correctly-resizing) container box.

An earlier version of this fix passed a `scale` directly into `AmbiguousMonomerRenderer`'s constructor. That distorted the layout: its root `<g>` transform is `translate(...) scale(...)`, which scales child coordinates around local `(0,0)` *before* translating — since R1/R2 sit much further from that origin than the body, they drifted disproportionately and ended up rendered underneath/detached from the body at any scale ≠ 1. Scaling the whole finished picture via CSS around its own center avoids this entirely.

## Check list
- [ ] unit-tests written
- [ ] e2e-tests written
- [ ] documentation updated

Fixes #5474


AlexeyGirin:

svvald (approved):

PR #11325, opened 2026-08-18
#11106 - Refactor: set-state-in-effect anti-pattern elimination from packages\ketcher-react\src\script\ui\views\components\MonomerCreationWizard\MonomerCreationWizard.tsx file

### LINKED ISSUE #11106: Refactor: set-state-in-effect anti-pattern elimination from `packages\ketcher-react\src\script\ui\views\components\MonomerCreationWizard\MonomerCreationWizard.tsx` file

yaroslav-b-io:
Two effects computed a value from state and called setState inside the effect — the anti-pattern that risks extra renders/loops.   
Fix: moved the guard + computation into a useMemo (pure, no side effects), leaving only a minimal useEffect to sync the           already-computed value to the imperative editor call and setState. Same behavior, no more state-as-trigger-for-effect pattern, no lint warnings.  

Issue - [#11317 ](https://github.com/epam/ketcher/issues/11106)

## Check list
- [ ] unit-tests written
- [ ] e2e-tests written
- [ ] documentation updated
- [X] PR name follows the pattern `#1234 – issue name`
- [X] branch name doesn't contain '#'
- [X] PR is linked with the issue


packages/ketcher-react/src/script/ui/views/components/MonomerCreationWizard/MonomerCreationWizard.tsx
```diff
@@ -1126,22 +1126,28 @@ const MonomerCreationWizardInternal = ({
...
+    hasActiveRnaPresetAtomValidationErrors,
     rnaPresetComponentStructures,
   ]);
 
+  useEffect(() => {
+    if (!rnaPresetProblematicAtomIds) {
+      return;
+    }
+
+    editor.setProblematicAtoms(rnaPresetProblematicAtomIds);
```
la-nuta: Master re-checked `editor?.render?.monomerCreationState` at commit time; now it's only checked during render. If the wizard closes between render and commit this hits `KetcherLogger.error('Can not set problematic atoms...')` instead of no-op'ing. Minor, but a real guard weakening.

packages/ketcher-react/src/script/ui/views/components/MonomerCreationWizard/MonomerCreationWizard.tsx
```diff
@@ -1190,20 +1196,24 @@ const MonomerCreationWizardInternal = ({
...
+    return inferPhosphatePosition(
       sugarAttachmentPoints,
       phosphateAttachmentPoints,
     );
-
-    handlePhosphatePositionChange(autoPhosphatePosition);
   }, [
     isRnaPresetType,
-    monomerCreationState?.assignedAttachmentPoints,
+    monomerCreationState,
```
la-nuta: Not equivalent. `assignedAttachmentPoints` is a Map mutated **in place** by the editor operations, so its identity never changes — the old dep was effectively inert. `monomerCreationState` is replaced by spread on `setConnectionAttachmentPoints`, `setVisibleAssignedAttachmentPoints`, `setProblematicAtoms` and friends, i.e. on every RNA-preset tab switch.

Failure: user manually picks phosphate position `5` where inference says `3`. Before, switching to the Sugar tab didn't re-run inference, so `5` stuck. Now the tab switch replaces `monomerCreationState`, inference recomputes `3`, the value changed → the effect fires → the manual pick is silently overwritten and the wrong position reaches `finishNewMonomersCreation`.

If you need to react to AP assignment, derive a stable key from the Map contents instead.

packages/ketcher-react/src/script/ui/views/components/MonomerCreationWizard/MonomerCreationWizard.tsx
```diff
@@ -1190,20 +1196,24 @@ const MonomerCreationWizardInternal = ({
...
     rnaPresetWizardState.phosphate.structure,
     rnaPresetWizardState.sugar.structure,
-    handlePhosphatePositionChange,
   ]);
 
+  useEffect(() => {
+    if (autoPhosphatePosition) {
```
la-nuta: `inferPhosphatePosition` always returns `'3' | '5'`, never `undefined` — so this guard only catches the memo's early return, and more importantly the value is a **primitive**. Memoizing it means the effect now only fires when the inferred value *changes*; master re-ran it on every sugar/phosphate structure change.

That matters because `handlePhosphatePositionChange` also clears the `phosphatePosition` error and two notifications. Failure: Save fails with the sugar-phosphate connection error, user fixes the structure, inference still says `3` → effect never fires → the stale error banner stays up until they manually toggle the position.

Note this pushes the *opposite* way from the dep widening above, so the firing moments differ from master in both directions.

AlexeyGirin:

svvald (approved):

PR #11326, opened 2026-08-18
#5345 - Replacement of any monomer (side connected to another one) on monomer of different type works wrong - system keeps side chain bond (it shouldn't)

### LINKED ISSUE #5345: Replacement of any monomer (side connected to another one) on monomer of different type works wrong - system keeps side chain bond (it shouldn't)
**Steps to Reproduce**
1. Go to Macromolecules mode - Flex mode
2. Load from HELM: `RNA1{[AmMC6T].[AmMC6T].[AmMC6T]}|RNA2{[AmMC6T].[AmMC6T].[AmMC6T]}$RNA2,RNA1,2:R3-2:R3|RNA2,RNA1,3:R3-3:R3|RNA2,RNA1,1:R3-1:R3$$$V2.0`
Image

3. Go to Sequence mode and select left to symbols in view mode 
Image

4. Go to Library - Petide tab and click on D-gGlu petide (E section)
5. Switch to Flex mode

**Actual behavior**
System connects D-gGlu to unsplit nucleotides
Image

**Expected behavior**
Confirmation window should appear
No connection should be kept if user accept
image

As per [requirement](https://github.com/epam/ketcher/issues/4878):

> .....4.3. If any of the selected monomers have side connections, and the monomer chosen from the library is not of the same type (types: peptides, chems, sugars, bases, phosphates, presets, nucleotides) as at least one selected monomer with a side connection
> or
> If the monomer chosen form the library is of the same type, but doesn't have all needed attachment points to establish all side connections (see 4.5. and 4.6.),
> .....4.5. If a peptide, chem, sugar, base, phosphate or a nucleotide is being replaced by a monomer of the same type, a side connection at the attachment point Rn (n>2) is preserved is the monomer chosen from the library also has Rn.
> (example: user selected, among others, a Cysteine (peptide) with an R3 side connection to Rm of some other monomer-> user chooses dGlutamine (peptide) from the library -> because dGlutamine has R3, Cysteine is replaced with dGlutamine and a side chain connection is established between R3 of dGlutamine and Rm of that other monomer)
> .....4.7. If both 4.2. and 4.3. are true, the first warning message should be the one from 4.2.

**Versions**
- Ketcher Version 2.24.0-rc.2 Build at 2024-08-14; 19:06:32
- Indigo Toolkit Version 1.23.0-rc.2.0-g6a2a7526a-x86_64-linux-gnu-11.2.1
- Chrome Version 127.0.6533.100 (Official Build) (64-bit)
- Win10

Related issue - https://github.com/epam/ketcher/issues/4878

andranikkk:
…omer-of-different-type-works-wrong---system-keeps-side-chain-bond-it-shouldnt

## Check list
- [ ] unit-tests written
- [ ] e2e-tests written
- [ ] documentation updated

### DIFF
--- a/packages/ketcher-core/src/application/editor/modes/SequenceMode.ts
+++ b/packages/ketcher-core/src/application/editor/modes/SequenceMode.ts
@@ -70,6 +70,7 @@ import { getNodeFromTwoStrandedNode } from 'domain/helpers/chains';
 import type { CoreEditor } from 'application/editor/Editor';
 import { MACROMOLECULES_BOND_TYPES } from 'application/editor/tools/types';
 import { KetMonomerClass } from 'application/formatters';
+import { monomerEntityFactory } from 'domain/helpers/monomerEntityFactory';
 import { registerMode } from './modesRegistry';
 
 const naturalAnalogues = uniq([
@@ -1921,6 +1922,35 @@ export class SequenceMode extends BaseMode {
     );
   }
 
+  private getMonomerClassForTypeComparison(
+    monomer: BaseMonomer | MonomerItemType,
+  ): KetMonomerClass {
+    if (!(monomer instanceof BaseMonomer)) {
+      const [, ketMonomerClass] = monomerEntityFactory(monomer);
+
+      return ketMonomerClass;
+    }
+
+    if (monomer instanceof AmbiguousMonomer) {
+      return monomer.monomerClass;
+    }
+
+    return this.getMonomerClassForTypeComparison(monomer.monomerItem);
+  }
+
+  // Side chain connections (Rn, n>2) are only preserved when the monomer
+  // being inserted is of the same type (peptide/chem/sugar/base/phosphate/
+  // nucleotide) as the monomer it replaces, per requirement 4.5.
+  private isSameMonomerTypeForSideChainConnection(
+    existingMonomer: BaseMonomer,
+    newMonomer: BaseMonomer | MonomerItemType,
+  ): boolean {
+    return (
+      this.getMonomerClassForTypeComparison(existingMonomer) ===
+      this.getMonomerClassForTypeComparison(newMonomer)
+    );
+  }
+
   isPasteAvailable(drawingEntitiesManager: DrawingEntitiesManager) {
     if (!this.isEditMode) {
       return true;
@@ -2111,11 +2141,16 @@ export class SequenceMode extends BaseMode {
 
     sideChainConnections?.forEach((sideConnectionData) => {
       const {
+        sourceMonomer,
         firstMonomerAttachmentPointName,
         secondMonomer,
         secondMonomerAttachmentPointName,
       } = sideConnectionData;
       if (
+        !this.isSameMonomerTypeForSideChainConnection(
+          sourceMonomer,
+          newMonomer,
+        ) ||
         !this.isConnectionPossible(
           newMonomer,
           firstMonomerAttachmentPointName,
@@ -2282,6 +2317,16 @@ export class SequenceMode extends BaseMode {
             : !bond.isBackBoneChainConnection))
       ) {
         return true;
+      }
+
+      if (
+        sideChainConnections &&
+        !this.isSameMonomerTypeForSideChainConnection(
+          selectedNode.monomer,
+          monomerItem,
+        )
+      ) {
+        return false;
       }
 
       return newMonomerAttachmentPoints.attachmentPointsList.includes(key);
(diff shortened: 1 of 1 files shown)


packages/ketcher-core/src/application/editor/modes/SequenceMode.ts
```diff
@@ -2284,6 +2319,16 @@ export class SequenceMode extends BaseMode {
         return true;
       }
 
+      if (
+        sideChainConnections &&
+        !this.isSameMonomerTypeForSideChainConnection(
+          selectedNode.monomer,
```
svvald: For Nucleotide it will return either Sugar or Phosphate and for Nucleoside it will return just Sugar yet there is also an RNA Base that could also have side chain connections and it seems to be missing here

packages/ketcher-core/src/application/editor/modes/SequenceMode.ts
```diff
@@ -2284,6 +2319,16 @@ export class SequenceMode extends BaseMode {
         return true;
       }
 
+      if (
```
svvald: This is now called for each iteration of `oldMonomerBonds.every` even though it doesn't depend on those entries, can we lift this up?

packages/ketcher-core/src/application/editor/modes/SequenceMode.ts
```diff
@@ -1921,6 +1922,35 @@ export class SequenceMode extends BaseMode {
...
+      const [, ketMonomerClass] = monomerEntityFactory(monomer);
+
+      return ketMonomerClass;
+    }
+
+    if (monomer instanceof AmbiguousMonomer) {
+      return monomer.monomerClass;
+    }
+
+    return this.getMonomerClassForTypeComparison(monomer.monomerItem);
```
la-nuta: Non-blocking: `monomerEntityFactory` already handles ambiguous items (it returns `AmbiguousMonomer.getMonomerClass(monomer.monomers)`, the same value stored in `AmbiguousMonomer.monomerClass`), so the helper collapses to one call. I ran `tsc --noEmit` and eslint on this variant — clean.

```suggestion
    const [, ketMonomerClass] = monomerEntityFactory(
      monomer instanceof BaseMonomer ? monomer.monomerItem : monomer,
    );

    return ketMonomerClass;
```

la-nuta: Non-blocking: the recursion can go — the `AmbiguousMonomer` branch has to stay (its `monomerItem` has no `monomers`, so passing it to the factory throws), but the other two cases collapse into one call. Verified on `PEPTIDE1{(C,K).A.A}|PEPTIDE2{C.A.A}$PEPTIDE1,PEPTIDE2,1:R3-1:R3$$$V2.0` → replace `(C,K)` with D-gGlu: no dialog, R3 bridge kept, same as your version.

```suggestion
    if (monomer instanceof AmbiguousMonomer) {
      return monomer.monomerClass;
    }

    const [, ketMonomerClass] = monomerEntityFactory(
      monomer instanceof BaseMonomer ? monomer.monomerItem : monomer,
    );

    return ketMonomerClass;
```

la-nuta (commented):
Verified locally on the issue's exact steps (HELM from #5345 → Sequence mode → select two symbols → Peptides → D-gGlu), PR applied onto current master:

- master: no dialog, both symbols replaced, all 3 R3–R3 side bonds kept (two now to D-gGlu) — the reported bug
- PR: "Side chain connections will be deleted during replacement" dialog appears; after Yes, the two bonds to D-gGlu are gone and the untouched AmMC6T–AmMC6T R3 bond remains

Same-type case (req. 4.5) unchanged: `PEPTIDE1{C.A.C}|PEPTIDE2{C.A.C}$…1:R3-1:R3…`, replacing a Cys with D-gGlu — no dialog, R3 bridge preserved, identical to master. Preset path already gates by class via `getPresetMonomerForPreservedSideChainConnection`. `tsc --noEmit` on ketcher-core: clean.

One ask: the branch is ~165 commits behind master and the last CI run has `part (9, 10)` cancelled / `merge_reports` failed — please pick up master (merge or rebase, whichever you prefer) so we get a full green run.

AlexeyGirin:
@copilot please make fix based on provided review comments

la-nuta (commented):
Re-verified on the new head `b06a5e49`: after merging onto master the file is byte-identical to the variant I tested (eslint, `tsc --noEmit`, and the issue / same-type / ambiguous-monomer scenarios at runtime), so the simplification is fine.

Issue #9693, opened 2026-04-07
RNA Builder: Default preset name changes unexpectedly after switching phosphate picker from 3' to 5'


MarinaKisacik:
Found while testing #9120 
**Steps to Reproduce:**

1. Switch to Macro mode
2. Expand the RNA builder in the RNA tab of the Monomer library
3. Select a sugar, a base, and a phosphate for the preset
4. Note the default preset name that appears in the name input field
5. Change the phosphate position from 3' to 5'
6. Observe the preset name field

**Actual Results**
The default preset name is changed

<img width="300" alt="Image" src="https://github.com/user-attachments/assets/d8ba2c18-b63e-4849-90aa-4d23a1181b8e" />

<img width="300" alt="Image" src="https://github.com/user-attachments/assets/77a8955e-327d-4259-a2c8-55cf2a1aa715" />

**Expected Result:**
The default preset name should remain unchanged, regardless of any changes to the phosphate position picker

<img width="600" alt="Image" src="https://github.com/user-attachments/assets/384b549d-bce9-4ec1-9f03-ff992ef082b8" />

**Found in Environment:**
- Ketcher Version 3.15.0-rc.1 Build at 2026-04-02; 15:41:30
- Indigo Version 1.43.0-rc.1.0-gd74653443-wasm32-wasm-clang-19.0.0
- Chrome Version 146.0.7680.178 (Official Build) (64-bit)
- Win11

PR #11327, opened 2026-08-19
#11287 - Refactor: react-hooks/exhaustive-deps rule violation clean up from packages\ketcher-macromolecules\src\components\monomerLibrary\RnaBuilder\RnaEditor\RnaEditorExpanded\RnaEditorExpanded.tsx file

### LINKED ISSUE #11287: Refactor: `react-hooks/exhaustive-deps` rule violation clean up from `packages\ketcher-macromolecules\src\components\monomerLibrary\RnaBuilder\RnaEditor\RnaEditorExpanded\RnaEditorExpanded.tsx` file

Example:
```ts

useEffect(() => {
  loadUser(userId);
}, []); // userId is missing
```

**Why is this a problem?**

This can cause:
- stale state or props;
- effects using old data;

**How is it usually fixed?**
```ts
useEffect(() => {
  loadUser(userId);
}, [userId]);
```

For a memoized callback:
```ts
}, [documentId, content]);
```

For a memoized value:
```ts
  [items, category],
);
```

**Problem locations:**
packages\ketcher-macromolecules\src\components\monomerLibrary\RnaBuilder\RnaEditor\RnaEditorExpanded\RnaEditorExpanded.tsx

LashaDevLasha:
Refactor: resolve react-hooks/exhaustive-deps violations in RnaEditorExpanded

Exhaustive-deps fixes

- Wrap onCancel and onUpdateSequence in useCallback with correct deps so they are stable references
- Add onCancel, onUpdateSequence, isSequenceEditInRNABuilderMode to the keyDown effect deps (effect already has cleanup via .remove())
- Add onCancel to the cancelSequenceEditInRNABuilderMode effect deps (added by #10409, same stale-closure shape as keyDown)
- Replace updatePresetMonomerGroup function with an inline setNewPreset functional updater; use sequenceSelectionRef to read the latest selection inside the monomer-group effect without triggering a dispatch → sequenceSelection → re-run loop
- Keep resolvePhosphatePosition as a plain function (updated each render) and track it via resolvePhosphatePositionRef; use activeMonomerGroupRef and isEditModeRef for the guard check — this prevents the monomer-group effect from firing on slot clicks (activeMonomerGroup changing) or layout-mode switches (resolvePhosphatePosition changing via isSequenceMode), which would drop editedName and regenerate the preset name incorrectly
- Monomer-group effect deps narrowed to [isSequenceEditInRNABuilderMode, activePresetMonomerGroup, dispatch]

With all effects clean, the two file-wide ESLint disable headers (exhaustive-deps and no-event-handler/no-chain-state-updates) are removed.

Bug fixes included

- keyDown effect: the previous dep array [editor, sequenceSelection, isSequenceEditInRNABuilderMode] left handleKeyDown with a stale onCancel closure after a preset switch — pressing Escape restored the previous preset instead of the current one. Fixed by listing onCancel as a dep.
- cancelSequenceEditInRNABuilderMode effect: same stale-closure shape introduced by #10409; onCancel was missing from deps. Fixed in the same pass.

## Check list
- [ ] unit-tests written
- [ ] e2e-tests written
- [ ] documentation updated

### DIFF
--- a/packages/ketcher-macromolecules/src/components/monomerLibrary/RnaBuilder/RnaEditor/RnaEditorExpanded/RnaEditorExpanded.tsx
+++ b/packages/ketcher-macromolecules/src/components/monomerLibrary/RnaBuilder/RnaEditor/RnaEditorExpanded/RnaEditorExpanded.tsx
@@ -69,7 +69,14 @@ import {
   selectEditor,
   selectIsSequenceEditInRNABuilderMode,
 } from 'state/common';
-import { ChangeEvent, KeyboardEvent, useEffect, useState } from 'react';
+import {
+  ChangeEvent,
+  KeyboardEvent,
+  useCallback,
+  useEffect,
+  useRef,
+  useState,
+} from 'react';
 import {
   generateSequenceSelectionGroupNames,
   generateSequenceSelectionName,
@@ -126,6 +133,9 @@ export const RnaEditorExpanded = ({
     selectActivePresetMonomerGroup,
   );
   const [newPreset, setNewPreset] = useState(activePreset);
+  const newPresetRef = useRef(newPreset);
+  newPresetRef.current = newPreset;
+
   const [selectedPhosphatePosition, setSelectedPhosphatePosition] = useState<
     RnaPhosphatePosition | undefined
   >(
@@ -134,54 +144,57 @@ export const RnaEditorExpanded = ({
       : undefined,
   );
 
-  const resolvePhosphatePosition = (
-    preset: typeof newPreset,
-  ): RnaPhosphatePosition | undefined => {
-    if (!preset?.phosphate) {
-      return undefined;
-    }
-
-    const {
-      is3PrimeAvailable: isRightPositionAvailable,
-      is5PrimeAvailable: isLeftPositionAvailable,
-    } = getPhosphatePositionAvailability(preset);
-
-    if (selectedPhosphatePosition === 'left' && isLeftPositionAvailable) {
-      return 'left';
-    }
-
-    if (selectedPhosphatePosition === 'right' && isRightPositionAvailable) {
-      return 'right';
-    }
-
-    if (preset.connections?.length) {
-      const presetPhosphatePosition = getRnaPresetPhosphatePosition(preset);
-
-      if (
-        (presetPhosphatePosition === 'left' && isLeftPositionAvailable) ||
-        (presetPhosphatePosition === 'right' && isRightPositionAvailable)
-      ) {
-        return presetPhosphatePosition;
+  const resolvePhosphatePosition = useCallback(
+    (preset: typeof newPreset): RnaPhosphatePosition | undefined => {
+      if (!preset?.phosphate) {
+        return undefined;
       }
-    }
 
-    if (isSequenceMode && isRightPositionAvailable) {
-      return 'right';
-    }
+      const {
+        is3PrimeAvailable: isRightPositionAvailable,
+        is5PrimeAvailable: isLeftPositionAvailable,
+      } = getPhosphatePositionAvailability(preset);
 
-    if (isLeftPositionAvailable && !isRightPositionAvailable) {
-      return 'left';
-    }
+      if (selectedPhosphatePosition === 'left' && isLeftPositionAvailable) {
+        return 'left';
+      }
 
-    if (isRightPositionAvailable && !isLeftPositionAvailable) {
-      return 'right';
-    }
+      if (selectedPhosphatePosition === 'right' && isRightPositionAvailable) {
+        return 'right';
+      }
 
-    return undefined;
-  };
+      if (preset.connections?.length) {
+        const presetPhosphatePosition = getRnaPresetPhosphatePosition(preset);
+
+        if (
+          (presetPhosphatePosition === 'left' && isLeftPositionAvailable) ||
+          (presetPhosphatePosition === 'right' && isRightPositionAvailable)
+        ) {
+          return presetPhosphatePosition;
+        }
+      }
+
+      if (isSequenceMode && isRightPositionAvailable) {
+        return 'right';
+      }
+
+      if (isLeftPositionAvailable && !isRightPositionAvailable) {
+        return 'left';
+      }
+
+      if (isRightPositionAvailable && !isLeftPositionAvailable) {
+        return 'right';
+      }
+
+      return undefined;
+    },
+    [selectedPhosphatePosition, isSequenceMode],
+  );
 
   // For sequence edit in RNA Builder mode
   const sequenceSelection = useAppSelector(selectSequenceSelection);
+  const sequenceSelectionRef = useRef(sequenceSelection);
+  sequenceSelectionRef.current = sequenceSelection;
   const sequenceSelectionName = useAppSelector(selectSequenceSelectionName);
   const isSequenceEditInRNABuilderMode = useAppSelector(
     selectIsSequenceEditInRNABuilderMode,
@@ -205,20 +218,6 @@ export const RnaEditorExpanded = ({
   const phosphatePositionDisabledTooltip = {
     left: 'Sugar must have R1, and phosphate must have R2.',
     right: 'Sugar must have R2, and phosphate must have R1.',
-  };
-
-  const updatePresetMonomerGroup = () => {
-    if (activePresetMonomerGroup) {
-      const groupName =
-        monomerGroupToPresetGroup[activePresetMonomerGroup.groupName];
-      const currentPreset = {
-        ...newPreset,
-        [groupName]: activePresetMonomerGroup.groupItem,
-      };
-      setNewPreset(currentPreset);
-      return currentPreset;
-    }
-    return newPreset;
   };
 
   useEffect(() => {
@@ -252,35 +251,51 @@ export const RnaEditorExpanded = ({
           monomerGroupToPresetGroup[activePresetMonomerGroup.groupName];
         const field = `${monomerType}Label`;
 
-        const updatedSequenceSelection = sequenceSelection.map((node) => {
-          // Do not set 'phosphateLabel' for Nucleoside if it is connected and selected with Phosphate
-          // Do not set 'sugarLabel', 'baseLabel' for Phosphate
-          if (
-            (node.isNucleosideConnectedAndSelectedWithPhosphate &&
-              field === 'phosphateLabel') ||
-            (node.type === Entities.Phosphate &&
-              (field === 'sugarLabel' || field === 'baseLabel'))
-          ) {
-            return node;
-          }
+        // sequenceSelectionRef.current avoids adding sequenceSelection to deps,
+        // which would cause a dispatch→sequenceSelection-change→re-run loop.
+        const updatedSequenceSelection = sequenceSelectionRef.current.map(
+          (node) => {
+            // Do not set 'phosphateLabel' for Nucleoside if it is connected and selected with Phosphate
+            // Do not set 'sugarLabel', 'baseLabel' for Phosphate
+            if (
+              (node.isNucleosideConnectedAndSelectedWithPhosphate &&
+                field === 'phosphateLabel') ||
+              (node.type === Entities.Phosphate &&
+                (field === 'sugarLabel' || field === 'baseLabel'))
+            ) {
+              return node;
+            }
 
-          return {
-            ...node,
-            [field]: activePresetMonomerGroup.groupItem.label,
-            rnaBaseMonomerItem:
-              activePresetMonomerGroup.groupName === 'Bases'
-                ? activePresetMonomerGroup.groupItem
-                : node.rnaBaseMonomerItem,
-          };
-        });
+            return {
+              ...node,
+              [field]: activePresetMonomerGroup.groupItem.label,
+              rnaBaseMonomerItem:
+                activePresetMonomerGroup.groupName === 'Bases'
+                  ? activePresetMonomerGroup.groupItem
+                  : node.rnaBaseMonomerItem,
+            };
+          },
+        );
 
         setIsSequenceSelectionUpdated(true);
         dispatch(setSequenceSelection(updatedSequenceSelection));
       } else {
-        const currentPreset = updatePresetMonomerGroup();
+        // Inlined updatePresetMonomerGroup — newPresetRef.current avoids adding
+        // newPreset to deps, which would cause a setNewPreset→re-run loop.
+        let currentPreset = newPresetRef.current;
+        if (activePresetMonomerGroup) {
+          const groupName =
+            monomerGroupToPresetGroup[activePresetMonomerGroup.groupName];
+          currentPreset = {
+            ...newPresetRef.current,
+            [groupName]: activePresetMonomerGroup.groupItem,
+          };
+          setNewPreset(currentPreset);
+        }
+
         const resolvedPhosphatePosition =
           resolvePhosphatePosition(currentPreset);
-        let presetFullName = newPreset?.name;
+        let presetFullName = newPresetRef.current?.name;
 
         if (!currentPreset.editedName) {
           presetFullName = selectPresetFullName({
@@ -296,9 +311,12 @@ export const RnaEditorExpanded = ({
       }
     }
   }, [
-    activePresetMonomerGroup?.groupItem,
+    activeMonomerGroup,
+    isEditMode,
     isSequenceEditInRNABuilderMode,
-    selectedPhosphatePosition,
+    activePresetMonomerGroup,
+    dispatch,
+    resolvePhosphatePosition,
   ]);
 
   const scrollToActiveItemInLibrary = (selectedGroup, selectedMonomer) => {
@@ -516,14 +534,14 @@ export const RnaEditorExpanded = ({
     );
   };
 
-  const onUpdateSequence = () => {
+  const onUpdateSequence = useCallback(() => {
     if (getCountOfNucleoelements(sequenceSelection) > 1) {
       dispatch(openModal('updateSequenceInRNABuilder'));
     } else {
       editor?.events.modifySequenceInRnaBuilder.dispatch(sequenceSelection);
       resetRnaBuilderAfterSequenceUpdate(dispatch, editor);
     }
-  };
+  }, [sequenceSelection, dispatch, editor]);
 
   const onSave = () => {
     const presetName = newPreset?.name;
@@ -570,7 +588,7 @@ export const RnaEditorExpanded = ({
     resetRnaBuilder(dispatch);
   };
 
-  const onCancel = () => {
+  const onCancel = useCallback(() => {
     if (isSequenceEditInRNABuilderMode) {
       resetRnaBuilderAfterSequenceUpdate(dispatch, editor);
     } else if (isActivePresetEmpty && presets.length > 0) {
@@ -585,7 +603,14 @@ export const RnaEditorExpanded = ({
       );
       resetRnaBuilder(dispatch);
     }
-  };
+  }, [
+    isSequenceEditInRNABuilderMode,
+    dispatch,
+    editor,
+    isActivePresetEmpty,
+    presets,
+    activePreset,
+  ]);
 
   const turnOnEditMode = () => {
     dispatch(setIsEditMode(true));
@@ -630,7 +655,7 @@ export const RnaEditorExpanded = ({
     return () => {
       editor?.events.keyDown.remove(handleKeyDown);
     };
-  }, [editor, sequenceSelection]);
+  }, [editor, onCancel, onUpdateSequence, isSequenceEditInRNABuilderMode]);
 
   let mainButton: JSX.Element;
   const isSaveButtonDisabled =
--- a/packages/ketcher-macromolecules/src/components/monomerLibrary/RnaBuilder/RnaEditor/RnaEditorExpanded/RnaEditorExpanded.tsx
+++ b/packages/ketcher-macromolecules/src/components/monomerLibrary/RnaBuilder/RnaEditor/RnaEditorExpanded/RnaEditorExpanded.tsx
@@ -1,5 +1,3 @@
-/* eslint-disable react-hooks/exhaustive-deps */
-/* eslint-disable react-you-might-not-need-an-effect/no-event-handler */
 /* eslint-disable react-hooks/set-state-in-effect */
 /****************************************************************************
  * Copyright 2021 EPAM Systems
@@ -147,52 +145,71 @@ export const RnaEditorExpanded = ({
       : undefined,
   );
 
-  const resolvePhosphatePosition = useCallback(
-    (preset: typeof newPreset): RnaPhosphatePosition | undefined => {
-      if (!preset?.phosphate) {
-        return undefined;
-      }
-
-      const {
-        is3PrimeAvailable: isRightPositionAvailable,
-        is5PrimeAvailable: isLeftPositionAvailable,
-      } = getPhosphatePositionAvailability(preset);
-
-      if (selectedPhosphatePosition === 'left' && isLeftPositionAvailable) {
-        return 'left';
-      }
-
-      if (selectedPhosphatePosition === 'right' && isRightPositionAvailable) {
-        return 'right';
-      }
-
-      if (preset.connections?.length) {
-        const presetPhosphatePosition = getRnaPresetPhosphatePosition(preset);
-
-        if (
-          (presetPhosphatePosition === 'left' && isLeftPositionAvailable) ||
-          (presetPhosphatePosition === 'right' && isRightPositionAvailable)
-        ) {
-          return presetPhosphatePosition;
-        }
-      }
-
-      if (isSequenceMode && isRightPositionAvailable) {
-        return 'right';
-      }
-
-      if (isLeftPositionAvailable && !isRightPositionAvailable) {
-        return 'left';
-      }
-
-      if (isRightPositionAvailable && !isLeftPositionAvailable) {
-        return 'right';
-      }
-
+  // Plain function — safe to call at render time and always reads the latest
+  // selectedPhosphatePosition / isSequenceMode from closure.
+  const resolvePhosphatePosition = (
+    preset: typeof newPreset,
+  ): RnaPhosphatePosition | undefined => {
+    if (!preset?.phosphate) {
       return undefined;
-    },
-    [selectedPhosphatePosition, isSequenceMode],
-  );
+    }
+
+    const {
+      is3PrimeAvailable: isRightPositionAvailable,
+      is5PrimeAvailable: isLeftPositionAvailable,
+    } = getPhosphatePositionAvailability(preset);
+
+    if (selectedPhosphatePosition === 'left' && isLeftPositionAvailable) {
+      return 'left';
+    }
+
+    if (selectedPhosphatePosition === 'right' && isRightPositionAvailable) {
+      return 'right';
+    }
+
+    if (preset.connections?.length) {
+      const presetPhosphatePosition = getRnaPresetPhosphatePosition(preset);
+
+      if (
+        (presetPhosphatePosition === 'left' && isLeftPositionAvailable) ||
+        (presetPhosphatePosition === 'right' && isRightPositionAvailable)
+      ) {
+        return presetPhosphatePosition;
+      }
+    }
+
+    if (isSequenceMode && isRightPositionAvailable) {
+      return 'right';
+    }
+
+    if (isLeftPositionAvailable && !isRightPositionAvailable) {
+      return 'left';
+    }
+
+    if (isRightPositionAvailable && !isLeftPositionAvailable) {
+      return 'right';
+    }
+
+    return undefined;
+  };
+
+  // Kept in sync every render so the monomer-group effect always calls the
+  // latest version without listing resolvePhosphatePosition as a dep (which
+  // would re-trigger on every selectedPhosphatePosition / isSequenceMode change
+  // and regenerate the preset name).
+  const resolvePhosphatePositionRef = useRef(resolvePhosphatePosition);
+  // eslint-disable-next-line react-hooks/refs
+  resolvePhosphatePositionRef.current = resolvePhosphatePosition;
+
+  // Guard refs — read inside the monomer-group effect without being listed as
+  // deps, preventing re-runs on slot clicks (activeMonomerGroup) or edit-mode
+  // toggle (isEditMode), which would drop editedName and regenerate the name.
+  const activeMonomerGroupRef = useRef(activeMonomerGroup);
+  // eslint-disable-next-line react-hooks/refs
+  activeMonomerGroupRef.current = activeMonomerGroup;
+  const isEditModeRef = useRef(isEditMode);
+  // eslint-disable-next-line react-hooks/refs
+  isEditModeRef.current = isEditMode;
 
   // For sequence edit in RNA Builder mode
   const sequenceSelection = useAppSelector(selectSequenceSelection);
@@ -248,7 +265,10 @@ export const RnaEditorExpanded = ({
   }, [dispatch, sequenceSelection]);
 
   useEffect(() => {
-    if (activeMonomerGroup !== RnaBuilderPresetsItem.Presets && isEditMode) {
+    if (
+      activeMonomerGroupRef.current !== RnaBuilderPresetsItem.Presets &&
+      isEditModeRef.current
+    ) {
       if (isSequenceEditInRNABuilderMode && activePresetMonomerGroup) {
         const monomerType =
           monomerGroupToPresetGroup[activePresetMonomerGroup.groupName];
@@ -292,7 +312,7 @@ export const RnaEditorExpanded = ({
               }
             : currentPreset;
           const resolvedPhosphatePosition =
-            resolvePhosphatePosition(updatedPreset);
+            resolvePhosphatePositionRef.current(updatedPreset);
           const presetFullName = updatedPreset.editedName
             ? updatedPreset.name
             : selectPresetFullName({
@@ -315,12 +335,16 @@ export const RnaEditorExpanded = ({
       }
     }
   }, [
-    activeMonomerGroup,
-    isEditMode,
+    // activeMonomerGroup and isEditMode are intentionally omitted — they are
+    // guard conditions read via refs. Including them would fire this effect on
+    // every slot click (activeMonomerGroup) or entering edit mode (isEditMode),
+    // which regenerates the preset name and drops editedName.
+    // resolvePhosphatePosition is intentionally omitted — it is stable (empty
+    // deps) and reads selectedPhosphatePosition/isSequenceMode via refs, so
+    // layout-mode switches no longer trigger this effect.
     isSequenceEditInRNABuilderMode,
     activePresetMonomerGroup,
     dispatch,
-    resolvePhosphatePosition,
   ]);
 
   const scrollToActiveItemInLibrary = (selectedGroup, selectedMonomer) => {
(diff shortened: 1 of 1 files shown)


packages/ketcher-macromolecules/src/components/monomerLibrary/RnaBuilder/RnaEditor/RnaEditorExpanded/RnaEditorExpanded.tsx
```diff
@@ -126,6 +133,9 @@ export const RnaEditorExpanded = ({
     selectActivePresetMonomerGroup,
   );
   const [newPreset, setNewPreset] = useState(activePreset);
+  const newPresetRef = useRef(newPreset);
+  newPresetRef.current = newPreset;
```
la-nuta: This ref exists only to keep `newPreset` out of the deps, and it's one of the two lines that trips `react-hooks/refs`. `setNewPreset((prev) => ...)` removes the dependency for real and satisfies both rules — and collapses the old double `setNewPreset` into one, which is a behavior improvement worth flagging in the description.

packages/ketcher-macromolecules/src/components/monomerLibrary/RnaBuilder/RnaEditor/RnaEditorExpanded/RnaEditorExpanded.tsx
```diff
@@ -134,54 +144,57 @@ export const RnaEditorExpanded = ({
...
-  };
+      return undefined;
+    },
+    [selectedPhosphatePosition, isSequenceMode],
+  );
 
   // For sequence edit in RNA Builder mode
   const sequenceSelection = useAppSelector(selectSequenceSelection);
+  const sequenceSelectionRef = useRef(sequenceSelection);
+  sequenceSelectionRef.current = sequenceSelection;
```
la-nuta: Same `react-hooks/refs` problem, and here the real signal is different: eslint already flags this branch as `no-event-handler`. The ref hides that rather than addressing it. Since only a library click triggers this branch, the clean fix is moving the `setSequenceSelection` dispatch into `handleItemSelection` (`RnaElements.tsx:71`). If that's out of scope, prefer an explicit `eslint-disable-next-line` with a rationale — same suppression, but greppable.

packages/ketcher-macromolecules/src/components/monomerLibrary/RnaBuilder/RnaEditor/RnaEditorExpanded/RnaEditorExpanded.tsx
```diff
@@ -296,9 +311,12 @@ export const RnaEditorExpanded = ({
       }
     }
   }, [
-    activePresetMonomerGroup?.groupItem,
+    activeMonomerGroup,
```
la-nuta: Adding `activeMonomerGroup`/`isEditMode` changes user-visible behavior.

Repro: create a preset → rename it `TestMonomers` → Save → **reload** → select → Edit → click the Sugar slot without picking a monomer. On reload `getPresets` rebuilds the preset field-by-field and drops `editedName`; the slot click now re-runs the effect, `activePresetMonomerGroup` is `null`, `!editedName` is true, and the name regenerates — `TestMonomers` becomes `25R(baA)bP`, and Save persists that. On master the effect doesn't fire for a slot click at all.

Same path fires on a **layout-mode switch**, since `resolvePhosphatePosition` is now a dep and closes over `isSequenceMode`.

E2E won't catch it: `rna-library.spec.ts:601` renames and saves but never clicks a slot after; the autofill case at :617 only uses a fresh empty preset.

Suggest keeping the original trigger set with a documented disable, or moving the guard values out of the effect entirely (compute in `selectGroup`/`handleItemSelection`) if the goal is a suppression-free file.

packages/ketcher-macromolecules/src/components/monomerLibrary/RnaBuilder/RnaEditor/RnaEditorExpanded/RnaEditorExpanded.tsx
```diff
@@ -296,9 +311,12 @@ export const RnaEditorExpanded = ({
...
+    isEditMode,
     isSequenceEditInRNABuilderMode,
-    selectedPhosphatePosition,
+    activePresetMonomerGroup,
```
la-nuta: Widening `?.groupItem` to the whole object also makes re-selecting the *same* monomer re-trigger: `RnaElements.tsx:94` dispatches a fresh object literal per click. Clicking the same base twice now re-dispatches `setSequenceSelection` with a new array, cascading into the `[dispatch, sequenceSelection]` effect. Idempotent, so nothing breaks — but it's churn the old dep deliberately avoided. Please keep `?.groupItem` or hoist it to a local.

packages/ketcher-macromolecules/src/components/monomerLibrary/RnaBuilder/RnaEditor/RnaEditorExpanded/RnaEditorExpanded.tsx
```diff
@@ -630,7 +655,7 @@ export const RnaEditorExpanded = ({
     return () => {
       editor?.events.keyDown.remove(handleKeyDown);
     };
-  }, [editor, sequenceSelection]);
+  }, [editor, onCancel, onUpdateSequence, isSequenceEditInRNABuilderMode]);
```
la-nuta: This is a genuine **bug fix**, not just a lint change — worth calling out rather than shipping inside a "pure refactor". The old deps left `handleKeyDown` holding a stale `onCancel`: switch the active preset without touching `sequenceSelection`, press Escape, and the builder reset to the *previous* preset. Both callbacks depend only on stable store refs, so no re-subscribe churn.

Please mention it in the description and add an E2E: select preset A → select B → Edit → Escape → assert B is restored.

packages/ketcher-macromolecules/src/components/monomerLibrary/RnaBuilder/RnaEditor/RnaEditorExpanded/RnaEditorExpanded.tsx
la-nuta: Heads-up: both render-phase ref writes fail lint once this merges. `react-hooks/refs` is `error` for `ketcher-macromolecules/**` and isn't covered by the file's disable header:

```
140:3  error  Cannot access refs during render   react-hooks/refs
200:3  error  Cannot access refs during render   react-hooks/refs
```

CI is green here only because this branch's base still used `.eslintrc.json`, which had no react-hooks rules at all.

For this one, `setNewPreset((prev) => ...)` removes the `newPreset` dependency without a ref and satisfies both rules. It also collapses the old double `setNewPreset` into one — worth mentioning in the description if you go that way.

packages/ketcher-macromolecules/src/components/monomerLibrary/RnaBuilder/RnaEditor/RnaEditorExpanded/RnaEditorExpanded.tsx
la-nuta: Same `react-hooks/refs` error here (200:3). Worth noting eslint already flags this branch as `no-event-handler` — it only runs in response to a library click — so moving the `setSequenceSelection` dispatch into `handleItemSelection` (`RnaElements.tsx:71`) would remove the branch rather than work around it. If that's out of scope, an explicit `eslint-disable-next-line` with a reason is at least greppable, unlike a ref.

packages/ketcher-macromolecules/src/components/monomerLibrary/RnaBuilder/RnaEditor/RnaEditorExpanded/RnaEditorExpanded.tsx
la-nuta: Adding `activeMonomerGroup` and `isEditMode` as triggers changes behaviour — the effect now fires on a slot click, where before it didn't.

Repro: create a preset → rename it `TestMonomers` → Save → **reload** → select it → Edit → click the Sugar slot without picking a monomer. `getPresets` (`helpers/getPreset.ts:96`) rebuilds the preset field-by-field and drops `editedName`, so the effect re-runs, `activePresetMonomerGroup` is `null`, `!editedName` is true, and the name regenerates to `25R(baA)bP` — which Save then persists.

Same path fires on a layout-mode switch, since `resolvePhosphatePosition` is now a dep and closes over `isSequenceMode`.

Could you keep the original trigger set (with a documented disable if needed), or move the guard values out of the effect entirely?

Also worth splitting out: the keyDown dep change is a real bug fix — the old deps left `handleKeyDown` holding a stale `onCancel`, so Escape after switching presets restored the previous one. Worth a line in the description and an e2e, rather than shipping silently inside a refactor.

packages/ketcher-macromolecules/src/components/monomerLibrary/RnaBuilder/RnaEditor/RnaEditorExpanded/RnaEditorExpanded.tsx
la-nuta: Both render-phase ref writes fail lint under the current config — I ran the PR head through it just now:

```
137:3  error  Cannot access refs during render   react-hooks/refs
197:3  error  Cannot access refs during render   react-hooks/refs
```

`react-hooks/refs` is `error` for `ketcher-macromolecules/**` and isn't covered by the file's disable header, so these survive whatever the rebase does. CI was green only because this branch's base predates the rule being active.

For this one, `setNewPreset((prev) => ...)` drops the `newPreset` dependency without needing a ref, and satisfies both rules. It also collapses the old double `setNewPreset` into one.

packages/ketcher-macromolecules/src/components/monomerLibrary/RnaBuilder/RnaEditor/RnaEditorExpanded/RnaEditorExpanded.tsx
la-nuta: Same `react-hooks/refs` error here (197:3). Worth noting eslint already flags this branch as `no-event-handler` — it only runs in response to a library click — so moving the `setSequenceSelection` dispatch into `handleItemSelection` (`RnaElements.tsx:71`) would remove the branch rather than work around it. If that's out of scope, an explicit `eslint-disable-next-line` with a reason is at least greppable, unlike a ref.

packages/ketcher-macromolecules/src/components/monomerLibrary/RnaBuilder/RnaEditor/RnaEditorExpanded/RnaEditorExpanded.tsx
la-nuta: Adding `activeMonomerGroup` and `isEditMode` as triggers changes behaviour — the effect now fires on a slot click, where before it didn't.

Repro: create a preset → rename it `TestMonomers` → Save → **reload** → select it → Edit → click the Sugar slot without picking a monomer. `getPresets` (`helpers/getPreset.ts:96`) rebuilds the preset field-by-field and drops `editedName`, so the effect re-runs, `activePresetMonomerGroup` is `null`, `!editedName` is true, and the name regenerates to `25R(baA)bP` — which Save then persists.

Same path fires on a layout-mode switch, since `resolvePhosphatePosition` is now a dep and closes over `isSequenceMode`.

Could you keep the original trigger set (with a documented disable if needed), or move the guard values out of the effect entirely?

Separately, the keyDown dep change is a real bug fix worth splitting out — the old deps left `handleKeyDown` holding a stale `onCancel`, so Escape after switching presets restored the previous one. Master still has that bug, and as noted above #10409 has now added a second effect with the same shape. Worth a line in the description and an e2e rather than shipping silently inside a refactor.

packages/ketcher-macromolecules/src/components/monomerLibrary/RnaBuilder/RnaEditor/RnaEditorExpanded/RnaEditorExpanded.tsx
la-nuta: Both render-phase ref writes fail lint under the current config — I ran the PR head through it just now:

```
137:3  error  Cannot access refs during render   react-hooks/refs
197:3  error  Cannot access refs during render   react-hooks/refs
```

`react-hooks/refs` is `error` for `ketcher-macromolecules/**` and isn't covered by the file's disable header, so these survive whatever the rebase does. CI was green only because this branch's base predates the rule being active.

For this one, `setNewPreset((prev) => ...)` drops the `newPreset` dependency without needing a ref, and satisfies both rules. It also collapses the old double `setNewPreset` into one.

packages/ketcher-macromolecules/src/components/monomerLibrary/RnaBuilder/RnaEditor/RnaEditorExpanded/RnaEditorExpanded.tsx
la-nuta: Same `react-hooks/refs` error here (197:3). Worth noting eslint already flags this branch as `no-event-handler` — it only runs in response to a library click — so moving the `setSequenceSelection` dispatch into `handleItemSelection` (`RnaElements.tsx:71`) would remove the branch rather than work around it. If that's out of scope, an explicit `eslint-disable-next-line` with a reason is at least greppable, unlike a ref.

packages/ketcher-macromolecules/src/components/monomerLibrary/RnaBuilder/RnaEditor/RnaEditorExpanded/RnaEditorExpanded.tsx
la-nuta: Adding `activeMonomerGroup` and `isEditMode` as triggers changes behaviour — the effect now fires on a slot click, where before it didn't.

Repro: create a preset → rename it `TestMonomers` → Save → **reload** → select it → Edit → click the Sugar slot without picking a monomer. `getPresets` (`helpers/getPreset.ts:96`) rebuilds the preset field-by-field and drops `editedName`, so the effect re-runs, `activePresetMonomerGroup` is `null`, `!editedName` is true, and the name regenerates to `25R(baA)bP` — which Save then persists.

Same path fires on a layout-mode switch, since `resolvePhosphatePosition` is now a dep and closes over `isSequenceMode`.

Could you keep the original trigger set (with a documented disable if needed), or move the guard values out of the effect entirely?

Separately, the keyDown dep change is a real bug fix worth splitting out — the old deps left `handleKeyDown` holding a stale `onCancel`, so Escape after switching presets restored the previous one. Master still has that bug, and as noted above #10409 has now added a second effect with the same shape. Worth a line in the description and an e2e rather than shipping silently inside a refactor.

packages/ketcher-macromolecules/src/components/monomerLibrary/RnaBuilder/RnaEditor/RnaEditorExpanded/RnaEditorExpanded.tsx
la-nuta: Both render-phase ref writes fail lint under the current config — I ran the PR head through it just now:

```
137:3  error  Cannot access refs during render   react-hooks/refs
197:3  error  Cannot access refs during render   react-hooks/refs
```

`react-hooks/refs` is `error` for `ketcher-macromolecules/**` and isn't covered by the file's disable header, so these survive whatever the rebase does. CI was green only because this branch's base predates the rule being active.

For this one, `setNewPreset((prev) => ...)` drops the `newPreset` dependency without needing a ref. I built that variant and linted it — the `137:3` error goes away and no `exhaustive-deps` error replaces it. It also collapses the old double `setNewPreset` into one.

packages/ketcher-macromolecules/src/components/monomerLibrary/RnaBuilder/RnaEditor/RnaEditorExpanded/RnaEditorExpanded.tsx
la-nuta, line 217: Same `react-hooks/refs` error here (197:3). Worth noting eslint already flags this branch as `no-event-handler` — it only runs in response to a library click — so moving the `setSequenceSelection` dispatch into `handleItemSelection` (`RnaElements.tsx:71`) would remove the branch rather than work around it. If that's out of scope, an explicit `eslint-disable-next-line react-hooks/refs` with a reason is at least greppable, unlike a ref — I checked, and it does clear this error.

packages/ketcher-macromolecules/src/components/monomerLibrary/RnaBuilder/RnaEditor/RnaEditorExpanded/RnaEditorExpanded.tsx
la-nuta: Adding `activeMonomerGroup` and `isEditMode` as triggers changes behaviour — the effect now fires on a slot click, where before it didn't.

Repro: create a preset → rename it `TestMonomers` → Save → **reload** → select it → Edit → click the Sugar slot without picking a monomer. `getPresets` (`helpers/getPreset.ts:96`) rebuilds the preset field-by-field and drops `editedName`, so the effect re-runs, `activePresetMonomerGroup` is `null`, `!editedName` is true, and the name regenerates to `25R(baA)bP` — which Save then persists.

Same path fires on a layout-mode switch, since `resolvePhosphatePosition` is now a dep and closes over `isSequenceMode`.

Reverting to the original trigger set does fix the behaviour, but it isn't free: I linted that variant and `exhaustive-deps` then errors on this effect, so a disable has to stay — which is the thing we'd like to delete from line 1. The widened set is what makes the file suppression-free.

So either accept a narrow inline disable with a documented reason (still better than the file-wide header), or move `activeMonomerGroup`/`isEditMode` out of the effect so they're no longer read inside it — that's the version that gets correct behaviour *and* a clean file.

Separately, the keyDown dep change is a real bug fix worth splitting out — the old deps left `handleKeyDown` holding a stale `onCancel`, so Escape after switching presets restored the previous one. Master still has that bug, and as noted above #10409 has now added a second effect with the same shape. Worth a line in the description and an e2e rather than shipping silently inside a refactor.

la-nuta (commented):
This one needs a pass over master before it can be reviewed properly — it's `CONFLICTING` now, and the conflict is with feature work rather than a stale base. Three commits have touched this file since the branch was cut:

- `308f8e10` (#11421, 24 Aug) — ESLint upgrade
- `053c0f17` (#10409, 30 Aug) — exit RNA Builder by clicking away or pressing Escape
- `5d87a528` (#10760, 2 Sep) — fix removing selection in Modify in RNA Builder

The last two land in the same region you're rewriting, so this is real resolution work, not a mechanical catch-up.

**Worth folding in while you're there: #10409 added a second effect with the same defect this PR fixes.** Near the bottom of the file master now has:

```ts
const handleCancel = () => onCancel();
editor?.events.cancelSequenceEditInRNABuilderMode.add(handleCancel);
// ...
}, [editor, isSequenceEditInRNABuilderMode]);
```

`onCancel` isn't in the deps — the same stale-closure shape as the keyDown effect. Better to fix both in one go than leave a matching bug next door.

On the `exhaustive-deps` header at line 1 — there's a trade-off worth naming, which I only found by building the variants and linting them:

- **With your widened dep set**, once both ref writes are fixed, `exhaustive-deps` reports nothing and line 1 becomes deletable (`no-chain-state-updates` also goes quiet, so part of line 2 too). That's what closes #11287.
- **With the original narrow set**, `exhaustive-deps` errors on the effect, so a disable has to stay — an inline one with a reason, rather than the file-wide header.

So "keep the narrow triggers" and "delete line 1" can't both happen as-is; see the inline note. Moving the guard values out of the effect would get both. Also minor: the ESLint upgrade adds a `prettier` complaint around `position ?? selectedPhosphatePosition ?? 'right'` that the resolution will pick up.

Two blockers inline, plus a coordination note.

**Coordination:** #11346 is still open and rewrites the same two regions for `set-state-in-effect`. It keeps `updatePresetMonomerGroup` (which you delete), keeps the `useEffect([activePreset])` (which you delete), and deliberately keeps its deps narrow where you widen them. Whichever lands second won't just conflict, it'll be broken — worth agreeing an order with @nickkurkan.

AlexeyGirin:

packages/ketcher-macromolecules/src/components/monomerLibrary/RnaBuilder/RnaEditor/RnaEditorExpanded/RnaEditorExpanded.tsx
LashaDevLasha: On the guard refs: Fixed via the latest-value ref pattern — activeMonomerGroupRef, isEditModeRef, and resolvePhosphatePositionRef are written synchronously each render; the effect reads them through .current without listing them as deps. The effect no longer fires on slot clicks or layout-mode switches, and the file-wide exhaustive-deps suppression is deleted — no inline disable needed because those values aren't in the dep array at all.

packages/ketcher-macromolecules/src/components/monomerLibrary/RnaBuilder/RnaEditor/RnaEditorExpanded/RnaEditorExpanded.tsx
LashaDevLasha: The inline disable with reason is already in the current commit. The handleItemSelection refactor is the right long-term direction but touches RnaElements.tsx and the data flow between components — I'd rather keep that in a follow-up PR than expand the scope here.

la-nuta (approved):
This addresses everything I raised, and the description is genuinely good — documenting both stale-closure fixes with the Escape-restores-the-previous-preset symptom is exactly right, and catching the #10409 effect in the same pass was more than I asked for.

Verified on `20db7401`: the file lints clean, and with its remaining header stripped the only violations left are three `set-state-in-effect` sites (242, 260, 301) — which are #11346's scope, not yours. So `exhaustive-deps` and both `you-might-not-need-an-effect` rules are fully cleared and their two file-wide headers are correctly gone. The dep-array comment explaining why `activeMonomerGroup`/`isEditMode` stay out is the kind of thing that stops this being re-broken later.

**Approving, conditional on one thing that isn't code: settle the landing order with #11346 first.** It keeps `updatePresetMonomerGroup` (you've inlined it) and deletes the `useEffect([activePreset])` (you've kept it), so whichever merges second is broken rather than merely conflicted. Since you're rebased on master, lint-clean and behaviourally fixed, and #11346 is still on a 21 Aug base, the cheaper order looks like this one landing first and #11346 rebasing onto it — its scope then becomes exactly those three `set-state-in-effect` lines. Worth confirming with @nickkurkan before merge.

**Suggestion, not blocking:** nothing pins the name-regeneration behaviour. This PR introduced that regression and then fixed it, and the guard is currently a comment — which doesn't fail CI. A test on the repro (rename → reload → Edit → click the Sugar slot → assert the name is unchanged) would stop the next dep edit from quietly reintroducing it.

**Follow-up idea, definitely not this PR:** the four `react-hooks/refs` disables all exist so an effect can read a latest value without re-triggering. `useEffectEvent` is a stable export in the React version here (19.2.8) and is React's sanctioned answer for precisely that, so it would likely let all four go. Worth an issue rather than more scope here — and if #11346's render-phase conversion lands on top, those refs may become unnecessary anyway, since render-phase code reads current values directly.

la-nuta (commented):
Follow-up to my approval — that was on `20db7401` when the branch was still `MERGEABLE`. It isn't any more, and the cause is worth knowing before you touch it.

`cb882e6b` (#11675, "Default preset name changes unexpectedly after switching phosphate picker from 3' to 5'") landed on master yesterday and rewrote the same effect this PR restructures:

- dropped `selectedPhosphatePosition` from that effect's dep array
- simplified the name computation to `selectPresetFullName(currentPreset)`, removing the `buildRnaPresetConnections` / `resolvedPhosphatePosition` wrapping — code your version still carries
- removed `no-chain-state-updates` from the file header

So the approval still stands on the review points, but it describes a state that master has since moved past, and part of what you're carrying is code that's now been deleted upstream. Whenever you next pick up master, that hunk is a judgement call rather than a mechanical resolution — #11675 is fixing a preset-name bug in the same area your `activeMonomerGroup`/`isEditMode` guard work was aimed at, so it's worth reading before deciding how to reconcile them.

#11346 is `CONFLICTING` for the same reason, so you're both behind #11675 rather than behind each other. The landing order we discussed still looks right to me — this one first, given it clears two of the three headers and #11346 currently clears none — but both of you should see #11675 before committing to that.

packages/ketcher-macromolecules/src/components/monomerLibrary/RnaBuilder/RnaEditor/RnaEditorExpanded/RnaEditorExpanded.tsx
```diff
@@ -249,59 +264,79 @@ export const RnaEditorExpanded = ({
...
+            resolvePhosphatePositionRef.current(updatedPreset);
+          const presetFullName = updatedPreset.editedName
+            ? updatedPreset.name
+            : selectPresetFullName({
+                ...updatedPreset,
+                connections: buildRnaPresetConnections(
+                  updatedPreset,
+                  resolvedPhosphatePosition,
+                ),
+              });
```
la-nuta: #11675 moved the phosphate-position logic *inside* `selectPresetFullName` — it dropped the `getRnaPresetPhosphatePosition` call (`rnaBuilderSlice.ts:447`), so the function no longer looks at `connections` at all. Building them here now feeds a field nothing reads.

To be clear, this isn't reintroducing #9693 — the computation is inert, not wrong. But it's dead weight, and removing it lets `resolvePhosphatePositionRef`, its `react-hooks/refs` disable, the comment above it and the dep-array note about it all go too (4 refs down to 3). Verified: the result still lints clean, and `resolvePhosphatePosition` / `buildRnaPresetConnections` stay in use on the save path.

```suggestion
          const presetFullName = updatedPreset.editedName
            ? updatedPreset.name
            : selectPresetFullName(updatedPreset);
```

packages/ketcher-macromolecules/src/components/monomerLibrary/RnaBuilder/RnaEditor/RnaEditorExpanded/RnaEditorExpanded.tsx
```diff
@@ -249,59 +264,79 @@ export const RnaEditorExpanded = ({
...
+    // activeMonomerGroup and isEditMode are intentionally omitted — they are
+    // guard conditions read via refs. Including them would fire this effect on
+    // every slot click (activeMonomerGroup) or entering edit mode (isEditMode),
+    // which regenerates the preset name and drops editedName.
+    // resolvePhosphatePosition is intentionally omitted — it is stable (empty
+    // deps) and reads selectedPhosphatePosition/isSequenceMode via refs, so
+    // layout-mode switches no longer trigger this effect.
     isSequenceEditInRNABuilderMode,
-    selectedPhosphatePosition,
+    activePresetMonomerGroup,
```
la-nuta: This is the one I'd think hardest about while resolving. Master's dep is `activePresetMonomerGroup?.groupItem`; the whole object churns identity on every click, because `RnaElements.tsx:110` dispatches a fresh `{ groupName, groupItem: monomer }` literal, `rnaBuilderSlice.ts:234` assigns it with no equality check, and the selector is a plain accessor. `groupItem` is the library item itself, so it stays stable.

Effect: clicking the *already selected* monomer re-runs this effect where master didn't. In Modify-in-RNA-Builder that takes the first branch, so it re-dispatches `setSequenceSelection` with a fresh `.map()` array plus `setIsSequenceSelectionUpdated(true)`, and that new array identity re-fires the `[dispatch, sequenceSelection]` effect above, which dispatches `setSequenceSelectionName`. A redundant cascade per repeat click — terminating, but real.

Narrowing it back to `?.groupItem` isn't open to you, though, since the body also reads `.groupName`:

```
318:6  error  React Hook useEffect has a missing dependency: 'activePresetMonomerGroup'  react-hooks/exhaustive-deps
```

which is exactly what this PR removes. So it needs a third option — a ref for `groupName` with `groupItem` as the dep, or memoizing the object at the dispatch site. Your call which; flagging it because the diff alone doesn't show the identity churn.

packages/ketcher-macromolecules/src/components/monomerLibrary/RnaBuilder/RnaEditor/RnaEditorExpanded/RnaEditorExpanded.tsx
la-nuta: #11675 deleted this: `selectPresetFullName` no longer reads `connections` (`rnaBuilderSlice.ts:447`). Take master's side here — `resolvePhosphatePositionRef`, its `react-hooks/refs` disable and the comments above it go with it.

```suggestion
          const presetFullName = updatedPreset.editedName
            ? updatedPreset.name
            : selectPresetFullName(updatedPreset);
```

packages/ketcher-macromolecules/src/components/monomerLibrary/RnaBuilder/RnaEditor/RnaEditorExpanded/RnaEditorExpanded.tsx
la-nuta: `RnaElements.tsx:110` dispatches a fresh `{ groupName, groupItem }` per click, so this dep churns identity where master's `?.groupItem` didn't: clicking the already-selected monomer re-runs the effect, re-dispatching `setSequenceSelection` in Modify-in-RNA-Builder.

`?.groupItem` alone won't do, since the body also reads `.groupName`:

```
318:6  error  React Hook useEffect has a missing dependency: 'activePresetMonomerGroup'  react-hooks/exhaustive-deps
```

So: a ref for `groupName`, or memoize the object at the dispatch site.

packages/ketcher-macromolecules/src/components/monomerLibrary/RnaBuilder/RnaEditor/RnaEditorExpanded/RnaEditorExpanded.tsx
la-nuta: #11675 deleted this: `selectPresetFullName` no longer reads `connections` (`rnaBuilderSlice.ts:447`). Take master's side here — `resolvePhosphatePositionRef`, its `react-hooks/refs` disable and the comments above it go with it.

```suggestion
          const presetFullName = updatedPreset.editedName
            ? updatedPreset.name
            : selectPresetFullName(updatedPreset);
```

packages/ketcher-macromolecules/src/components/monomerLibrary/RnaBuilder/RnaEditor/RnaEditorExpanded/RnaEditorExpanded.tsx
la-nuta: `RnaElements.tsx:110` dispatches a fresh `{ groupName, groupItem }` per click, so this dep churns identity where master's `?.groupItem` didn't — clicking the already-selected monomer re-runs the effect and re-dispatches `setSequenceSelection` in Modify-in-RNA-Builder.

`?.groupItem` alone won't fix it (the body reads `.groupName` and twice tests the object for truthiness), and `!` is out — `no-non-null-assertion` is on. What passes lint + `tsc`: hoist `activePresetMonomerGroup?.groupItem` into a local, dep on that and use it for both truthiness checks, then read `groupName` through a ref.

la-nuta (commented):
Conflicts with master since #11675 (2026-09-08) touched this effect — the header block and the monomer-group effect. Two things to get right while resolving.

packages/ketcher-macromolecules/src/components/monomerLibrary/RnaBuilder/RnaEditor/RnaEditorExpanded/RnaEditorExpanded.tsx
LashaDevLasha: removed resolvePhosphatePositionRef and simplified to selectPresetFullName(updatedPreset).

packages/ketcher-macromolecules/src/components/monomerLibrary/RnaBuilder/RnaEditor/RnaEditorExpanded/RnaEditorExpanded.tsx
LashaDevLasha: hoisted activePresetMonomerGroup?.groupItem into activePresetGroupItem, dep on that, and read groupName through activePresetGroupNameRef

packages/ketcher-macromolecules/src/components/monomerLibrary/RnaBuilder/RnaEditor/RnaEditorExpanded/RnaEditorExpanded.tsx
```diff
@@ -249,48 +265,72 @@ export const RnaEditorExpanded = ({
...
-    if (activeMonomerGroup !== RnaBuilderPresetsItem.Presets && isEditMode) {
-      if (isSequenceEditInRNABuilderMode && activePresetMonomerGroup) {
+    if (
+      activeMonomerGroupRef.current !== RnaBuilderPresetsItem.Presets &&
+      isEditModeRef.current
+    ) {
+      if (isSequenceEditInRNABuilderMode && activePresetGroupItem) {
         const monomerType =
-          monomerGroupToPresetGroup[activePresetMonomerGroup.groupName];
+          monomerGroupToPresetGroup[activePresetGroupNameRef.current ?? ''];
```
la-nuta: `monomerGroupToPresetGroup['']` is `undefined`, so an empty ref silently gives `field = 'undefinedLabel'` here and an `undefined` computed key at line 311 — a corrupt preset rather than a throw.

I traced it as unreachable today (`groupName` and `groupItem` are always dispatched together, and the ref is rewritten every render), so this is about not hiding the case if that stops being true. Guarding costs nothing and lints + typechecks clean:

```suggestion
      const activePresetGroupName = activePresetGroupNameRef.current;
      if (
        isSequenceEditInRNABuilderMode &&
        activePresetGroupItem &&
        activePresetGroupName
      ) {
        const monomerType = monomerGroupToPresetGroup[activePresetGroupName];
```

Same at line 311 — `activePresetGroupItem && activePresetGroupName` in the ternary, then index with the local.

la-nuta (commented):
Looks good — rebased cleanly, lints clean, and both earlier points are addressed. Nothing further from me.

packages/ketcher-macromolecules/src/components/monomerLibrary/RnaBuilder/RnaEditor/RnaEditorExpanded/RnaEditorExpanded.tsx
```diff
@@ -183,8 +191,30 @@ export const RnaEditorExpanded = ({
     return undefined;
   };
 
+  // Guard refs — read inside the monomer-group effect without being listed as
+  // deps, preventing re-runs on slot clicks (activeMonomerGroup) or edit-mode
+  // toggle (isEditMode), which would drop editedName and regenerate the name.
+  const activeMonomerGroupRef = useRef(activeMonomerGroup);
+  // eslint-disable-next-line react-hooks/refs -- guard ref: latest value readable in effects without triggering re-runs on slot clicks
```
AlexeyGirin: Please eliminate eslint-disable

AlexeyGirin (changes requested):

packages/ketcher-macromolecules/src/components/monomerLibrary/RnaBuilder/RnaEditor/RnaEditorExpanded/RnaEditorExpanded.tsx
LashaDevLasha: Done

PR #11332, opened 2026-08-19
#11284 - Refactor: react-hooks/exhaustive-deps rule violation clean up from packages\ketcher-macromolecules\src\components\SequenceSyncEditModeButton\SequenceSyncEditModeButton.tsx file

### LINKED ISSUE #11284: Refactor: `react-hooks/exhaustive-deps` rule violation clean up from `packages\ketcher-macromolecules\src\components\SequenceSyncEditModeButton\SequenceSyncEditModeButton.tsx` file

Example:
```ts

useEffect(() => {
  loadUser(userId);
}, []); // userId is missing
```

**Why is this a problem?**

This can cause:
- stale state or props;
- effects using old data;

**How is it usually fixed?**
```ts
useEffect(() => {
  loadUser(userId);
}, [userId]);
```

For a memoized callback:
```ts
}, [documentId, content]);
```

For a memoized value:
```ts
  [items, category],
);
```

**Problem locations:**
packages\ketcher-macromolecules\src\components\SequenceSyncEditModeButton\SequenceSyncEditModeButton.tsx

LashaDevLasha:
…atisfy exhaustive-deps rule

## Check list
- [ ] unit-tests written
- [ ] e2e-tests written
- [ ] documentation updated

### DIFF
--- a/packages/ketcher-macromolecules/src/components/SequenceSyncEditModeButton/SequenceSyncEditModeButton.tsx
+++ b/packages/ketcher-macromolecules/src/components/SequenceSyncEditModeButton/SequenceSyncEditModeButton.tsx
@@ -61,6 +61,7 @@ export const SequenceSyncEditModeButton = () => {
   const handleClick = () => {
     const isSequenceSyncEditModeNewState = !isSequenceSyncEditMode;
 
+    isSequenceSyncEditModeRef.current = isSequenceSyncEditModeNewState;
     setIsSequenceSyncEditMode(isSequenceSyncEditModeNewState);
     editor?.events.toggleIsSequenceSyncEditMode.dispatch(
       isSequenceSyncEditModeNewState,
--- a/packages/ketcher-macromolecules/src/components/SequenceSyncEditModeButton/SequenceSyncEditModeButton.tsx
+++ b/packages/ketcher-macromolecules/src/components/SequenceSyncEditModeButton/SequenceSyncEditModeButton.tsx
@@ -14,7 +14,7 @@
  * limitations under the License.
  ***************************************************************************/
 
-import { useEffect, useState } from 'react';
+import { useEffect, useRef, useState } from 'react';
 import { useAppSelector, useLayoutMode } from 'hooks';
 import { selectEditor } from 'state/common';
 import styled from '@emotion/styled';
@@ -55,9 +55,12 @@ export const SequenceSyncEditModeButton = () => {
     Boolean(editor?.drawingEntitiesManager?.hasAntisenseChains),
   );
 
+  const isSequenceSyncEditModeRef = useRef(isSequenceSyncEditMode);
+
   const handleClick = () => {
     const isSequenceSyncEditModeNewState = !isSequenceSyncEditMode;
 
+    isSequenceSyncEditModeRef.current = isSequenceSyncEditModeNewState;
     setIsSequenceSyncEditMode(isSequenceSyncEditModeNewState);
     editor?.events.toggleIsSequenceSyncEditMode.dispatch(
       isSequenceSyncEditModeNewState,
@@ -75,7 +78,7 @@ export const SequenceSyncEditModeButton = () => {
 
       if (isSequenceMode && hasAntisenseChains) {
         editor?.events.toggleIsSequenceSyncEditMode.dispatch(
-          isSequenceSyncEditMode,
+          isSequenceSyncEditModeRef.current,
         );
       }
     };
@@ -86,7 +89,7 @@ export const SequenceSyncEditModeButton = () => {
     return () => {
       editor?.events.modelChange.remove(updateHasAntisenseChains);
     };
-  }, [editor, isSequenceMode, isSequenceSyncEditMode]);
+  }, [editor, isSequenceMode]);
 
   return isSequenceMode && hasAtLeastOneAntisense ? (
     <StyledButton
--- a/packages/ketcher-macromolecules/src/components/SequenceSyncEditModeButton/SequenceSyncEditModeButton.tsx
+++ b/packages/ketcher-macromolecules/src/components/SequenceSyncEditModeButton/SequenceSyncEditModeButton.tsx
@@ -14,7 +14,7 @@
  * limitations under the License.
  ***************************************************************************/
 
-import { useEffect, useState } from 'react';
+import { useEffect, useRef, useState } from 'react';
 import { useAppSelector, useLayoutMode } from 'hooks';
 import { selectEditor } from 'state/common';
 import styled from '@emotion/styled';
@@ -55,6 +55,9 @@ export const SequenceSyncEditModeButton = () => {
     Boolean(editor?.drawingEntitiesManager?.hasAntisenseChains),
   );
 
+  const isSequenceSyncEditModeRef = useRef(isSequenceSyncEditMode);
+  isSequenceSyncEditModeRef.current = isSequenceSyncEditMode;
+
   const handleClick = () => {
     const isSequenceSyncEditModeNewState = !isSequenceSyncEditMode;
 
@@ -82,10 +85,10 @@ export const SequenceSyncEditModeButton = () => {
   useEffect(() => {
     if (isSequenceMode && hasAtLeastOneAntisense) {
       editor?.events.toggleIsSequenceSyncEditMode.dispatch(
-        isSequenceSyncEditMode,
+        isSequenceSyncEditModeRef.current,
       );
     }
-  }, [isSequenceMode, hasAtLeastOneAntisense]);
+  }, [isSequenceMode, hasAtLeastOneAntisense, editor]);
 
   return isSequenceMode && hasAtLeastOneAntisense ? (
     <StyledButton
--- a/packages/ketcher-macromolecules/src/components/SequenceSyncEditModeButton/SequenceSyncEditModeButton.tsx
+++ b/packages/ketcher-macromolecules/src/components/SequenceSyncEditModeButton/SequenceSyncEditModeButton.tsx
@@ -14,7 +14,7 @@
  * limitations under the License.
  ***************************************************************************/
 
-import { useEffect, useRef, useState } from 'react';
+import { useEffect, useState } from 'react';
 import { useAppSelector, useLayoutMode } from 'hooks';
 import { selectEditor } from 'state/common';
 import styled from '@emotion/styled';
@@ -55,13 +55,9 @@ export const SequenceSyncEditModeButton = () => {
     Boolean(editor?.drawingEntitiesManager?.hasAntisenseChains),
   );
 
-  const isSequenceSyncEditModeRef = useRef(isSequenceSyncEditMode);
-  isSequenceSyncEditModeRef.current = isSequenceSyncEditMode;
-
   const handleClick = () => {
     const isSequenceSyncEditModeNewState = !isSequenceSyncEditMode;
 
-    isSequenceSyncEditModeRef.current = isSequenceSyncEditModeNewState;
     setIsSequenceSyncEditMode(isSequenceSyncEditModeNewState);
     editor?.events.toggleIsSequenceSyncEditMode.dispatch(
       isSequenceSyncEditModeNewState,
@@ -79,7 +75,7 @@ export const SequenceSyncEditModeButton = () => {
 
       if (isSequenceMode && hasAntisenseChains) {
         editor?.events.toggleIsSequenceSyncEditMode.dispatch(
-          isSequenceSyncEditModeRef.current,
+          isSequenceSyncEditMode,
         );
       }
     };
@@ -90,15 +86,7 @@ export const SequenceSyncEditModeButton = () => {
     return () => {
       editor?.events.modelChange.remove(updateHasAntisenseChains);
     };
-  }, [editor]);
-
-  useEffect(() => {
-    if (isSequenceMode && hasAtLeastOneAntisense) {
-      editor?.events.toggleIsSequenceSyncEditMode.dispatch(
-        isSequenceSyncEditModeRef.current,
-      );
-    }
-  }, [isSequenceMode, hasAtLeastOneAntisense, editor]);
+  }, [editor, isSequenceMode, isSequenceSyncEditMode]);
 
   return isSequenceMode && hasAtLeastOneAntisense ? (
     <StyledButton
--- a/packages/ketcher-macromolecules/src/components/SequenceSyncEditModeButton/SequenceSyncEditModeButton.tsx
+++ b/packages/ketcher-macromolecules/src/components/SequenceSyncEditModeButton/SequenceSyncEditModeButton.tsx
@@ -14,7 +14,7 @@
  * limitations under the License.
  ***************************************************************************/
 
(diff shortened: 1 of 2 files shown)


AlexeyGirin:

packages/ketcher-macromolecules/src/components/SequenceSyncEditModeButton/SequenceSyncEditModeButton.tsx
```diff
@@ -55,6 +55,12 @@ export const SequenceSyncEditModeButton = () => {
     Boolean(editor?.drawingEntitiesManager?.hasAntisenseChains),
   );
 
+  const isSequenceSyncEditModeRef = useRef(isSequenceSyncEditMode);
```
la-nuta: Nit: the extra effect isn't needed for a latest-value ref. Either assign in the component body:

```ts
const isSequenceSyncEditModeRef = useRef(isSequenceSyncEditMode);
isSequenceSyncEditModeRef.current = isSequenceSyncEditMode;
```

or just write it in `handleClick` (the only place the state ever changes), which also closes the staleness window noted below. Both avoid a second render-commit cycle.

packages/ketcher-macromolecules/src/components/SequenceSyncEditModeButton/SequenceSyncEditModeButton.tsx
```diff
@@ -75,7 +81,7 @@ export const SequenceSyncEditModeButton = () => {
 
       if (isSequenceMode && hasAntisenseChains) {
         editor?.events.toggleIsSequenceSyncEditMode.dispatch(
-          isSequenceSyncEditMode,
+          isSequenceSyncEditModeRef.current,
```
la-nuta: Worth noting a subtlety this introduces: `handleClick` dispatches to the editor synchronously, *before* React re-renders and before the ref-sync effect above runs — so for the duration of that editor call `isSequenceSyncEditModeRef.current` still holds the pre-click value.

Today that's harmless, because `modelChange` is only dispatched from `EditorHistory.update/undo/redo` and the sync-mode toggle path goes through `renderersContainer.update()` instead, so nothing can observe the stale ref inside that window.

But the previous deps array gave us a free self-correction (the effect re-ran and re-dispatched the fresh value); we're giving that up. If sync-mode toggling ever starts recording a history command, the handler would re-dispatch the old value and flip sync mode back, with nothing to correct it. Cheap insurance: also update the ref in `handleClick`, right next to `setIsSequenceSyncEditMode`.

packages/ketcher-macromolecules/src/components/SequenceSyncEditModeButton/SequenceSyncEditModeButton.tsx
```diff
@@ -86,7 +92,7 @@ export const SequenceSyncEditModeButton = () => {
     return () => {
       editor?.events.modelChange.remove(updateHasAntisenseChains);
     };
-  }, [editor, isSequenceMode, isSequenceSyncEditMode]);
+  }, [editor, isSequenceMode]);
```
la-nuta: Could we add a test for this? `SequenceSyncEditModeButton.test.tsx` currently only asserts visibility, so it passes identically before and after this change — the dispatch semantics being refactored here are completely uncovered.

Two cases would pin it down: (1) clicking the button dispatches `toggleIsSequenceSyncEditMode` exactly once with the new value, and (2) a `modelChange` fired *after* a toggle dispatches the *latest* value, not the value captured at subscribe time. The `FakeSubscription` harness already in the file makes both easy.

packages/ketcher-macromolecules/src/components/SequenceSyncEditModeButton/SequenceSyncEditModeButton.tsx
la-nuta: Nit: a latest-value ref doesn't need an effect — `isSequenceSyncEditModeRef.current = isSequenceSyncEditMode;` in the component body does it in one line, without a second render-commit cycle. Setting it in `handleClick` instead also closes the staleness window noted below.

packages/ketcher-macromolecules/src/components/SequenceSyncEditModeButton/SequenceSyncEditModeButton.tsx
la-nuta: `handleClick` dispatches to the editor synchronously, before React re-renders and before the ref-sync effect runs — so the ref is stale for the duration of that call.

Harmless today (`modelChange` only fires from `EditorHistory.update/undo/redo`, and the toggle path goes through `renderersContainer.update()`), but the old deps array gave us a free self-correction that we're giving up. If toggling ever records a history command, the handler re-dispatches the old value and nothing corrects it. Cheap insurance: set the ref in `handleClick` too.

packages/ketcher-macromolecules/src/components/SequenceSyncEditModeButton/SequenceSyncEditModeButton.tsx
la-nuta: Could we add a test? The existing suite only asserts visibility, so it passes identically before and after — the dispatch semantics being refactored are uncovered. Two cases: a click dispatches once with the new value, and a `modelChange` after a toggle dispatches the *latest* value. The `FakeSubscription` harness makes both easy.

packages/ketcher-macromolecules/src/components/SequenceSyncEditModeButton/SequenceSyncEditModeButton.tsx
la-nuta: If this PR does stay, the ref is heavier than the job needs. `isSequenceSyncEditModeRef.current = isSequenceSyncEditMode;` in the component body does the same in one line without a second render-commit cycle.

It also opens a small window: `handleClick` dispatches to the editor synchronously, before React re-renders and before this effect runs, so the ref is stale for the duration of that call. Harmless today — `modelChange` only fires from `EditorHistory.update/undo/redo` and the toggle path goes through `renderersContainer.update()` — but the old deps array gave us a free self-correction that this gives up. Setting the ref in `handleClick` avoids both problems.

packages/ketcher-macromolecules/src/components/SequenceSyncEditModeButton/SequenceSyncEditModeButton.tsx
la-nuta: This line is the actual value in the PR: removing `isSequenceSyncEditMode` stops the effect re-running on toggle, which is what eliminates the second `toggleIsSequenceSyncEditMode.dispatch` and the second `SequenceMode.initialize()` per click.

It's a genuine improvement — it just isn't what #11284 asked for, and it's being delivered via a ref that only exists to keep the dep array short. Guarding `turnOnSyncEditMode`/`turnOffSyncEditMode` against a no-op transition gets the same result without touching this component.

If we do keep the change here, it needs a test — `SequenceSyncEditModeButton.test.tsx` currently only asserts visibility, so it passes identically before and after and the dispatch semantics stay uncovered. Two cases would pin it: a click dispatches exactly once with the new value, and a `modelChange` fired after a toggle dispatches the *latest* value rather than the one captured at subscribe time. The `FakeSubscription` harness in the file already supports both.

packages/ketcher-macromolecules/src/components/SequenceSyncEditModeButton/SequenceSyncEditModeButton.tsx
la-nuta: If this PR does stay, the ref is heavier than the job needs. `isSequenceSyncEditModeRef.current = isSequenceSyncEditMode;` in the component body does the same in one line without a second render-commit cycle.

It also opens a small window: `handleClick` dispatches to the editor synchronously, before React re-renders and before this effect runs, so the ref is stale for the duration of that call. Harmless today — `modelChange` only fires from `EditorHistory.update/undo/redo` and the toggle path goes through `renderersContainer.update()` — but the old deps array gave us a free self-correction that this gives up. Setting the ref in `handleClick` avoids both problems.

packages/ketcher-macromolecules/src/components/SequenceSyncEditModeButton/SequenceSyncEditModeButton.tsx
la-nuta: This line is the actual value in the PR: removing `isSequenceSyncEditMode` stops the effect re-running on toggle, which is what eliminates the second `toggleIsSequenceSyncEditMode.dispatch` and the second `SequenceMode.initialize()` per click.

It's a genuine improvement — it just isn't what #11284 asked for, and it's being delivered via a ref that only exists to keep the dep array short. Guarding `turnOnSyncEditMode`/`turnOffSyncEditMode` against a no-op transition gets the same result without touching this component.

If we do keep the change here, it needs a test — `SequenceSyncEditModeButton.test.tsx` currently only asserts visibility, so it passes identically before and after and the dispatch semantics stay uncovered. Two cases would pin it: a click dispatches exactly once with the new value, and a `modelChange` fired after a toggle dispatches the *latest* value rather than the one captured at subscribe time. The `FakeSubscription` harness in the file already supports both.

packages/ketcher-macromolecules/src/components/SequenceSyncEditModeButton/SequenceSyncEditModeButton.tsx
la-nuta, line 58: If this PR does stay, the ref is heavier than the job needs. `isSequenceSyncEditModeRef.current = isSequenceSyncEditMode;` in the component body does the same in one line without a second render-commit cycle.

It also opens a small window: `handleClick` dispatches to the editor synchronously, before React re-renders and before this effect runs, so the ref is stale for the duration of that call. Harmless today — `modelChange` only fires from `EditorHistory.update/undo/redo` and the toggle path goes through `renderersContainer.update()` — but the old deps array gave us a free self-correction that this gives up. Setting the ref in `handleClick` avoids both problems.

packages/ketcher-macromolecules/src/components/SequenceSyncEditModeButton/SequenceSyncEditModeButton.tsx
la-nuta, line 95: This line is the actual value in the PR: removing `isSequenceSyncEditMode` stops the effect re-running on toggle, which is what eliminates the second `toggleIsSequenceSyncEditMode.dispatch` and the second `SequenceMode.initialize()` per click.

It's a genuine improvement — it just isn't what #11284 asked for, and it's being delivered via a ref that only exists to keep the dep array short. Guarding `turnOnSyncEditMode`/`turnOffSyncEditMode` against a no-op transition gets the same result without touching this component.

If we do keep the change here, it needs a test — `SequenceSyncEditModeButton.test.tsx` currently only asserts visibility, so it passes identically before and after and the dispatch semantics stay uncovered. Two cases would pin it: a click dispatches exactly once with the new value, and a `modelChange` fired after a toggle dispatches the *latest* value rather than the one captured at subscribe time. The `FakeSubscription` harness in the file already supports both.

la-nuta (commented):
**I think #11284 is already fixed, and this PR can be closed — but there's a real finding buried in it that's worth keeping.**

**The lint violation is gone on master.** #10806 (`e16cf3ee`, Aug 20) merged the two effects into one and listed all three deps; the violating effect was the old `}, [isSequenceMode, hasAtLeastOneAntisense]);`, which was missing `editor` and `isSequenceSyncEditMode`. ESLint on master's copy of this file now exits clean, and I confirmed the rule is genuinely active for this path — dropping a dep from master's version reproduces `89:6 error React Hook useEffect has a missing dependency: 'isSequenceSyncEditMode'`. So this isn't a false negative.

**What this PR still does is remove a duplicate dispatch.** On master one button click dispatches twice: once from `handleClick`, then again because the state change re-runs the effect, whose immediate `updateHasAntisenseChains()` re-dispatches the same value. Both land in `turnOnSyncEditMode`/`turnOffSyncEditMode`, which call `initialize(false, false, false)` **unconditionally** — so every toggle currently costs two full sequence re-initializes.

**But the better fix is one level down.** An early return in `turnOnSyncEditMode`/`turnOffSyncEditMode` when `_isSyncEditMode` already equals the target value fixes the redundant re-init for *every* caller, needs no ref, and has no staleness window. This component isn't necessarily the only source of a no-op dispatch.

Suggested resolution: close #11284 as already-fixed by #10806, and open a fresh issue for the double `initialize()` against `SequenceMode`. If you'd rather land the win here instead, see the inline notes.

packages/ketcher-macromolecules/src/components/SequenceSyncEditModeButton/SequenceSyncEditModeButton.tsx
LashaDevLasha: updated the ref in handleClick as you suggested. The body-update approach is cleaner, but the react-hooks/refs rule now active on this path flags ref.current = value during render as an error, so handleClick is the way to go.

packages/ketcher-macromolecules/src/components/SequenceSyncEditModeButton/SequenceSyncEditModeButton.tsx
LashaDevLasha: Done — added both tests you described. A click dispatches exactly once with the toggled value, and a modelChange fired after a toggle dispatches the ref's latest value rather than the stale closure capture.

la-nuta (approved):
Both points addressed, and you were right to correct me on the first one — I led with the component-body ref write, and `react-hooks/refs` does flag that. I confirmed it: injecting `isSequenceSyncEditModeRef.current = isSequenceSyncEditMode;` into the render body reproduces the error, so `handleClick` is the correct home for it. It also closes the staleness window properly, since the ref is now updated before the synchronous editor dispatch rather than after the next commit.

Verified before approving: the file lints clean with no disables, and two positive controls confirm the rules are actually running on this path (a render-phase ref write trips `react-hooks/refs`; dropping `editor` from the deps trips `exhaustive-deps`). Master has 2 tests and this head has 4, so the new cases were added rather than replacing the existing visibility ones, and CI is 19/19 including the jest job. The second test is the one that matters — spying after the click so it isolates the `modelChange` dispatch and pins that it carries the latest value rather than a stale closure capture.

Approving. One thing still worth settling separately, not blocking: the `exhaustive-deps` violation #11284 describes was already fixed on master by #10806 (`e16cf3ee`), which merged the two effects and listed all three deps — ESLint on master's copy of this file is clean today. What this PR actually delivers is removing the duplicate `toggleIsSequenceSyncEditMode.dispatch` per click, and with it a second full `SequenceMode.initialize()`. That's a real win and I'm happy for it to land here, but it'd be worth re-titling to match, and #11284 itself looks closeable. Longer term, an early return in `turnOnSyncEditMode`/`turnOffSyncEditMode` when `_isSyncEditMode` already equals the target would fix the redundant re-init for every caller rather than just this button — probably its own issue.

rrodionov91 (approved):

PR #11333, opened 2026-08-19
#11283 Refactor: react-hooks/exhaustive-deps rule violation clean up from packages\ketcher-macromolecules\src\components\SequenceTypeGroupButton\SequenceTypeGroupButton.tsx file

### LINKED ISSUE #11283: Refactor: `react-hooks/exhaustive-deps` rule violation clean up from `packages\ketcher-macromolecules\src\components\SequenceTypeGroupButton\SequenceTypeGroupButton.tsx` file

Example:
```ts

useEffect(() => {
  loadUser(userId);
}, []); // userId is missing
```

**Why is this a problem?**

This can cause:
- stale state or props;
- effects using old data;

**How is it usually fixed?**
```ts
useEffect(() => {
  loadUser(userId);
}, [userId]);
```

For a memoized callback:
```ts
}, [documentId, content]);
```

For a memoized value:
```ts
  [items, category],
);
```

**Problem locations:**
packages\ketcher-macromolecules\src\components\SequenceTypeGroupButton\SequenceTypeGroupButton.tsx

LashaDevLasha:
…o satisfy exhaustive-deps rule

## Check list
- [ ] unit-tests written
- [ ] e2e-tests written
- [ ] documentation updated


packages/ketcher-macromolecules/src/components/SequenceTypeGroupButton/SequenceTypeGroupButton.tsx
```diff
@@ -112,11 +112,11 @@ export const SequenceTypeGroupButton = () => {
       editor?.events.selectMode.remove(onToggleSequenceMode);
       editor?.events.changeSequenceTypeEnterMode.remove(onChangeSequenceType);
     };
-  }, [editor]);
+  }, [editor, dispatch, onToggleSequenceMode]);
 
   useEffect(() => {
     onToggleSequenceMode(layoutMode);
-  }, [layoutMode]);
+  }, [layoutMode, onToggleSequenceMode]);
```
svvald: This effect is not actually needed, it transitively calls set-state thus violating another eslint rule we have – please consider removing it

packages/ketcher-macromolecules/src/components/SequenceTypeGroupButton/SequenceTypeGroupButton.tsx
svvald: Moreover, we can drastically simplify this component now –  we don't need to mirror the current mode state here, `useLayoutMode` is sufficient enough and we already incorporated it in some other components such as `SequenceSyncEditModeButton`

So you can just derive it from `const isSequenceMode = layoutMode === 'sequence-layout-mode'`
It will allow you to remove this `useState`, the second unnecessary effect, remove `selectMode` add/remove pair from the first effect and remove `onToggleSequenceMode` handler

svvald (changes requested):

svvald (approved):

PR #11342, opened 2026-08-19
#11115 - Refactor: set-state-in-effect anti-pattern elimination from packages\ketcher-macromolecules\src\components\monomerLibrary\RnaBuilder\RnaEditor\RnaEditor.tsx file

### LINKED ISSUE #11115: Refactor: set-state-in-effect anti-pattern elimination from `packages\ketcher-macromolecules\src\components\monomerLibrary\RnaBuilder\RnaEditor\RnaEditor.tsx` file
**Problem:** 
1. Duplicate source of truth
2. Extra render cycle:
- Render
- Effect runs
- setState runs
- Second render

**Why is this an issue?**

**How can it be fixed?**

```ts
function Component({data}) {

  useEffect(() => {
  }, [data]);
}

function Component() {

  useEffect(() => {
  }, []);
}

  useEffect(() => {
  }, [rawData]);
}

  useEffect(() => {
  }, [selectedId, items]);
}
```

**Examples of correct code:**
```ts
function Tooltip() {
  const ref = useRef(null);

  useLayoutEffect(() => {
    setTooltipHeight(height);
  }, []);
}

// ✅ Calculate during render
}
```

**Problem locations:**
packages\ketcher-macromolecules\src\components\monomerLibrary\RnaBuilder\RnaEditor\RnaEditor.tsx:50

nickkurkan:
## Summary
- Fixes #11115 (closes #11115)
- `RnaEditor.tsx` had a `useEffect` that mixed a pure state sync (`setExpanded(true)` derived from `activePreset`/`isEditMode`) with a real side effect (dispatching `createNewPreset` / `setActiveRnaBuilderItem` when there is no active preset).
- Following the established pattern in this repo (see MeasureInput.tsx and RnaElementsAccordionView.tsx / #11113), the pure sync now happens during render: a `prevActivePreset` value is tracked in state and compared against the current `activePreset` on every render; when it changes, `setExpanded(true)` is called directly in the render body if the condition holds.
- The remaining `useEffect` now only performs the real side effect (dispatching redux actions) and is guarded by the same condition (`!activePreset`) as before, with `dispatch` added to the dependency array since it is used inside.
- No behavior change: the two dispatch calls still fire exactly when `activePreset` is falsy, and `expanded` is still only ever set to `true` under the same condition as before.

## Test plan
- [x] `cd packages/ketcher-macromolecules && npx eslint src/components/monomerLibrary/RnaBuilder/RnaEditor/RnaEditor.tsx` — clean (0 problems). Note: on `master` this file already had 2 pre-existing `react-hooks/exhaustive-deps` warnings; this change removes one of them (for the effect that was touched) without introducing any new warnings, and the other pre-existing warning (on the unrelated `recalculateRnaBuilderValidations` effect) is untouched and out of scope.
- [x] `npx tsc --noEmit -p tsconfig.json` in `packages/ketcher-macromolecules` — no new errors reference `RnaEditor.tsx` (there is a pre-existing, unrelated environment error: `Cannot find type definition file for 'react-dom'`, reproducible on `master` before this change).
- [x] No `RnaEditor.test.*` file exists in this folder (only `RnaEditorExpanded.test.tsx`, a different component) — no automated test to run for this file, and adding one is out of scope per the task.
- [ ] Manual verification in the running app: confirm the RNA Builder editor still auto-expands when a preset with a name is selected or edit mode is entered, and still creates a new preset / sets the active builder item to "Presets" when there is no active preset (I could not exercise this manually in this environment).


AlexeyGirin:

svvald (approved):

PR #11343, opened 2026-08-19
#11116 - Refactor: set-state-in-effect anti-pattern elimination from packages\ketcher-macromolecules\src\components\monomerLibrary\monomerLibraryList\MonomerList.tsx file

### LINKED ISSUE #11116: Refactor: set-state-in-effect anti-pattern elimination from `packages\ketcher-macromolecules\src\components\monomerLibrary\monomerLibraryList\MonomerList.tsx` file
**Problem:** 
1. Duplicate source of truth
2. Extra render cycle:
- Render
- Effect runs
- setState runs
- Second render

**Why is this an issue?**

**How can it be fixed?**

```ts
function Component({data}) {

  useEffect(() => {
  }, [data]);
}

function Component() {

  useEffect(() => {
  }, []);
}

  useEffect(() => {
  }, [rawData]);
}

  useEffect(() => {
  }, [selectedId, items]);
}
```

**Examples of correct code:**
```ts
function Tooltip() {
  const ref = useRef(null);

  useLayoutEffect(() => {
    setTooltipHeight(height);
  }, []);
}

// ✅ Calculate during render
}
```

**Problem locations:**
packages\ketcher-macromolecules\src\components\monomerLibrary\monomerLibraryList\MonomerList.tsx:85

nickkurkan:
## Summary
- Closes #11116.
- `MonomerList` had a `useEffect` that called `setSelectedMonomers('')` whenever `activeTool !== 'monomer'`, purely to sync local state from the `activeTool` selector — a classic set-state-in-effect anti-pattern (extra render cycle after the tool change).
- Replaced it with the established house-style fix (see #11113 / RnaElementsAccordionView): track `prevActiveTool` in state and adjust `selectedMonomers` directly in the render body when `activeTool` changes, instead of via `useEffect`.
- No behavior change: `selectedMonomers` is still reset to `''` exactly when `activeTool` transitions away from `'monomer'`, just one render cycle earlier.

## Test plan
- [x] `npx eslint src/components/monomerLibrary/monomerLibraryList/MonomerList.tsx` — clean.
- [x] `npx tsc --noEmit -p tsconfig.json` in `packages/ketcher-macromolecules` — no new errors for this file.
- [ ] `MonomerList.test.tsx` — could not run locally: `jest-environment-jsdom` throws `TypeError: Cannot read properties of undefined (reading 'html')` on this checkout. Verified this is a pre-existing environment issue unrelated to this change (reproduces identically on unmodified `master`/pre-refactor code via `git stash`). Needs a manual test run in a working CI/local environment to fully confirm.

### DIFF
--- a/packages/ketcher-macromolecules/src/components/monomerLibrary/monomerLibraryList/MonomerList.tsx
+++ b/packages/ketcher-macromolecules/src/components/monomerLibrary/monomerLibraryList/MonomerList.tsx
@@ -14,7 +14,7 @@
  * limitations under the License.
  ***************************************************************************/
 
-import { useEffect, useState } from 'react';
+import { useState } from 'react';
 import { MonomerGroup } from '../monomerLibraryGroup';
 import { useAppSelector } from 'hooks';
 import { MonomerListContainer } from './styles';
@@ -78,12 +78,14 @@ const MonomerList = ({
     ? selectAmbiguousMonomersInFavorites(monomers)
     : selectAmbiguousMonomersInCategory(monomers, MonomerGroups.PEPTIDES);
   const [selectedMonomers, setSelectedMonomers] = useState('');
+  const [prevActiveTool, setPrevActiveTool] = useState(activeTool);
 
-  useEffect(() => {
+  if (activeTool !== prevActiveTool) {
+    setPrevActiveTool(activeTool);
     if (activeTool !== 'monomer') {
       setSelectedMonomers('');
     }
-  }, [activeTool]);
+  }
 
   return (
     <MonomerListContainer>
--- a/packages/ketcher-macromolecules/src/components/monomerLibrary/monomerLibraryList/MonomerList.tsx
+++ b/packages/ketcher-macromolecules/src/components/monomerLibrary/monomerLibraryList/MonomerList.tsx
@@ -14,7 +14,6 @@
  * limitations under the License.
  ***************************************************************************/
 
-import { useState } from 'react';
 import { MonomerGroup } from '../monomerLibraryGroup';
 import { useAppSelector } from 'hooks';
 import { MonomerListContainer } from './styles';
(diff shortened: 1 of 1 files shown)


packages/ketcher-macromolecules/src/components/monomerLibrary/monomerLibraryList/MonomerList.tsx
```diff
@@ -78,12 +78,14 @@ const MonomerList = ({
...
+  const [prevActiveTool, setPrevActiveTool] = useState(activeTool);
 
-  useEffect(() => {
+  if (activeTool !== prevActiveTool) {
+    setPrevActiveTool(activeTool);
     if (activeTool !== 'monomer') {
       setSelectedMonomers('');
     }
-  }, [activeTool]);
+  }
```
svvald: It seems that refactoring prospects here are even better – `selectedMonomers` is a dead remnants of what was called `Monomer` and `RnaPreset` tools. Previously we selected them in library and placed on canvas. Now this flow doesn't exist – we use drag-and-drop and auto-chain tools for that. So this is dead code, we can safely remove `selectedMonomers` state and this whole logic related to tool selection and state reset

svvald (changes requested):

packages/ketcher-macromolecules/src/components/monomerLibrary/monomerLibraryList/MonomerList.tsx
nickkurkan: Confirmed — grepped the whole component tree, setSelectedMonomers is only ever called with '' (in the reset effect itself); no code path sets it to a real monomer key anymore. Removed selectedMonomers, the reset effect, and the now-unused activeTool selector instead of refactoring the effect. Pushed in bd9ce78706.

svvald (approved):

PR #11344, opened 2026-08-19
#11118 - Refactor: set-state-in-effect anti-pattern elimination from packages\ketcher-macromolecules\src\components\modal\Settings\Settings.tsx file

### LINKED ISSUE #11118: Refactor: set-state-in-effect anti-pattern elimination from `packages\ketcher-macromolecules\src\components\modal\Settings\Settings.tsx` file
**Problem:** 
1. Duplicate source of truth
2. Extra render cycle:
- Render
- Effect runs
- setState runs
- Second render

**Why is this an issue?**

**How can it be fixed?**

```ts
function Component({data}) {

  useEffect(() => {
  }, [data]);
}

function Component() {

  useEffect(() => {
  }, []);
}

  useEffect(() => {
  }, [rawData]);
}

  useEffect(() => {
  }, [selectedId, items]);
}
```

**Examples of correct code:**
```ts
function Tooltip() {
  const ref = useRef(null);

  useLayoutEffect(() => {
    setTooltipHeight(height);
  }, []);
}

// ✅ Calculate during render
}
```

**Problem locations:**
packages\ketcher-macromolecules\src\components\modal\Settings\Settings.tsx:54

nickkurkan:
## Summary
Closes #11118.

`Settings.tsx` had a `useEffect` that, on `settingsService`/`isModalOpen` changes, both (a) synced `currentSettings`/`initialSettings` from `settingsService.getSettings()` and (b) subscribed to `settingsService` for live updates. Part (a) is a classic set-state-in-effect anti-pattern (extra render, brief stale state before the effect runs); part (b) is a legitimate external subscription that needs `useEffect` with cleanup.

- Replaced the state-sync part with the "adjusting state when a dependency changes" render-time pattern already used elsewhere in this codebase (see #11320, #11113): track `prevSettingsService`/`prevIsModalOpen` in state, and when either differs from the current value, call `setCurrentSettings`/`setInitialSettings` directly during render instead of via an effect.
- Kept the `useEffect` for the `settingsService.subscribe(...)` call only, since that's a real external subscription with cleanup, not derivable state.
- No behavior change intended: settings still populate from the service exactly when the modal opens (or the service becomes available), still re-sync on external settings updates, and Apply/Cancel/Reset semantics are untouched.

## Test plan
- [x] `npx eslint src/components/modal/Settings/Settings.tsx` in `packages/ketcher-macromolecules` — clean
- [x] `npx tsc --noEmit -p tsconfig.json` in `packages/ketcher-macromolecules` — no new errors for this file
- [x] No existing `Settings.test.*` file found for this component, so no automated test run was possible (out of scope to add per task instructions)
- [ ] Manually verify in the running app: open the Settings modal and confirm it still shows the current settings on open, external settings changes (e.g. via ACS style or reset) still reflect in the open modal, and Apply/Cancel/Reset behave exactly as before

### DIFF
--- a/packages/ketcher-macromolecules/src/components/modal/Settings/Settings.tsx
+++ b/packages/ketcher-macromolecules/src/components/modal/Settings/Settings.tsx
@@ -44,27 +44,14 @@ export const Settings = ({ isModalOpen, onClose }: RequiredModalProps) => {
     useState<SettingsFormValue | null>(null);
   const [expandedGroups, setExpandedGroups] = useState<string[]>(['general']);
   const [isLoading, setIsLoading] = useState(false);
-  const [prevSettingsService, setPrevSettingsService] =
-    useState(settingsService);
-  const [prevIsModalOpen, setPrevIsModalOpen] = useState(isModalOpen);
-
-  if (
-    settingsService !== prevSettingsService ||
-    isModalOpen !== prevIsModalOpen
-  ) {
-    setPrevSettingsService(settingsService);
-    setPrevIsModalOpen(isModalOpen);
-
-    if (settingsService && isModalOpen) {
-      const coreSettings = settingsService.getSettings();
-      const formSettings = normalizeSettingsForForm(coreSettings);
-      setCurrentSettings(formSettings);
-      setInitialSettings(formSettings);
-    }
-  }
 
   useEffect(() => {
     if (!settingsService || !isModalOpen) return;
+
+    const coreSettings = settingsService.getSettings();
+    const formSettings = normalizeSettingsForForm(coreSettings);
+    setCurrentSettings(formSettings);
+    setInitialSettings(formSettings);
 
     return settingsService.subscribe((newCoreSettings) => {
       const newFormSettings = normalizeSettingsForForm(newCoreSettings);
--- a/packages/ketcher-macromolecules/src/components/modal/Settings/Settings.tsx
+++ b/packages/ketcher-macromolecules/src/components/modal/Settings/Settings.tsx
@@ -44,14 +44,27 @@ export const Settings = ({ isModalOpen, onClose }: RequiredModalProps) => {
     useState<SettingsFormValue | null>(null);
   const [expandedGroups, setExpandedGroups] = useState<string[]>(['general']);
   const [isLoading, setIsLoading] = useState(false);
+  const [prevSettingsService, setPrevSettingsService] =
+    useState(settingsService);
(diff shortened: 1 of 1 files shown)


packages/ketcher-macromolecules/src/components/modal/Settings/Settings.tsx
```diff
@@ -44,15 +44,28 @@ export const Settings = ({ isModalOpen, onClose }: RequiredModalProps) => {
...
   const [expandedGroups, setExpandedGroups] = useState<string[]>(['general']);
   const [isLoading, setIsLoading] = useState(false);
+  const [prevSettingsService, setPrevSettingsService] =
+    useState(settingsService);
+  const [prevIsModalOpen, setPrevIsModalOpen] = useState(isModalOpen);
+
+  if (
+    settingsService !== prevSettingsService ||
+    isModalOpen !== prevIsModalOpen
+  ) {
```
dorogusha: This potentially breaks the Settings modal on every open.

Settings is fully unmounted/remounted each time the modal opens/closes — ModalContainer returns null when modalName is null, and closeModal() resets state.name = null. So on every open, this is a fresh mount with isModalOpen and settingsService already at their "open" values.

Since prevSettingsService/prevIsModalOpen are initialized via useState(settingsService) / useState(isModalOpen), on the very first render prev === current by construction, so this condition is false on mount and the sync block never runs. currentSettings stays null forever, so if (!currentSettings) return null; (line 165) fires and the modal renders nothing.

The original useEffect didn't have this problem because effects always run once after mount, regardless of whether "dependencies changed." The "adjust state when a dependency changes" pattern only catches changes between renders of an already-mounted component — it's not a drop-in replacement for effects that must also run on mount.

Repro: open Settings — the modal should now render blank/empty. Please verify manually (the PR's test-plan checkbox for this was left unchecked).

dorogusha (changes requested):

packages/ketcher-macromolecules/src/components/modal/Settings/Settings.tsx
nickkurkan: Confirmed — reproduced exactly as you described: prevSettingsService/prevIsModalOpen are seeded from the same values on first render, so the sync block never fires on mount and the modal renders blank (ModalContainer fully unmounts Settings on close, so every open is a fresh mount). Reverted to initializing currentSettings/initialSettings inside the existing subscribe useEffect (which already runs on mount), removed the prev* state entirely. Pushed in 216c2f6928. Thanks for catching this before merge.

PR #11345, opened 2026-08-19
#11117 - Refactor: set-state-in-effect anti-pattern elimination from packages\ketcher-macromolecules\src\components\monomerLibrary\monomerLibraryItem\hooks\useDisabledForSequenceMode.ts file

### LINKED ISSUE #11117: Refactor: set-state-in-effect anti-pattern elimination from `packages\ketcher-macromolecules\src\components\monomerLibrary\monomerLibraryItem\hooks\useDisabledForSequenceMode.ts` file
**Problem:** 
1. Duplicate source of truth
2. Extra render cycle:
- Render
- Effect runs
- setState runs
- Second render

**Why is this an issue?**

**How can it be fixed?**

```ts
function Component({data}) {

  useEffect(() => {
  }, [data]);
}

function Component() {

  useEffect(() => {
  }, []);
}

  useEffect(() => {
  }, [rawData]);
}

  useEffect(() => {
  }, [selectedId, items]);
}
```

**Examples of correct code:**
```ts
function Tooltip() {
  const ref = useRef(null);

  useLayoutEffect(() => {
    setTooltipHeight(height);
  }, []);
}

// ✅ Calculate during render
}
```

**Problem locations:**
packages\ketcher-macromolecules\src\components\monomerLibrary\monomerLibraryItem\hooks\useDisabledForSequenceMode.ts:23

nickkurkan:
## Summary
- Closes #11117
- `useDisabledForSequenceMode` previously synced a purely derived boolean into `useState` via a `useEffect` that recomputed it on every dependency change. The value is 100% derivable from the hook's inputs (redux selectors + the `item`/`groupName` args) on every render, so this collapses into a plain derived value computed directly in the hook body — no `useState`/`useEffect` needed.
- Both selector hooks are still called unconditionally above the first early `return`, so hook order is unchanged.
- This removes the extra render that used to occur before `isDisabled` reflected the correct value (matching the anti-pattern elimination effort already applied elsewhere in this repo, e.g. `MeasureInput.tsx`, `RnaElementsAccordionView.tsx`).
- Also removes two now-dead `eslint-disable` comments (`react-hooks/set-state-in-effect`, `react-you-might-not-need-an-effect/no-event-handler`) that `e16cf3e` (#10806) added to this file on master. The rebase kept them silently; with no effect and no state left they were suppressions for the exact rules this PR exists to satisfy. `eslint` now passes on the file with none.

## Behavior
Same boolean logic per `MonomerGroups` branch. Two differences, both intended:

**1. First-render timing.** Consumers previously saw `false` on the first render in RNA-builder mode and the correct value one render later. They now get the correct value immediately — this is the "temporary incorrect state before effect syncs it" #11117 asks to remove.

**2. The fallthrough** (no matching group) went from silently keeping the previous state to an explicit `return false`.

An earlier version of this description called that branch "unreachable in practice". That was wrong. `groupName` is optional (`monomerLibraryItem/types.ts:20`) and is `undefined` on live paths: `MonomerList.tsx:82`/`:105` (the peptides/CHEM/favorites library), and `RnaElementsTabsView.tsx:164`, where the ambiguous-monomer `<MonomerGroup>` renders inside the RNA builder with no `groupName` — so the branch runs with `isSequenceEditInRNABuilderMode === true`.

There is still no regression, for a different reason: the old code's retained value on that path could only ever be the `useState<boolean>(false)` initial value, since `groupName` is a fixed prop per `MonomerItem` instance and no call site mutates it. So both versions yield `false`. That is now pinned by tests rather than argued in a PR body.

## Test plan
- [x] `eslint` on the hook — clean, with no suppressions
- [x] `tsc --noEmit` in `packages/ketcher-macromolecules` — clean
- [x] `prettier --check` — clean
- [x] Added 2 tests for the fallthrough: `groupName === undefined` and `MonomerGroups.PEPTIDES`, both with `MonomerCaps = {}` so they fail if either matches a restricted branch
- [x] Mutation-checked those tests — flipping the fallthrough to `return true` fails exactly the 2 new cases and none of the 9 existing ones, confirming the branch was previously uncovered
- [x] Full package unit suite: 36 suites, 160 passed, 2 pre-existing skips
- [x] Reviewed the only call site (`MonomerItem.tsx:66`) — usage is `useDisabledForSequenceMode(item, groupName) || disabled`, read inline

### DIFF
--- a/packages/ketcher-macromolecules/src/components/monomerLibrary/monomerLibraryItem/hooks/useDisabledForSequenceMode.ts
+++ b/packages/ketcher-macromolecules/src/components/monomerLibrary/monomerLibraryItem/hooks/useDisabledForSequenceMode.ts
@@ -2,7 +2,6 @@
 /* eslint-disable react-hooks/set-state-in-effect */
 import { MonomerGroups, MonomerItemType } from 'ketcher-core';
 import { useSelector } from 'react-redux';
-import { useEffect, useState } from 'react';
 import { selectIsSequenceFirstsOnlyNucleotidesSelected } from 'state/rna-builder';
 import { useAppSelector } from 'hooks';
 import { selectIsSequenceEditInRNABuilderMode } from 'state/common';
@@ -17,43 +16,26 @@ const useDisabledForSequenceMode = (
   const isSequenceFirstsOnlyNucleoelementsSelected = useSelector(
     selectIsSequenceFirstsOnlyNucleotidesSelected,
   );
-  const [isDisabled, setIsDisabled] = useState<boolean>(false);
 
-  useEffect(() => {
-    if (!isSequenceEditInRNABuilderMode) return setIsDisabled(false);
+  if (!isSequenceEditInRNABuilderMode) return false;
 
-    if (groupName === MonomerGroups.BASES) {
-      setIsDisabled(!item?.props?.MonomerCaps?.R1);
-    } else if (groupName === MonomerGroups.PHOSPHATES) {
-      setIsDisabled(
-        !(item?.props?.MonomerCaps?.R1 && item?.props?.MonomerCaps?.R2),
+  if (groupName === MonomerGroups.BASES) {
+    return !item?.props?.MonomerCaps?.R1;
+  } else if (groupName === MonomerGroups.PHOSPHATES) {
+    return !(item?.props?.MonomerCaps?.R1 && item?.props?.MonomerCaps?.R2);
+  } else if (groupName === MonomerGroups.SUGARS) {
+    if (isSequenceFirstsOnlyNucleoelementsSelected) {
+      return !(item?.props?.MonomerCaps?.R3 && item?.props?.MonomerCaps?.R2);
+    } else {
+      return !(
+        item?.props?.MonomerCaps?.R3 &&
+        item?.props?.MonomerCaps?.R2 &&
+        item?.props?.MonomerCaps?.R1
       );
-    } else if (groupName === MonomerGroups.SUGARS) {
-      if (isSequenceFirstsOnlyNucleoelementsSelected) {
-        setIsDisabled(
-          !(item?.props?.MonomerCaps?.R3 && item?.props?.MonomerCaps?.R2),
-        );
-      } else {
-        setIsDisabled(
-          !(
-            item?.props?.MonomerCaps?.R3 &&
-            item?.props?.MonomerCaps?.R2 &&
-            item?.props?.MonomerCaps?.R1
-          ),
-        );
-      }
     }
-  }, [
-    groupName,
-    isSequenceEditInRNABuilderMode,
-    isSequenceFirstsOnlyNucleoelementsSelected,
-    item?.props?.MonomerCaps?.R1,
-    item?.props?.MonomerCaps?.R2,
-    item?.props?.MonomerCaps?.R3,
-    setIsDisabled,
-  ]);
+  }
 
-  return isDisabled;
+  return false;
 };
 
 export default useDisabledForSequenceMode;
--- a/packages/ketcher-macromolecules/src/components/monomerLibrary/monomerLibraryItem/hooks/useDisabledForSequenceMode.ts
+++ b/packages/ketcher-macromolecules/src/components/monomerLibrary/monomerLibraryItem/hooks/useDisabledForSequenceMode.ts
@@ -1,6 +1,5 @@
 import { MonomerGroups, MonomerItemType } from 'ketcher-core';
 import { useSelector } from 'react-redux';
-import { useEffect, useState } from 'react';
 import { selectIsSequenceFirstsOnlyNucleotidesSelected } from 'state/rna-builder';
 import { useAppSelector } from 'hooks';
 import { selectIsSequenceEditInRNABuilderMode } from 'state/common';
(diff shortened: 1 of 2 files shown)


packages/ketcher-macromolecules/src/components/monomerLibrary/monomerLibraryItem/hooks/useDisabledForSequenceMode.ts
```diff
@@ -15,43 +14,26 @@ const useDisabledForSequenceMode = (
   const isSequenceFirstsOnlyNucleoelementsSelected = useSelector(
     selectIsSequenceFirstsOnlyNucleotidesSelected,
   );
-  const [isDisabled, setIsDisabled] = useState<boolean>(false);
 
-  useEffect(() => {
-    if (!isSequenceEditInRNABuilderMode) return setIsDisabled(false);
+  if (!isSequenceEditInRNABuilderMode) return false;
```
la-nuta: One behavior change worth naming rather than calling this a pure no-op: previously the hook returned `false` for one commit on mount before the effect corrected it. `MonomerItem` uses that to gate `onClick` and the favorite-star/autochain controls, so a monomer that should be disabled was briefly clickable. This PR closes that window — good, just not a no-op.

packages/ketcher-macromolecules/src/components/monomerLibrary/monomerLibraryItem/hooks/useDisabledForSequenceMode.ts
```diff
@@ -15,43 +14,26 @@ const useDisabledForSequenceMode = (
...
-    isSequenceFirstsOnlyNucleoelementsSelected,
-    item?.props?.MonomerCaps?.R1,
-    item?.props?.MonomerCaps?.R2,
-    item?.props?.MonomerCaps?.R3,
-    setIsDisabled,
-  ]);
+  }
 
-  return isDisabled;
+  return false;
```
la-nuta: This turns the old implicit "keep previous state" fallthrough into an explicit `false`. Unreachable in practice (`MonomerItem` is keyed per monomer, so `groupName` never changes for an instance) and it matches the old initial value — but it's the one contract this PR redefines and it's untested. Worth a one-line comment plus a test for `groupName === undefined`.

packages/ketcher-macromolecules/src/components/monomerLibrary/monomerLibraryItem/hooks/useDisabledForSequenceMode.ts
la-nuta: The old code returned `isDisabled === false` for one commit on mount, then the effect corrected it. `MonomerItem` consumes that value to spread the click handler (`MonomerItem.tsx:153`, `{...(!isDisabled ? { onClick } : {})}`) and to gate the favourite-star and autochain controls (`:173`) — so a monomer that should be disabled was briefly clickable, and a click landing in that first commit would fire `selectMonomer`.

This PR closes that window, which is the outcome we want. Please just reword the description so it isn't described as behaviour-neutral.

packages/ketcher-macromolecules/src/components/monomerLibrary/monomerLibraryItem/hooks/useDisabledForSequenceMode.ts
la-nuta: This changes the fallthrough from "keep the previous state" to "always `false`", and nothing covers it.

It's unreachable today — `MonomerItem` is keyed by `getMonomerUniqueKey(monomer)` (`MonomerGroup.tsx:131`) under a `MonomerGroup` keyed by `groupTitle` (`RnaElementsAccordionView.tsx:158`), so `groupName` is fixed for the lifetime of any instance — and it matches the old `useState<boolean>(false)` initial value, so mount behaviour is unchanged. But that reasoning currently lives nowhere.

Please add a one-line comment recording it, plus a case in `useDisabledForSequenceMode.test.tsx` for `groupName === undefined` (or a non-RNA group). The existing 10 tests pass identically before and after this PR, so nothing pins the new contract.

la-nuta, line 36: Could we get a test for this branch? It's the one contract the PR redefines — the fallthrough goes from "keep the previous state" to "always `false`" — and the existing 9 cases never exercise `groupName === undefined` or a non-RNA group, so nothing pins it.

Your reasoning in the description is right (unreachable in practice, and it matches the old `useState<boolean>(false)` initial value), so this is just about making that durable in code rather than in the PR body.

la-nuta (commented):
This kills **two** suppressions rather than one, which is worth collecting while you're here.

Master carries these on lines 1-2 of this file (added by #10806 after your branch was cut, so they're not in the diff — but they come along on merge):

```
/* eslint-disable react-you-might-not-need-an-effect/no-event-handler */
```

Your change makes both dead. Linting the merged shape reports:

```
1:1  warning  Unused eslint-disable directive (no problems were reported from 'react-you-might-not-need-an-effect/no-event-handler')
2:1  warning  Unused eslint-disable directive (no problems were reported from 'react-hooks/set-state-in-effect')
```

(Both were genuinely firing before — master's body with the headers stripped errors at `21:49` and `29:16`.) So once you've picked up master — merge or rebase, whichever you prefer — deleting both lines would leave the file with no suppressions at all.

One small ask inline; otherwise this looks good, and the description does a nice job of naming the removed extra render rather than glossing it as a pure no-op.

nickkurkan (commented):
Reviewed the diff and the surrounding code. The refactor itself is correct — the two selector hooks (`useAppSelector`, `useSelector`) are still called unconditionally *above* the first early return, so there's no Rules-of-Hooks problem, which is the usual way this kind of state→derived-value conversion goes wrong. Logic is a faithful 1:1 translation of the old branches.

Three things:

**1. Stale eslint-disable comments after rebase (blocking-ish).**
This branch is based on a commit that predates e16cf3e (#10806, "raise ESLint to latest compatible versions"), which added these two lines to the top of the file on `master`:

```js
/* eslint-disable react-you-might-not-need-an-effect/no-event-handler */
```

The PR shows as MERGEABLE, so git will happily keep both lines and take your body — leaving a file with zero effects and zero state carrying two suppressions for exactly the rules this PR exists to satisfy. Please rebase on `master` and delete them; otherwise the next person greps for `set-state-in-effect` suppressions and finds a false positive.

**2. +1 to @la-nuta on the fallthrough test.**
Worth adding, and I'd suggest covering both shapes: `groupName === undefined` and an explicit non-RNA group (`MonomerGroups.PEPTIDES` / `NUCLEOTIDES`), both asserting `false` while `isSequenceEditInRNABuilderMode` is `true`. The old code's fallthrough was genuinely stale-state-dependent (a `MonomerItem` whose `groupName` changed `BASES` → `PEPTIDES` would keep `isDisabled === true`), so the new `return false` is a real fix, not just a tidy-up — that's exactly the kind of thing that deserves a pinned test.

Heads up when you add them: the test file mutates a single shared `monomer` object and never resets `mockUseSelector` between cases, so ordering matters. Set both `mockUseAppSelector` / `monomer.props.MonomerCaps` explicitly in the new cases rather than inheriting.

**3. Worth calling out in the description: the first-render behavior does change.**
Previously `MonomerItem` rendered one frame with `isDisabled === false` before the effect corrected it — for a disabled monomer that frame was clickable and draggable (`useLibraryItemDrag`). Now it's correct on the first render. That's the improvement, but it's an observable change, so if any Playwright test asserts on the pre-effect frame it'd be worth a CI check.

Optional nit: every branch now returns, so the `else if` chain plus the trailing `return false` could flatten into plain `if`s or a `switch (groupName)`. Non-blocking, take it or leave it.

packages/ketcher-macromolecules/src/components/monomerLibrary/monomerLibraryItem/hooks/useDisabledForSequenceMode.ts
nickkurkan: Done in 71d6889 — two cases added: `groupName === undefined` and `MonomerGroups.PEPTIDES`, both with `MonomerCaps = {}` so they'd fail if either matched a restricted branch rather than falling through.

To check they aren't vacuous, I mutated the fallthrough to `return true` and reran: exactly those 2 failed, all 9 originals passed. So your read was right on the nose — nothing in the existing suite touched this branch.

One correction to my own earlier comment, and to the reasoning in the PR description you were kind enough to endorse: **the branch isn't unreachable.** `groupName` is optional (`monomerLibraryItem/types.ts:20`) and is genuinely `undefined` on live paths — `MonomerList.tsx:82`/`:105` (the whole peptides/CHEM/favorites library), and more to the point `RnaElementsTabsView.tsx:164`, where the ambiguous-monomer `<MonomerGroup>` is rendered *inside the RNA builder* with no `groupName`. So the fallthrough runs with `isSequenceEditInRNABuilderMode === true`, which is exactly the state this hook exists for.

The conclusion still holds — old and new both yield `false` there, so no regression — but for a different reason than "never reached". It's "reached routinely, and already `false`". Which makes your ask for a pinned test better-founded than either of us put it. I've updated the PR description to say that instead.

Also folded in while rebasing: this branch was based before e16cf3e (#10806), which added `set-state-in-effect` and `no-event-handler` suppressions to this file on master. The rebase merged clean and kept both — leaving dead suppressions for the exact rules this PR exists to satisfy. Removed; `eslint` now passes on the file with none.

Full package suite is green locally (36 suites, 160 passed), which also closes the box I'd left unchecked in the description.

rrodionov91 (approved):

AlexeyGirin:

Issue #11236, opened 2026-08-14
Refactor: `react-you-might-not-need-an-effect/no-chain-state-updates` rule violation clean up from `packages\ketcher-macromolecules\src\components\monomerLibrary\RnaBuilder\RnaEditor\RnaEditorExpanded\RnaEditorExpanded.tsx` file


AlexeyGirin:
**What problem does `react-you-might-not-need-an-effect/no-chain-state-updates` detect?**
It detects chains of state updates where one state update causes a render, and an effect then performs another state update based on the new state or derived data.

Typical example:
```ts
const [firstName, setFirstName] = useState('');
const [lastName, setLastName] = useState('');
const [fullName, setFullName] = useState('');

useEffect(() => {
  setFullName(`${firstName} ${lastName}`);
}, [firstName, lastName]);
```

The rule also targets multiple related state updates that are unnecessarily coordinated through effects.

**Why is this a problem?**
The extra state update creates an avoidable render cycle:
- An input state changes.
- React renders with outdated derived state.
- The effect runs after rendering.
- The effect updates another state variable.
- React renders again with the correct value.

This can cause:
- unnecessary renders;
- briefly inconsistent UI;
- more complicated data flow;
- synchronization bugs;
- additional effect dependency problems;
- harder-to-understand component behavior.

React recommends deriving values during render when they can be calculated from existing props or state. Effects are intended mainly for synchronizing with external systems, not for keeping related internal state variables synchronized.

**How is it usually fixed?**
Remove the redundant state and calculate the derived value directly:
```
const [firstName, setFirstName] = useState('');
const [lastName, setLastName] = useState('');

const fullName = `${firstName} ${lastName}`;
```

For expensive calculations, use useMemo when there is a measured need:
```
const filteredItems = useMemo(
  () => expensiveFilter(items, query),
  [items, query],
);
```

If several values must change as one logical operation, update them together in the same event handler:
```
const handleReset = () => {
  setFirstName('');
  setLastName('');
};
```

For complex state transitions, use a reducer:
```
const [state, dispatch] = useReducer(reducer, initialState);
```

The general principle is: avoid using an effect to calculate or synchronize state that can be derived during render or updated directly at the point where the user action occurs.

**Problem locations:**
packages\ketcher-macromolecules\src\components\monomerLibrary\RnaBuilder\RnaEditor\RnaEditorExpanded\RnaEditorExpanded.tsx

PR #11346, opened 2026-08-19
#11114 - Refactor: set-state-in-effect anti-pattern elimination from packages\ketcher-macromolecules\src\components\monomerLibrary\RnaBuilder\RnaEditor\RnaEditorExpanded\RnaEditorExpanded.tsx file

### LINKED ISSUE #11114: Refactor: set-state-in-effect anti-pattern elimination from `packages\ketcher-macromolecules\src\components\monomerLibrary\RnaBuilder\RnaEditor\RnaEditorExpanded\RnaEditorExpanded.tsx` file
**Problem:** 
1. Duplicate source of truth
2. Extra render cycle:
- Render
- Effect runs
- setState runs
- Second render

**Why is this an issue?**

**How can it be fixed?**

```ts
function Component({data}) {

  useEffect(() => {
  }, [data]);
}

function Component() {

  useEffect(() => {
  }, []);
}

  useEffect(() => {
  }, [rawData]);
}

  useEffect(() => {
  }, [selectedId, items]);
}
```

**Examples of correct code:**
```ts
function Tooltip() {
  const ref = useRef(null);

  useLayoutEffect(() => {
    setTooltipHeight(height);
  }, []);
}

// ✅ Calculate during render
}
```

**Problem locations:**
packages\ketcher-macromolecules\src\components\monomerLibrary\RnaBuilder\RnaEditor\RnaEditorExpanded\RnaEditorExpanded.tsx:227
packages\ketcher-macromolecules\src\components\monomerLibrary\RnaBuilder\RnaEditor\RnaEditorExpanded\RnaEditorExpanded.tsx:245
packages\ketcher-macromolecules\src\components\monomerLibrary\RnaBuilder\RnaEditor\RnaEditorExpanded\RnaEditorExpanded.tsx:279

nickkurkan:
## Summary
Closes #11114.

`RnaEditorExpanded.tsx` had three `useEffect` blocks that called `setState` purely to sync/derive local state, matching the anti-pattern flagged in this series of refactors:

- **`newPreset` / `selectedPhosphatePosition` sync from `activePreset`**: converted to the render-time "track previous prop value, setState directly in render body when it differs" pattern (same idiom used in `MeasureInput` / `RnaElementsAccordionView`). No more effect needed for this sync.
- **`sequenceSelectionGroupNames` derived from `sequenceSelection`**: the derivation itself moved to the same render-time pattern. The `dispatch(setSequenceSelectionName(...))` call (a real side effect on the redux store) stays in a `useEffect`, now stripped of the state-derivation code it previously carried.
- **`newPreset` update when a monomer is picked in the library while editing a preset**: this was one branch of a combined effect. That branch (the else branch, no dispatch involved) is now computed directly during render using the same dependency-change tracking. The other branch — applying a library selection onto sequence nodes, which dispatches `setSequenceSelection` and is a genuine side effect — stays in a `useEffect`; only the `setNewPreset` derived-state part was extracted out of it.

No behavior changes are intended — same trigger conditions, same computed values, just less indirection through an extra render/effect cycle.

## Test plan
- [x] `cd packages/ketcher-macromolecules && npx eslint src/components/monomerLibrary/RnaBuilder/RnaEditor/RnaEditorExpanded/RnaEditorExpanded.tsx` — down from 7 warnings (0 errors) to 3 warnings, all pre-existing/out of scope for this issue: two style advisories on the remaining side-effect `useEffect` (`no-event-handler`, `no-chain-state-updates`, both already present before this change, on the effect that dispatches `setSequenceSelection`) and one unrelated pre-existing missing-deps warning on the keyDown-handling effect further down the file (untouched by this change). No new warnings/errors introduced.
- [x] `npx tsc --noEmit -p tsconfig.json` in `packages/ketcher-macromolecules` — no errors reported for this file.
- [ ] `RnaEditorExpanded.test.tsx` exists but its suite currently fails to run in this environment with `TypeError: Cannot read properties of undefined (reading 'html')` from `jest-environment-jsdom` — confirmed this is a pre-existing environment issue unrelated to this change (reproduces identically on `origin/master` with no changes applied).
- [ ] Manually verify in the running app that RNA preset editing (name/phosphate position sync, monomer group selection) and sequence selection (Modify in RNA Builder mode) still update correctly.


packages/ketcher-macromolecules/src/components/monomerLibrary/RnaBuilder/RnaEditor/RnaEditorExpanded/RnaEditorExpanded.tsx
```diff
@@ -221,80 +233,116 @@ export const RnaEditorExpanded = ({
...
-        ? getRnaPresetPhosphatePosition(activePreset)
-        : undefined,
-    );
-  }, [activePreset]);
+  const [
+    prevSequenceSelectionForGroupNames,
+    setPrevSequenceSelectionForGroupNames,
+  ] = useState(sequenceSelection);
+
+  if (sequenceSelection !== prevSequenceSelectionForGroupNames) {
```
la-nuta: Splitting this changes commit ordering: the group names now apply during render while `setSequenceSelectionName` still dispatches post-commit, where both used to land in one batched re-render. Result is one frame in "Modify in RNA Builder" where the group slots show the new selection but the name input still shows the old one. Cosmetic, but an observable change in a PR that claims none.

packages/ketcher-macromolecules/src/components/monomerLibrary/RnaBuilder/RnaEditor/RnaEditorExpanded/RnaEditorExpanded.tsx
```diff
@@ -221,80 +233,116 @@ export const RnaEditorExpanded = ({
...
+  );
+  const [prevIsSequenceEditForPresetSync, setPrevIsSequenceEditForPresetSync] =
+    useState(isSequenceEditInRNABuilderMode);
+  const [
+    prevSelectedPhosphatePositionForPresetSync,
+    setPrevSelectedPhosphatePositionForPresetSync,
+  ] = useState(selectedPhosphatePosition);
+
+  if (
+    activePresetMonomerGroup?.groupItem !== prevGroupItemForPresetSync ||
```
la-nuta: **Blocker.** The `else` branch used to run on **mount** and no longer does — all three trackers are seeded with the current values, so this block is a no-op on first render. That's dropped behavior, not a timing change, and it's reachable: `Tabs.tsx` renders only the selected tab panel, so switching library tabs unmounts this component, as does collapsing the header. Neither path resets `activePresetMonomerGroup`, `activeRnaBuilderItem` or `isEditMode`.

Repro: Edit a preset → click the Sugars slot → pick sugar `5A6` → switch to Peptides and back. On master the mount run re-applies `5A6` and recomputes the name. Now nothing runs: the name reverts, Save can flip back to disabled, and `onSave` persists the preset **without the sugar the user picked** — while the slot still displays it, so the UI contradicts what's saved.

Fix: seed the trackers with sentinels that can't equal the first real values (or a `hasRun` ref) so the block still fires once on mount. Worth an E2E for tab-switch-during-edit — no existing RNA-Builder spec covers a remount mid-edit.

packages/ketcher-macromolecules/src/components/monomerLibrary/RnaBuilder/RnaEditor/RnaEditorExpanded/RnaEditorExpanded.tsx
```diff
@@ -221,80 +233,116 @@ export const RnaEditorExpanded = ({
...
-              activePresetMonomerGroup.groupName === 'Bases'
-                ? activePresetMonomerGroup.groupItem
-                : node.rnaBaseMonomerItem,
-          };
+    if (
+      activeMonomerGroup !== RnaBuilderPresetsItem.Presets &&
+      isEditMode &&
+      !(isSequenceEditInRNABuilderMode && activePresetMonomerGroup)
+    ) {
+      const currentPreset = updatePresetMonomerGroup();
```
la-nuta: Design note: this isn't the sanctioned "adjust state when a prop changes" case (the `activePreset` block is) — it's an event response ("user picked a monomer") pushed into render. It calls `updatePresetMonomerGroup()`, which itself calls `setNewPreset`, then sets it again. The rule goes quiet but the design is unchanged, and a helper that mutates state from render is harder to follow.

Structurally: apply the group onto the draft preset in the handler that dispatches `setActivePresetMonomerGroup` (`RnaElements.tsx:94`). If that's out of scope, please leave a comment so this doesn't get re-filed.

packages/ketcher-macromolecules/src/components/monomerLibrary/RnaBuilder/RnaEditor/RnaEditorExpanded/RnaEditorExpanded.tsx
```diff
@@ -221,80 +233,116 @@ export const RnaEditorExpanded = ({
...
+        rnaBaseMonomerItem:
+          activePresetMonomerGroup.groupName === 'Bases'
+            ? activePresetMonomerGroup.groupItem
+            : node.rnaBaseMonomerItem,
+      };
+    });
+
+    setIsSequenceSelectionUpdated(true);
+    dispatch(setSequenceSelection(updatedSequenceSelection));
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
la-nuta: This file has no eslint disables on the PR base, and #11327 exists specifically to fix `exhaustive-deps` here — adding a disable on this exact hook silences the warning that PR is meant to resolve. Either fix the deps or leave it visible for #11327.

packages/ketcher-macromolecules/src/components/monomerLibrary/RnaBuilder/RnaEditor/RnaEditorExpanded/RnaEditorExpanded.tsx
la-nuta: I think this loses the mount run. The three trackers above are seeded with the *current* values, so on first render all three comparisons are `x !== x` → false and the block is skipped — whereas the old effect ran on mount as well as on changes.

That's reachable, because this component unmounts routinely: `Tabs.tsx` renders only the selected tab panel, and `RnaEditor.tsx` unmounts it when the header collapses. Neither resets `activePresetMonomerGroup` / `activeRnaBuilderItem` / `isEditMode`.

Repro: edit a preset → click the Sugars slot → pick a sugar → switch to Peptides and back. Before, the mount run re-applied the sugar and recomputed the name. Now nothing runs, so the name reverts and Save would persist the preset without the monomer — while the slot still shows it.

Seeding the trackers with sentinels that can't equal the first values (or a `hasRun` ref) should restore it.

packages/ketcher-macromolecules/src/components/monomerLibrary/RnaBuilder/RnaEditor/RnaEditorExpanded/RnaEditorExpanded.tsx
la-nuta: This disable can just go — it isn't doing anything. Linting the merged shape reports:

```
348:5  warning  Unused eslint-disable directive (no problems were reported from 'react-hooks/exhaustive-deps')
```

The deps below already satisfy the rule.

packages/ketcher-macromolecules/src/components/monomerLibrary/RnaBuilder/RnaEditor/RnaEditorExpanded/RnaEditorExpanded.tsx
la-nuta, line 280: I think this loses the mount run. The three trackers are seeded with the *current* values, so on first render every comparison is `x !== x` and the block is skipped — where the old effect ran on mount too.

That's reachable because the component unmounts routinely: `Tabs.tsx` renders only the selected tab panel, and `RnaEditor.tsx` unmounts it when the header collapses. Neither resets `activePresetMonomerGroup` / `activeRnaBuilderItem` / `isEditMode`.

Repro: edit a preset → click the Sugars slot → pick a sugar → switch to Peptides and back. Before, the mount run re-applied the sugar and recomputed the name; now nothing runs, so the name reverts and Save would persist the preset without the monomer — while the slot still shows it.

A first-run flag restores it, kept as state rather than a ref so it stays in the same render-phase pattern you're already using:

```ts
const [hasRunPresetSync, setHasRunPresetSync] = useState(false);

if (
  !hasRunPresetSync ||
  activePresetMonomerGroup?.groupItem !== prevGroupItemForPresetSync ||
  // …
) {
  setHasRunPresetSync(true);
  // …
}
```

Worth noting a `useRef` flag does *not* work here — `hasRun.current = true` during render trips `react-hooks/refs` (`Cannot access refs during render`), which is `error` for this package. I checked both variants against the config; the state version lints clean.

packages/ketcher-macromolecules/src/components/monomerLibrary/RnaBuilder/RnaEditor/RnaEditorExpanded/RnaEditorExpanded.tsx
la-nuta: This disable can go — it isn't doing anything. Linting the merged shape reports `348:5 warning Unused eslint-disable directive (no problems were reported from 'react-hooks/exhaustive-deps')`. The deps below already satisfy the rule.

la-nuta (commented):
Two inline notes, plus a coordination heads-up.

**Coordination:** #11327 rewrites the same two regions of this file for `exhaustive-deps` — it deletes `updatePresetMonomerGroup` (which you keep and call) and widens the deps you deliberately keep narrow. It's gone `CONFLICTING` since master picked up #10409 and #10760, so it needs a rebase regardless; worth agreeing an order with @LashaDevLasha so whichever lands second isn't left broken.

Also, minor: the 24 Aug ESLint upgrade adds a `prettier` complaint around `position ?? selectedPhosphatePosition ?? 'right'` that you'll pick up on the next merge from master.

packages/ketcher-macromolecules/src/components/monomerLibrary/RnaBuilder/RnaEditor/RnaEditorExpanded/RnaEditorExpanded.tsx
nickkurkan: Removed the inline react-hooks/exhaustive-deps suppression in a65d78203e. The remaining file-level directives are inherited from current master and still cover the other intentional effects in this component.

packages/ketcher-macromolecules/src/components/monomerLibrary/RnaBuilder/RnaEditor/RnaEditorExpanded/RnaEditorExpanded.tsx
nickkurkan: Fixed in a65d78203e. Added the suggested hasRunPresetSync state guard so preset synchronization runs once on mount/remount even when the tracked values are initialized to their current values. Added a regression test covering an active sugar selection being applied to the preset name on initial render.

la-nuta (commented):
Both points are fixed — thanks. The `hasRunPresetSync` state guard is exactly the shape that lints clean (a `useRef` flag would have tripped `react-hooks/refs`), and the initial-render regression test is more than I asked for.

**Heads-up before you do any more work on this: master moved again yesterday and it lands in the same code.** `cb882e6b` (#11675, "Default preset name changes unexpectedly after switching phosphate picker from 3' to 5'") rewrote the effect you're restructuring:

- dropped `selectedPhosphatePosition` from that effect's dep array
- simplified the name computation to `selectPresetFullName(currentPreset)`, removing the `buildRnaPresetConnections` / `resolvedPhosphatePosition` wrapping
- removed `no-chain-state-updates` from the file header

That's why this went back to `CONFLICTING` right after your merge. It also makes the resolution a judgement call rather than a textual one — your tracker block still tracks `selectedPhosphatePosition`, which master deliberately stopped doing. Worth deciding what that should mean for the mount-sync guard before picking the conflict apart, whenever you next pick up master.

Same applies to #11327, which is now `CONFLICTING` for the same reason — so neither of you is behind the other, you're both behind #11675.

One observation on scope, no action needed now: with all three file headers stripped, this branch still reports six violations, including a `set-state-in-effect` at 345 and three `exhaustive-deps`. So this PR isn't yet in a position to delete any of the headers. #11327 does clear two of the three — which is another reason the order we discussed (that one first, this one rebasing onto it with a narrower target) still looks like the cheaper path, if you and @LashaDevLasha still agree once you've both seen #11675.

la-nuta (commented):
Needs a rebase before this can be reviewed — and the rebase isn't mechanical.

**1. Master rewrote this same effect region.** `aa7556bcc3` (#11236, PR #11377) landed 4 days ago and takes a different approach to overlapping logic: it derives `isSequenceSelectionUpdated` during render via `lastAppliedPresetMonomerGroupItem` / `wasEditMode`, where this branch uses an `appliedPresetSync` state object plus `activeMonomerGroupRef` / `isEditModeRef`. A 3-way merge leaves two conflict regions and neither side is obviously right — it needs a decision, not a resolution. The `hasRunPresetSync` guard added in `a65d7820` may be redundant once master's render-phase reset is in.

While you're in there: master's trackers are seeded from current values (`useState(isEditMode)`), so `if (isEditMode !== wasEditMode)` is false on the first render and that block doesn't run on mount — where the old effect did. This component remounts on tab switches, so "first render" recurs. Worth deciding deliberately which behaviour you want rather than inheriting it.

**2. Only line 3 is this ticket's.** The branch deletes all three file-level headers, but `exhaustive-deps` (line 1) belongs to #11287 — @LashaDevLasha's #11327, still open against this same file — and `no-event-handler` (line 2) to its own issue. On current master all three are load-bearing; stripping them reports:

```
258:5, 276:5, 312:31  error  Calling setState synchronously within an effect can trigger cascading renders
282:65, 315:14        error  react-you-might-not-need-an-effect/no-event-handler
322:6, 657:6, 667:6   error  react-hooks/exhaustive-deps
```

Your head lints clean with no headers, but that's measured against the old base. After picking up master, re-check each one and keep the two that aren't yours.

**3. Landing order.** #11327 targets this file too. Whichever of you lands second inherits the conflict, so worth agreeing an order between you.

AlexeyGirin:

la-nuta (commented):
Re-checked after `b69947c3` — the headers are still gone and the file lints clean on current master, so my earlier "keep the two that aren't yours" no longer applies.

[PR 11327](https://github.com/epam/ketcher/pull/11327) is 27 commits behind and predates [11236](https://github.com/epam/ketcher/issues/11236), so it can't be usefully rebased — this PR is the one that lands. The third header is [11255](https://github.com/epam/ketcher/issues/11255) (PR [11409](https://github.com/epam/ketcher/pull/11409) closed 8 Sep, issue still open).

Suggest closing [11287](https://github.com/epam/ketcher/issues/11287) and [11255](https://github.com/epam/ketcher/issues/11255) as delivered here.

la-nuta (approved):

PR #11350, opened 2026-08-19
#11272 - Refactor: react-hooks/exhaustive-deps rule violation clean up from packages\ketcher-react\src\hooks\useClickOutside.ts file

### LINKED ISSUE #11272: Refactor: `react-hooks/exhaustive-deps` rule violation clean up from `packages\ketcher-react\src\hooks\useClickOutside.ts` file

Example:
```ts

useEffect(() => {
  loadUser(userId);
}, []); // userId is missing
```

**Why is this a problem?**

This can cause:
- stale state or props;
- effects using old data;

**How is it usually fixed?**
```ts
useEffect(() => {
  loadUser(userId);
}, [userId]);
```

For a memoized callback:
```ts
}, [documentId, content]);
```

For a memoized value:
```ts
  [items, category],
);
```

**Problem locations:**
packages\ketcher-react\src\hooks\useClickOutside.ts

skorphil:
## Summary
- Increase robustness of useClickOutside by following the existing project pattern used in useInterval/useRequestAnimationFrame: keep the latest callback in a ref synchronized via a `[callback]` sync effect to avoid stale closures and listener churn.
- Add an SSR guard at the top of the effect to skip attaching DOM listeners when `document` is not available.

## Details
- Listener effect depends only on `[targetRef]`; `targetRef` is expected to be stable.
- Tests added for the hook (outside/inside click, updated callback, no re-attach on callback change, unmount cleanup, null-ref). Adjustments made to satisfy Testing Library and ESLint rules in the new tests.

## Note on convention
- Intention to propose the “latest-callback via sync ref” as a Memory Bank convention for hooks that accept external callbacks (already used in ketcher-react: useInterval, useRequestAnimationFrame). 

## Limitations
- This PR is scoped strictly to the useClickOutside hook and its tests. Broader type/unit issues in the package are out-of-scope and will be handled separately.

Closes #11272

---
<details><summary>Manual testing report</summary>
<p>

Locate the Target Component
1. In the Ketcher toolbar (left side), click the **Text Tool** (icon with the letter **'T'**).
2. Click anywhere on the drawing canvas to open the **Text Editor** modal.
3. Locate the **Font Size** control (the number, e.g., "13", next to the "Font Size" label).

## Manual Testing Checklist

- [x] **Open Menu**: Click the Font Size number.
  - *Expected*: Dropdown with font sizes (4-144) appears.
- [x] **Click Outside**: Click on the "Text:" label or any empty space in the modal.
  - *Expected*: The Font Size dropdown closes.
- [x] **Apply Size**: Open dropdown and click on a size (e.g., "24").
  - *Expected*: The dropdown closes; the size "24" is applied to selected text or new text.
- [x] **Toggle Button**: Click the Font Size number when the menu is open.
  - *Expected*: The dropdown closes (toggles off).
- [x] **Overlay Click**: Open dropdown and click the dark overlay outside the modal.
  - *Expected*: The dropdown (and likely the modal) closes.
  - *Philipp S comment*: Only dropdown closes
- [x] **Rapid Toggling**: Click the Font Size number quickly multiple times.
  - *Expected*: Menu state remains synced, no flickering or getting stuck.
- [x] **Selection Persistence**: Select text in the editor, open the dropdown, and click outside.
  - *Expected*: Dropdown closes; text selection remains active.

</p>
</details>


skorphil:
I currently have issues with accessing Sonar (chatting with support)
Alexey G told me `You can ignore sonar problems so far - it is not mandaroty for merging yet`

packages/ketcher-react/src/hooks/useClickOutside.ts
```diff
@@ -1,20 +1,38 @@
...
+  const callbackRef = useRef(callback);
 
+  // Keep the latest callback to avoid re-subscribing when only callback changes
   useEffect(() => {
+    callbackRef.current = callback;
+  }, [callback]);
+
+  useEffect(() => {
+    // SSR guard: skip attaching listeners when document is not available
+    if (typeof document === 'undefined') return;
```
svvald: It's unnecessary as effects do not run on server side anyway

svvald (changes requested):

packages/ketcher-react/src/hooks/useClickOutside.ts
skorphil: Ah, i see

svvald (approved):

PR #11356, opened 2026-08-19
#11252 - Refactor: react-you-might-not-need-an-effect/no-event-handler rule violation clean up from packages\ketcher-macromolecules\src\components\ZoomControls\ZoomInput.tsx file

### LINKED ISSUE #11252: Refactor: `react-you-might-not-need-an-effect/no-event-handler` rule violation clean up from `packages\ketcher-macromolecules\src\components\ZoomControls\ZoomInput.tsx` file

Typical example:
```ts

useEffect(() => {
  if (submitted) {
    sendForm();
  }
}, [submitted]);

const handleSubmit = () => {
  setSubmitted(true);
};
```

**Why is this a problem?**
- React renders again.

- an extra render;
- delayed execution;
- dependency-array problems;

**How is it usually fixed?**
```ts
const handleSubmit = () => {
  sendForm();
  setSubmitted(true);
};
```

```ts
const submitForm = () => {
  sendForm();
  setSubmitted(true);
};

```

**Problem locations:**
packages\ketcher-macromolecules\src\components\ZoomControls\ZoomInput.tsx

Shushanik01:
Fixed the react-you-might-not-need-an-effect/no-event-handler violation in ZoomInput.tsx (/packages/ketcher-macromolecules/src/components/ZoomControls/ZoomInput.tsx):

Root cause: ZoomInput had a useEffect reacting to its currentZoom prop that reached back into the DOM (updateInputString, conditional .select()). That prop only ever changes from one place — the subscribeOnZoomEvent callback in the parent ZoomControls.tsx — so the effect was really just relaying an event the parent already knows about.

Fix: removed that effect from ZoomInput and moved its logic directly into the parent's subscribeOnZoomEvent callback, which already owns both inputRef and the zoom value. ZoomInput now sets its initial display via defaultValue={${Math.round(currentZoom)}%} (safe since the dropdown unmounts/remounts ZoomInput on each open, per MUI Popover's default keepMounted={false}). The mount-focus effect was left untouched since it wasn't flagged.

Verified with eslint: the no-event-handler warning is gone; the two remaining warnings (exhaustive-deps on ZoomTool.instance and hotkeysShortcuts) are pre-existing and out of scope for this ticket. tsc --noEmit shows no new type errors.

## Check list
-  [ - ] unit-tests written
- [ - ] e2e-tests written
- [ - ] documentation updated
- [ x ] PR name follows the pattern `#1234 – issue name`
- [ x ] branch name doesn't contain '#'
- [ x ] PR is linked with the issue
- [ x ] base branch (master or release/xx) is correct
- [ x ] task status changed to "Code review"
- [ x ] reviewers are notified about the pull request


AlexeyGirin:

packages/ketcher-macromolecules/src/components/ZoomControls/ZoomInput.tsx
```diff
@@ -84,6 +73,7 @@ export const ZoomInput = ({
     <StyledInput
       ref={inputRef}
       data-testid="zoom-value"
+      defaultValue={`${Math.round(currentZoom)}%`}
```
dorogusha, line 76: Not critical. Just improvement.
`currentZoom `is already rounded before it gets in `setCurrentZoom(Math.round(...))` in the parent, so this `Math.round` is redundant.

dorogusha (approved):

PR #11363, opened 2026-08-20
#11280 - Refactor: react-hooks/exhaustive-deps rule violation clean up from packages\ketcher-macromolecules\src\components\ZoomControls\ZoomControls.tsx file

### LINKED ISSUE #11280: Refactor: `react-hooks/exhaustive-deps` rule violation clean up from `packages\ketcher-macromolecules\src\components\ZoomControls\ZoomControls.tsx` file

Example:
```ts

useEffect(() => {
  loadUser(userId);
}, []); // userId is missing
```

**Why is this a problem?**

This can cause:
- stale state or props;
- effects using old data;

**How is it usually fixed?**
```ts
useEffect(() => {
  loadUser(userId);
}, [userId]);
```

For a memoized callback:
```ts
}, [documentId, content]);
```

For a memoized value:
```ts
  [items, category],
);
```

**Problem locations:**
packages\ketcher-macromolecules\src\components\ZoomControls\ZoomControls.tsx

HelenaIsh:

## Check list
- [ ] unit-tests written
- [ ] e2e-tests written
- [ ] documentation updated

### DIFF
--- a/packages/ketcher-macromolecules/src/components/ZoomControls/ZoomControls.tsx
+++ b/packages/ketcher-macromolecules/src/components/ZoomControls/ZoomControls.tsx
@@ -44,7 +44,7 @@ export const ZoomControls = () => {
     ZoomTool?.instance?.subscribeOnZoomEvent(() => {
       setCurrentZoom(Math.round(ZoomTool?.instance?.getZoomLevel() * 100));
     });
-  }, [ZoomTool?.instance]);
+  }, []);
 
   const onZoomSubmit = useCallback(() => {
     const inputEl = inputRef.current;
--- a/packages/ketcher-macromolecules/src/components/ZoomControls/ZoomControls.tsx
+++ b/packages/ketcher-macromolecules/src/components/ZoomControls/ZoomControls.tsx
@@ -40,11 +40,13 @@ export const ZoomControls = () => {
   const containerRef = useRef<HTMLDivElement>(null);
   const inputRef = useRef<HTMLInputElement>(null);
 
+  const zoomTool = ZoomTool?.instance;
+
   useEffect(() => {
-    ZoomTool?.instance?.subscribeOnZoomEvent(() => {
-      setCurrentZoom(Math.round(ZoomTool?.instance?.getZoomLevel() * 100));
+    zoomTool?.subscribeOnZoomEvent(() => {
+      setCurrentZoom(Math.round(zoomTool?.getZoomLevel() * 100));
     });
-  }, []);
+  }, [zoomTool]);
 
   const onZoomSubmit = useCallback(() => {
     const inputEl = inputRef.current;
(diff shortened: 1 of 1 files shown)


packages/ketcher-macromolecules/src/components/ZoomControls/ZoomControls.tsx
```diff
@@ -40,11 +40,13 @@ export const ZoomControls = () => {
...
+  const zoomTool = ZoomTool?.instance;
+
   useEffect(() => {
-    ZoomTool?.instance?.subscribeOnZoomEvent(() => {
-      setCurrentZoom(Math.round(ZoomTool?.instance?.getZoomLevel() * 100));
+    zoomTool?.subscribeOnZoomEvent(() => {
+      setCurrentZoom(Math.round(zoomTool?.getZoomLevel() * 100));
     });
-  }, [ZoomTool?.instance]);
+  }, [zoomTool]);
```
svvald: This silences the linter yet the `zoomTool` doesn't become reactive here anyway so effect won't be triggered properly.
`zoomTool` is exposed in `CoreEditor` object, a proper way would be to have the following:

```
const editor = useAppSelector(selectEditor);
const zoomTool = editor?.zoomTool;
```

packages/ketcher-macromolecules/src/components/ZoomControls/ZoomControls.tsx
```diff
@@ -68,15 +70,15 @@ export const ZoomControls = () => {
...
   };
 
   const onZoomOut = () => {
-    ZoomTool.instance.zoomOut();
+    zoomTool.zoomOut();
   };
 
   const onZoomReset = () => {
-    ZoomTool.instance.resetZoom();
+    zoomTool?.resetZoom();
```
svvald: These handlers are imperative, not reactive and `ZoomTool.instance` is a getter which return an actual instance on the call site, so here it's better to revert to `ZoomTool.instance` usage instead

packages/ketcher-macromolecules/src/components/ZoomControls/ZoomControls.tsx
svvald, line 62: Also it was missing initially but it would be nice to add a cleanup for this effect while we're here

svvald (changes requested):

packages/ketcher-macromolecules/src/components/ZoomControls/ZoomControls.tsx
HelenaIsh: thanks, fixed

packages/ketcher-macromolecules/src/components/ZoomControls/ZoomControls.tsx
HelenaIsh: reverted to use ZoomTool.instance

packages/ketcher-macromolecules/src/components/ZoomControls/ZoomControls.tsx
HelenaIsh: added clean up

svvald: Please check the implementation, I'm not sure that `subscribeOnZoomEvent` returns an `unsubscribe` function actually

svvald (changes requested):

packages/ketcher-macromolecules/src/components/ZoomControls/ZoomControls.tsx
HelenaIsh: indeed, used `unsubscribeOnZoomEvent` for cleaning up

svvald (approved):

PR #11368, opened 2026-08-20
#11221 - Refactor: Reading from or writing to ref.current during React render problem clean up from packages\ketcher-react\src\script\ui\views\toolbars\TopToolbar\ZoomControls.tsx file

### LINKED ISSUE #11221: Refactor: Reading from or writing to `ref.current` during React render problem clean up from `packages\ketcher-react\src\script\ui\views\toolbars\TopToolbar\ZoomControls.tsx` file
**Problem:**
Examples:
```ts
function Component() {
  const valueRef = useRef(0);

}
```

**Why is this an issue?**

**How can it be fixed?**
```ts
function Component() {

  useEffect(() => {
    inputRef.current?.focus();
  }, []);

}
```

For event-driven access:
```ts
const handleClick = () => {
  ref.current?.focus();
};
```

```ts

return (
    {count}
  </button>
);
```

**Problem locations:**
packages\ketcher-react\src\script\ui\views\toolbars\TopToolbar\ZoomControls.tsx

Arshakyan11:
Fixed the react-hooks violation caused by reading ref.current during render in ZoomControls.tsx (/packages/ketcher-react/src/script/ui/views/toolbars/TopToolbar/ZoomControls.tsx).

Root cause: ZoomControls passed containerRef.current directly to the Dropdown component during render via anchorEl={containerRef.current}. This reads mutable ref state during React's render phase, which can lead to inconsistent behavior with Strict Mode, concurrent rendering, and React Compiler optimizations.

Fix: introduced an anchorEl state to store the container DOM element. The ref is now accessed only inside the onExpand event handler, where containerRef.current is assigned to anchorEl. The dropdown then uses the state value via anchorEl={anchorEl}. The state is also cleared when the dropdown closes.

Verified: manually tested opening/closing the Zoom dropdown, changing the zoom value, Zoom In, Zoom Out, and resetting to 100%. The full ketcher-react test suite passes: 54 test suites passed, 371 tests passed. Prettier, Stylelint, ESLint, TypeScript, circular dependency checks, and git diff --check also pass.

## Check list
- [ ] unit-tests written
- [ ] e2e-tests written
- [ ] documentation updated

### DIFF
--- a/packages/ketcher-react/src/script/ui/views/toolbars/TopToolbar/ZoomControls.tsx
+++ b/packages/ketcher-react/src/script/ui/views/toolbars/TopToolbar/ZoomControls.tsx
@@ -126,6 +126,7 @@ export const ZoomControls = ({
   shortcuts,
 }: ZoomProps) => {
   const [isExpanded, setIsExpanded] = useState<boolean>(false);
+  const [anchorEl, setAnchorEl] = useState<HTMLElement | null>(null);
   const containerRef = useRef<HTMLDivElement>(null);
   const inputRef = useRef<HTMLInputElement>(null);
 
@@ -149,9 +150,11 @@ export const ZoomControls = ({
       onZoomSubmit();
     }
     setIsExpanded(false);
+    setAnchorEl(null);
   };
 
   const onExpand = () => {
+    setAnchorEl(containerRef.current);
     setIsExpanded(true);
   };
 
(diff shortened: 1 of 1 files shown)


dorogusha (approved):

AlexeyGirin (approved):
Approved by Roman
Image

PR #11369, opened 2026-08-20
#11235 - Refactor: react-you-might-not-need-an-effect/no-chain-state-updates rule violation clean up from packages\ketcher-macromolecules\src\components\preview\Preview.tsx file

### LINKED ISSUE #11235: Refactor: `react-you-might-not-need-an-effect/no-chain-state-updates` rule violation clean up from `packages\ketcher-macromolecules\src\components\preview\Preview.tsx` file

Typical example:
```ts

useEffect(() => {
}, [firstName, lastName]);
```

**Why is this a problem?**
- An input state changes.

This can cause:
- unnecessary renders;
- briefly inconsistent UI;
- more complicated data flow;
- synchronization bugs;

**How is it usually fixed?**
```

```

```
  [items, query],
);
```

```
const handleReset = () => {
  setFirstName('');
  setLastName('');
};
```

```
```

**Problem locations:**
packages\ketcher-macromolecules\src\components\preview\Preview.tsx

nickkurkan:
## Summary
Closes #11235.

`Preview.tsx` had a `useEffect` that positioned the preview element via direct DOM style manipulation (a genuine, layout-dependent side effect) but also called `setIsPreviewVisible(true/false)` purely as a function of `preview.type`/`preview.style`. That state was never read in render — it was only checked back inside the same effect — so the `setState` call was a textbook `no-chain-state-updates` violation: derived bookkeeping routed through `useState`/`useEffect` instead of being computed/tracked directly.

- Replaced `const [isPreviewVisible, setIsPreviewVisible] = useState(false)` with `const isPreviewVisibleRef = useRef(false)`, since the value is only used as internal bookkeeping inside the effect and never drives rendering.
- Updated the two call sites (`setIsPreviewVisible(true)` / `setIsPreviewVisible(false)`) to mutate `isPreviewVisibleRef.current` instead.
- Removed the now-unused `useState` import.
- No behavior change intended: the positioning logic, style resets, and preview rendering are untouched. This also removes a redundant re-render that previously occurred on every effect run without ever being observed in the output.

## Test plan
- [x] `npx eslint src/components/preview/Preview.tsx` in `packages/ketcher-macromolecules` — 0 errors (1 pre-existing, unrelated `react-hooks/exhaustive-deps` warning); the `no-chain-state-updates` violation no longer fires
- [x] `npx tsc --noEmit -p tsconfig.json` in `packages/ketcher-macromolecules` — same pre-existing `Cannot find type definition file for 'react-dom'` environment error occurs identically on `master` (verified via `git stash`), so no new type errors are attributable to this change
- [x] No existing `Preview.test.*` file for this component, so no automated test run was possible (out of scope to add per task instructions)
- [ ] Manually verify in the running app: hover/select a monomer, preset, bond, ambiguous monomer, and text element to confirm the preview tooltip still appears/disappears and positions itself identically to before

### DIFF
--- a/packages/ketcher-macromolecules/src/components/preview/Preview.tsx
+++ b/packages/ketcher-macromolecules/src/components/preview/Preview.tsx
@@ -14,7 +14,7 @@
  * limitations under the License.
  ***************************************************************************/
 
-import { useEffect, useRef, useState } from 'react';
+import { useEffect, useRef } from 'react';
 import { useSelector } from 'react-redux';
 import styled from '@emotion/styled';
 import { ZoomTool } from 'ketcher-core';
@@ -37,7 +37,7 @@ const PreviewContainer = styled.div`
 export const Preview = () => {
   const preview = useAppSelector(selectShowPreview);
   const previewRef = useRef<HTMLDivElement>(null);
-  const [isPreviewVisible, setIsPreviewVisible] = useState(false);
+  const isPreviewVisibleRef = useRef(false);
   const editor = useSelector(selectEditor);
 
   useEffect(() => {
@@ -47,7 +47,7 @@ export const Preview = () => {
 
     if (preview?.type) {
       previewRef.current.setAttribute('style', '');
-      setIsPreviewVisible(true);
+      isPreviewVisibleRef.current = true;
 
       const PREVIEW_OFFSET = 5;
 
@@ -107,8 +107,8 @@ export const Preview = () => {
           canvasWrapperRight - previewWidth - SCROLL_BAR_OFFSET
         }px`;
       }
-    } else if (isPreviewVisible) {
-      setIsPreviewVisible(false);
+    } else if (isPreviewVisibleRef.current) {
+      isPreviewVisibleRef.current = false;
       previewRef.current.setAttribute('style', '');
     }
   }, [preview]);
--- a/packages/ketcher-macromolecules/src/components/preview/Preview.tsx
+++ b/packages/ketcher-macromolecules/src/components/preview/Preview.tsx
@@ -14,7 +14,7 @@
  * limitations under the License.
  ***************************************************************************/
 
-import { useEffect, useRef } from 'react';
+import { useLayoutEffect, useRef } from 'react';
 import { useSelector } from 'react-redux';
 import styled from '@emotion/styled';
 import { ZoomTool } from 'ketcher-core';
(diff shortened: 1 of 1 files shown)


packages/ketcher-macromolecules/src/components/preview/Preview.tsx
```diff
@@ -37,7 +37,7 @@ const PreviewContainer = styled.div`
 export const Preview = () => {
   const preview = useAppSelector(selectShowPreview);
   const previewRef = useRef<HTMLDivElement>(null);
-  const [isPreviewVisible, setIsPreviewVisible] = useState(false);
+  const isPreviewVisibleRef = useRef(false);
   const editor = useSelector(selectEditor);
 
   useEffect(() => {
```
svvald: As we're on refactoring side, let's use `useLayoutEffect `here to avoid paint flash

packages/ketcher-macromolecules/src/components/preview/Preview.tsx
```diff
@@ -107,8 +107,8 @@ export const Preview = () => {
           canvasWrapperRight - previewWidth - SCROLL_BAR_OFFSET
         }px`;
       }
-    } else if (isPreviewVisible) {
-      setIsPreviewVisible(false);
+    } else if (isPreviewVisibleRef.current) {
+      isPreviewVisibleRef.current = false;
       previewRef.current.setAttribute('style', '');
     }
```
svvald: I wonder do we even need this ref in the component? I assume if `preview?.type` is truthy is means that preview is actually visible, otherwise it's not hence this whole effect function might be reduced down to simple `if (preview?.type) { ... } else { ... }` statement, with no additional flags. Can you check that please?

svvald (changes requested):

packages/ketcher-macromolecules/src/components/preview/Preview.tsx
nickkurkan: Done — switched to useLayoutEffect since this measures and writes DOM layout synchronously and should run before paint. Pushed in 2b5b373ab0.

packages/ketcher-macromolecules/src/components/preview/Preview.tsx
nickkurkan: Checked — you're right, isPreviewVisibleRef only gated a redundant (but idempotent) setAttribute('style', '') call in the else branch, so it added no real behavior. Removed it and call setAttribute unconditionally in the else branch instead. Pushed in 2b5b373ab0.

svvald (approved):

PR #11370, opened 2026-08-20
#11246 - Refactor: react-you-might-not-need-an-effect/no-event-handler rule violation clean up from packages\ketcher-react\src\script\ui\dialog\template\TemplateTable.tsx file

### LINKED ISSUE #11246: Refactor: `react-you-might-not-need-an-effect/no-event-handler` rule violation clean up from `packages\ketcher-react\src\script\ui\dialog\template\TemplateTable.tsx` file

Typical example:
```ts

useEffect(() => {
  if (submitted) {
    sendForm();
  }
}, [submitted]);

const handleSubmit = () => {
  setSubmitted(true);
};
```

**Why is this a problem?**
- React renders again.

- an extra render;
- delayed execution;
- dependency-array problems;

**How is it usually fixed?**
```ts
const handleSubmit = () => {
  sendForm();
  setSubmitted(true);
};
```

```ts
const submitForm = () => {
  sendForm();
  setSubmitted(true);
};

```

**Problem locations:**
packages\ketcher-react\src\script\ui\dialog\template\TemplateTable.tsx

nickkurkan:
## Summary
Closes #11246.

- The `react-you-might-not-need-an-effect/no-event-handler` lint rule flagged the progressive-preview-hydration effect in `TemplateTable.tsx`: it watched the `previewRenderedCount` state (and the `templates` prop) purely to decide whether to schedule another idle/animation-frame batch, i.e. it re-derived an action from a state value it had just set itself instead of scheduling that action directly at the point the state changed.
- Refactored the effect into a self-contained recursive scheduler: each idle/animation-frame callback now updates the count and immediately schedules the next batch itself, rather than relying on a separate effect keyed on `previewRenderedCount` to notice the change and react to it. The effect now only depends on `templates` (still legitimate, since it syncs with the external idle/animation-frame scheduler) and cleans up any pending idle/animation-frame callback on unmount or template change.
- No behavior change: initial preview count, batch size, progressive hydration via `requestIdleCallback`/`requestAnimationFrame`, and cleanup semantics are all preserved exactly as before.

## Test plan
- `npx eslint src/script/ui/dialog/template/TemplateTable.tsx` (from `packages/ketcher-react`) — the `no-event-handler` warnings are gone; only the pre-existing, unrelated `no-explicit-any` warning remains.
- `npx tsc --noEmit -p tsconfig.json` (from `packages/ketcher-react`) — no new errors introduced (remaining errors are pre-existing and unrelated to this file: missing `miew`/`miew-react` modules and a generated schema file).
- No existing unit test file covers `TemplateTable.tsx`; none added per task scope.
- Manually verified the component's data flow: initial preview batch size, progressive batch growth, and cleanup on unmount/template swap are unchanged.

### DIFF
--- a/packages/ketcher-react/src/script/ui/dialog/template/TemplateTable.tsx
+++ b/packages/ketcher-react/src/script/ui/dialog/template/TemplateTable.tsx
@@ -53,6 +53,15 @@ interface TemplateTableProps {
 // then progressively mount heavy SVG previews with low-priority updates.
 const INITIAL_PREVIEW_COUNT = 4;
 const PREVIEW_BATCH_SIZE = 4;
+// How long a batch may wait for idle time before the browser is asked to run
+// it regardless. Capped at 100ms because a longer deadline (500ms) leaves a
+// pause that is perceptible before the previews appear.
+const PREVIEW_IDLE_TIMEOUT_MS = 100;
+
+// Single source of truth for where preview hydration starts, so the initial
+// render, the reset on templates change, and the batch scheduler cannot drift.
+const getInitialPreviewCount = (templateCount: number) =>
+  Math.min(INITIAL_PREVIEW_COUNT, templateCount);
 
 const TemplateTable: FC<TemplateTableProps> = (props) => {
   const {
@@ -67,7 +76,7 @@ const TemplateTable: FC<TemplateTableProps> = (props) => {
 
   const [prevTemplates, setPrevTemplates] = useState(templates);
   const [previewRenderedCount, setPreviewRenderedCount] = useState(() =>
-    Math.min(INITIAL_PREVIEW_COUNT, templates.length),
+    getInitialPreviewCount(templates.length),
   );
   const [, startTransition] = useTransition();
   const [containerSize, setContainerSize] = useState<{
@@ -78,7 +87,7 @@ const TemplateTable: FC<TemplateTableProps> = (props) => {
 
   if (templates !== prevTemplates) {
     setPrevTemplates(templates);
-    setPreviewRenderedCount(Math.min(INITIAL_PREVIEW_COUNT, templates.length));
+    setPreviewRenderedCount(getInitialPreviewCount(templates.length));
   }
 
   // Calculate container size once to avoid getBoundingClientRect() calls during rendering
@@ -132,13 +141,15 @@ const TemplateTable: FC<TemplateTableProps> = (props) => {
       };
 
       if (typeof requestIdle === 'function') {
-        idleCallbackId = requestIdle(runBatch, { timeout: 500 });
+        idleCallbackId = requestIdle(runBatch, {
+          timeout: PREVIEW_IDLE_TIMEOUT_MS,
+        });
       } else {
         frameId = window.requestAnimationFrame(runBatch);
       }
     };
 
-    scheduleNext(Math.min(INITIAL_PREVIEW_COUNT, templates.length));
+    scheduleNext(getInitialPreviewCount(templates.length));
 
     return () => {
       cancelled = true;
--- a/packages/ketcher-react/src/script/ui/dialog/template/TemplateTable.tsx
+++ b/packages/ketcher-react/src/script/ui/dialog/template/TemplateTable.tsx
@@ -105,38 +105,48 @@ const TemplateTable: FC<TemplateTableProps> = (props) => {
     return () => resizeObserver.disconnect();
   }, []);
 
+  // Progressive preview hydration reacts to the idle/frame scheduler (an
+  // external system), so each scheduled batch schedules the next one
+  // directly instead of relying on an effect that watches the count state.
   useEffect(() => {
-    if (previewRenderedCount >= templates.length) {
-      return;
-    }
+    let cancelled = false;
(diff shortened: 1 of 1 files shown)


packages/ketcher-react/src/script/ui/dialog/template/TemplateTable.tsx
```diff
@@ -105,38 +105,48 @@ const TemplateTable: FC<TemplateTableProps> = (props) => {
...
-            Math.min(prev + PREVIEW_BATCH_SIZE, templates.length),
-          );
-        });
-      });
-      return () => window.cancelAnimationFrame(frameId);
-    }
-  }, [previewRenderedCount, templates.length, startTransition]);
+
+      if (typeof requestIdle === 'function') {
+        idleCallbackId = requestIdle(runBatch, { timeout: 500 });
```
Jesovile: Why timeout is 500? According UI research 300ms is a max time for shooth rendering.
Do we have a reason for that?

Jesovile: With 500 I see this rendering (pause and rendering after).
I've checked it with 100 - much better from user's perspective. Almost no delay.

Can we set 100 instead of 500?

packages/ketcher-react/src/script/ui/dialog/template/TemplateTable.tsx
```diff
@@ -105,38 +105,48 @@ const TemplateTable: FC<TemplateTableProps> = (props) => {
...
+      const runBatch = () => {
+        if (cancelled) {
+          return;
+        }
+        const nextCount = Math.min(
+          currentCount + PREVIEW_BATCH_SIZE,
+          templates.length,
+        );
+        startTransition(() => setPreviewRenderedCount(nextCount));
+        scheduleNext(nextCount);
```
dorogusha, line 140: Nice fix for the no-event-handler violation — the recursive scheduler approach is the right direction. Two things worth considering, though:

1. Complexity trade-off

The original effect was a single, flat if + one requestIdleCallback/requestAnimationFrame call per batch — easy to follow in one read. The new version introduces two nested closures (scheduleNext → runBatch → scheduleNext) plus three pieces of local state to track (cancelled, idleCallbackId, frameId), where before there was just one callback handle. We're trading a lint warning for a meaningfully harder-to-follow control flow. It might be worth asking whether the lint rule is actually flagging a real problem here, or a false positive worth a targeted eslint-disable with a comment explaining why this effect legitimately needs to react to its own state (progressive hydration paced by an external scheduler is arguably a valid exception to the rule's intent). If we do go with the recursive rewrite, I'd at least like to simplify it — see below.

2. Duplicated "initial count" formula

The new effect re-derives the starting point of the batch scheduler with Math.min(INITIAL_PREVIEW_COUNT, templates.length) instead of reading the actual previewRenderedCount state. That formula is now duplicated in three places in this file (the useState initializer, the reset on templates change at line ~81, and this effect), and nothing enforces that they stay in sync. If the reset logic above ever changes (e.g. initial count starts depending on container size), this effect's assumption would silently drift from the real state.

We can avoid the duplication - and simplify the closures a bit - by reading the current progress through a ref instead of recomputing it, so the effect never needs to know the initial-count formula at all:

`const previewRenderedCountRef = useRef(previewRenderedCount);
previewRenderedCountRef.current = previewRenderedCount;

useEffect(() => {
  let cancelled = false;
  let idleCallbackId: number | undefined;
  let frameId: number | undefined;

  const scheduleNext = () => {
    if (previewRenderedCountRef.current >= templates.length) {
      return;
    }

    const runBatch = () => {
      if (cancelled) {
        return;
      }
      const nextCount = Math.min(
        previewRenderedCountRef.current + PREVIEW_BATCH_SIZE,
        templates.length,
      );
      previewRenderedCountRef.current = nextCount;
      startTransition(() => setPreviewRenderedCount(nextCount));
      scheduleNext();
    };

    if (typeof window.requestIdleCallback === 'function') {
      idleCallbackId = window.requestIdleCallback(runBatch, { timeout: 500 });
    } else {
      frameId = window.requestAnimationFrame(runBatch);
    }
  };

  scheduleNext();

  return () => {
    cancelled = true;
    if (idleCallbackId !== undefined)
      window.cancelIdleCallback?.(idleCallbackId);
    if (frameId !== undefined) window.cancelAnimationFrame(frameId);
  };
}, [templates, startTransition]);`

This doesn't remove the extra nesting entirely (that's inherent to the recursive-scheduler approach), but it does:
- Drop the duplicated INITIAL_PREVIEW_COUNT formula - the effect always continues from whatever previewRenderedCount actually is, so it can't drift out of sync with the reset logic above.
- Drop the currentCount parameter threaded through the closures (one less thing to track per call).
- Still satisfy react-hooks/exhaustive-deps and no-event-handler without an eslint-disable, since refs are exempt from the dependency array.

What do you think? Including whether an eslint-disable with a clear justification might actually be the simpler, more honest fix here compared to restructuring the whole effect.

dorogusha (approved):

AlexeyGirin:
@nickkurkan please refactor, return batch rendering to previous state and enchance it if possible

PR #11372, opened 2026-08-20
#11240 - Refactor: react-you-might-not-need-an-effect/no-event-handler rule violation clean up from packages\ketcher-react\src\script\ui\views\modal\components\document\Open\components\CDXStructuresViewer\CDXStructuresViewer.tsx file

### LINKED ISSUE #11240: Refactor: `react-you-might-not-need-an-effect/no-event-handler` rule violation clean up from `packages\ketcher-react\src\script\ui\views\modal\components\document\Open\components\CDXStructuresViewer\CDXStructuresViewer.tsx` file

Typical example:
```ts

useEffect(() => {
  if (submitted) {
    sendForm();
  }
}, [submitted]);

const handleSubmit = () => {
  setSubmitted(true);
};
```

**Why is this a problem?**
- React renders again.

- an extra render;
- delayed execution;
- dependency-array problems;

**How is it usually fixed?**
```ts
const handleSubmit = () => {
  sendForm();
  setSubmitted(true);
};
```

```ts
const submitForm = () => {
  sendForm();
  setSubmitted(true);
};

```

**Problem locations:**
packages\ketcher-react\src\script\ui\views\modal\components\document\Open\components\CDXStructuresViewer\CDXStructuresViewer.tsx

nickkurkan:
## Summary
- Closes #11240.
- `CDXStructuresViewer.tsx` had two `useEffect` hooks whose logic only ever ran as a reaction to the component's own selection-click handler (`setSelectedIndex`) and to the state it later set (`itemsMap`), which is the exact anti-pattern flagged by `react-you-might-not-need-an-effect/no-event-handler`.
- The effect that called `inputHandler` based on `itemsMap`/`selectedIndex`, and the effect that kicked off `getImage` fetch for the selected index, are removed. That logic now runs directly inside a new `handleSelectStructure` click handler and inside the `getImage` promise callbacks (checked against the current selection via a ref, since structure parsing is async).
- A single `useEffect` remains, scoped to loading the currently selected structure whenever a new `structList` prop arrives (e.g. a new file is opened) — this is a legitimate reaction to an external prop/async parsing, not to a local event, so it stays as an effect.
- No behavior change: structure selection, loading spinner, and error display all work identically to before.

## Test plan
- `npx eslint src/script/ui/views/modal/components/document/Open/components/CDXStructuresViewer/CDXStructuresViewer.tsx` from `packages/ketcher-react` — the `no-event-handler` warnings for state-driven effects are gone (only a pre-existing, legitimate prop-driven async-loading warning remains).
- `npx tsc --noEmit -p tsconfig.json` from `packages/ketcher-react` — no new type errors introduced by this change.
- No existing unit tests cover this component; manually reasoned through selection flow (click → immediate load/inputHandler call) and async load flow (parseStruct resolves after selection may have changed) to confirm parity with prior behavior.

### DIFF
--- a/packages/ketcher-react/src/script/ui/views/modal/components/document/Open/components/CDXStructuresViewer/CDXStructuresViewer.tsx
+++ b/packages/ketcher-react/src/script/ui/views/modal/components/document/Open/components/CDXStructuresViewer/CDXStructuresViewer.tsx
@@ -14,7 +14,7 @@
  * limitations under the License.
  ***************************************************************************/
 
-import { useEffect, useState } from 'react';
+import { useEffect, useRef, useState } from 'react';
 import { useSelector } from 'react-redux';
 import { MenuList } from '@mui/material';
 import MenuItem from '@mui/material/MenuItem';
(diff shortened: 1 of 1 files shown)


packages/ketcher-react/src/script/ui/views/modal/components/document/Open/components/CDXStructuresViewer/CDXStructuresViewer.tsx
```diff
@@ -49,36 +49,63 @@ export const CDXStructuresViewer = ({
   const [selectedIndex, setSelectedIndex] = useState(0);
   const [itemsMap, setItemsMap] = useState<itemsMapInterface>({});
   const loading = !!structList[selectedIndex] && !itemsMap[selectedIndex];
+  const selectedIndexRef = useRef(selectedIndex);
+  selectedIndexRef.current = selectedIndex;
+  const itemsMapRef = useRef(itemsMap);
+  itemsMapRef.current = itemsMap;
```
dorogusha, line 55: Nit: mutating refs during render

Writing to .current directly in the render body is technically outside React's documented ref usage (refs are meant to be read/written in event handlers or effects, not during render). It happens to work here since the write is idempotent and has no rendering side effects, but it's a fragile pattern — under React's Strict Mode double-invocation or future concurrent rendering changes it could behave unexpectedly.

Suggest syncing these in a useEffect instead:

```
useEffect(() => {
  selectedIndexRef.current = selectedIndex;
}, [selectedIndex]);

useEffect(() => {
  itemsMapRef.current = itemsMap;
}, [itemsMap]);
```

This keeps the "latest value" ref pattern but moves the mutation to a place React explicitly allows.

packages/ketcher-react/src/script/ui/views/modal/components/document/Open/components/CDXStructuresViewer/CDXStructuresViewer.tsx
```diff
@@ -49,36 +49,63 @@ export const CDXStructuresViewer = ({
...
 
-  useEffect(() => {
-    if (!itemsMap[selectedIndex] || itemsMap[selectedIndex].error) {
+  const notifyInputHandler = (item?: item) => {
+    if (!item || item.error) {
```
dorogusha, line 58: Nit: parameter shadows the item type

const notifyInputHandler = (item?: item) => {

The parameter name item shadows the item type it's typed with. Compiles fine, but reads confusingly. Consider renaming the parameter, e.g. structItem?: item.

dorogusha (changes requested):

dorogusha (approved):

AlexeyGirin:

PR #11373, opened 2026-08-20
#11244 - Refactor: react-you-might-not-need-an-effect/no-event-handler rule violation clean up from packages\ketcher-react\src\script\ui\views\components\ContextMenu\menuItems\BondMenuItems.tsx file

### LINKED ISSUE #11244: Refactor: `react-you-might-not-need-an-effect/no-event-handler` rule violation clean up from `packages\ketcher-react\src\script\ui\views\components\ContextMenu\menuItems\BondMenuItems.tsx` file

Typical example:
```ts

useEffect(() => {
  if (submitted) {
    sendForm();
  }
}, [submitted]);

const handleSubmit = () => {
  setSubmitted(true);
};
```

**Why is this a problem?**
- React renders again.

- an extra render;
- delayed execution;
- dependency-array problems;

**How is it usually fixed?**
```ts
const handleSubmit = () => {
  sendForm();
  setSubmitted(true);
};
```

```ts
const submitForm = () => {
  sendForm();
  setSubmitted(true);
};

```

**Problem locations:**
packages\ketcher-react\src\script\ui\views\components\ContextMenu\menuItems\BondMenuItems.tsx

nickkurkan:
## Summary
Closes #11244.

`BondMenuItems.tsx` had an ESLint warning from `react-you-might-not-need-an-effect/no-event-handler`: the `useEffect` that derives `bondData`/`isBondBetweenMonomers` from `props.propsFromTrigger` contained an outer `if (bondIds.length > 0 && editor)` block with no matching `else`, which the rule treats as prop-driven logic that belongs in an explicit branch rather than an implicit fallthrough.

The fix adds an explicit `else` branch that resets `bondData` and `isBondBetweenMonomers` when there are no bond ids / no editor, mirroring the reset already done for the "bond not found" case inside the inner `if (bond)`. This satisfies the rule without altering when/how the bond context menu displays its items — the menu is only visible when a real bond context is active, so this else branch only executes while the component is mounted but hidden (e.g. a different context menu is open), which does not affect observable behavior.

This PR does **not** touch the separate `no-adjust-state-on-prop-change` (set-state-in-effect) violations on the same effect, which are being handled by #11107 / PR #11348.

## Test plan
- `npx eslint src/script/ui/views/components/ContextMenu/menuItems/BondMenuItems.tsx` (from `packages/ketcher-react`) — `no-event-handler` warning no longer fires; only the pre-existing `no-adjust-state-on-prop-change` warnings (owned by #11107) remain.
- `npx tsc --noEmit -p tsconfig.json` (from `packages/ketcher-react`) — no new errors introduced by this change.
- No dedicated unit test file exists for `BondMenuItems.tsx`; behavior verified by code inspection (no observable change to bond context menu rendering).

### DIFF
--- a/packages/ketcher-react/src/script/ui/views/components/ContextMenu/menuItems/BondMenuItems.tsx
+++ b/packages/ketcher-react/src/script/ui/views/components/ContextMenu/menuItems/BondMenuItems.tsx
@@ -67,6 +67,9 @@ const BondMenuItems: FC<MenuItemsProps<BondsContextMenuProps>> = (props) => {
         setBondData(null);
         setIsBondBetweenMonomers(false);
       }
+    } else {
+      setBondData(null);
+      setIsBondBetweenMonomers(false);
     }
   }, [props.propsFromTrigger, ketcherId]);
 
--- a/packages/ketcher-react/src/script/ui/views/components/ContextMenu/menuItems/BondMenuItems.tsx
+++ b/packages/ketcher-react/src/script/ui/views/components/ContextMenu/menuItems/BondMenuItems.tsx
@@ -9,7 +9,12 @@ import useBondSGroupAttach from '../hooks/useBondSGroupAttach';
 import useBondSGroupEdit from '../hooks/useBondSGroupEdit';
 import useBondTypeChange from '../hooks/useBondTypeChange';
 import useDelete from '../hooks/useDelete';
-import { formatTitle, getNonQueryBondNames, queryBondNames } from '../utils';
+import {
+  formatTitle,
+  getNonQueryBondNames,
+  isBondBetweenMonomers,
+  queryBondNames,
+} from '../utils';
 import type {
   BondsContextMenuProps,
   ItemEventParams,
@@ -19,7 +24,7 @@ import { getIconName, Icon } from 'components';
 import { useChangeBondDirection } from '../hooks/useChangeBondDirection';
 import { useAppContext } from 'src/hooks/useAppContext';
 import HighlightMenu from 'src/script/ui/action/highlightColors/HighlightColors';
-import { ketcherProvider, MonomerMicromolecule } from 'ketcher-core';
+import { ketcherProvider } from 'ketcher-core';
 
 type Params = ItemEventParams<BondsContextMenuProps>;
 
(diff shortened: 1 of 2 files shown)


dorogusha (approved):
Approving. Verified the state reset only happens while the bond context menu is hidden (it stays mounted via react-contexify, only shown when propsFromTrigger has valid bondIds), so there's no observable UI impact — confirms the PR's own claim rather than just taking it at face value. No new tests needed given the scope.

packages/ketcher-react/src/script/ui/views/components/ContextMenu/menuItems/BondMenuItems.tsx
```diff
@@ -67,6 +67,9 @@ const BondMenuItems: FC<MenuItemsProps<BondsContextMenuProps>> = (props) => {
         setBondData(null);
         setIsBondBetweenMonomers(false);
       }
+    } else {
+      setBondData(null);
+      setIsBondBetweenMonomers(false);
```
svvald: This seems like a rule bypass code rather than a real fix, this branch is dead and will never be called

packages/ketcher-react/src/script/ui/views/components/ContextMenu/menuItems/BondMenuItems.tsx
```diff
@@ -67,6 +67,9 @@ const BondMenuItems: FC<MenuItemsProps<BondsContextMenuProps>> = (props) => {
...
     }
   }, [props.propsFromTrigger, ketcherId]);
```
svvald: Essentially we don't even need this effect – there is no need in putting both`bondData` and `isBondBetweenMonomers` to states and let alone changing them in effect. They could be easily derived on render-time – as re-render is triggered by both reactive `ketcherId` change from `AppContext` and own props change

svvald (changes requested):

packages/ketcher-react/src/script/ui/views/components/ContextMenu/menuItems/BondMenuItems.tsx
nickkurkan: Fair point on the branch being technically reachable rather than strictly dead (it fires when bondIds is empty or editor is momentarily undefined) — but it's moot once the state+effect go away per your other comment. Removed both useState calls and derived bondData/isBondBetweenMonomers directly during render instead. Pushed in ab8a8e86f2.

packages/ketcher-react/src/script/ui/views/components/ContextMenu/menuItems/BondMenuItems.tsx
nickkurkan: Agreed and done — bondData and isBondBetweenMonomers are now computed directly in the render body from props.propsFromTrigger.bondIds + the already-reactive editor/ketcherId, no state or effect needed. Pushed in ab8a8e86f2.

AlexeyGirin:

packages/ketcher-react/src/script/ui/views/components/ContextMenu/menuItems/BondMenuItems.tsx
```diff
@@ -45,30 +40,27 @@ const BondMenuItems: FC<MenuItemsProps<BondsContextMenuProps>> = (props) => {
...
-        setIsBondBetweenMonomers(false);
-      }
-    }
-  }, [props.propsFromTrigger, ketcherId]);
+    return (
+      beginAtomSgroup instanceof MonomerMicromolecule &&
+      endAtomSgroup instanceof MonomerMicromolecule &&
+      beginAtomSgroup !== endAtomSgroup
+    );
+  })();
```
svvald: Let's extract it into an util function outside of the component instead of this long IIFE

svvald (changes requested):

svvald (approved):

PR #11375, opened 2026-08-20
#11238 - Refactor: react-you-might-not-need-an-effect/no-event-handler rule violation clean up from packages\ketcher-react\src\script\ui\views\toolbars\ToolbarGroupItem\ToolbarMultiToolItem\usePortalStyle.ts file

### LINKED ISSUE #11238: Refactor: `react-you-might-not-need-an-effect/no-event-handler` rule violation clean up from `packages\ketcher-react\src\script\ui\views\toolbars\ToolbarGroupItem\ToolbarMultiToolItem\usePortalStyle.ts` file

Typical example:
```ts

useEffect(() => {
  if (submitted) {
    sendForm();
  }
}, [submitted]);

const handleSubmit = () => {
  setSubmitted(true);
};
```

**Why is this a problem?**
- React renders again.

- an extra render;
- delayed execution;
- dependency-array problems;

**How is it usually fixed?**
```ts
const handleSubmit = () => {
  sendForm();
  setSubmitted(true);
};
```

```ts
const submitForm = () => {
  sendForm();
  setSubmitted(true);
};

```

**Problem locations:**
packages\ketcher-react\src\script\ui\views\toolbars\ToolbarGroupItem\ToolbarMultiToolItem\usePortalStyle.ts

nickkurkan:
## Summary
- Closes #11238.
- `usePortalStyle.ts` was tripping the `react-you-might-not-need-an-effect/no-event-handler` ESLint rule: the effect body opened with a bare `if (!ref.current) return;` guard, and the rule's heuristic flags any `if` (without an `else`) inside an effect whose test traces back to a prop/hook-parameter as an "event handler in disguise".
- The effect itself is legitimately needed — it measures the toolbar item's DOM position via `getBoundingClientRect()` after render/resize/reposition, which cannot be computed synchronously during render. Removing the effect was not an option.
- Fix: restructured the null guard from an early `return` into an `if (ref.current) { ... } else { ... }` block. This keeps the exact same runtime behavior (position is recalculated only once the ref is attached, and the previous computed style is left untouched otherwise) while giving the `if` an explicit `else` branch, which is outside the pattern this rule flags.
- No change in observable behavior: portal positioning for open/close/resize/reposition works identically.

## Test plan
- `npx eslint src/script/ui/views/toolbars/ToolbarGroupItem/ToolbarMultiToolItem/usePortalStyle.ts` (from `packages/ketcher-react`) — `no-event-handler` warning no longer fires (only the pre-existing, out-of-scope `no-derived-state` warning remains, which reflects a legitimately-needed DOM-measurement effect).
- `npx tsc --noEmit -p tsconfig.json` (from `packages/ketcher-react`) — no new type errors introduced by this change (remaining errors are pre-existing environment issues unrelated to this file).
- No existing unit tests cover this hook/component; none were added per task scope.

### DIFF
--- a/packages/ketcher-react/src/script/ui/views/toolbars/ToolbarGroupItem/ToolbarMultiToolItem/usePortalStyle.ts
+++ b/packages/ketcher-react/src/script/ui/views/toolbars/ToolbarGroupItem/ToolbarMultiToolItem/usePortalStyle.ts
@@ -33,29 +33,29 @@ function usePortalStyle([
   const [portalStyle, setPortalStyle] = useState<CSSProperties>({});
 
   useEffect(() => {
-    if (!ref.current) {
-      return;
+    if (ref.current) {
+      const editorRect = document
+        .querySelector(rootElementSelector ?? KETCHER_ROOT_NODE_CSS_SELECTOR)
+        ?.getBoundingClientRect() ?? { top: 0, left: 0 };
+      const menuItemRect = ref.current.getBoundingClientRect();
+
+      const spaceBetween = 4;
+      const alignmentOffset = 2;
+      const top =
+        menuItemRect.top -
+        editorRect.top +
+        (isTop ? spaceBetween + menuItemRect.height : 0) -
+        (isTop ? 0 : alignmentOffset);
+      const left =
+        menuItemRect.left -
+        editorRect.left +
+        (isTop ? 0 : spaceBetween + menuItemRect.width) -
+        (isTop ? alignmentOffset : 0);
+
+      setPortalStyle({ top: `${top}px`, left: `${left}px` });
+    } else {
+      // ref not attached to the DOM yet; keep the previously computed style
     }
-
-    const editorRect = document
-      .querySelector(rootElementSelector ?? KETCHER_ROOT_NODE_CSS_SELECTOR)
-      ?.getBoundingClientRect() ?? { top: 0, left: 0 };
-    const menuItemRect = ref.current.getBoundingClientRect();
-
-    const spaceBetween = 4;
-    const alignmentOffset = 2;
-    const top =
-      menuItemRect.top -
-      editorRect.top +
-      (isTop ? spaceBetween + menuItemRect.height : 0) -
-      (isTop ? 0 : alignmentOffset);
-    const left =
-      menuItemRect.left -
-      editorRect.left +
-      (isTop ? 0 : spaceBetween + menuItemRect.width) -
-      (isTop ? alignmentOffset : 0);
-
-    setPortalStyle({ top: `${top}px`, left: `${left}px` });
   }, [ref, isOpen, isTop, rootElementSelector]);
 
   return [portalStyle];
--- a/packages/ketcher-react/src/script/ui/views/toolbars/ToolbarGroupItem/ToolbarMultiToolItem/usePortalStyle.ts
+++ b/packages/ketcher-react/src/script/ui/views/toolbars/ToolbarGroupItem/ToolbarMultiToolItem/usePortalStyle.ts
@@ -14,7 +14,12 @@
  * limitations under the License.
  ***************************************************************************/
 
-import { type CSSProperties, type RefObject, useEffect, useState } from 'react';
+import {
+  type CSSProperties,
+  type RefObject,
+  useLayoutEffect,
+  useState,
+} from 'react';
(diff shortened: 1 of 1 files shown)


packages/ketcher-react/src/script/ui/views/toolbars/ToolbarGroupItem/ToolbarMultiToolItem/usePortalStyle.ts
```diff
@@ -33,29 +33,29 @@ function usePortalStyle([
...
+        menuItemRect.left -
+        editorRect.left +
+        (isTop ? 0 : spaceBetween + menuItemRect.width) -
+        (isTop ? alignmentOffset : 0);
 
-    setPortalStyle({ top: `${top}px`, left: `${left}px` });
+      setPortalStyle({ top: `${top}px`, left: `${left}px` });
+    } else {
+      // ref not attached to the DOM yet; keep the previously computed style
+    }
```
dorogusha: It's not critical, but perhaps empty `else` statement should be removed?

dorogusha (changes requested):

packages/ketcher-react/src/script/ui/views/toolbars/ToolbarGroupItem/ToolbarMultiToolItem/usePortalStyle.ts
```diff
@@ -33,29 +33,29 @@ function usePortalStyle([
...
   useEffect(() => {
-    if (!ref.current) {
-      return;
-    }
-
-    const editorRect = document
-      .querySelector(rootElementSelector ?? KETCHER_ROOT_NODE_CSS_SELECTOR)
-      ?.getBoundingClientRect() ?? { top: 0, left: 0 };
-    const menuItemRect = ref.current.getBoundingClientRect();
+    if (ref.current) {
```
svvald: It seems that the only change it makes is inverting the if statement condition without incorporating any reals changes. It seems that eslint rule just fires false-positively here, I'd suggest to revert changes back to early return and use a suppression comment as well as a comment on it being false positive

packages/ketcher-react/src/script/ui/views/toolbars/ToolbarGroupItem/ToolbarMultiToolItem/usePortalStyle.ts
```diff
@@ -33,29 +33,29 @@ function usePortalStyle([
   const [portalStyle, setPortalStyle] = useState<CSSProperties>({});
 
   useEffect(() => {
```
svvald: As we're on the refactoring side – `useLayoutEffect` would fit better in order to avoid paint flash

svvald (changes requested):

packages/ketcher-react/src/script/ui/views/toolbars/ToolbarGroupItem/ToolbarMultiToolItem/usePortalStyle.ts
nickkurkan: Removed — reverted to the original early-return guard clause instead of the inverted if/empty-else, so there's no dangling empty else anymore. Pushed in 164d573903.

packages/ketcher-react/src/script/ui/views/toolbars/ToolbarGroupItem/ToolbarMultiToolItem/usePortalStyle.ts
nickkurkan: Agreed, confirmed by linting both versions — the original guard clause was fine and the refactor just inverted the condition to satisfy the rule literally without changing behavior. Reverted to the early-return form and added a suppression comment explaining it's a false positive (portalStyle depends on DOM measurements only available post-layout, so it can't be computed synchronously during render). Pushed in 164d573903.

packages/ketcher-react/src/script/ui/views/toolbars/ToolbarGroupItem/ToolbarMultiToolItem/usePortalStyle.ts
nickkurkan: Done — switched to useLayoutEffect to avoid a paint flash when the portal opens. Pushed in 164d573903.

dorogusha (approved):

svvald (approved):

PR #11377, opened 2026-08-20
#11236 - Refactor: react-you-might-not-need-an-effect/no-chain-state-updates rule violation clean up from packages\ketcher-macromolecules\src\components\monomerLibrary\RnaBuilder\RnaEditor\RnaEditorExpanded\RnaEditorExpanded.tsx

### LINKED ISSUE #11236: Refactor: `react-you-might-not-need-an-effect/no-chain-state-updates` rule violation clean up from `packages\ketcher-macromolecules\src\components\monomerLibrary\RnaBuilder\RnaEditor\RnaEditorExpanded\RnaEditorExpanded.tsx` file

nickkurkan:
## Summary
Closes #11236.

`RnaEditorExpanded.tsx` had a `useEffect` (keyed on `activePresetMonomerGroup?.groupItem`, `isSequenceEditInRNABuilderMode`, `selectedPhosphatePosition`) that, when editing a sequence selection in the RNA Builder, computed an updated sequence selection, dispatched it to redux, **and** called `setIsSequenceSelectionUpdated(true)`. That last call is a pure derived-state update chained inside the effect with no genuine external synchronization of its own, tripping `react-you-might-not-need-an-effect/no-chain-state-updates`.

- Removed the `setIsSequenceSelectionUpdated(true)` call from inside the effect.
- Added a `lastAppliedPresetMonomerGroupItem` state variable that tracks the previously applied monomer group item, and moved the `isSequenceSelectionUpdated` update to the render body using the "adjust state during render" pattern (comparing the previous tracked value against the current `activePresetMonomerGroup?.groupItem` and calling `setState` directly when they differ), matching the pattern already used elsewhere in this codebase for this exact rule (see #11113, #11320).
- The `dispatch(setSequenceSelection(...))` call remains inside the `useEffect`, since it's a genuine side effect (redux dispatch), not derivable state.
- No behavior change intended: `isSequenceSelectionUpdated` still flips to `true` under exactly the same condition (editing a sequence selection in the RNA Builder and a monomer group item being applied), which still gates the "Update" button the same way.

## Test plan
- [x] `npx eslint src/components/monomerLibrary/RnaBuilder/RnaEditor/RnaEditorExpanded/RnaEditorExpanded.tsx` in `packages/ketcher-macromolecules` — the `no-chain-state-updates` warning is gone; only pre-existing, unrelated warnings remain (`no-derived-state`, `no-event-handler`, `exhaustive-deps`, tracked by other issues)
- [x] `npx tsc --noEmit -p tsconfig.json` in `packages/ketcher-macromolecules` — no errors
- [x] `npx jest src/components/monomerLibrary/RnaBuilder/RnaEditor/RnaEditorExpanded/RnaEditorExpanded.test.tsx` — 3/3 tests pass, snapshots match

### DIFF
--- a/packages/ketcher-macromolecules/src/components/monomerLibrary/RnaBuilder/RnaEditor/RnaEditorExpanded/RnaEditorExpanded.tsx
+++ b/packages/ketcher-macromolecules/src/components/monomerLibrary/RnaBuilder/RnaEditor/RnaEditorExpanded/RnaEditorExpanded.tsx
@@ -188,10 +188,34 @@ export const RnaEditorExpanded = ({
   );
   const [isSequenceSelectionUpdated, setIsSequenceSelectionUpdated] =
     useState<boolean>(false);
+  // Tracks the previously applied monomer group item so that
+  // `isSequenceSelectionUpdated` can be derived during render (see below)
+  // instead of via a `useEffect` chained state update.
+  const [
+    lastAppliedPresetMonomerGroupItem,
+    setLastAppliedPresetMonomerGroupItem,
+  ] = useState(activePresetMonomerGroup?.groupItem);
   const [sequenceSelectionGroupNames, setSequenceSelectionGroupNames] =
     useState<SequenceSelectionGroupNames | undefined>(
       generateSequenceSelectionGroupNames(sequenceSelection),
     );
+  // Adjust state during render (instead of chaining a setState call inside
+  // the useEffect below): once the user picks a new monomer group item while
+  // editing a sequence selection, mark the selection as updated so the
+  // "Update" button becomes enabled.
+  if (
+    activeMonomerGroup !== RnaBuilderPresetsItem.Presets &&
+    isEditMode &&
+    isSequenceEditInRNABuilderMode &&
+    activePresetMonomerGroup?.groupItem &&
+    activePresetMonomerGroup.groupItem !== lastAppliedPresetMonomerGroupItem
+  ) {
+    setLastAppliedPresetMonomerGroupItem(activePresetMonomerGroup.groupItem);
+    if (!isSequenceSelectionUpdated) {
+      setIsSequenceSelectionUpdated(true);
+    }
+  }
+
   const phosphatePosition = resolvePhosphatePosition(newPreset);
   const { is3PrimeAvailable, is5PrimeAvailable } =
     getPhosphatePositionAvailability(newPreset || {});
@@ -274,7 +298,6 @@ export const RnaEditorExpanded = ({
           };
         });
 
-        setIsSequenceSelectionUpdated(true);
         dispatch(setSequenceSelection(updatedSequenceSelection));
       } else {
         const currentPreset = updatePresetMonomerGroup();
--- a/packages/ketcher-macromolecules/src/components/monomerLibrary/RnaBuilder/RnaEditor/RnaEditorExpanded/RnaEditorExpanded.tsx
+++ b/packages/ketcher-macromolecules/src/components/monomerLibrary/RnaBuilder/RnaEditor/RnaEditorExpanded/RnaEditorExpanded.tsx
@@ -199,11 +199,22 @@ export const RnaEditorExpanded = ({
     useState<SequenceSelectionGroupNames | undefined>(
       generateSequenceSelectionGroupNames(sequenceSelection),
     );
+  // Tracks whether the component was in edit mode on the previous render, so
+  // that re-entering edit mode (e.g. after Cancel) can be detected during
+  // render and used to discard stale tracking from a previous edit session.
+  const [wasEditMode, setWasEditMode] = useState(isEditMode);
   // Adjust state during render (instead of chaining a setState call inside
-  // the useEffect below): once the user picks a new monomer group item while
-  // editing a sequence selection, mark the selection as updated so the
-  // "Update" button becomes enabled.
-  if (
+  // the useEffect below). Re-entering edit mode starts a fresh session: reset
+  // the tracked monomer group item and the "updated" flag so state left over
+  // from a previous, cancelled edit session can't leak in and prematurely
+  // enable the "Update" button.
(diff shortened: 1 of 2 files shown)


packages/ketcher-macromolecules/src/components/monomerLibrary/RnaBuilder/RnaEditor/RnaEditorExpanded/RnaEditorExpanded.tsx
```diff
@@ -188,10 +188,34 @@ export const RnaEditorExpanded = ({
...
+  // Adjust state during render (instead of chaining a setState call inside
+  // the useEffect below): once the user picks a new monomer group item while
+  // editing a sequence selection, mark the selection as updated so the
+  // "Update" button becomes enabled.
+  if (
+    activeMonomerGroup !== RnaBuilderPresetsItem.Presets &&
+    isEditMode &&
+    isSequenceEditInRNABuilderMode &&
+    activePresetMonomerGroup?.groupItem &&
+    activePresetMonomerGroup.groupItem !== lastAppliedPresetMonomerGroupItem
```
dorogusha, line 211: **Correctness: this isn't equivalent to the old effect's trigger condition.**

The removed `useEffect` only re-ran when `activePresetMonomerGroup?.groupItem`, `isSequenceEditInRNABuilderMode`, or `selectedPhosphatePosition` changed — `isEditMode`/`activeMonomerGroup` were read inside it but were never dependencies. This new `if` block re-evaluates on **every** render and reads `isEditMode`/`activeMonomerGroup` directly (lines 207-208).

Neither `lastAppliedPresetMonomerGroupItem` nor `isSequenceSelectionUpdated` is reset by `resetRnaBuilder`/`resetRnaBuilderAfterSequenceUpdate` (they only touch Redux state), and `RnaEditorExpanded` doesn't unmount between Cancel/Edit cycles (see `RnaEditor.tsx`, mount is gated only by `expanded && activePreset`). So a concrete failure case: user edits a sequence selection, picks a monomer (`isSequenceSelectionUpdated` → true, `lastAppliedPresetMonomerGroupItem` → X), cancels. Later, in the same mounted instance, they re-enter edit mode on a *different* selection where `activePresetMonomerGroup.groupItem` still happens to differ from the stale `X` — this condition fires purely from `isEditMode` flipping false→true, with no monomer picked in the new session, enabling the "Update" button prematurely.

This contradicts the PR description's "no behavior change intended" / "exactly the same condition" claim. Suggest resetting these two local state values whenever edit mode is (re-)entered (or wiring them into `resetRnaBuilder`/`resetRnaBuilderAfterSequenceUpdate`), and adding a regression test that toggles edit mode off/on without re-picking a monomer to assert "Update" stays disabled.

packages/ketcher-macromolecules/src/components/monomerLibrary/RnaBuilder/RnaEditor/RnaEditorExpanded/RnaEditorExpanded.tsx
```diff
@@ -188,10 +188,34 @@ export const RnaEditorExpanded = ({
...
+  ) {
+    setLastAppliedPresetMonomerGroupItem(activePresetMonomerGroup.groupItem);
+    if (!isSequenceSelectionUpdated) {
+      setIsSequenceSelectionUpdated(true);
```
dorogusha, line 215: nit: `if (!isSequenceSelectionUpdated) { setIsSequenceSelectionUpdated(true); }` is equivalent to calling `setIsSequenceSelectionUpdated(true)` unconditionally — React bails out same-value `setState` calls via `Object.is`. Not a bug, just unnecessary.

dorogusha:
Reviewed with code-quality, performance, security, and documentation-accuracy passes.

- **Security & performance:** no issues. This is local UI state only, and the render-time `setState` pattern converges in one bounded extra render pass with no loop risk.
- **Approach:** removing the chained `setState` from the `useEffect` correctly fixes the `no-chain-state-updates` lint violation, and the general "adjust state during render" idiom matches how this rule is already handled elsewhere in the codebase.
- **Concern (see inline comment):** the replacement render-body condition reads `isEditMode`/`activeMonomerGroup` non-reactively on every render, which isn't equivalent to the old effect's dependency-gated trigger. This looks like it can prematurely enable the "Update" button in a real (if narrow) scenario, which contradicts the "no behavior change intended" claim in the description. No test was added to cover this path — worth adding a regression test alongside a fix.

nickkurkan:
Thanks for the thorough review — good catch.

Pushed a fix (9de91b1) that addresses the correctness issue: I added a `wasEditMode` render-time tracker so re-entering edit mode (false→true) is detected and resets `lastAppliedPresetMonomerGroupItem`/`isSequenceSelectionUpdated`, so stale state from a previous, cancelled edit session can no longer leak in and prematurely enable "Update". Also dropped the redundant `if (!isSequenceSelectionUpdated)` guard per the nit.

Added a regression test that reproduces the exact scenario from your comment (pick a monomer → Update enabled → Cancel → re-enter edit mode on a different selection without picking a new monomer → asserts Update stays disabled). Confirmed it fails without the fix and passes with it.

`tsc --noEmit` and the full `RnaEditorExpanded.test.tsx` suite (4/4) pass.

dorogusha (approved):

AlexeyGirin:

PR #11379, opened 2026-08-20
#11234 - Refactor: react-you-might-not-need-an-effect/no-chain-state-updates rule violation clean up from packages\ketcher-react\src\script\ui\views\components\MonomerCreationWizard\MonomerCreationWizard.tsx file

### LINKED ISSUE #11234: Refactor: `react-you-might-not-need-an-effect/no-chain-state-updates` rule violation clean up from `packages\ketcher-react\src\script\ui\views\components\MonomerCreationWizard\MonomerCreationWizard.tsx` file

Typical example:
```ts

useEffect(() => {
}, [firstName, lastName]);
```

**Why is this a problem?**
- An input state changes.

This can cause:
- unnecessary renders;
- briefly inconsistent UI;
- more complicated data flow;
- synchronization bugs;

**How is it usually fixed?**
```

```

```
  [items, query],
);
```

```
const handleReset = () => {
  setFirstName('');
  setLastName('');
};
```

```
```

**Problem locations:**
packages\ketcher-react\src\script\ui\views\components\MonomerCreationWizard\MonomerCreationWizard.tsx

nickkurkan:
## Summary
- Closes #11234.
- `MonomerCreationWizard.tsx` had a `useEffect` (recomputing RNA-preset atom-ownership validation highlights) whose only job, once the highlighted problematic atoms became empty, was to call `setHasActiveRnaPresetAtomValidationErrors(false)` — a value derived purely from other state (`rnaPresetComponentStructures`, the active-errors flag itself). This effect depended on the very state it wrote, producing a one-render-late "chain" update flagged by `react-you-might-not-need-an-effect/no-chain-state-updates`.
- Fix: the problematic-atom set is now computed inline via `useMemo` (`problematicAtomIds`). The flag reset is applied immediately during render ("adjust state during render" pattern) by comparing the freshly computed value against the previous one, instead of being deferred through a follow-up effect. A small remaining `useEffect` only syncs the memoized value to the external (non-React) editor render state via `editor.setProblematicAtoms`, which is a legitimate effect use (external-system sync), not a `setState` chain.
- No behavior change: highlights and the active-errors flag are still recomputed/cleared exactly when component structures change while validation errors are active.
- Note: two other pre-existing `no-event-handler` warnings in this file are unrelated and out of scope for this PR (tracked separately under #11243).

## Test plan
- `npx eslint src/script/ui/views/components/MonomerCreationWizard/MonomerCreationWizard.tsx` (from `packages/ketcher-react`): confirmed the `no-chain-state-updates` warning is gone.
- `npx tsc --noEmit -p tsconfig.json` (from `packages/ketcher-react`): no new errors introduced by this change (pre-existing unrelated errors remain, verified present without this change too).
- Confirmed no unit test file directly covers this component (only its pure `.utils` helpers, unaffected); the jest suite currently fails to boot in this sandbox for an unrelated environment reason (`jest-environment-jsdom`), verified to happen identically on `master` without this change.


Jesovile (commented):
This changes behavior when an RNA preset ownership error is corrected. If problematicAtomIds becomes empty, the render-phase update sets hasActiveRnaPresetAtomValidationErrors to false. 

React rerenders before committing, so problematicAtomIds becomes null, and the effect never calls editor.setProblematicAtoms(new Set()). Previously, that call cleared the red highlighting from corrected atoms.                                                                                                                     
                                                                                                                                                                                  Could we keep validation active until the next submit/reset and always synchronize the calculated set, including an empty set? This avoids chained effect state updates while   preserving the previous visible behavior.  

My llm provide this scenario of reproducing:
`
- submit an invalid Nucleotide preset with unassigned or multiply assigned atoms
- correct the component assignments
- observe that the old problematic-atom   highlighting remains until another submit/reset. 
- `
Could you please check it?

dorogusha (approved):

nickkurkan:
Good catch, thanks — confirmed the repro. Pushed a fix in 8d444596cf: `hasActiveRnaPresetAtomValidationErrors` is no longer cleared during render. It now only resets on submit/discard (`handleSubmit`/`handleDiscard`), so `problematicAtomIds` never collapses to `null` mid-correction, and every recompute — including one that resolves to an empty set — is synced to `editor.setProblematicAtoms` via the effect. This restores the pre-refactor clearing behavior while keeping the render-derived computation (no chained effect state updates).

AlexeyGirin:

Jesovile (approved):

rrodionov91 (approved):

AlexeyGirin:

PR #11380, opened 2026-08-20
#11242 - Refactor: react-you-might-not-need-an-effect/no-event-handler rule violation clean up from packages\ketcher-react\src\script\ui\views\components\MonomerCreationWizard\RnaPresetTabs.tsx file

### LINKED ISSUE #11242: Refactor: `react-you-might-not-need-an-effect/no-event-handler` rule violation clean up from `packages\ketcher-react\src\script\ui\views\components\MonomerCreationWizard\RnaPresetTabs.tsx` file

Typical example:
```ts

useEffect(() => {
  if (submitted) {
    sendForm();
  }
}, [submitted]);

const handleSubmit = () => {
  setSubmitted(true);
};
```

**Why is this a problem?**
- React renders again.

- an extra render;
- delayed execution;
- dependency-array problems;

**How is it usually fixed?**
```ts
const handleSubmit = () => {
  sendForm();
  setSubmitted(true);
};
```

```ts
const submitForm = () => {
  sendForm();
  setSubmitted(true);
};

```

**Problem locations:**
packages\ketcher-react\src\script\ui\views\components\MonomerCreationWizard\RnaPresetTabs.tsx

nickkurkan:
## Summary
Closes #11242.

- `react-you-might-not-need-an-effect/no-event-handler` flagged two effects in `RnaPresetTabs.tsx` that were largely reacting to `selectedTab`, a state value set only by the tab-click handler (`handleChange`) in this same file:
  - one effect re-applied canvas highlights and cleared the selection after a tab switch,
  - the other re-synced the visible/connection attachment points after a tab switch.
- Moved the tab-switch behavior directly into `handleChange` (and extracted `syncConnectionAttachmentPoints` as a reusable callback it now calls), so the tab-click event handles its own consequences instead of a redundant effect reacting to the resulting state.
- The remaining reactivity in both effects is now driven only by `wizardState`/`struct`/`phosphatePosition` (props that can also change asynchronously, independent of any click in this component — e.g. a component's structure gets marked via the Editor and flows back down through a window event handled by the parent), read via a `selectedTabRef` so the effects no longer list `selectedTab` as a dependency. This eliminates the effect-as-event-handler pattern while keeping the effects that must genuinely react to external/async updates.
- No observable behavior change: tab switching, highlighting, attachment-point syncing, and selection clearing all work identically (verified via the existing Jest suite's assertions on `handleChange`/tab-click behavior).

## Test plan
- [x] `npx eslint src/script/ui/views/components/MonomerCreationWizard/RnaPresetTabs.tsx` — no-event-handler warnings reduced from 3 to 1 (the remaining one is for an effect that genuinely reacts to an async prop update originating outside this component, which is scoped to a different, in-progress fix in the parent component).
- [x] `npx tsc --noEmit -p tsconfig.json` — no new errors (pre-existing unrelated errors only).
- [ ] `npx jest RnaPresetTabs.test.tsx` — could not run in this environment due to a pre-existing `jest-environment-jsdom`/`jsdom` version mismatch affecting all test files in this package, unrelated to this change.

### DIFF
--- a/packages/ketcher-react/src/script/ui/views/components/MonomerCreationWizard/RnaPresetTabs.tsx
+++ b/packages/ketcher-react/src/script/ui/views/components/MonomerCreationWizard/RnaPresetTabs.tsx
@@ -60,8 +60,8 @@ interface IRnaPresetTabsProps {
   wizardState: RnaPresetWizardState;
   editor: Editor;
   wizardStateDispatch: (action: RnaPresetWizardAction) => void;
-  phosphatePosition: '3' | '5' | undefined;
-  onPhosphatePositionChange: (position: '3' | '5') => void;
+  phosphatePosition: PhosphatePosition | undefined;
+  onPhosphatePositionChange: (position: PhosphatePosition) => void;
   /** User-overridden leaving atom labels for connection APs, keyed by
    * "<componentKey>:<apName>". Persists across tab switches. */
   connectionLeavingAtoms?: Map<string, AtomLabel>;
@@ -141,19 +141,19 @@ export const RnaPresetTabs = (props: IRnaPresetTabsProps) => {
       wizardState,
       struct,
       'base',
-      phosphatePosition as PhosphatePosition | undefined,
+      phosphatePosition,
     ),
     sugar: getConnectionAttachmentPointsForRnaPresetComponent(
       wizardState,
       struct,
       'sugar',
-      phosphatePosition as PhosphatePosition | undefined,
+      phosphatePosition,
     ),
     phosphate: getConnectionAttachmentPointsForRnaPresetComponent(
       wizardState,
       struct,
       'phosphate',
-      phosphatePosition as PhosphatePosition | undefined,
+      phosphatePosition,
     ),
   };
   const readonlyComponentAttachmentPoints = {
@@ -225,7 +225,7 @@ export const RnaPresetTabs = (props: IRnaPresetTabsProps) => {
         wizardState,
         struct,
         activeComponentKey,
-        phosphatePosition as PhosphatePosition | undefined,
+        phosphatePosition,
       );
       editor.setConnectionAttachmentPoints(connectionAtomIds);
     },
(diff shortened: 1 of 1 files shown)


Jesovile (approved):

packages/ketcher-react/src/script/ui/views/components/MonomerCreationWizard/RnaPresetTabs.tsx
```diff
@@ -201,9 +206,45 @@ export const RnaPresetTabs = (props: IRnaPresetTabsProps) => {
...
+      if (!activeComponentKey) {
+        editor.setConnectionAttachmentPoints(new Map());
+        return;
+      }
+
+      const connectionAtomIds = getConnectionAttachmentPointAtomIdsForComponent(
+        wizardState,
+        struct,
+        activeComponentKey,
+        phosphatePosition as PhosphatePosition | undefined,
```
dorogusha, line 228: Redundant type assertion

`phosphatePosition` is already typed as `'3' | '5' | undefined` on the props interface, and `PhosphatePosition` (from RnaPresetAttachmentPointValidation.ts) is defined as `'3' | '5'`. So `PhosphatePosition | undefined` is structurally identical to the prop's own type - TypeScript already accepts `phosphatePosition` here without any cast.

Suggest dropping the as `PhosphatePosition | undefined` and instead typing the prop with the shared type directly:

```
import { type PhosphatePosition, ... } from './RnaPresetAttachmentPointValidation';

interface IRnaPresetTabsProps {
  ...
  phosphatePosition: PhosphatePosition | undefined;
  onPhosphatePositionChange: (position: PhosphatePosition) => void;
  ...
}
```

dorogusha (changes requested):

packages/ketcher-react/src/script/ui/views/components/MonomerCreationWizard/RnaPresetTabs.tsx
nickkurkan: Good catch, done in 26220b0597.

dorogusha (approved):

PR #11381, opened 2026-08-20
#11243 - Refactor: react-you-might-not-need-an-effect/no-event-handler rule violation clean up from packages\ketcher-react\src\script\ui\views\components\MonomerCreationWizard\MonomerCreationWizard.tsx file

### LINKED ISSUE #11243: Refactor: `react-you-might-not-need-an-effect/no-event-handler` rule violation clean up from `packages\ketcher-react\src\script\ui\views\components\MonomerCreationWizard\MonomerCreationWizard.tsx` file

Typical example:
```ts

useEffect(() => {
  if (submitted) {
    sendForm();
  }
}, [submitted]);

const handleSubmit = () => {
  setSubmitted(true);
};
```

**Why is this a problem?**
- React renders again.

- an extra render;
- delayed execution;
- dependency-array problems;

**How is it usually fixed?**
```ts
const handleSubmit = () => {
  sendForm();
  setSubmitted(true);
};
```

```ts
const submitForm = () => {
  sendForm();
  setSubmitted(true);
};

```

**Problem locations:**
packages\ketcher-react\src\script\ui\views\components\MonomerCreationWizard\MonomerCreationWizard.tsx

nickkurkan:
## Summary
- Closes #11243.
- `MonomerCreationWizard.tsx` had two `useEffect`s that existed purely to react to a prop and immediately dispatch/setState, instead of running that logic where the prop's underlying value is actually decided:
  - An effect keyed on `monomerCreationState?.hasDefaultAttachmentPoints` whose only job was to add the `defaultAttachmentPoints` notification. This value is only ever known once, when the wizard mounts (it never changes afterwards), so it is now folded into the existing lazy `useReducer` initializer (`getInitialWizardStateForEdit`) — the same pattern already used in this file for `editInstanceInitialValues` — instead of being applied one render late via an effect.
  - An effect keyed on `monomerCreationState?.assignedAttachmentPoints` plus the sugar/phosphate structures, whose only job was to compute the auto-inferred phosphate position and call `handlePhosphatePositionChange`. The inferred value is now computed inline via `useMemo`, and the resulting state/notification updates are applied immediately during render when the derived value changes ("adjusting state during render", per the React docs), rather than deferred through an effect.
- This is independent from the PR for #11234 (`no-chain-state-updates`, https://github.com/epam/ketcher/pull/11379) — both touch this same file but different, non-overlapping effects, so a small rebase may be needed if both merge.
- Note: one remaining `no-event-handler` warning in this file (on the RNA-preset atom-ownership-highlight effect) is intertwined with the `no-chain-state-updates` violation fixed in #11234's PR and is left untouched here — it also genuinely reacts to an external, non-single-handler-driven system (canvas structure edits), matching the documented exception for keeping an effect.
- No behavior change: the notification and the auto-inferred phosphate position are still applied at the same points in time.

## Test plan
- `npx eslint src/script/ui/views/components/MonomerCreationWizard/MonomerCreationWizard.tsx` (from `packages/ketcher-react`): confirmed both fixed `no-event-handler` warnings are gone (the one remaining, unrelated warning is explained above).
- `npx tsc --noEmit -p tsconfig.json` (from `packages/ketcher-react`): no new errors introduced by this change.
- Confirmed no unit test file directly covers this component (only its pure `.utils` helpers, unaffected); the jest suite currently fails to boot in this sandbox for an unrelated environment reason (`jest-environment-jsdom`), verified to happen identically on `master` without this change.


dorogusha (commented):
I noticed, the `no-chain-state-updates` rule this PR fixes is already suppressed for the whole file by the disable comment on line 2. Since that suppression remains untouched after this change, the refactor doesn't actually resolve any lint/CI violation — the rule was never firing here to begin with.

packages/ketcher-react/src/script/ui/views/components/MonomerCreationWizard/MonomerCreationWizard.tsx, lines 1-4 (not part of this PR's diff - already on master before this change, added in #10806 "Refactor: raise ESLint to latest compatible versions across workspaces"):
 /* eslint-disable react-hooks/exhaustive-deps */
/* eslint-disable react-you-might-not-need-an-effect/no-event-handler, react-you-might-not-need-an-effect/no-chain-state-updates */
/* eslint-disable react-hooks/immutability */


 Could you please either:
1. Rebase master to your branch and remove `no-chain-state-updates` from that line-2 disable comment as part of this PR (and address/annotate any other violations it then surfaces in this file), so the fix is verifiably real, or
2. Update PR description and reconsider whether the added complexity (extra state variable, render-phase setState, extra effect) is worth it given there's no behavior change and no tooling benefit.

AlexeyGirin:

nickkurkan:
@dorogusha Good catch — confirmed, `no-chain-state-updates` was indeed dead weight in that disable line since day one of this PR.

Went with option 1: rebased onto master, removed `no-chain-state-updates` from the file-level disable comment (kept `no-event-handler`, which this PR still legitimately needs). With the blanket suppression narrowed, `--no-inline-config` surfaces exactly one real violation for that rule — the `setHasActiveRnaPresetAtomValidationErrors(false)` call inside the RNA-preset atom-ownership-highlight effect. That effect is the one already called out in the PR description as a legitimate exception (it synchronizes with canvas structure edits, not a single-handler-driven event), so I annotated it with a scoped `eslint-disable-next-line` + justification instead of refactoring it away, following the existing convention used elsewhere in the codebase (e.g. `usePortalStyle.ts`).

Confirmed `npx eslint MonomerCreationWizard.tsx` is clean and `tsc --noEmit` has no new errors. Pushed as 920a1397cd.

dorogusha (approved):

AlexeyGirin:

AlexeyGirin:

PR #11383, opened 2026-08-20
 #11225 - Refactor: Reading from or writing to ref.current during React render problem clean up from packages\ketcher-react\src\script\ui\component\form\colorPicker\ColorPicker.tsx file

### LINKED ISSUE #11225: Refactor: Reading from or writing to `ref.current` during React render problem clean up from `packages\ketcher-react\src\script\ui\component\form\colorPicker\ColorPicker.tsx` file
**Problem:**
Examples:
```ts
function Component() {
  const valueRef = useRef(0);

}
```

**Why is this an issue?**

**How can it be fixed?**
```ts
function Component() {

  useEffect(() => {
    inputRef.current?.focus();
  }, []);

}
```

For event-driven access:
```ts
const handleClick = () => {
  ref.current?.focus();
};
```

```ts

return (
    {count}
  </button>
);
```

**Problem locations:**
packages\ketcher-react\src\script\ui\component\form\colorPicker\ColorPicker.tsx

AnzhelaMartsinovich:
Related Issue - [11225](https://github.com/epam/ketcher/issues/11225)

Replaced the trigger element ref in `ColorPicker` with state populated through a callback ref.

The `Popover` now receives the trigger element without reading `ref.current` during render. The existing behavior remains unchanged.

## Check list
- [ ] unit-tests written
- [ ] e2e-tests written
- [ ] documentation updated


dorogusha (approved):

packages/ketcher-react/src/script/ui/component/form/colorPicker/ColorPicker.tsx
svvald: Not quite related to the problematic ref but still a worthy cleanup – `clickThrottleTimeoutRef` is never written which renders it and a effect below unnecessary

svvald (changes requested):

svvald (approved):

PR #11388, opened 2026-08-20
#11278 - Refactor: react-hooks/exhaustive-deps rule violation clean up from packages\ketcher-macromolecules\src\contexts\RootSizeContext.tsx file

### LINKED ISSUE #11278: Refactor: `react-hooks/exhaustive-deps` rule violation clean up from `packages\ketcher-macromolecules\src\contexts\RootSizeContext.tsx` file

Example:
```ts

useEffect(() => {
  loadUser(userId);
}, []); // userId is missing
```

**Why is this a problem?**

This can cause:
- stale state or props;
- effects using old data;

**How is it usually fixed?**
```ts
useEffect(() => {
  loadUser(userId);
}, [userId]);
```

For a memoized callback:
```ts
}, [documentId, content]);
```

For a memoized value:
```ts
  [items, category],
);
```

**Problem locations:**
packages\ketcher-macromolecules\src\contexts\RootSizeContext.tsx

HelenaIsh:

## Check list
- [ ] unit-tests written
- [ ] e2e-tests written
- [ ] documentation updated


packages/ketcher-macromolecules/src/contexts/RootSizeContext.tsx
```diff
@@ -32,23 +32,19 @@ export const RootSizeProvider = ({
...
-
+    const debouncedHandleResize = debounce(handleResize, 100);
     window.addEventListener('resize', debouncedHandleResize);
 
     return () => {
       window.removeEventListener('resize', debouncedHandleResize);
+      debouncedHandleResize.cancel();
     };
-  }, [debouncedHandleResize]);
+  }, [handleResize]);
```
svvald: I explored a bit and it seems that instead of silencing lint for the first effect we might get rid of it (as well as changing the second one) by using a more universal approach with ResizeObserver – you can just create an effect and attach the observer to the `rootRef` element, add a small debounce and `setSize` accordingly when. `rootRef` points to the EditorWrapper component so it will for on _each_ resize - not only window resize or macromolecules mode turn on but also for example on monomer library open/close

svvald (changes requested):

AlexeyGirin:

packages/ketcher-macromolecules/src/contexts/RootSizeContext.tsx
HelenaIsh: I tried to switch to ResizeObserver, but...
1. ResizeObserver seems not to cover macromolecules-mode switching: you can see how many visual regression we have in pipeline for [3c271d5](https://github.com/epam/ketcher/pull/11388/commits/3c271d5441b8c90874545fe2cfe32512c347bce0) commit.
2. In some test, where we change window size, we also have regression without  `window.addEventListener('resize', debouncedHandleResize);` see pipeline [0d29690](https://github.com/epam/ketcher/pull/11388/commits/0d2969024d8c0d2439b98f203843c37fddfd2164)
```
await browser.newContext({ deviceScaleFactor: 2.5 });
await page.setViewportSize({ width: 435, height: 291 });
```

So, my suggestion is to keep `window.addEventListener('resize', debouncedHandleResize);`

svvald (approved):

PR #11389, opened 2026-08-20
#11277 - Refactor: react-hooks/exhaustive-deps rule violation clean up from packages\ketcher-macromolecules\src\hooks\useIndigoVersionToRedux.ts file

### LINKED ISSUE #11277: Refactor: `react-hooks/exhaustive-deps` rule violation clean up from `packages\ketcher-macromolecules\src\hooks\useIndigoVersionToRedux.ts` file

Example:
```ts

useEffect(() => {
  loadUser(userId);
}, []); // userId is missing
```

**Why is this a problem?**

This can cause:
- stale state or props;
- effects using old data;

**How is it usually fixed?**
```ts
useEffect(() => {
  loadUser(userId);
}, [userId]);
```

For a memoized callback:
```ts
}, [documentId, content]);
```

For a memoized value:
```ts
  [items, category],
);
```

**Problem locations:**
packages\ketcher-macromolecules\src\hooks\useIndigoVersionToRedux.ts

HelenaIsh:

## Check list
- [ ] unit-tests written
- [ ] e2e-tests written
- [ ] documentation updated

### DIFF
--- a/packages/ketcher-macromolecules/src/hooks/useIndigoVersionToRedux.ts
+++ b/packages/ketcher-macromolecules/src/hooks/useIndigoVersionToRedux.ts
@@ -1,30 +1,28 @@
 import { useEffect } from 'react';
 import { IndigoProvider } from 'ketcher-react';
-import { useAppDispatch, useAppSelector } from './stateHooks';
-import { setAppMeta, selectAppMeta } from 'state/common/editorSlice';
+import { useAppDispatch } from './stateHooks';
+import { setIndigoVersion } from 'state/common/editorSlice';
 
 export function useIndigoVersionToRedux() {
   const dispatch = useAppDispatch();
-  const app = useAppSelector(selectAppMeta);
 
   useEffect(() => {
     async function fetchIndigoInfo() {
       const indigo = IndigoProvider.getIndigo();
-      if (indigo?.info) {
-        try {
-          const info = await indigo.info();
-          dispatch(
-            setAppMeta({
-              ...app,
-              indigoVersion: info.indigoVersion ?? '',
-            }),
-          );
-        } catch (_e) {
-          // ignore
-        }
+
+      if (!indigo?.info) {
+        return;
+      }
+
+      try {
+        const info = await indigo.info();
+
+        dispatch(setIndigoVersion(info.indigoVersion ?? ''));
+      } catch (_e) {
+        // ignore
       }
     }
+
     fetchIndigoInfo();
-    // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [dispatch]);
 }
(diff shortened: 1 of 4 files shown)


packages/ketcher-macromolecules/src/state/common/editorSlice.ts
```diff
@@ -230,6 +230,9 @@ export const editorSlice: Slice<EditorState> = createSlice({
     setAppMeta: (state, action: PayloadAction<AppMeta>) => {
       state.app = action.payload;
     },
```
svvald: This is a dead code now, could be removed

packages/ketcher-macromolecules/src/hooks/useIndigoVersionToRedux.ts
```diff
@@ -1,30 +1,28 @@
...
+      if (!indigo?.info) {
+        return;
+      }
+
+      try {
+        const info = await indigo.info();
+
+        dispatch(setIndigoVersion(info.indigoVersion ?? ''));
+      } catch (_e) {
+        // ignore
```
svvald: While we're here perhaps we could address this silent catch, the bare minimum is to log something via KetcherLogger

packages/ketcher-macromolecules/src/hooks/useIndigoVersionToRedux.ts
```diff
@@ -1,30 +1,28 @@
 import { useEffect } from 'react';
 import { IndigoProvider } from 'ketcher-react';
-import { useAppDispatch, useAppSelector } from './stateHooks';
-import { setAppMeta, selectAppMeta } from 'state/common/editorSlice';
+import { useAppDispatch } from './stateHooks';
+import { setIndigoVersion } from 'state/common/editorSlice';
 
 export function useIndigoVersionToRedux() {
```
svvald, line 7: Also I checked and it seems that this hook is called (and the version fetch is invoked) on each `About` window opening, perhaps we could change the call site to fetch this data on macro editor init?

svvald (changes requested):

packages/ketcher-macromolecules/src/state/common/editorSlice.ts
HelenaIsh: done

packages/ketcher-macromolecules/src/hooks/useIndigoVersionToRedux.ts
HelenaIsh: added KetcherLogger

packages/ketcher-macromolecules/src/hooks/useIndigoVersionToRedux.ts
HelenaIsh: moved hook call to EditorEvent

svvald (approved):

PR #11390, opened 2026-08-20
#11218 - Refactor: Use memo rule violation clean up from packages\ketcher-macromolecules\src\components\monomerLibrary\monomerLibraryGroup\MonomerGroup.tsx file

### LINKED ISSUE #11218: Refactor: Use memo rule violation clean up from `packages\ketcher-macromolecules\src\components\monomerLibrary\monomerLibraryGroup\MonomerGroup.tsx` file
**Problem:**

Example:
```ts
useMemo(() => {
}, []);
```

**Why is this an issue?**

**How can it be fixed?**
There are three common fixes.

Return the calculated value:
```ts
const total = useMemo(() => {
  return calculateTotal(items);
}, [items]);
```

```ts
const total = useMemo(
  [items],
);
```

```ts
useEffect(() => {
  logItems(items);
}, [items]);
```

```ts
```

**Problem locations:**
packages\ketcher-macromolecules\src\components\monomerLibrary\monomerLibraryGroup\MonomerGroup.tsx

nickkurkan:
Closes #11218

## How did you fix the issue?

`MonomerGroup.tsx` created its debounced preview handler with:

```ts
const debouncedShowPreview = useCallback(
  debounce((p) => dispatchShowPreview(p), 500),
  [dispatchShowPreview],
);
```

Here `debounce(...)` is invoked eagerly as an argument expression on every render (a side effect that allocates a new debounced closure), and `useCallback` was only memoizing the reference to whichever debounced instance happened to be created on the render where dependencies last changed — the wrong hook for deriving/caching this value.

This mirrors the exact anti-pattern already fixed for `RootSizeContext.tsx` in #11387, which introduced a reusable `useDebouncedCallback` hook (`packages/ketcher-macromolecules/src/hooks/useDebouncedCallback.ts`) that correctly wraps `debounce()` inside `useMemo`, returning the memoized debounced function and recomputing it only when its dependencies change.

I replaced the local `useCallback(debounce(...), [...])` construct in `MonomerGroup.tsx` with a call to that existing `useDebouncedCallback` hook, removing the now-unused `debounce` import. No behavioral change — same 500ms debounce is preserved.

## Testing

- `npx eslint` on the changed file — no errors.
- `npx tsc --noEmit` for the `ketcher-macromolecules` package — passes.
- Existing unit test `MonomerGroup.test.tsx` currently fails to run in this environment due to a pre-existing, unrelated issue (missing native `canvas` 2D context bindings for jsdom/paper.js); confirmed this failure is identical on unmodified `master`, so it is not caused by this change.


Jesovile (approved):

AlexeyGirin (approved):
Approved by Roman
Image

AlexeyGirin:

PR #11391, opened 2026-08-20
#11208 – Refactor: useMemo anti-patterns clean up in MacromoleculePropertiesWindow.tsx

### LINKED ISSUE #11208: Refactor: useMemo anti-patterns clean up in `packages\ketcher-macromolecules\src\components\macromoleculeProperties\MacromoleculePropertiesWindow.tsx` file
**Problem:**

**Why is this an issue?**

Problems:

- requests;
- modifying the DOM;
- writing to localStorage;
- subscriptions;
- calling setState;
- logging and analytics.

**How can it be fixed?**
Problematic code:
```ts
  doSomething();
}, [value]);
```

```ts
return items.map(transform);
}, [items]);
```

or shorter:

If you need a side effect:
```ts
useEffect(() => {
  doSomething();
}, [value]);
```

```ts
const handleClick = () => {
  doSomething();
};
```

```ts
  doSomething();
}, [value]);
```

**Problem locations:**

### LINKED ISSUE #11219: Refactor: Use memo rule violation clean up from `packages\ketcher-macromolecules\src\components\macromoleculeProperties\MacromoleculePropertiesWindow.tsx` file
**Problem:**

Example:
```ts
useMemo(() => {
}, []);
```

**Why is this an issue?**

**How can it be fixed?**
There are three common fixes.

Return the calculated value:
```ts
const total = useMemo(() => {
  return calculateTotal(items);
}, [items]);
```

```ts
const total = useMemo(
  [items],
);
```

```ts
useEffect(() => {
  logItems(items);
}, [items]);
```

```ts
```

**Problem locations:**

nickkurkan:
Closes #11208
Closes #11219

## How did you fix the issue?

`MacromoleculePropertiesWindow` used `useMemo` to derive two small JSX fragments: `grossFormula` and `molecularMassValue`. Both callbacks did return a value correctly, but the memoization itself was unnecessary — the computations are cheap (just building a couple of `<span>`/`<div>` elements from already-derived data), so wrapping them in `useMemo` added indirection without a real performance benefit, which is exactly the "useMemo used where it isn't needed" anti-pattern flagged by both issues.

Fix:
- Replaced both `useMemo` calls with plain inline conditional expressions computed directly in the component body (same null-check + JSX as before, no `useEffect`/side effects involved — this was never a side-effect case, just an unneeded memoization).
- Removed the now-unused `useMemo` import.

No other `useMemo` usages exist in this file (the only two instances are the ones changed here).

Behavior is unchanged — this is a pure refactor per the `refactor` label on both issues.

## Testing

- `eslint` on the changed file: same 3 pre-existing `react-hooks/exhaustive-deps` warnings (unrelated code), 0 new warnings/errors.
- `tsc --noEmit` for `ketcher-macromolecules`: clean.
- `prettier --check` on the changed file: passes.
- No unit test suite exists for this component; the relevant E2E coverage lives in `ketcher-autotests/tests/specs/Macromolecule-editor/Calculate-Properties/calculate-properties.spec.ts`, which exercises the gross formula and molecular mass display this change touches — not re-run in this environment (requires a full standalone build + Playwright), but the change is a like-for-like value computation with no behavioral difference.

## Check list
- [x] unit-tests written (N/A — no unit test suite exists for this component)
- [ ] e2e-tests written (N/A — existing suite already covers this behavior; not re-run in this environment)
- [ ] documentation updated


Jesovile (approved):

AlexeyGirin:

Issue #11251, opened 2026-08-15
Refactor: `react-you-might-not-need-an-effect/no-event-handler` rule violation clean up from `packages\ketcher-react\src\Editor.tsx` file


AlexeyGirin:
**What problem does `react-you-might-not-need-an-effect/no-event-handler` detect?**
It detects event-specific logic that is placed inside a useEffect instead of being executed directly in the event handler that caused the action.

Typical example:
```ts
const [submitted, setSubmitted] = useState(false);

useEffect(() => {
  if (submitted) {
    sendForm();
  }
}, [submitted]);

const handleSubmit = () => {
  setSubmitted(true);
};
```

The effect is being used to react to a user event indirectly through state.

**Why is this a problem?**
This creates an unnecessary chain:
- The user performs an action.
- The event handler updates state.
- React renders again.
- The effect runs after rendering.
- The effect performs the action.

The action could usually happen directly in the event handler. Using an effect can lead to:
- an extra render;
- delayed execution;
- more complicated control flow;
- duplicated or unexpected side effects;
- unnecessary state variables;
- dependency-array problems;
- difficulty understanding what caused the operation.

Effects are primarily intended for synchronizing React with external systems, not for handling user interactions.

**How is it usually fixed?**
Move the event-specific logic directly into the event handler:
```ts
const handleSubmit = () => {
  sendForm();
  setSubmitted(true);
};
```

If the logic is shared, extract it into a regular function:
```ts
const submitForm = () => {
  sendForm();
  setSubmitted(true);
};

return <form onSubmit={submitForm}>...</form>;
```

An effect may still be appropriate when the behavior must happen in response to an external system or when it should run after every relevant render regardless of which event caused it.

The general principle is: actions caused by user interactions belong in event handlers; effects belong to synchronization with external systems.

**Problem locations:**
packages\ketcher-react\src\Editor.tsx

PR #11393, opened 2026-08-20
#11274 - Refactor: react-hooks/exhaustive-deps rule violation clean up from packages\ketcher-react\src\Editor.tsx file

### LINKED ISSUE #11274: Refactor: `react-hooks/exhaustive-deps` rule violation clean up from `packages\ketcher-react\src\Editor.tsx` file

Example:
```ts

useEffect(() => {
  loadUser(userId);
}, []); // userId is missing
```

**Why is this a problem?**

This can cause:
- stale state or props;
- effects using old data;

**How is it usually fixed?**
```ts
useEffect(() => {
  loadUser(userId);
}, [userId]);
```

For a memoized callback:
```ts
}, [documentId, content]);
```

For a memoized value:
```ts
  [items, category],
);
```

**Problem locations:**
packages\ketcher-react\src\Editor.tsx

skorphil:
This PR resolves ESLint `react-hooks/exhaustive-deps` violations in `packages/ketcher-react/src/Editor.tsx` to ensure hook reliability and prevent stale closure issues.

**Key Changes:**
- **Exhaustive Dependencies**: Added all missing reactive values to the dependency arrays of `useEffect` hooks for mode switching and initialization.
- **onInit Guard**: Implemented a `useRef` based guard to ensure the `onInit` callback is executed exactly once per `ketcher.id`, even if the prop identity changes. This satisfies the exhaustive-deps rule for `onInit` without causing duplicate triggers.
- **Stability**: Wrapped `togglePolymerEditor` in `useCallback` to maintain identity across renders.
- **Type Safety**: Removed unsupported props (`disableMacromoleculesEditor`, `monomersLibraryUpdate`, `monomersLibraryReplace`) from the `MicromoleculesEditor` mount to fix type-level violations.
- **Compliance**: Added scoped ESLint disables for `react-you-might-not-need-an-effect` with comments justifying the effect-based approach due to current architectural constraints (lazy-loading and cross-package initialization).

Closes #11274

## Check list
- [ ] unit-tests written
- [ ] e2e-tests written
- [x] documentation updated


skorphil:
@code-quality-reviewer @security-code-reviewer Please review this PR which addresses react-hooks/exhaustive-deps violations in Editor.tsx.

packages/ketcher-react/src/Editor.tsx
```diff
@@ -113,19 +149,35 @@ export const Editor = (props: Props) => {
...
       ketcher &&
       moleculesEditor &&
-      (macromoleculesEditor || props.disableMacromoleculesEditor)
+      (macromoleculesEditor || disableMacromoleculesEditor)
     ) {
-      if (ketcherProvider.getIndexById(ketcher.id) !== -1) {
-        props.onInit?.(ketcher);
+      if (
+        ketcherProvider.getIndexById(ketcher.id) !== -1 &&
+        calledForIdRef.current !== ketcher.id
```
NataliaLoginova: Could we verify that a new Ketcher instance can never be initialized with the same ketcher.id while this Editor component remains mounted?

NataliaLoginova (approved):

packages/ketcher-react/src/Editor.tsx
skorphil: As far as I understood, id generated by lodash.uniqueId() which just a counter and it technically can't be non-unique in a single session. https://github.com/lodash/lodash/blob/4.18.1/lodash.js#L16317 Only if app were reloaded, but in this case Editor.tsx will be remounted

However, we can implement additional guard by comparing current `ketcher` object reference with a new one. I guess. This way it will be replaced with new one even if it has same id

NataliaLoginova: Let''s implement this solution:

However, we can implement additional guard by comparing current ketcher object reference with a new one. I guess. This way it will be replaced with new one even if it has same id

NataliaLoginova (changes requested):

skorphil:
I have updated the implementation in `Editor.tsx` to use **object identity comparison** (`initializedKetcherRef.current !== ketcher`) instead of checking the ID string.

This provides an additional defensive layer: even if two different Ketcher instances were to somehow share the same ID string, the component will recognize the new object reference and correctly re-initialize, keeping the consumer synchronized with the `ketcherProvider`.

packages/ketcher-react/src/Editor.tsx
skorphil: > Let''s implement this solution:
> 
> However, we can implement additional guard by comparing current ketcher object reference with a new one. I guess. This way it will be replaced with new one even if it has same id

Implemented check by reference in latest commit

skorphil:
Claude Risk Assessment: Removing `calledForIdRef.current !== ketcher.id Check`

**Current State**
- ✅ Already removed from implementation
- Only using: initializedKetcherRef.current !== ketcher (object identity)

**Risks of Removal**
🟢 LOW RISK - Safe to Remove
Why:
1. Object identity is stricter - If initializedKetcherRef.current !== ketcher, they're different objects, regardless of ID
2. lodash.uniqueId() guarantees uniqueness - IDs won't collide in a session
3. Edge case handled correctly - If IDs somehow collided, object identity would still detect the new instance and correctly re-initialize
Only scenario where ID check differs from identity check:
- New Ketcher object created with same ID as previous one
- Identity check: ✅ Detects different object → calls onInit (CORRECT)
- ID check: ❌ Sees same ID → skips onInit (WRONG - would miss new instance)

**Verdict**
✅ Safe to remove - Object identity check is superior and handles all cases correctly, including the theoretical ID collision case where the ID check would fail.

skorphil (commented):
Implemented Object reference check.

AI risk assessment proved that check by ID can be safely omitted if we replacing it with reference check

NataliaLoginova (approved):

skorphil:
Merged updated CI to re-test playwright. No ticket code was changed

la-nuta (approved):
Verified the post-merge shape locally: ESLint clean (no unused directives), tsc clean, and a runtime comparison against master (init/onInit/mode-toggle traces) shows identical behaviour.

packages/ketcher-react/src/Editor.tsx
```diff
@@ -95,16 +111,34 @@ export const Editor = (props: Props) => {
...
-  }, [macromoleculesEditor]);
+  }, [macromoleculesEditor, togglePolymerEditor]);
 
   useEffect(() => {
     return () => {
       window.isPolymerEditorTurnedOn = false;
     };
   }, []);
 
+  /* eslint-disable react-you-might-not-need-an-effect/no-event-handler */
```
la-nuta: Non-blocking, and strictly #11251's rule rather than this ticket's, but since this PR is what turned the file-level `no-event-handler` header into two block disables: here is a shape that needs neither the disable nor the guard refs. Everything this effect reacts to is set in `togglePolymerEditor`, so the switch can happen there:

```tsx
import { flushSync } from 'react-dom';
// ...
  const togglePolymerEditor = useCallback(
    (toggleValue: boolean) => {
      if (toggleValue === showPolymerEditor) {
        return;
      }

      // Commit the display flip before switching: switchToMacromolecules()
      // measures the macro canvas and focusCliparea() focuses the micro editor,
      // and both are display:none until this state is rendered.
      flushSync(() => {
        setShowPolymerEditor(toggleValue);
      });
      window.isPolymerEditorTurnedOn = toggleValue;

      if (!moleculesEditor || !macromoleculesEditor) {
        return;
      }

      if (toggleValue) {
        moleculesEditor.closeMonomerCreationWizard?.();
        macromoleculesEditor.switchToMacromolecules();
      } else {
        macromoleculesEditor.switchToMicromolecules();
        moleculesEditor.focusCliparea();
      }
    },
    [showPolymerEditor, moleculesEditor, macromoleculesEditor],
  );
```

This effect, `prevShowPolymerEditorRef` and `editorsReadyRef` then go away. The `toggleValue === showPolymerEditor` early return is what master's `setState` bail-out gave for free: a repeated `ketcher.switchToMoleculesMode()` must not re-run `switchToMicromolecules()`.

`flushSync` is load-bearing. I ran both builds in `example` (standalone): with it, `switchToMacromolecules()` measures the canvas at 872×644 and `focusCliparea()` lands on `.cliparea`; without it the canvas is measured at 0×0 while still `display:none` and focus stays on `<body>`. ESLint (`--report-unused-disable-directives`) and `tsc` are clean on this variant.

One trade-off: React 19's `flushSync` does not flush (and logs an error) when called from inside a render or commit, which includes passive effects, so a consumer calling `ketcher.switchToMacromoleculesMode()` from their own `useEffect` would get the hidden-canvas behaviour. If that matters, leaving this to #11251 is fine.

packages/ketcher-react/src/Editor.tsx
```diff
@@ -113,19 +147,39 @@ export const Editor = (props: Props) => {
         moleculesEditor?.focusCliparea();
       }
     }
-  }, [showPolymerEditor]);
+  }, [showPolymerEditor, moleculesEditor, macromoleculesEditor]);
+  /* eslint-enable react-you-might-not-need-an-effect/no-event-handler */
 
+  /* eslint-disable react-you-might-not-need-an-effect/no-event-handler */
```
la-nuta: Also #11251's. I don't think this one can move into the handlers without changing behaviour: `onInitMoleculesEditor` / `onInitMacromoleculesEditor` are captured once (`MicromoleculesEditor` in its `[]` init effect, the macro `Editor` via `createEditor` under `[dispatch]`), so neither can see the other editor's state, and `onInit` would become the mount-time prop instead of the current one. Merging two independently-arriving async sources is what an effect is for, so the block disable is honest here. Nothing to change.

packages/ketcher-react/src/Editor.tsx
```diff
@@ -49,19 +54,30 @@ const MacromoleculesEditorComponent = lazy(
 >;
 
 export const Editor = (props: Props) => {
+  const {
+    onInit,
+    disableMacromoleculesEditor,
+    monomersLibraryUpdate,
+    monomersLibraryReplace,
+    ...restProps
+  } = props;
```
la-nuta: Non-blocking. Pulling `onInit` and `disableMacromoleculesEditor` out is exactly what the rule needs — it was asking for the whole `props` object. But `monomersLibraryUpdate` and `monomersLibraryReplace` come out too, so `{...restProps}` on line 229 stops passing all three to `MicromoleculesEditor`, and from there to `init({ ...props, element, appRoot })`.

That's harmless in practice — none of the three has a reader in `ketcher-react/src` outside this file, so they were only ever dead keys on the config object. But the PR description gives the reason as fixing type-level violations, and there aren't any: `npx tsc --noEmit -p packages/ketcher-react/tsconfig.json` is clean on master.

Keeping the spread identical to master is a smaller diff and takes the question off the table:

```tsx
  const { onInit, disableMacromoleculesEditor } = props;
```

with line 229 back to `{...props}` and lines 214-215 back to `props.monomersLibraryUpdate` / `props.monomersLibraryReplace`. I built that variant and ran it: ESLint clean with `--report-unused-disable-directives`, `tsc --noEmit` clean against a clean baseline, still 0 `exhaustive-deps` directives.

(No suggestion block since the edit spans three separate places and a partial apply wouldn't compile.)

packages/ketcher-react/src/Editor.tsx
la-nuta: Non-blocking, and strictly #11251's rule rather than this ticket's — but this PR deletes master's file-level `/* eslint-disable react-you-might-not-need-an-effect/no-event-handler */` and replaces it with these two block pairs (122/151 and 153/182). Both forms are load-bearing: dropping either re-raises the errors, so nothing is actually closed, the suppression just moves into another ticket's territory.

Two side effects worth knowing about: the justification comments pre-empt #11251's analysis, and the widened deps take the file from 6 `no-event-handler` violations to 8 (new ones on `showPolymerEditor` and on `ketcher`), so #11251 gets bigger rather than smaller.

Simplest is to leave master's line-2 header where it was and drop these four block directives — that shape lints and typechecks clean too, and keeps this PR inside its own ticket. Happy either way if you'd rather keep the narrower scoping.

packages/ketcher-react/src/Editor.tsx
la-nuta: Non-blocking. Pulling `onInit` and `disableMacromoleculesEditor` out is what the rule needs — it was asking for the whole `props` object. But `monomersLibraryUpdate` and `monomersLibraryReplace` come out too, so `{...restProps}` on line 229 stops passing all three to `MicromoleculesEditor`, and from there into `init({ ...props, element, appRoot })`.

No behaviour rides on it — none of the three has a reader in `ketcher-react/src` outside this file. But the PR description gives the reason as fixing type-level violations, and there aren't any: `npx tsc --noEmit -p packages/ketcher-react/tsconfig.json` is clean on master.

Destructuring only what the hooks use keeps the spread identical to master:

```tsx
  const { onInit, disableMacromoleculesEditor } = props;
```

with line 229 back to `{...props}` and lines 214-215 back to `props.monomersLibraryUpdate` / `props.monomersLibraryReplace`. I ran that variant: ESLint and `tsc --noEmit` clean, still 0 `exhaustive-deps` directives.

PR #11397, opened 2026-08-21
#11273 - Refactor: react-hooks/exhaustive-deps rule violation clean up from packages\ketcher-react\src\MicromoleculesEditor.tsx file  

### LINKED ISSUE #11273: Refactor: `react-hooks/exhaustive-deps` rule violation clean up from `packages\ketcher-react\src\MicromoleculesEditor.tsx` file

Example:
```ts

useEffect(() => {
  loadUser(userId);
}, []); // userId is missing
```

**Why is this a problem?**

This can cause:
- stale state or props;
- effects using old data;

**How is it usually fixed?**
```ts
useEffect(() => {
  loadUser(userId);
}, [userId]);
```

For a memoized callback:
```ts
}, [documentId, content]);
```

For a memoized value:
```ts
  [items, category],
);
```

**Problem locations:**
packages\ketcher-react\src\MicromoleculesEditor.tsx

LashaDevLasha:

## Check list
- [ ] unit-tests written
- [ ] e2e-tests written
- [ ] documentation updated

### DIFF
--- a/packages/ketcher-react/src/MicromoleculesEditor.tsx
+++ b/packages/ketcher-react/src/MicromoleculesEditor.tsx
@@ -1,4 +1,3 @@
-/* eslint-disable react-hooks/exhaustive-deps */
 /* eslint-disable react-you-might-not-need-an-effect/no-event-handler */
 /****************************************************************************
  * Copyright 2021 EPAM Systems
@@ -55,6 +54,7 @@ function MicromoleculesEditor(props: Readonly<EditorProps>) {
 
   const setServerRef = useRef<(structService: StructService) => void>(() => {});
   const structServiceProvider = props.structServiceProvider;
+  const ketcherId = props.ketcherId;
 
   const rootElRef = useRef<HTMLDivElement>(null);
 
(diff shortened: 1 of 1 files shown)


packages/ketcher-react/src/MicromoleculesEditor.tsx
```diff
@@ -63,39 +63,50 @@ function MicromoleculesEditor(props: Readonly<EditorProps>) {
...
+      ketcherId,
+      structServiceProvider,
       setServerRef.current,
     );
+    // ketcherId is intentionally omitted: after mount, this effect only re-runs when
+    // structServiceProvider changes. The mount run is intentionally skipped via the
+    // ketcherId guard above, since the parent sets ketcherId via the onSetKetcherId
+    // callback after initKetcher completes. Making ketcherId reactive would cause an
+    // extra reinitializeApi call at that point, before the builder is ready for it.
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
NataliaLoginova, line 77: Improve comment. When the ketcherId change comes back via props and triggers a new render/effect, ketcherBuilderRef.current should already contain the builder.

// ketcherId is intentionally omitted because changing it after initialization
// should not trigger reinitializeApi. The API should only be reinitialized when
// structServiceProvider changes.

NataliaLoginova (changes requested):

packages/ketcher-react/src/MicromoleculesEditor.tsx
LashaDevLasha: updated

NataliaLoginova (approved):

la-nuta (approved):
Verified the post-merge shape locally: ESLint clean (no unused directives), tsc clean, and a runtime comparison against master (init/onInit/mode-toggle traces) shows identical behaviour.

packages/ketcher-react/src/MicromoleculesEditor.tsx
```diff
@@ -111,6 +120,7 @@ function MicromoleculesEditor(props: Readonly<EditorProps>) {
       });
     };
     // TODO: provide the list of dependencies after implementing unsubscribe function
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
AlexeyGirin, line 123: Please eliminate that eslint-disable

AlexeyGirin (changes requested):

PR #11401, opened 2026-08-21
#11270 - Refactor: react-hooks/exhaustive-deps rule violation clean up from packages\ketcher-react\src\script\ui\App\App.tsx file  

### LINKED ISSUE #11270: Refactor: `react-hooks/exhaustive-deps` rule violation clean up from `packages\ketcher-react\src\script\ui\App\App.tsx` file

Example:
```ts

useEffect(() => {
  loadUser(userId);
}, []); // userId is missing
```

**Why is this a problem?**

This can cause:
- stale state or props;
- effects using old data;

**How is it usually fixed?**
```ts
useEffect(() => {
  loadUser(userId);
}, [userId]);
```

For a memoized callback:
```ts
}, [documentId, content]);
```

For a memoized value:
```ts
  [items, category],
);
```

**Problem locations:**
packages\ketcher-react\src\script\ui\App\App.tsx

LashaDevLasha:

## Check list
- [ ] unit-tests written
- [ ] e2e-tests written
- [ ] documentation updated

### DIFF
--- a/ketcher-autotests/tests/specs/Macromolecule-editor/Sequence-Mode/sequence-mode-edit.spec.ts
+++ b/ketcher-autotests/tests/specs/Macromolecule-editor/Sequence-Mode/sequence-mode-edit.spec.ts
@@ -46,7 +46,6 @@ import { SequenceSymbolOption } from '@tests/pages/constants/contextMenu/Constan
 import { MacromoleculesTopToolbar } from '@tests/pages/macromolecules/MacromoleculesTopToolbar';
 import { LayoutMode } from '@tests/pages/constants/macromoleculesTopToolbar/Constants';
 import { MonomerPreviewTooltip } from '@tests/pages/macromolecules/canvas/MonomerPreviewTooltip';
-import { CommonTopRightToolbar } from '@tests/pages/common/CommonTopRightToolbar';
 
 async function hoverMouseOverMonomer(page: Page, monomer: Monomer, nth = 0) {
   await CommonLeftToolbar(page).bondTool(MacroBondTool.Single);
--- a/packages/ketcher-react/src/script/ui/App/App.tsx
+++ b/packages/ketcher-react/src/script/ui/App/App.tsx
@@ -1,4 +1,3 @@
-/* eslint-disable react-hooks/exhaustive-deps */
 /****************************************************************************
  * Copyright 2021 EPAM Systems
  *
@@ -95,6 +94,7 @@ const App = (props: Props) => {
       // Ideally is to remove ketcher instance in cleanup function of the most parent component (MicromoleculesEditor, or Editor, depends on usage)
       ketcherProvider.removeKetcherInstance(ketcherId);
     };
+    // eslint-disable-next-line react-hooks/exhaustive-deps
   }, []);
 
   // Temporary workaround: add proper types for Editor
--- a/packages/ketcher-react/src/script/ui/App/App.tsx
+++ b/packages/ketcher-react/src/script/ui/App/App.tsx
@@ -94,8 +94,7 @@ const App = (props: Props) => {
       // Ideally is to remove ketcher instance in cleanup function of the most parent component (MicromoleculesEditor, or Editor, depends on usage)
       ketcherProvider.removeKetcherInstance(ketcherId);
     };
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, []);
+  }, [checkServer, dispatch, ketcherId]);
 
   // Temporary workaround: add proper types for Editor
   const Editor = ConnectedEditor as React.ComponentType<{


packages/ketcher-react/src/script/ui/App/App.tsx
```diff
@@ -95,6 +94,7 @@ const App = (props: Props) => {
       // Ideally is to remove ketcher instance in cleanup function of the most parent component (MicromoleculesEditor, or Editor, depends on usage)
       ketcherProvider.removeKetcherInstance(ketcherId);
     };
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
la-nuta: Narrowing the file-wide disable is the right move, and suppressing beats adding the deps here: `dispatch` is stable, `checkServer` is a bound action creator from `connect`, and `ketcherId` is fixed for this App tree's lifetime. Listing them would make a mount/unmount effect re-runnable — and its cleanup calls `removeKetcherInstance` and re-dispatches `initLib([])`.

The issue asks that suppression be "a deliberate decision", so please record that in a comment above the disable.

ketcher-autotests/tests/specs/Macromolecule-editor/Sequence-Mode/sequence-mode-edit.spec.ts
```diff
@@ -46,7 +46,6 @@ import { SequenceSymbolOption } from '@tests/pages/constants/contextMenu/Constan
 import { MacromoleculesTopToolbar } from '@tests/pages/macromolecules/MacromoleculesTopToolbar';
 import { LayoutMode } from '@tests/pages/constants/macromoleculesTopToolbar/Constants';
 import { MonomerPreviewTooltip } from '@tests/pages/macromolecules/canvas/MonomerPreviewTooltip';
-import { CommonTopRightToolbar } from '@tests/pages/common/CommonTopRightToolbar';
```
la-nuta: For other reviewers: dead import, not a weakened test — zero usages in this 1633-line spec, added without one by #10806 (`e16cf3ee`) the day before. `noUnusedLocals: true` makes it a real `tsc --noEmit` error, so the removal is warranted. Please mention this drive-by fix in the description.

packages/ketcher-react/src/script/ui/App/App.tsx
la-nuta, line 97: Suppression isn't needed here — the three deps are all stable identities:

- `dispatch` — `useAppDispatch()`, stable by Redux contract
- `checkServer` — bound via `connect(null, { checkServer })` object-form `mapDispatchToProps` (`App.container.ts:6`); react-redux memoizes these
- `ketcherId` — a parameter of `initApp` (`initApp.tsx:35`) passed into `AppContext.Provider` (`:72-74`), so it's a closure constant for the whole tree

So the effect still runs exactly once if you list them:

```suggestion
  }, [checkServer, dispatch, ketcherId]);
```

(delete the `eslint-disable-next-line` above it as well). I linted this variant against the repo config — clean, no disable required.

This is also the safer form: under the suppression, if `ketcherId` ever becomes dynamic the effect silently holds a stale id, while the cleanup calls `ketcherProvider.removeKetcherInstance(ketcherId)`. With real deps it re-initialises correctly.

la-nuta (commented):
Two things to change before this lands.

**1. Use the real dependencies instead of suppressing.** All three values the rule wants are stable identities, so listing them is behaviour-identical and removes the need for a disable comment entirely — see the inline note.

**2. Mention the `sequence-mode-edit.spec.ts` import removal in the PR description.** A one-line diff in an unrelated package reads as churn without a note explaining it's a dead import left by #10806.

packages/ketcher-react/src/script/ui/App/App.tsx
LashaDevLasha: Done — replaced the suppression with the real deps [checkServer, dispatch, ketcherId].

Jesovile (approved):

AlexeyGirin:
Not approved by Roman
изображение

la-nuta (approved):
Thanks for switching to real dependencies — the file now carries no suppressions at all, which is a better outcome than the targeted disable I originally suggested.

Verified before approving: the head lints clean under the repo config with no disables in the file, and a positive control (dropping `ketcherId` from the array) reproduces `97:6 error React Hook useEffect has a missing dependency: 'ketcherId'`, so the rule is genuinely running and satisfied rather than absent. All three values are stable identities — `dispatch` from `useAppDispatch()`, `checkServer` from `connect`'s object-form `mapDispatchToProps`, and `ketcherId` a closure constant from `initApp` — so the effect still runs exactly once and the mount/unmount behaviour is unchanged. Master hasn't touched `App.tsx` since #10806, and the merged shape differs from master by exactly the dependency array.

Approving. One non-blocking request: the description is currently just the bare template, so the `sequence-mode-edit.spec.ts` import removal has no explanation attached — worth a line noting it's a dead import left by #10806 (`e16cf3ee`) that breaks `tsc --noEmit` under `noUnusedLocals`, so it doesn't look like unrelated churn in master's history after squash-merge.

PR #11403, opened 2026-08-21
#11250 - Refactor: react-you-might-not-need-an-effect/no-event-handler rule violation clean up from packages\ketcher-react\src\MicromoleculesEditor.tsx file

### LINKED ISSUE #11250: Refactor: `react-you-might-not-need-an-effect/no-event-handler` rule violation clean up from `packages\ketcher-react\src\MicromoleculesEditor.tsx` file

Typical example:
```ts

useEffect(() => {
  if (submitted) {
    sendForm();
  }
}, [submitted]);

const handleSubmit = () => {
  setSubmitted(true);
};
```

**Why is this a problem?**
- React renders again.

- an extra render;
- delayed execution;
- dependency-array problems;

**How is it usually fixed?**
```ts
const handleSubmit = () => {
  sendForm();
  setSubmitted(true);
};
```

```ts
const submitForm = () => {
  sendForm();
  setSubmitted(true);
};

```

**Problem locations:**
packages\ketcher-react\src\MicromoleculesEditor.tsx

Karen1704:
…olation clean up from

## Check list
- [ ] unit-tests written
- [ ] e2e-tests written
- [ ] documentation updated


packages/ketcher-react/src/MicromoleculesEditor.tsx
```diff
@@ -63,12 +62,15 @@ function MicromoleculesEditor(props: Readonly<EditorProps>) {
   });
 
   useEffect(() => {
-    if (!props.ketcherId) {
+    // This effect synchronizes the editor with an external service provider;
+    // it is not triggered by a user event.
+    // eslint-disable-next-line react-you-might-not-need-an-effect/no-event-handler
```
Jesovile: Seems that we create the same issue :-( 
Why can't we add ketcherId in deps?

Karen1704:
I’ve updated the file and ran the tests locally,  all tests are passing now

PR #11410, opened 2026-08-21
#11215 - Refactor: Preserve manual memoization rule violation problem clean up from packages\ketcher-macromolecules\src\components\preview\components\BondPreview\BondPreview.tsx file

### LINKED ISSUE #11215: Refactor: Preserve manual memoization rule violation problem clean up from `packages\ketcher-macromolecules\src\components\preview\components\BondPreview\BondPreview.tsx` file
**Problem:**
Manually added memoization, such as `useMemo` and `useCallback`, can not be safely preserved by the React Compiler.
Code where the compiler may be unable to preserve the existing memoization semantics. Typical causes include:
- a dependency array that does not match the values used by the memoized function;
- patterns that make the memoized value unsafe or impossible to analyze;
- mutations or other operations that undermine the assumptions of memoization;
- manually memoized code that is incompatible with compiler optimizations.

**Why is this an issue?**
Memoization is often used to preserve referential identity:
```ts
const value = useMemo(() => createValue(data), [data]);
```

Other code may rely on value remaining the same object while data has not changed. For example:
- React.memo may use the reference to skip rendering;
- an effect may use reference equality to decide whether to run;
- a child component may depend on a stable callback;
- caches or subscriptions may depend on stable values.

If manual memoization is lost or changed unexpectedly, it can cause:
- unnecessary renders;
- effects or subscriptions running more often;
- broken performance assumptions;
- behavior changes caused by unstable object or function references.

At the same time, incorrect memoization can also produce stale values if dependencies are incomplete. The rule helps ensure that manually optimized code remains valid when React Compiler processes it.

**How can it be fixed?**
The usual approach is to make the memoized code compiler-compatible:
- provide a complete and correct dependency array;
- avoid mutating values created by useMemo;
- keep the callback or calculation pure;
- avoid complex dynamic patterns inside useMemo and useCallback;
- move calculations into a simple pure function when appropriate;
- remove unnecessary manual memoization if the React Compiler can handle it automatically.

For example:
```ts
const handleSelect = useCallback(() => {
  selectItem(selectedId);
}, [selectedId]);
```

Instead of:
```ts
const handleSelect = useCallback(() => {
  selectItem(selectedId);
}, []);
```

If memoization is not actually needed, removing it is often the cleanest fix:
```ts
const handleSelect = () => {
  selectItem(selectedId);
};
```

Disabling the rule should generally be reserved for a deliberate case where the pattern is known to be correct but cannot be understood by the compiler.

**Problem locations:**
packages\ketcher-macromolecules\src\components\preview\components\BondPreview\BondPreview.tsx

nickkurkan:
## Summary
Closes #11215

`BondPreview.tsx` used `useMemo` to build a brand new emotion `styled(Container)` component on every `style` change, then rendered that dynamically created component type as the JSX root tag (`<ContainerDynamic>`). Creating a component definition inside `useMemo` and returning it as a value used for JSX is exactly the kind of "complex dynamic pattern" that the React Compiler's `react-hooks/preserve-manual-memoization` rule flags as unsafe/impossible to analyze — the file even had `/* eslint-disable react-hooks/preserve-manual-memoization */` at the top to suppress it.

Fix: apply the positioning styles (`top`/`left`/`right`) directly via the `style` prop on the static `Container` component instead of dynamically generating a new styled component. This removes the need for `useMemo` entirely (no component-identity churn, no eslint-disable needed) while preserving the exact same rendered positioning behavior.

- Removed the `useMemo`-wrapped `ContainerDynamic` styled-component factory.
- Removed the now-unnecessary `/* eslint-disable react-hooks/preserve-manual-memoization */` comment and unused `useMemo` import.
- Render `<Container style={{ top, left, right }}>` directly.

## Test plan
- [x] `prettier --check` passes on the changed file.
- [x] `tsc --noEmit` shows no errors related to this file (a pre-existing, unrelated `react-dom` type-defs error exists in this environment on master too).
- [x] No dedicated unit tests exist for `BondPreview.tsx`; verified manually that the JSX/props change is behavior-preserving (same CSS properties applied, same conditional presence via optional style values).
- Note: this sandbox's local `node_modules` has a stale `eslint-plugin-react-hooks@4.6.0` (package.json specifies `^7.1.1`), so the new compiler rules (including `preserve-manual-memoization`) can't be executed locally here; this is reproducible identically on `master` with no code changes, so it's a pre-existing environment issue unrelated to this change and should resolve once CI installs the pinned dependency versions.


Jesovile (approved):

AlexeyGirin:

PR #11411, opened 2026-08-21
#11220, #11209 - Refactor: fix useMemo anti-patterns in EditorEvents.tsx

### LINKED ISSUE #11209: Refactor: useMemo anti-patterns clean up in `packages\ketcher-macromolecules\src\EditorEvents.tsx` file
**Problem:**

**Why is this an issue?**

Problems:

- requests;
- modifying the DOM;
- writing to localStorage;
- subscriptions;
- calling setState;
- logging and analytics.

**How can it be fixed?**
Problematic code:
```ts
  doSomething();
}, [value]);
```

```ts
return items.map(transform);
}, [items]);
```

or shorter:

If you need a side effect:
```ts
useEffect(() => {
  doSomething();
}, [value]);
```

```ts
const handleClick = () => {
  doSomething();
};
```

```ts
  doSomething();
}, [value]);
```

**Problem locations:**
packages\ketcher-macromolecules\src\EditorEvents.tsx

### LINKED ISSUE #11220: Refactor: Use memo rule violation clean up from `packages\ketcher-macromolecules\src\EditorEvents.tsx` file
**Problem:**

Example:
```ts
useMemo(() => {
}, []);
```

**Why is this an issue?**

**How can it be fixed?**
There are three common fixes.

Return the calculated value:
```ts
const total = useMemo(() => {
  return calculateTotal(items);
}, [items]);
```

```ts
const total = useMemo(
  [items],
);
```

```ts
useEffect(() => {
  logItems(items);
}, [items]);
```

```ts
```

**Problem locations:**
packages\ketcher-macromolecules\src\EditorEvents.tsx

nickkurkan:
## Summary
- Removes the stale file-level `/* eslint-disable react-hooks/use-memo */` in `packages/ketcher-macromolecules/src/EditorEvents.tsx`. It was added during the ESLint version bump (#10806) and, once the plugin resolves correctly, produces an "unused eslint-disable directive" warning — there is no violation left for it to suppress on its own.
- Converts `debouncedShowPreview` from `useCallback(debounce((p) => dispatchShowPreview(p), 500), [dispatchShowPreview])` to `useMemo(() => debounce((p) => dispatchShowPreview(p), 500), [dispatchShowPreview])`.
  - `react-hooks/use-memo` flags this because `useCallback`'s first argument must be an inline function expression, but here the code passes the *result* of calling `debounce(...)`, not a function literal.
  - The real intent is to memoize the *value* returned by `debounce(...)` (a debounced function), which is exactly what `useMemo` is for — so this isn't an event handler or side effect, just a mismatched hook choice.
  - Behavior is unchanged: `debouncedShowPreview` is still recreated only when `dispatchShowPreview` changes, still callable as a function, and `.cancel()` still works the same way in `handleClosePreview`.

No other `useMemo`/`useCallback` calls in the file needed changes — the rest of the file already uses `useCallback`/`useEffect` correctly for subscriptions and handlers.

Closes #11209, Closes #11220

## Test plan
- [x] `npm run test:eslint -- src/EditorEvents.tsx` in `packages/ketcher-macromolecules` — 0 errors, 0 warnings for this file (previously showed an "unused eslint-disable directive" warning before this fix, and an error when the disable was removed without the useMemo fix).
- [x] `npm run test:types` in `packages/ketcher-macromolecules` — passes.
- [ ] No dedicated unit tests exist for `EditorEvents.tsx`; this is a behavior-preserving lint cleanup only.


AlexeyGirin:

PR #11414, opened 2026-08-21
#11208, #11219 - Refactor: fix useMemo anti-pattern in MacromoleculePropertiesWindow

### LINKED ISSUE #11208: Refactor: useMemo anti-patterns clean up in `packages\ketcher-macromolecules\src\components\macromoleculeProperties\MacromoleculePropertiesWindow.tsx` file

### LINKED ISSUE #11219: Refactor: Use memo rule violation clean up from `packages\ketcher-macromolecules\src\components\macromoleculeProperties\MacromoleculePropertiesWindow.tsx` file

nickkurkan:
## Summary
- Fixes the `react-hooks/use-memo` (React Compiler) rule violation in `MacromoleculePropertiesWindow.tsx`: `debouncedRecalculateMacromoleculeProperties` was created with `useCallback(debounce(fn, 500), [])`, i.e. the argument passed to `useCallback` was the *result of calling* `debounce()` rather than an inline function expression — exactly the "useMemo/useCallback used incorrectly" pattern described in the issues.
- Replaced it with `useState`'s lazy initializer: `const [debouncedRecalculateMacromoleculeProperties] = useState(() => debounce(fn, 500))`. `debounce()` returns a stateful function (it holds its own timer), not a pure derived value, so it isn't a good fit for `useMemo` either — converting to `useMemo(() => debounce(fn, 500), [])` trades the `use-memo` error for a `react-hooks/preserve-manual-memoization` error, since the compiler can't verify it's safe to recompute. `useState`'s initializer runs exactly once, giving the same "create once, for the life of the component" semantics as the previous `useCallback(..., [])` without misusing a memoization hook for a side-effecting value.
- The two genuine `useMemo` calls in the file (`grossFormula`, `molecularMassValue`) were already correct (pure computation, proper `return`) and were left untouched.
- Removed the now-unneeded `/* eslint-disable react-hooks/use-memo */` directive, and the stale `react-hooks/set-state-in-effect` disable that eslint reported as unused once the `use-memo` issue was fixed (that anti-pattern was already fixed in #11303).
- Dropped the now-unused `useCallback` import.

Closes #11208, Closes #11219

## Verification
- `eslint` (project's real `test:eslint` script) on the file and the whole `ketcher-macromolecules` package: 0 errors/warnings.
- `tsc --noEmit` (`test:types`) for `ketcher-macromolecules`: passes.
- `prettier --check`: passes.
- `ketcher-macromolecules` unit tests (`test:unit`): 36/36 suites pass (no existing tests target this component directly).
- Verified with a standalone reproduction using `eslint-plugin-react-hooks@7.1.1`'s React Compiler-based `use-memo` rule that the specific violation is fixed and no new `preserve-manual-memoization`/`refs` errors are introduced beyond the pre-existing, already-disabled `react-hooks/refs` case.

## Test plan
- [x] Confirm the two remaining `useMemo` calls in the file (gross formula, molecular mass) are unchanged and still lint-clean.
- [x] Confirm opening/using the Macromolecule Properties window (measurement-unit changes, value inputs) is unaffected — the debounced recalculation function's identity and timing are unchanged, only how it's created.
- [x] `npm run test:eslint`, `npm run test:types`, `npm run test:unit` for `ketcher-macromolecules` all green.


AlexeyGirin:

AlexeyGirin:
Not approved by Roman
изображение

AlexeyGirin:

PR #11430, opened 2026-08-24
#11261 Refactor: react-hooks/exhaustive-deps rule violation clean up from packages\ketcher-react\src\script\ui\views\modal\components\process\Check\Check.tsx file

### LINKED ISSUE #11261: Refactor: `react-hooks/exhaustive-deps` rule violation clean up from `packages\ketcher-react\src\script\ui\views\modal\components\process\Check\Check.tsx` file

Example:
```ts

useEffect(() => {
  loadUser(userId);
}, []); // userId is missing
```

**Why is this a problem?**

This can cause:
- stale state or props;
- effects using old data;

**How is it usually fixed?**
```ts
useEffect(() => {
  loadUser(userId);
}, [userId]);
```

For a memoized callback:
```ts
}, [documentId, content]);
```

For a memoized value:
```ts
  [items, category],
);
```

**Problem locations:**
packages\ketcher-react\src\script\ui\views\modal\components\process\Check\Check.tsx

LashaDevLasha:
…p from packages\ketcher-react\src\script\ui\views\modal\components\process\Check\Check.tsx file

## Check list
- [ ] unit-tests written
- [ ] e2e-tests written
- [ ] documentation updated


packages/ketcher-react/src/script/ui/views/modal/components/process/Check/Check.tsx
```diff
@@ -229,9 +229,13 @@ const CheckDialog: FC<CheckDialogProps> = (props) => {
...
-    handleCheck();
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, []);
+    setIsStructureChecking(false);
+    onCheck(checkState.checkOptions).then(() => {
+      setIsStructureChecking(true);
+      setLastCheckDate(new Date());
+      setIsCheckedWithNewSettings(true);
+    });
+  }, [checkState, onCheck]);
```
Jesovile: I'm not sure in onCheck stability and identity. It's just a prop, and in case if regular arrow function is passed, this useEffect will be triggered on each re-render.

On one hand, it's good for fast-check from outer scope (fail-fast approach)
On the other hand, we can handle it with refs (as in [this PR](https://github.com/epam/ketcher/pull/11429))

I keep this decision on your side. As for me, both are well enough.

Jesovile (approved):

packages/ketcher-react/src/script/ui/views/modal/components/process/Check/Check.tsx
LashaDevLasha: Went with the ref approach — onCheck is created in mapDispatchToProps referencing ownProps.onCancel, so it gets a new reference whenever parent props change. Using a ref ensures the effect only re-runs when checkState actually changes, while still always calling the latest onCheck

packages/ketcher-react/src/script/ui/views/modal/components/process/Check/Check.tsx
```diff
@@ -215,6 +221,11 @@ const CheckDialog: FC<CheckDialogProps> = (props) => {
   const [isCheckedWithNewSettings, setIsCheckedWithNewSettings] =
     useState(false);
 
+  const onCheckRef = useRef(onCheck);
+  useEffect(() => {
+    onCheckRef.current = onCheck;
+  }, [onCheck]);
```
la-nuta: Replace the latest-`onCheck` ref with a ref that holds the whole mount-time call, so the effect and the Check button stay one code path and the deps can go back to `[]`. Pairs with the suggestion on the effect below — apply both.

```suggestion
  const initialCheckRef = useRef(() => onCheck(result.checkOptions));
```

packages/ketcher-react/src/script/ui/views/modal/components/process/Check/Check.tsx
```diff
@@ -229,9 +240,13 @@ const CheckDialog: FC<CheckDialogProps> = (props) => {
...
-    handleCheck();
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, []);
+    setIsStructureChecking(false);
+    onCheckRef.current(checkState.checkOptions).then(() => {
+      setIsStructureChecking(true);
+      setLastCheckDate(new Date());
+      setIsCheckedWithNewSettings(true);
+    });
+  }, [checkState]);
```
la-nuta:
```suggestion
  useEffect(() => {
    // On mount `isStructureChecking` is already false, so there is nothing to reset.
    initialCheckRef.current().then(() => {
      setIsStructureChecking(true);
      setLastCheckDate(new Date());
      setIsCheckedWithNewSettings(true);
    });
  }, []);
```

`initialCheckRef`'s initializer captures the first render's `onCheck` and `result`, which is precisely master's mount semantics, and the remaining `setState`s are inside `.then()` so `set-state-in-effect` does not apply. With this the line-1 `/* eslint-disable react-hooks/set-state-in-effect */` can go too.

packages/ketcher-react/src/script/ui/views/modal/components/process/Check/Check.tsx
la-nuta, line 227: The mount path and the button path need to send the same options, so capture the mount-time call itself rather than the latest `onCheck`. Pairs with the suggestion on the effect below — apply both.

```suggestion
  const initialCheckRef = useRef(() => onCheck(result.checkOptions));
```

packages/ketcher-react/src/script/ui/views/modal/components/process/Check/Check.tsx
la-nuta, line 249:
```suggestion
  useEffect(() => {
    // On mount `isStructureChecking` is already false, so there is nothing to reset.
    initialCheckRef.current().then(() => {
      setIsStructureChecking(true);
      setLastCheckDate(new Date());
      setIsCheckedWithNewSettings(true);
    });
  }, []);
```

`initialCheckRef`'s initializer captures the first render's `onCheck` and `result`, which is exactly what master's `}, []);` closure did — so mount behaviour is unchanged, including sending `result.checkOptions`. The remaining `setState`s are inside `.then()`, where `set-state-in-effect` does not apply.

Dropping `setIsStructureChecking(false)` here is the part that belongs to #11101 rather than #11261 — it is a no-op on mount (`useState(false)`), but it is what makes line 1 dead, so it is worth agreeing with #11300 before this lands.

Verified: `eslint` clean once line 1 goes, and `1:1 Unused eslint-disable directive` while it stays.

la-nuta (commented):
Two changes, both in the new effect — and a scope point that decides how this lands.

**1. The effect sends different options than the Check button.** On master the mount effect was `handleCheck();`, the same function the button calls, so there was one code path. The new effect sends `checkState.checkOptions` — the persisted redux options — while `handleCheck` (line 231) still sends `result.checkOptions`, the live form state. They agree today only by accident: `formsState.check` ships no `result`, so `result` falls back to `checkState` on the first render.

**2. `}, [checkState])` widens a mount-only effect.** `optionsReducer` replaces `state.options.check` on `SAVE_CHECK_OPTS`, dispatched by this dialog's own `onApply` (line 324) — so the dep array now asks for a second `/check` POST on Apply. It doesn't fire today only because the same handler also dispatches `MODAL_CLOSE` and React unmounts in the same batch. The intent is `[]`.

**This necessarily overlaps #11101.** Any ref-based fix for the deps also kills the line-1 `/* eslint-disable react-hooks/set-state-in-effect */`, which is #11101's deliverable (@LashaDevLasha's #11300, also approved). The suggestions below drop `setIsStructureChecking(false)` from the mount path — it is already `false` there — which genuinely removes the violation:

```
1:1  error  Unused eslint-disable directive (no problems were reported from 'react-hooks/set-state-in-effect')
```

I also tried keeping `handleCheck` behind a ref (`useRef(handleCheck)`) to preserve the single code path. It lints clean, but it kills line 1 too — by hiding the `setState` behind `.current()` rather than removing it, which would close #11101 only cosmetically. So there is no version of this that stays inside #11261 alone.

Cleanest route: apply these two, delete line 1, and close #11300 as folded in here. Otherwise land this PR first and rebase #11300 on top — the reverse order leaves `setIsStructureChecking(false)` in an effect body with line 1 already deleted:

```
242:5  react-hooks/set-state-in-effect  Calling setState synchronously within an effect can trigger cascading renders
```

Either way this needs agreeing with #11300 — neither PR can see it from its own diff.

packages/ketcher-react/src/script/ui/views/modal/components/process/Check/Check.tsx
LashaDevLasha: Applied. Replaced onCheckRef + sync effect with initialCheckRef, the ref initializer captures the first render's onCheck and result.checkOptions, which is exactly what the original }, [] closure did on master

packages/ketcher-react/src/script/ui/views/modal/components/process/Check/Check.tsx
LashaDevLasha: Applied both suggestions. Removed setIsStructureChecking(false) (it's already false at mount, useState(false)), switched to initialCheckRef.current(), and scoped the dep array to []. Dropped the /* eslint-disable react-hooks/set-state-in-effect */ directive on line 1,  it's now an unused directive. This folds in #11300's deliverable, so #11300 can be closed as resolved here

la-nuta (approved):

PR #11431, opened 2026-08-24
#11268 Refactor: react-hooks/exhaustive-deps rule violation clean up from packages\ketcher-react\src\script\ui\component\form\colorPicker\ColorPickerContent.tsx file

### LINKED ISSUE #11268: Refactor: `react-hooks/exhaustive-deps` rule violation clean up from `packages\ketcher-react\src\script\ui\component\form\colorPicker\ColorPickerContent.tsx` file

Example:
```ts

useEffect(() => {
  loadUser(userId);
}, []); // userId is missing
```

**Why is this a problem?**

This can cause:
- stale state or props;
- effects using old data;

**How is it usually fixed?**
```ts
useEffect(() => {
  loadUser(userId);
}, [userId]);
```

For a memoized callback:
```ts
}, [documentId, content]);
```

For a memoized value:
```ts
  [items, category],
);
```

**Problem locations:**
packages\ketcher-react\src\script\ui\component\form\colorPicker\ColorPickerContent.tsx

LashaDevLasha:
Replace terse inline "-- run once on mount" note with a full explanation of why each dep is intentionally omitted: pendingColor would cause re-seeding on every color selection; customColors/settings would create a feedback loop since the effect itself updates them via updateSettings.

## Check list
- [ ] unit-tests written
- [ ] e2e-tests written
- [ ] documentation updated


packages/ketcher-react/src/script/ui/component/form/colorPicker/ColorPickerContent.tsx
```diff
@@ -64,17 +70,21 @@ function ColorPickerContent({
...
-    const newColors = addCustomColor(customColors, pendingColor);
+    hasSeededRef.current = true;
+    const newColors = addCustomColor(
+      customColorsAtOpenRef.current,
+      pendingColorAtOpenRef.current,
+    );
     updateSettings({ colorPickerCustomColors: newColors });
-    // eslint-disable-next-line react-hooks/exhaustive-deps -- run once on mount
-  }, []);
+  }, [settings, updateSettings]);
```
la-nuta: The `hasSeededRef` guard is the part that does the work. The two value refs only matter if the effect ever runs on a pass later than mount — and that is exactly the pass where reading them is wrong: if `settings` is `null` on the first render, `customColors` falls back to `DEFAULT_CUSTOM_COLORS` (`[]`), `customColorsAtOpenRef` freezes that empty array, and when settings arrive the seed writes `[pendingColor]` over the user's saved list. On mount they are by definition equal to the live values, so they buy nothing.

I couldn't show that null-at-mount path is actually reachable — `useSettings` lazy-initialises from `ketcherProvider.getKetcher(ketcherId)?.settingsService`, which is present by the time a picker opens — so this is a latent hazard rather than a live bug. But the version without the refs is shorter and has no such edge:

```suggestion
  const hasSeededRef = useRef(false);

  // Seed the currently-selected color into the custom-color list once, when
  // the picker opens (ColorPickerContent mounts fresh on every open).
  useEffect(() => {
    if (settings === null || hasSeededRef.current) {
      return;
    }
    hasSeededRef.current = true;
    const newColors = addCustomColor(customColors, pendingColor);
    updateSettings({ colorPickerCustomColors: newColors });
  }, [customColors, pendingColor, settings, updateSettings]);
```

I built that variant and linted it — clean, plain and with `--report-unused-disable-directives`. `updateSettings` is a `useCallback` keyed on `ketcherId`, so it is stable; `settings` and `customColors` do churn after the seed writes, but `hasSeededRef` short-circuits those re-runs, so it still fires exactly once per open.

la-nuta: The `hasSeededRef` guard is the part that does the work. The two value refs only matter if the effect ever runs on a pass later than mount — and that is exactly the pass where reading them is wrong: if `settings` is `null` on the first render, `customColors` falls back to `DEFAULT_CUSTOM_COLORS` (`[]`), `customColorsAtOpenRef` freezes that empty array, and when settings arrive the seed writes `[pendingColor]` over the user's saved list. On mount they are by definition equal to the live values, so they buy nothing.

I couldn't show that null-at-mount path is actually reachable — `useSettings` lazy-initialises from `ketcherProvider.getKetcher(ketcherId)?.settingsService`, which is present by the time a picker opens — so this is a latent hazard rather than a live bug. But the version without the refs is shorter and has no such edge:

```suggestion
  const hasSeededRef = useRef(false);

  // Seed the currently-selected color into the custom-color list once, when
  // the picker opens (ColorPickerContent mounts fresh on every open).
  useEffect(() => {
    if (settings === null || hasSeededRef.current) {
      return;
    }
    hasSeededRef.current = true;
    const newColors = addCustomColor(customColors, pendingColor);
    updateSettings({ colorPickerCustomColors: newColors });
  }, [customColors, pendingColor, settings, updateSettings]);
```

I built that variant and linted it — clean, plain and with `--report-unused-disable-directives`. `updateSettings` is a `useCallback` keyed on `ketcherId`, so it is stable; `settings` and `customColors` do churn after the seed writes, but `hasSeededRef` short-circuits those re-runs, so it still fires exactly once per open.

la-nuta, line 87: Non-blocking. `hasSeededRef` already does the work here, so the two value refs can go:

```suggestion
  const hasSeededRef = useRef(false);

  // Seed the currently-selected color into the custom-color list once, when
  // the picker opens (ColorPickerContent mounts fresh on every open).
  useEffect(() => {
    if (hasSeededRef.current) {
      return;
    }
    hasSeededRef.current = true;
    if (settings === null) {
      return;
    }
    const newColors = addCustomColor(customColors, pendingColor);
    updateSettings({ colorPickerCustomColors: newColors });
  }, [customColors, pendingColor, settings, updateSettings]);
```

The guard order is the reason to prefer this shape. On master the deps are `[]`, so if `settings` is `null` on the first render the effect returns and never runs again. Here `settings` is a dep, so the effect runs again when settings arrive and seeds from the refs frozen at mount. Setting the flag before the null check keeps master's exactly-once semantics.

To be clear, that is not a live bug — I originally wrote it up as one and that was wrong. `useSettings` lazy-initialises from `ketcherProvider.getKetcher(ketcherId)?.settingsService` during render, and its subscribe effect bails permanently when no service is found, so `settings` cannot go from `null` to non-null while the picker is mounted. This is about keeping the shape simple and master-equivalent, not fixing a defect.

Verified: `eslint` clean plain and with `--report-unused-disable-directives`, and `tsc --noEmit` on `ketcher-react` clean.

The suggestion also restores the comment above the effect — it is the only place the mounts-fresh-per-open invariant is written down (`ColorPicker.tsx:85` has no `keepMounted`).

la-nuta (commented):
One non-blocking simplification, plus a note.

**1. `ColorPickerContent.tsx:73-87`** — `pendingColorAtOpenRef` and `customColorsAtOpenRef` are redundant; `hasSeededRef` already does the work, and dropping them also restores master's exactly-once guard order. See the inline comment — take it or leave it, it isn't blocking.

**2.** The PR description describes a different change ("replace terse inline note with a full explanation of why each dep is intentionally omitted") than the diff does. Probably a leftover from an earlier approach.

*(Edited: I first raised the seeding path as a defect. It isn't reachable — corrected in the inline note.)*

packages/ketcher-react/src/script/ui/component/form/colorPicker/ColorPickerContent.tsx
LashaDevLasha: Applied. Dropped pendingColorAtOpenRef and customColorsAtOpenRef — hasSeededRef alone is sufficient for exactly-once semantics. The guard order is now: set the flag first, then check settings === null, so if settings is absent on the first run the effect can never re-seed on a subsequent run even with settings in the dep array. Dep array is now exhaustive (customColors, pendingColor, settings, updateSettings), eslint-disable directive removed, and the mount-per-open comment is restored

la-nuta (approved):

PR #11433, opened 2026-08-24
#11266 Refactor: react-hooks/exhaustive-deps rule violation clean up from packages\ketcher-react\src\script\ui\dialog\template\useSaltsAndSolvets.ts file

### LINKED ISSUE #11266: Refactor: `react-hooks/exhaustive-deps` rule violation clean up from `packages\ketcher-react\src\script\ui\dialog\template\useSaltsAndSolvets.ts` file

Example:
```ts

useEffect(() => {
  loadUser(userId);
}, []); // userId is missing
```

**Why is this a problem?**

This can cause:
- stale state or props;
- effects using old data;

**How is it usually fixed?**
```ts
useEffect(() => {
  loadUser(userId);
}, [userId]);
```

For a memoized callback:
```ts
}, [documentId, content]);
```

For a memoized value:
```ts
  [items, category],
);
```

**Problem locations:**
packages\ketcher-react\src\script\ui\dialog\template\useSaltsAndSolvets.ts

LashaDevLasha:
…p from packages\ketcher-react\src\script\ui\dialog\template\useSaltsAndSolvets.ts file

## Check list
- [ ] unit-tests written
- [ ] e2e-tests written
- [ ] documentation updated


AlexeyGirin:

packages/ketcher-react/src/script/ui/dialog/template/useSaltsAndSolvets.ts
```diff
@@ -12,12 +18,17 @@ export default function useSaltsAndSolvents(
   saltsAndSolvents: Template[],
   filter: string,
 ) {
-  const [isFirstRender, setIsFirstRender] = useState(true);
+  const isFirstRenderRef = useRef(true);
+  const filterRef = useRef(filter);
```
Jesovile: It's strange for me to see refs for string prop.
Do we realy need this? If filter change, we need to recall filterFGLib, don't we?

packages/ketcher-react/src/script/ui/dialog/template/useSaltsAndSolvets.ts
```diff
@@ -12,12 +18,17 @@ export default function useSaltsAndSolvents(
...
   const timerId = useRef<null | ReturnType<typeof setTimeout>>(null);
   const [filteredSaltsAndSolvents, setFilteredSaltsAndSolvents] = useState(
     saltsAndSolvents[SALTS_AND_SOLVENTS],
   );
 
+  useLayoutEffect(() => {
```
Jesovile: In the first second it looks like strange way that cancels ol refs benefits. But... I see useLayoutEffect, not just useEffect.

Actually, in my opinion, it makes code more complex and descrease maintainability.

Is it just a fix of deps or some part of business logic?

packages/ketcher-react/src/script/ui/dialog/template/useSaltsAndSolvets.ts
```diff
@@ -39,13 +50,14 @@ export default function useSaltsAndSolvents(
...
+      filterFGLib(saltsAndSolvents, filterRef.current)[SALTS_AND_SOLVENTS] ??
+      [];
     addToSaSWithBatches(filteredSaS);
   }, [saltsAndSolvents, addToSaSWithBatches]);
 
   useEffect(() => {
-    if (isFirstRender) {
-      setIsFirstRender(false);
+    if (isFirstRenderRef.current) {
+      isFirstRenderRef.current = false;
```
Jesovile: I believe, this logic might be packed to a custom hook and reused across the application.
Can we consider such solution?

packages/ketcher-react/src/script/ui/dialog/template/useSaltsAndSolvets.ts
LashaDevLasha: Removed filterRef and the useLayoutEffect syncing it — merged both effects into one. Now filter is a direct dependency, so filterFGLib is recalled on every change. Also removed isFirstRenderRef since the single effect resets state and clears the timeout on every run, making the first-render guard unnecessary

packages/ketcher-react/src/script/ui/dialog/template/useSaltsAndSolvets.ts
LashaDevLasha: removed it entirely. The useLayoutEffect only existed to sync filterRef, and both are gone now. It was just a workaround to avoid adding filter to the deps, not actual business logic

packages/ketcher-react/src/script/ui/dialog/template/useSaltsAndSolvets.ts
LashaDevLasha: Removed the pattern entirely by merging the two effects into one — isFirstRenderRef is no longer needed at all, so no custom hook required

packages/ketcher-react/src/script/ui/dialog/template/useSaltsAndSolvets.ts
```diff
@@ -1,6 +1,6 @@
-/* eslint-disable react-hooks/exhaustive-deps */
 /* eslint-disable react-you-might-not-need-an-effect/no-event-handler */
 /* eslint-disable react-hooks/set-state-in-effect */
+/* eslint-disable react-hooks/immutability */
```
la-nuta: This line is new on the branch, and it should not be. #11212 (`c0adc5193d`) removed exactly this directive from master on 24 Aug after fixing the `.splice()` mutation, and your branch already contains that commit — re-adding it re-suppresses a rule master is clean on. Lint confirms it does nothing:

```
3:1  error  Unused eslint-disable directive (no problems were reported from 'react-hooks/immutability')
```

Same for lines 1 and 2 after this change, so the whole block can go:

```suggestion
```

packages/ketcher-react/src/script/ui/dialog/template/useSaltsAndSolvets.ts
```diff
@@ -38,22 +37,12 @@ export default function useSaltsAndSolvents(
   }, []);
 
   useEffect(() => {
+    clearTimeout(timerId.current as unknown as number);
+    setFilteredSaltsAndSolvents([]);
```
la-nuta: Behaviour change worth confirming before this lands. On master the `filter` path was the *second* effect, which replaced state with the whole filtered array in one `setState` — typing in the search box updated the list at once, and batching only ever applied to mount (and library reload).

Now every `filter` change goes through `addToSaSWithBatches`, so the list renders the first 16 rows and then grows by 16 every 300 ms. `salts-and-solvents.sdf` has ~135 entries, so clearing the search box takes ~2.4 s to fill in, and each further keystroke restarts the fill from 16.

Repro: Templates dialog → **Salts and Solvents** tab → type `a`. On master all matches appear together; here they arrive in 16-row steps.

If that timing wasn't intended, this restores it and lints clean — I built the variant and ran it, with and without the file headers, and `react-hooks/refs` is `off` for `ketcher-react` so the ref write is fine:

```ts
const hasBatchedRef = useRef(false);

useEffect(() => {
  clearTimeout(timerId.current as unknown as number);
  const filteredSaS =
    filterFGLib(saltsAndSolvents, filter)[SALTS_AND_SOLVENTS] ?? [];
  if (hasBatchedRef.current) {
    setFilteredSaltsAndSolvents(filteredSaS);
    return;
  }
  hasBatchedRef.current = true;
  setFilteredSaltsAndSolvents([]);
  addToSaSWithBatches(filteredSaS);
}, [saltsAndSolvents, filter, addToSaSWithBatches]);
```

If batching on every filter *is* the intent, say so and I'll drop this — but then it's a product change riding on a lint refactor and Jesovile should sign it off.

Jesovile (approved):

packages/ketcher-react/src/script/ui/dialog/template/useSaltsAndSolvets.ts
la-nuta, line 3: This line is new on the branch, and it shouldn't be. #11212 (`c0adc5193d`) removed exactly this directive from master on 24 Aug after fixing the `.splice()` mutation, and your branch already contains that commit — re-adding it re-suppresses a rule master is clean on. Lint confirms it does nothing:

```
3:1  warning  Unused eslint-disable directive (no problems were reported from 'react-hooks/immutability')
```

```suggestion
```

Lines 1 and 2 are dead after this change too, but they belong to #11245 and #11108 — worth leaving for those PRs rather than closing their tickets from here.

packages/ketcher-react/src/script/ui/dialog/template/useSaltsAndSolvets.ts
la-nuta, line 41: Behaviour change worth confirming before this lands. On master the `filter` path was the *second* effect, which replaced state with the whole filtered array in one `setState` — typing in the search box updated the list at once, and batching only ever applied to mount (and library reload).

Now every `filter` change goes through `addToSaSWithBatches`, so the list renders the first 16 rows and then grows by 16 every 300 ms. `salts-and-solvents.sdf` has ~135 entries, so clearing the search box takes ~2.4 s to fill in, and each further keystroke restarts the fill from 16.

Repro: Templates dialog → **Salts and Solvents** tab → type `a`. On master all matches appear together; here they arrive in 16-row steps.

If that timing wasn't intended, this restores it and lints clean — I built the variant and ran it, with and without the file headers, and `react-hooks/refs` is `off` for `ketcher-react` so the ref write is fine:

```ts
const hasBatchedRef = useRef(false);

useEffect(() => {
  clearTimeout(timerId.current as unknown as number);
  const filteredSaS =
    filterFGLib(saltsAndSolvents, filter)[SALTS_AND_SOLVENTS] ?? [];
  if (hasBatchedRef.current) {
    setFilteredSaltsAndSolvents(filteredSaS);
    return;
  }
  hasBatchedRef.current = true;
  setFilteredSaltsAndSolvents([]);
  addToSaSWithBatches(filteredSaS);
}, [saltsAndSolvents, filter, addToSaSWithBatches]);
```

If batching on every filter *is* the intent, say so and I'll drop this — but then it's a product change riding on a lint refactor, and worth calling out as one.

la-nuta (commented):
Three things before it lands.

**1. Drop the new `/* eslint-disable react-hooks/immutability */` on line 3.** It's new on this branch, and #11212 (`c0adc5193d`) removed exactly that directive from master on 24 Aug after fixing the `.splice()` mutation — a commit your branch already contains. See the inline note.

Lines 1 and 2 go dead once this lands as well:

```
1:1  warning  Unused eslint-disable directive (no problems were reported from 'react-you-might-not-need-an-effect/no-event-handler')
2:1  warning  Unused eslint-disable directive (no problems were reported from 'react-hooks/set-state-in-effect')
```

but those are #11245's and #11108's deliverables (#11374, #11376), so leaving them is the right call here — flagging them only so they aren't a surprise.

**2. Behaviour: filtering is now batched.** See the inline note on line 41.

**3. Landing order — two other open PRs rewrite this same hook.**

- #11376 (@ALexanderSPB, #11108) merges the two effects into one exactly as you do, but without `setFilteredSaltsAndSolvents([])`, so filtered results append to the previous list. Whichever lands second conflicts on the same block.
- #11374 (@nickkurkan, #11245) keeps both effects with `[saltsAndSolvents, addToSaSWithBatches]` and deletes the headers, so landing it after this one reopens #11266. Linted:

```
  36:6  error  React Hook useEffect has a missing dependency: 'filter'  react-hooks/exhaustive-deps
```

It also still carries the pre-#11212 `.splice()` body. Worth agreeing an order — this PR first, with the other two rebased or closed on top.

packages/ketcher-react/src/script/ui/dialog/template/useSaltsAndSolvets.ts
LashaDevLasha: Done — removed. Confirmed it was new on this branch and lint reports it as an unused directive

packages/ketcher-react/src/script/ui/dialog/template/useSaltsAndSolvets.ts
LashaDevLasha: Batching on every filter change was unintentional. Fixed with hasBatchedRef: the first run (mount) clears and batch-loads, all subsequent runs (filter changes) call setFilteredSaltsAndSolvents directly so results appear immediately — matching master behaviour

la-nuta (approved):

rrodionov91 (approved):

AlexeyGirin:

AlexeyGirin:

PR #11440, opened 2026-08-24
#11262 Refactor: react-hooks/exhaustive-deps rule violation clean up from packages\ketcher-react\src\script\ui\views\modal\components\document\Open\components\CDXStructuresViewer\CDXStructuresViewer.tsx file

### LINKED ISSUE #11262: Refactor: `react-hooks/exhaustive-deps` rule violation clean up from `packages\ketcher-react\src\script\ui\views\modal\components\document\Open\components\CDXStructuresViewer\CDXStructuresViewer.tsx` file

Example:
```ts

useEffect(() => {
  loadUser(userId);
}, []); // userId is missing
```

**Why is this a problem?**

This can cause:
- stale state or props;
- effects using old data;

**How is it usually fixed?**
```ts
useEffect(() => {
  loadUser(userId);
}, [userId]);
```

For a memoized callback:
```ts
}, [documentId, content]);
```

For a memoized value:
```ts
  [items, category],
);
```

**Problem locations:**
packages\ketcher-react\src\script\ui\views\modal\components\document\Open\components\CDXStructuresViewer\CDXStructuresViewer.tsx

LashaDevLasha:
…p from packages\ketcher-react\src\script\ui\views\modal\components\document\Open\components\CDXStructuresViewer\CDXStructuresViewer.tsx file

## Check list
- [ ] unit-tests written
- [ ] e2e-tests written
- [ ] documentation updated

### DIFF
--- a/packages/ketcher-react/src/script/ui/views/modal/components/document/Open/components/CDXStructuresViewer/CDXStructuresViewer.tsx
+++ b/packages/ketcher-react/src/script/ui/views/modal/components/document/Open/components/CDXStructuresViewer/CDXStructuresViewer.tsx
@@ -79,11 +79,13 @@ export const CDXStructuresViewer = ({
     [server],
   );
 
+  const currentStructItem = structList[selectedIndex];
+
   useEffect(() => {
-    if (structList[selectedIndex] && !itemsMap[selectedIndex]) {
-      getImage(structList[selectedIndex], selectedIndex);
+    if (currentStructItem && !itemsMap[selectedIndex]) {
+      getImage(currentStructItem, selectedIndex);
     }
-  }, [getImage, itemsMap, selectedIndex, structList]);
+  }, [getImage, itemsMap, selectedIndex, currentStructItem]);
 
   const renderStructure = (structure: item) => {
     if (loading) {
--- a/packages/ketcher-react/src/script/ui/views/modal/components/document/Open/components/CDXStructuresViewer/CDXStructuresViewer.tsx
+++ b/packages/ketcher-react/src/script/ui/views/modal/components/document/Open/components/CDXStructuresViewer/CDXStructuresViewer.tsx
@@ -1,4 +1,3 @@
-/* eslint-disable react-hooks/exhaustive-deps */
 /* eslint-disable react-you-might-not-need-an-effect/no-event-handler */
 /* eslint-disable react-hooks/set-state-in-effect */
 /****************************************************************************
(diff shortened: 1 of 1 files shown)


packages/ketcher-react/src/script/ui/views/modal/components/document/Open/components/CDXStructuresViewer/CDXStructuresViewer.tsx
```diff
@@ -61,27 +60,30 @@ export const CDXStructuresViewer = ({
...
+    },
+    [server],
+  );
 
   useEffect(() => {
     if (structList[selectedIndex] && !itemsMap[selectedIndex]) {
       getImage(structList[selectedIndex], selectedIndex);
     }
-  }, [itemsMap, selectedIndex]);
+  }, [getImage, itemsMap, selectedIndex, structList]);
```
Jesovile: I'm not shure in structList identity. It's a prop, and regular array may be passed.
We can use stringifyed json string instead of the object. Strings are strings, and this json will be updated only if content is updated.

May it be a solution or I'm completely on the Moon?

LashaDevLasha: Extracted structList[selectedIndex] into a currentStructItem string variable and used that as the dependency instead strings compare by value, so no identity issue

la-nuta (commented):
Nothing blocking in the diff itself — two notes on things outside it.

**1. Landing order vs #11372 (@nickkurkan).** It rewrites this same effect as `}, [structList]);` plus an inline `exhaustive-deps` disable that is load-bearing:

```
101:6  error  React Hook useEffect has a missing dependency: 'getImage'
```

So if #11372 lands after this PR, #11262 is silently reopened. This one first, then #11372 rebased without reintroducing that disable. It needs work regardless — its head still fails its own rule: `98:9 error Avoid using props and effects as an event handler … "structList"`.

**2. The `+/* eslint-disable react-hooks/set-state-in-effect */` line in the diff is a phantom.** Master dropped it in #11420 (`308f8e1003`) and `git merge-file` resolves it away with no conflict. Nothing to do — just don't hand-resolve it back in when you pick up master (merge or rebase, whichever you prefer).

la-nuta (approved):

Jesovile (approved):

rrodionov91 (approved):

PR #11442, opened 2026-08-24
#11275 - Refactor: react-hooks/exhaustive-deps rule violation clean up from packages\ketcher-macromolecules\src\hooks\useZoomTransform.ts file

### LINKED ISSUE #11275: Refactor: `react-hooks/exhaustive-deps` rule violation clean up from `packages\ketcher-macromolecules\src\hooks\useZoomTransform.ts` file

Example:
```ts

useEffect(() => {
  loadUser(userId);
}, []); // userId is missing
```

**Why is this a problem?**

This can cause:
- stale state or props;
- effects using old data;

**How is it usually fixed?**
```ts
useEffect(() => {
  loadUser(userId);
}, [userId]);
```

For a memoized callback:
```ts
}, [documentId, content]);
```

For a memoized value:
```ts
  [items, category],
);
```

**Problem locations:**
packages\ketcher-macromolecules\src\hooks\useZoomTransform.ts

mariam-khutuashvili:
The effect listed `ZoomTool.instance`, a static that the rule correctly reports as an unnecessary dependency — mutating it doesn't re-render, so it isn't a reactive value.

```
32:6  error  React Hook useEffect has an unnecessary dependency: 'ZoomTool.instance'
```

**Simply dropping it, as the rule suggests, would have been a regression.** This hook's consumers — `RulerArea` and `DragGhost` — are children of the macromolecules `Editor`, and React runs child effects **before** parent effects. The editor, and with it `ZoomTool.instance`, is created in an effect of that parent (`Editor.tsx` → `createEditor` → `ZoomTool.initInstance`), so on first mount the instance is still empty and the effect returns without subscribing. Listing the static is what made the effect *retry*: once the editor landed and the tree re-rendered, the value differed and the subscription was established. With `[]` it would run once on mount and never subscribe, silently leaving the ruler and drag ghost without zoom updates.

So the fix depends on the editor from the store instead:

```diff
+const editor = useAppSelector(selectEditor);
-}, [ZoomTool.instance]);
+}, [editor]);
```

That is the actual signal that the instance exists, and also what replaces it, so the subscription is established and re-established at exactly the right moments while the dependency is a genuine reactive value. Both consumers already read the same value through `selectEditor`, so no new store coupling is introduced.

This also resolves the existing `// TODO: ... should better rely on some promise/init event`, so that comment is removed along with the file-level `eslint-disable`.

### Verification

On ESLint 9.39.5 / eslint-plugin-react-hooks 7.1.1:
- `--print-config` confirms `exhaustive-deps` resolves to `error` for this file
- removing the suppression **before** the fix reported exactly this violation
- the fixed file reports no problems across all rules
- restoring the static dependency makes the rule fire again, confirming it's still enforced
- prettier clean; ketcher-macromolecules jest: 35 suites / 155 tests pass

The 11 pre-existing type errors in ketcher-macromolecules and the 2 failing `SequenceSyncEditModeButton` tests were checked against clean master and reproduce unchanged — unrelated to this PR.

## Check list
- [ ] unit-tests written
- [ ] e2e-tests written
- [ ] documentation updated

### DIFF
--- a/packages/ketcher-macromolecules/src/hooks/useZoomTransform.ts
+++ b/packages/ketcher-macromolecules/src/hooks/useZoomTransform.ts
@@ -1,9 +1,15 @@
-/* eslint-disable react-hooks/exhaustive-deps */
 import { useEffect, useState } from 'react';
 import { ZoomTransform } from 'd3';
 import { ZoomTool } from 'ketcher-core';
+import { selectEditor } from 'state/common';
+import { useAppSelector } from './stateHooks';
 
 export const useZoomTransform = () => {
+  // ZoomTool.instance is created by the editor, which is set up in an effect of
+  // a parent component, so it is still empty when this hook first runs. The
+  // editor from the store is what tells us the instance now exists, and it is
+  // also what replaces it, so the subscription is (re)established from it.
+  const editor = useAppSelector(selectEditor);
   const [transform, setTransform] = useState<ZoomTransform>(
     new ZoomTransform(1, 0, 0),
   );
@@ -29,8 +35,7 @@ export const useZoomTransform = () => {
     return () => {
       zoom.unsubscribeOnZoomEvent(zoomEventHandler);
     };
-    // TODO: Perhaps it's not the best approach, should better rely on some promise/init event
-  }, [ZoomTool.instance]);
+  }, [editor]);
 
   return transform;
 };
--- a/packages/ketcher-macromolecules/src/hooks/useZoomTransform.ts
+++ b/packages/ketcher-macromolecules/src/hooks/useZoomTransform.ts
@@ -4,10 +4,6 @@ import { selectEditor } from 'state/common';
 import { useAppSelector } from './stateHooks';
 
 export const useZoomTransform = () => {
-  // The editor creates its zoom tool while being constructed in an effect of a
-  // parent component, so it is still absent when this hook first runs.
-  // Depending on the tool itself subscribes as soon as it exists and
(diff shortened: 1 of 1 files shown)


packages/ketcher-macromolecules/src/hooks/useZoomTransform.ts
```diff
@@ -1,9 +1,15 @@
...
 import { ZoomTool } from 'ketcher-core';
+import { selectEditor } from 'state/common';
+import { useAppSelector } from './stateHooks';
 
 export const useZoomTransform = () => {
+  // ZoomTool.instance is created by the editor, which is set up in an effect of
+  // a parent component, so it is still empty when this hook first runs. The
+  // editor from the store is what tells us the instance now exists, and it is
+  // also what replaces it, so the subscription is (re)established from it.
+  const editor = useAppSelector(selectEditor);
```
svvald, line 12: Perhaps we could read `const zoomTool = editor.zoomTool` then and use it in effect, this will make it reactive and resolve it properly

svvald (dismissed):

packages/ketcher-macromolecules/src/hooks/useZoomTransform.ts
mariam-khutuashvili: Done — reading `editor?.zoomTool` and depending on that. Needed the optional chaining since `selectEditor` is typed `CoreEditor | undefined`.

AlexeyGirin (approved):
Approved by @rrodionov91 
Image

PR #11512, opened 2026-08-31
#11231 - Refactor: Reading from or writing to ref.current during React render problem clean up from packages\ketcher-macromolecules\src\components\modal\monomerConnection\MonomerConnections.tsx file

### LINKED ISSUE #11231: Refactor: Reading from or writing to `ref.current` during React render problem clean up from `packages\ketcher-macromolecules\src\components\modal\monomerConnection\MonomerConnections.tsx` file
**Problem:**
Examples:
```ts
function Component() {
  const valueRef = useRef(0);

}
```

**Why is this an issue?**

**How can it be fixed?**
```ts
function Component() {

  useEffect(() => {
    inputRef.current?.focus();
  }, []);

}
```

For event-driven access:
```ts
const handleClick = () => {
  ref.current?.focus();
};
```

```ts

return (
    {count}
  </button>
);
```

**Problem locations:**
packages\ketcher-macromolecules\src\components\modal\monomerConnection\MonomerConnections.tsx

AnzhelaMartsinovich:
Replaced refs used for storing the initial monomer attachment points and attachment point availability with lazily initialized state.

MonomerConnections no longer reads ref.current during render. The obsolete react-hooks/refs ESLint suppression and unused useRef import were also removed.

The existing connection and reconnection behavior remains unchanged.

## Check list
- [ ] unit-tests written
- [ ] e2e-tests written
- [ ] documentation updated

### DIFF
--- a/packages/ketcher-macromolecules/src/components/modal/monomerConnection/MonomerConnections.tsx
+++ b/packages/ketcher-macromolecules/src/components/modal/monomerConnection/MonomerConnections.tsx
@@ -1,10 +1,9 @@
-/* eslint-disable react-hooks/refs */
 import styled from '@emotion/styled';
 import { ActionButton } from 'components/shared/actionButton';
 import { Modal } from 'components/shared/modal';
 import { useAppSelector } from 'hooks';
 import { selectEditor } from 'state/common';
-import { useRef, useState } from 'react';
+import { useState } from 'react';
 import {
   AttachmentPoint,
   AttachmentPointName as AttachmentPointNameComponent,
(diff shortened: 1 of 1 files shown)


packages/ketcher-macromolecules/src/components/modal/monomerConnection/MonomerConnections.tsx
```diff
@@ -70,14 +69,15 @@ const MonomerConnection = ({
...
+  const [initialSecondMonomerAttachmentPoint] = useState(
+    () => polymerBond?.secondMonomerAttachmentPoint,
   );
-  const hasFreeAttachmentPointsRef = useRef(
-    firstMonomer?.hasFreeAttachmentPoint ||
+  const [hasFreeAttachmentPoints] = useState(
+    () =>
+      firstMonomer?.hasFreeAttachmentPoint ||
       secondMonomer?.hasFreeAttachmentPoint,
   );
```
svvald: Do we even need to have these fields as states? They don't have a `set...` function so we might as well just derive these values from props in render time, right?

svvald (changes requested):

packages/ketcher-macromolecules/src/components/modal/monomerConnection/MonomerConnections.tsx
AnzhelaMartsinovich: right, fixed it.

AnzhelaMartsinovich: Good point, I tried it. However, in this case polymerBond and the monomers are mutable domain objects. Selecting another attachment point mutates them before the dialog is closed, so deriving the values on every render loses the original connection and prevents Cancel from restoring it correctly.

I’ve updated the implementation to keep an explicitly named initialConnection snapshot initialized once when the dialog opens. This avoids reading ref.current during render while preserving the existing Cancel and Reconnect behavior.

svvald (approved):

PR #11513, opened 2026-08-31
#11232 - Refactor: Reading from or writing to ref.current during React render problem clean up from packages\ketcher-macromolecules\src\components\macromoleculeProperties\MacromoleculePropertiesWindow.tsx file

### LINKED ISSUE #11232: Refactor: Reading from or writing to `ref.current` during React render problem clean up from `packages\ketcher-macromolecules\src\components\macromoleculeProperties\MacromoleculePropertiesWindow.tsx` file
**Problem:**
Examples:
```ts
function Component() {
  const valueRef = useRef(0);

}
```

**Why is this an issue?**

**How can it be fixed?**
```ts
function Component() {

  useEffect(() => {
    inputRef.current?.focus();
  }, []);

}
```

For event-driven access:
```ts
const handleClick = () => {
  ref.current?.focus();
};
```

```ts

return (
    {count}
  </button>
);
```

**Problem locations:**

AnzhelaMartsinovich:
Moved the creation of the debounced macromolecule properties recalculation handler from the render phase to an effect.

Refs are now accessed only inside effects and deferred event handlers. This keeps the component render pure while preserving a single stable debounce instance and ensuring that the latest recalculation callback is used.

Added cleanup to cancel pending debounced calls when the component is unmounted. The obsolete react-hooks/refs and react-hooks/use-memo ESLint suppressions were removed.

The existing macromolecule properties recalculation behavior remains unchanged.

## Check list
- [ ] unit-tests written
- [ ] e2e-tests written
- [ ] documentation updated

### DIFF
--- a/packages/ketcher-macromolecules/src/components/macromoleculeProperties/MacromoleculePropertiesWindow.tsx
+++ b/packages/ketcher-macromolecules/src/components/macromoleculeProperties/MacromoleculePropertiesWindow.tsx
@@ -980,10 +980,10 @@ export const MacromoleculePropertiesWindow = () => {
   const recalculateMacromoleculeProperties =
     useRecalculateMacromoleculeProperties();
   const skipDataFetch = !isMacromoleculesPropertiesWindowOpened;
-  const debouncedRecalculateMacromoleculeProperties = useDebouncedCallback(
-    recalculateMacromoleculeProperties,
-    500,
-  );
+  const {
+    debouncedCallback: debouncedRecalculateMacromoleculeProperties,
+    cancel: cancelDebouncedRecalculateMacromoleculeProperties,
+  } = useDebouncedCallback(recalculateMacromoleculeProperties, 500);
 
   useEffect(() => {
     if (recalculatePropertiesHandler) {
@@ -1030,14 +1030,14 @@ export const MacromoleculePropertiesWindow = () => {
   // re-runs the effect above and schedules a debounced call; cancel it so
   // only this immediate calculation actually runs.
   useEffect(() => {
-    debouncedRecalculateMacromoleculeProperties.cancel();
+    cancelDebouncedRecalculateMacromoleculeProperties();
     recalculateMacromoleculeProperties(skipDataFetch);
   }, [
     unipositiveIonsMeasurementUnit,
     oligonucleotidesMeasurementUnit,
     skipDataFetch,
-    debouncedRecalculateMacromoleculeProperties,
     recalculateMacromoleculeProperties,
+    cancelDebouncedRecalculateMacromoleculeProperties,
   ]);
 
   // The properties object is re-parsed from the Indigo response on every
--- a/packages/ketcher-macromolecules/src/components/macromoleculeProperties/MacromoleculePropertiesWindow.tsx
+++ b/packages/ketcher-macromolecules/src/components/macromoleculeProperties/MacromoleculePropertiesWindow.tsx
@@ -1,5 +1,4 @@
 /* eslint-disable react-hooks/exhaustive-deps */
-/* eslint-disable react-hooks/use-memo */
 /****************************************************************************
  * Copyright 2021 EPAM Systems
  *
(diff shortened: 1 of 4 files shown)


packages/ketcher-macromolecules/src/components/macromoleculeProperties/MacromoleculePropertiesWindow.tsx
```diff
@@ -993,13 +990,29 @@ export const MacromoleculePropertiesWindow = () => {
   const recalculateMacromoleculePropertiesRef = useRef<
     (shouldSkip?: boolean) => void
   >(recalculateMacromoleculeProperties);
+  const debouncedRecalculateMacromoleculePropertiesRef = useRef<
+    ReturnType<typeof debounce> | undefined
+  >(undefined);
   const debouncedRecalculateMacromoleculeProperties = useCallback(
```
svvald: This does stack a third layer of indirection (ref → ref → `useCallback` proxy → effect) just to get one stable debounced callback. Worth considering a small reusable `useDebouncedCallback(fn, delay)` hook instead — it would collapse this to fewer moving parts and be reusable elsewhere (e.g. `HydrophobicityChart`'s `ResizeObserver`/`debounce` further down in this same file has the same "recreated every render, never cancelled" issue).

Separately, `debouncedRecalculateMacromoleculePropertiesRef` (holds the actual debounce instance) and `debouncedRecalculateMacromoleculeProperties` (the stable proxy that forwards to it) are easy to conflate at a glance since they differ only by the `Ref` suffix. A more distinct name for the ref (e.g. `debounceInstanceRef`) would reduce misread risk.

packages/ketcher-macromolecules/src/components/macromoleculeProperties/MacromoleculePropertiesWindow.tsx
```diff
@@ -993,13 +990,29 @@ export const MacromoleculePropertiesWindow = () => {
...
-    debounce((shouldSkip?: boolean) => {
-      recalculateMacromoleculePropertiesRef.current(shouldSkip);
-    }, 500),
+    (shouldSkip?: boolean) => {
+      debouncedRecalculateMacromoleculePropertiesRef.current?.(shouldSkip);
+    },
     [],
   );
 
+  useEffect(() => {
```
svvald: `debouncedRecalculateMacromoleculePropertiesRef.current?.cancel()` (and the `?.()` call in the proxy above) silently no-ops if invoked before this effect has populated the ref. That's safe today only because effect declaration order guarantees this mount effect runs before the other effects that reference the ref — but nothing documents that ordering dependency. A short comment here would help a future refactor (e.g. extracting this into a custom hook, or reordering effects) not silently break it.

svvald (commented):
Reviewed with code-quality, performance, documentation-accuracy, and security passes.

**Security:** no findings — this is a self-contained internal refactor with no new input handling, network calls, or DOM writes.

**Correctness/performance:** the core fix is sound. I traced the effect ordering explicitly — the new mount effect that creates the debounce instance is declared before every effect that dereferences it, so there's no dropped first call in practice. This also closes a real pre-existing gap: the old code never cancelled the debounce on unmount, so a pending 500ms call could fire and dispatch after teardown. The `.cancel()` in the effect cleanup fixes that. Removing the two `eslint-disable` comments (`react-hooks/refs`, `react-hooks/use-memo`) is good hygiene — worth noting for context that neither is an actual rule ID in the currently-installed `eslint-plugin-react-hooks@4.3.0`, so they were already dead/no-op suppressions.

Left two non-blocking inline notes on maintainability (the indirection is more layered than the problem strictly requires, and the ordering invariant it relies on isn't documented). Nothing here blocks merging.

packages/ketcher-macromolecules/src/components/macromoleculeProperties/MacromoleculePropertiesWindow.tsx
AnzhelaMartsinovich: thanks

svvald (approved):

PR #11528, opened 2026-09-01
#7328 Version update, resulting in errors

### LINKED ISSUE #7328: Version update, resulting in errors
I updated the version of ketcher-core and ketcher-react to 3.4 in the react project, and the following error occurred. However, there is no such problem in version 3.2. I am not sure whether it is affected by the new features added in the new version.

Image

Image

Image

Image

LashaDevLasha:
…tive SVG dimensions

## Check list
- [ ] unit-tests written
- [ ] e2e-tests written
- [ ] documentation updated

### DIFF
--- a/packages/ketcher-core/src/application/editor/tools/Zoom.ts
+++ b/packages/ketcher-core/src/application/editor/tools/Zoom.ts
@@ -430,7 +430,12 @@ export class ZoomTool implements BaseTool {
       return;
     }
 
-    this.zoom?.transform(this.canvasWrapper, new ZoomTransform(1, 0, 0));
+    try {
+      this.zoom?.transform(this.canvasWrapper, new ZoomTransform(1, 0, 0));
+    } catch {
+      // SVG canvas may have relative dimensions (e.g. width="100%") that
+      // d3 cannot resolve to an absolute pixel value — skip the reset.
+    }
   }
 
   observeCanvasResize = () => {
--- a/packages/ketcher-core/src/application/editor/tools/Zoom.ts
+++ b/packages/ketcher-core/src/application/editor/tools/Zoom.ts
@@ -111,7 +111,7 @@ export class ZoomTool implements BaseTool {
 
   initActions() {
     this.zoom = zoom<SVGSVGElement, void>()
-      .extent(() => {
+      .extent((): [[number, number], [number, number]] => {
         const rect = this.canvasWrapper.node()?.getBoundingClientRect();
         return [
           [0, 0],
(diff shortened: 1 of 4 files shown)


packages/ketcher-core/src/application/editor/tools/Zoom.ts
```diff
@@ -430,7 +430,12 @@ export class ZoomTool implements BaseTool {
       return;
     }
 
-    this.zoom?.transform(this.canvasWrapper, new ZoomTransform(1, 0, 0));
+    try {
+      this.zoom?.transform(this.canvasWrapper, new ZoomTransform(1, 0, 0));
+    } catch {
+      // SVG canvas may have relative dimensions (e.g. width="100%") that
+      // d3 cannot resolve to an absolute pixel value — skip the reset.
+    }
```
dorogusha: I think the diagnosis is right, but I'd suggest fixing it a bit differently.

The actual throw happens inside d3-zoom's Gesture constructor, which runs on every zoom.transform call (including resetZoom, but also zoomIn/zoomOut, wheel, drag, pinch):

```
function Gesture(that, args) {
  ...
  this.extent = extent.apply(that, args); // <- here
}
```

`defaultExtent()` falls back to e.width.baseVal.value / e.height.baseVal.value whenever the SVG has no viewBox attribute. Our #polymer-editor-canvas (packages/ketcher-macromolecules/src/Editor.tsx) is rendered with width="100%" height="100%" and no viewBox, so defaultExtent hits the percentage-based branch — and resolving a percentage SVGAnimatedLength.baseVal.value throws when the element momentarily has no layout context (detached/hidden/mid-remount), which is the classic "Could not resolve relative length in the given context" error.

Since this happens inside Gesture's constructor, it's not specific to resetZoom - zoomIn/zoomOut, scrollbar dragging, and wheel zoom all create a Gesture too and should be able to throw the same way. Wrapping only resetZoom in try/catch fixes the one call site we happened to hit, but the same crash is likely still reachable from normal zoom/pan interactions.

I'd rather fix it at the source: give the zoom behavior a custom .extent() in initActions() that reads the size via getBoundingClientRect() instead of relying on baseVal.value - getBoundingClientRect() never throws, it just returns zeros for an unattached/zero-size element:

```
this.zoom = zoom<SVGSVGElement, void>()
  .extent(() => {
    const rect = this.canvasWrapper.node()?.getBoundingClientRect();
    return [[0, 0], [rect?.width ?? 0, rect?.height ?? 0]];
  })
  .scaleExtent([this.MINZOOMSCALE, this.MAXZOOMSCALE])
  ...

```
That removes the need for the try/catch entirely and covers all zoom/pan paths, not just resetZoom.

While you're in there, `initScrollBars()` and `scrollTo()` read .width.baseVal.value / .height.baseVal.value too, unguarded - initScrollBars() already computes getBoundingClientRect() for boundingBox/wrapperBoundingBox right above, so switching those to rect.width/rect.height would fix the same fragility there and keep the method consistent.

dorogusha (approved):

rrodionov91 (approved):

PR #11536, opened 2026-09-01
#10954 - Refactor: Explicit/implicit Any type clean up from packages/ketcher-core/src/application/editor/actions/atom.ts file

### LINKED ISSUE #10954: Refactor: Explicit/implicit Any type clean up from `packages/ketcher-core/src/application/editor/actions/atom.ts` file
**Problem:** 
Unspecified any.

**Why is this an issue?**

**How can it be fixed?**

**Problem locations:**
packages/ketcher-core/src/application/editor/actions/atom.ts

mariam-khutuashvili:

Removes the file-level `no-explicit-any` suppression and types the three explicit `any`s plus the five untyped signatures, all from existing types: `ReStruct`, `Point`, `Partial<AtomAttributes>`, `Pile<number>`, `Action`, `AtomAttributeName`.

The `as` casts are replaced by narrowing rather than moved around, per the previous review. `FragmentAdd.frid` and `AtomAdd.data.aid` are nullable, so the operations are held in locals and asserted after `perform()`; `execute()` always assigns both, so those asserts can't fire.

Two spots needed care to keep behaviour identical:

- **S-group context** — a missing `context` matched none of the not-expanded ones and fell through, so it reads with an empty-string fallback. An early return would have inverted that and skipped every s-group without one.
- **`mergeSgroups`** gets arrays from bond/template but a `Pile` (a `Set`) from `mergeFragmentsIfNeeded`, and lodash `without` returns `[]` for a `Set` — so that path adds no atoms today. The array case is selected explicitly and the `Pile` case keeps producing nothing, with a comment. Worth raising separately as a bug rather than fixing it silently here.

One deliberate delta: asserting `frid` turns a silent skip into a throw, but only for a non-existent atom id — `Atom.fragment` is `number` defaulting to `-1`, so it's always a number for a real atom. That's the premise the review stated, and it's what lets the cast go.

### Verification

`tsc` — **0 errors in ketcher-core, ketcher-react and ketcher-standalone** (`fromAtomsAttrs` alone has 40 call sites across both); macromolecules unchanged at its 13 pre-existing errors. eslint/prettier clean, restoring an `any` makes the rule fire again, ketcher-core suite passes at **822 tests**, ketcher-react identical to master.

## Check list
- [ ] unit-tests written
- [ ] e2e-tests written
- [ ] documentation updated

### DIFF
--- a/packages/ketcher-core/src/application/editor/actions/atom.ts
+++ b/packages/ketcher-core/src/application/editor/actions/atom.ts
@@ -32,6 +32,7 @@ import { fromRGroupFragment, fromUpdateIfThen } from './rgroup';
 import { fromBondStereoUpdate } from './bondStereo';
 
 import { Action } from './action';
+import { isNumber } from 'lodash';
 import { without } from 'lodash/fp';
 import type { Pile } from 'domain/entities/pile';
 import type { Point } from 'domain/entities/vec2';
@@ -165,9 +166,9 @@ export function mergeFragmentsIfNeeded(
 ) {
   const frid = atomGetAttr(restruct, srcId, 'fragment');
   const frid2 = atomGetAttr(restruct, dstId, 'fragment');
-  assert(typeof frid === 'number', `Fragment of atom ${srcId} was not found`);
+  assert(isNumber(frid), `Fragment of atom ${srcId} was not found`);
 
-  if (frid2 !== frid && typeof frid2 === 'number') {
+  if (frid2 !== frid && isNumber(frid2)) {
     const struct = restruct.molecule;
 
     const rgid = RGroup.findRGroupByFragment(struct.rgroups, frid2);
--- a/packages/ketcher-core/src/application/editor/actions/atom.ts
+++ b/packages/ketcher-core/src/application/editor/actions/atom.ts
@@ -1,4 +1,3 @@
-/* eslint-disable @typescript-eslint/no-explicit-any */
 /****************************************************************************
  * Copyright 2021 EPAM Systems
  *
@@ -15,7 +14,7 @@
  * limitations under the License.
  ***************************************************************************/
 
-import { Atom } from 'domain/entities/atom';
+import { Atom, type AtomAttributes } from 'domain/entities/atom';
 import { MonomerMicromolecule } from 'domain/entities/monomerMicromolecule';
 import { RGroup } from 'domain/entities/rgroup';
 import {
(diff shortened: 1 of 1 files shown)


packages/ketcher-core/src/application/editor/actions/atom.ts
```diff
@@ -140,11 +157,17 @@ export function fromAtomsFragmentAttr(restruct, aids, newfrid) {
...
+  restruct: ReStruct,
+  srcId: number,
+  dstId: number,
+) {
+  const frid = atomGetAttr(restruct, srcId, 'fragment');
   const frid2 = atomGetAttr(restruct, dstId, 'fragment');
+  assert(typeof frid === 'number', `Fragment of atom ${srcId} was not found`);
 
-  if (frid2 !== frid && typeof frid === 'number' && typeof frid2 === 'number') {
+  if (frid2 !== frid && typeof frid2 === 'number') {
```
svvald: Let's use `isNumber` from lodash for `frid2` here

packages/ketcher-core/src/application/editor/actions/atom.ts
```diff
@@ -140,11 +157,17 @@ export function fromAtomsFragmentAttr(restruct, aids, newfrid) {
...
-  const frid = atomGetAttr(restruct, srcId, 'fragment') as number;
+export function mergeFragmentsIfNeeded(
+  action: Action,
+  restruct: ReStruct,
+  srcId: number,
+  dstId: number,
+) {
+  const frid = atomGetAttr(restruct, srcId, 'fragment');
   const frid2 = atomGetAttr(restruct, dstId, 'fragment');
+  assert(typeof frid === 'number', `Fragment of atom ${srcId} was not found`);
```
svvald: And same here

svvald (changes requested):

packages/ketcher-core/src/application/editor/actions/atom.ts
mariam-khutuashvili: Done — both now use isNumber from lodash. Verified the assert still narrows frid to number: removing it makes tsc fail at getFragmentIds(frid), so the type predicate flows through the assertion signature exactly as typeof did.

packages/ketcher-core/src/application/editor/actions/atom.ts
mariam-khutuashvili: Done — both now use isNumber from lodash. Verified the assert still narrows frid to number: removing it makes tsc fail at getFragmentIds(frid), so the type predicate flows through the assertion signature exactly as typeof did.

svvald (approved):

